PR #54, opened 2018-10-10
Fix: You can rate your own routes

### LINKED ISSUE #48: You can rate your own routes
Bit of an oversight, this should not be possible :)

mltnhm:
Fixes #48

### DIFF
--- a/app/Http/Controllers/APIDungeonRouteController.php
+++ b/app/Http/Controllers/APIDungeonRouteController.php
@@ -167,6 +167,10 @@ class APIDungeonRouteController extends Controller
     {
         $value = $request->get('rating', -1);
 
+        if ($dungeonroute->isOwnedByCurrentUser()) {
+            return response()->setStatusCode(403);
+        }
+
         if ($value > 0) {
             $user = Auth::user();
 
--- a/app/Models/DungeonRoute.php
+++ b/app/Models/DungeonRoute.php
@@ -451,6 +451,15 @@ class DungeonRoute extends Model
         return $result;
     }
 
+    /**
+     * @return bool
+     */
+    public function isOwnedByCurrentUser()
+    {
+        $user = Auth::user();
+        return $this->author_id === $user->id;
+    }
+
     public static function boot()
     {
         parent::boot();
--- a/app/Models/DungeonRoute.php
+++ b/app/Models/DungeonRoute.php
@@ -456,6 +456,10 @@ class DungeonRoute extends Model
      */
     public function isOwnedByCurrentUser()
     {
+        if (!Auth::check()) {
+            return false;
+        }
+
         $user = Auth::user();
         return $this->author_id === $user->id;
     }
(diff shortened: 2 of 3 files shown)


app/Http/Controllers/APIDungeonRouteController.php
```diff
@@ -167,6 +167,10 @@ function rate(Request $request, DungeonRoute $dungeonroute)
     {
         $value = $request->get('rating', -1);
 
+        if ($dungeonroute->isOwnedByCurrentUser()) {
+            return response()->setStatusCode(403);
```
Wotuu: This is more of a personal thing, but I prefer to have a single point of exit in my code which is located at the end of the function. Unless it'd really make my code ugly I try to adhere to it. I think it's best if you try the same, unless you got a good argument against it of course. Otherwise I'm just gonna change it myself since I've been doing it for so long now :)

app/Models/DungeonRoute.php
```diff
@@ -451,6 +451,15 @@ public function isFavoritedByCurrentUser()
         return $result;
     }
 
+    /**
+     * @return bool
+     */
+    public function isOwnedByCurrentUser()
+    {
+        $user = Auth::user();
+        return $this->author_id === $user->id;
```
Wotuu: You also need to check if `$user` is null (user not logged in).

Wotuu:
Thanks for the contribution! Those two little things and it should be good to go!

app/Http/Controllers/APIDungeonRouteController.php
mltnhm: I usually place most of my constraints at the beginning to prevent any unnecessary database calls etc. So if anything I'd personally move it up to go above the `$value` assignment.

Wotuu: What I try to do (but not always I guess) is have a clear happy flow, then put any exceptions in elses. Example:

```
if( happy1 ){
    if( happy2) {
        happy();
    } else {
        error2();
    }
} else {
    error1();
}
```
In the end you return the result below all that. I dislike having to find a bunch of return paths once functions start to grow bigger where it turns out that halfway through are a bunch of return statements.

Of course as you say you should exit prior to doing unnecessary database updates. I just prefer to if chain with a single point of return rather than a return at the beginning :).

I'm not bashing your code by the way, I'm trying to find a common ground for both of us so we continue in the same style etc, otherwise we're going to end up diverging and that's never good.

Wotuu: Btw, I'm going to have to push my local changes now as I'm about to leave work, I'm quite sure you're going to get merge conflicts since I already fixed this bug. As I said before, I'll start working more organized so that this won't happen too often in the future.

Wotuu: I did some more thinking, and I think it's okay to put your constraints at the top of the function and return then. Just because I wouldn't do it that way shouldn't mean you can't. It's not a bad practice by any standard so it's fine :)

PR #283, opened 2020-05-13
Development

Wotuu:


Wotuu:
Spoken context from the user when starting this work: This session should work on all three related issues just filed on Katie's own Argus repo, using sub-agents: issue 283 (no dedicated window reserved for Argus self-work), issue 284 (stop asking for window confirmation when the user doesn't care which window is used), and issue 285 (session busy status is stale — windows reported occupied are often actually free). Address all three in this session. Make sure the working directory is the Argus repository itself, not any other project repo.

Issue #537, opened 2020-12-28
Add indicated key level range for routes and allow filtering based on it


Wotuu:
This way people can indicate their key is for levels <10 for example, so they don't take Prideful into account.

PR #565, opened 2021-01-02
#537 Keystone level implementation

greatman:
I implemented the Keystone level on dungeon routes.

It defaults to a 10, the level list is from 2 to 30.

The next best feature to add for this would be to hide the affix in the table that are not active on that key level (Example only Tyrannical on a +2). Didn't find yet how to do it.

I'm greatman on the discord server if you want to chat :)

New page:
image

Edit page:
image

Search Page (It does a >=):
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,4 @@ yarn-error.log
 .env
 .phpstorm.meta.php
 _ide_helper.php
+/storage/debugbar/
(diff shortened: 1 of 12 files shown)


Wotuu:
As discussed on Discord, thanks a lot for your contribution! Just checked it out and ran into some issues. The missing class, but I'm also missing a migration for the `keystone_level` for the `dungeon_routes` table :)

greatman:
Hmm, must of gotten lost when I was fixing the commit history. Fixing it!

Wotuu:
Hi, I'm going to close this PR - for the redesign I've touched a lot of this code and it isn't quite compatible anymore with what I've rewritten so it'll be quicker for me to start over. Thanks a lot either way :)

PR #614, opened 2021-01-13
Adding docker-compose

triat:
Here is a first version of what I could do with docker and docker-compose. This seems to work on my linux. It's certainly not 100% ready yet, but this is already a base and this can evolve through time.

There is still some issue with the installation and some manual work which I'd like to simplify when I understand better the way it works


Wotuu:
Looks good! I will keep this PR open for a while until I have time to check out the docker-compose things myself. Keep me updated on your progress as you've been doing, I'd definitely like to move to this solution!

Issue #3093, opened 2026-02-02
Heatmap selection shows the new season, but the dungeon list is empty


Wotuu:
This is correct - but it should then not show the empty list at all.

PR #3094, opened 2026-02-02
#3093 Heatmap selection no longer shows empty seasons (because no dun…

Wotuu:
…geon is enabled for this season yet).


Wotuu:
Adding this to the release since the release had some issues compiling. This can still be inserted.

Issue #2528, opened 2024-10-01
Move Split floors/facade toggle outside of settings menu


Wotuu:
Explore page can't toggle between the two now which is annoying at best

PR #3245, opened 2026-06-02
#2528 Move facade/split floors toggle to left sidebar

Wotuu:
## Summary

- Moves the facade/split floors display toggle from the map settings modal to the left sidebar, making it accessible on all map pages including Explore (which has no pulls sidebar)
- The toggle only renders when the dungeon's mapping version has `facade_enabled = true`, so it's invisible for dungeons that don't support it
- Clicking the button saves the preference (cookie + AJAX for logged-in users) and reloads the page
- Removed the toggle from the map settings modal since it now has a dedicated place in the sidebar

## Test plan

- [ ] Open a dungeon with `facade_enabled = true` on the Explore page — confirm the new button appears in the left sidebar
- [ ] Click the button — page reloads and the dungeon renders in the toggled mode
- [ ] Open a dungeon without `facade_enabled` — confirm the button does not appear
- [ ] Confirm the facade style setting is no longer shown in the map settings modal


Wotuu (commented):
All good

PR #3256, opened 2026-06-05
#2483 recognize affixes better for low level arc runs

### LINKED ISSUE #2483: ARC does not recognize affix groups if you didn't do a max-level dungeon
Some dungeons now just have Fortified for example. It should still recognize this and save routes accordingly.

Wotuu:
Closes #2483


Wotuu (commented):
Ok

PR #3285, opened 2026-06-12
#3275 Real-time live-sessions

### LINKED ISSUE #3275: Make live-sessions truly real-time and automatic
Live-sessions currently work by users manually selecting a pull, and then which enemies they overpulled. Keystone.guru then calculates which enemies can be removed from the route (by checking pulled skippable enemies) and adjusting the route while still reaching 100% enemy forces.

Assuming we can get a stream of combat log information from the players doing the run, we can see what's happening and do the following:

a) Render player locations live on the map
b) Hide killed enemies (using the auto-route creator to determine what enemies were killed)
c) automatically suggest corrections should the players go off-route
d) show other things we can find in the combat log, but that'll come later

A combat log can be 100s of thousands of lines long, and be generated over the course of 30-60 minutes. Hence the need for filtering. We can have over a dozen to a 100 concurrent sessions, each generating incoming combat log lines.

Step 1: we need a standalone NodeJS app that can read a combat log file, perform filters based on the combat log lines (filters undefined for now, but we need a system to be able to add filters), and send the log lines that pass the filters to the main Keystone.guru site
Step 1.1: we need a way to feed this tool a completed combat log, and have it be fed events at an accellerated time, configurable, x2, x5, x10 speed. That way we can debug things repeatedly without having to go in-game and generate a combat log live. There are timestamps on each combat log line to achieve this.
Step 2: We need a receiver on the Keystone.guru side that will receive the incoming events (server-side reverb receiver? other plain websocket solution?)
Step 3: The "current session data" needs to be persisted (what enemies have been killed, which enemies are part of the route but should no longer be killed, where are the players currently on the map, for starters). The moment someone refreshes their page, they should be right back where they left off
Step 4: Events need to be fired to the web browser, which will respond in kind by updating where the players are etc.
Step 5: The auto-route creator needs to be used to determine what enemies are killed as the combat log lines flow in. Currently ARC is stateless, as in, it performs its actions in one-shot, instead of being trickle-fed data over the course of multiple requests. This will need to be addressed somehow.

### LINKED ISSUE #3277: Live sessions: ingestion endpoint for raw filtered combat-log lines
Part of #3275 (live combat-log streaming for live sessions). **Plan: Issue 3 of 8.**

Authenticated REST endpoint on keystone.guru that accepts batches of raw combat-log lines for a given live session and persists them to a raw buffer. Near-real-time is not required (≤~30s end-to-end delay acceptable); transport is batched REST POST.

## Scope
- New route under the existing API v1 internal-team group in `routes/api.php` (e.g. `POST /api/v1/combatlog/livesession/{liveSession}/events`), throttled like the existing `api-combatlog-*` limiters.
- Controller alongside `app/Http/Controllers/Api/V1/InternalTeam/Combatlog/APICombatLogController.php` (e.g. `APILiveSessionCombatLogController@store`). Body: `{ lines: string[] }` (+ optional batch sequence number for ordering/idempotency).
- Form Request (`exists`-validated `liveSession` via cached getter, matching CLAUDE.md pattern) authorizing the caller against the live session / its dungeon route.
- Persist appended raw lines to a per-session buffer on the **main DB**; this raw buffer is the source of truth for ARC re-runs.
  - **Size/compression:** an accumulated buffer can reach **multiple megabytes** per session. Combat logs compress extremely well, so store the buffer **compressed** — e.g. a gzipped blob column written/read via `gzencode`/`gzdecode` (or MySQL `COMPRESS()` / InnoDB page compression). The diff in the ARC issue needs to read the full buffer back, so favour app-side gzip of one `longblob`/`mediumblob` per session over a row-per-line table.

## Tests
Happy path, auth failure, expired/missing session, malformed body, round-trip of the compressed buffer.

## Dependencies
Depends on the existing `LiveSession` model. Unblocks the Node streamer (real target), the state schema, and the incremental ARC issue.

## Conventions
Run all commands in Docker. Form Request validation (string-rule style), `exists` + cached getter for IDs, PHPUnit feature tests with factories, `__()` keys in `lang/en_US`, `sprintf` over concatenation, stage new files with `git add`, finish with `composer run fix` + `composer run analyse`. No DB foreign keys.

### LINKED ISSUE #3278: Live sessions: combat-state schema + persistence service
Part of #3275 (live combat-log streaming for live sessions). **Plan: Issue 4 of 8.**

Main-DB tables capturing the "current session data" so a browser refresh restores exactly where the user was.

## Scope (new tables on the main connection, keyed by `live_session_id`, no FKs)
- **Killed enemies** — which mapped enemies have been detected killed. **Key by `npc_id` + `mdt_id`, NOT `enemy_id`.** Although `kill_zone_enemies` recently re-gained an `enemy_id` column, `enemy_id`s are **recalculated when a route's mapping version is upgraded**, which can happen mid-session for popular routes (we don't want to block upgrades during live sessions). `npc_id`+`mdt_id` stays stable across mapping-version upgrades, exactly like `OverpulledEnemy` (see `LiveSession::getEnemies()` joining on `npc_id`+`mdt_id`). Resolve back to live `Enemy` rows at read time via the same join.
- **Obsolete enemies** — enemies in the route that should no longer be killed (already a concept in `MapContextLiveSession` via route correction; persist the live equivalent).
- **Player positions** — latest known LatLng + floor per player/character, updated as the log flows in.
- A `LiveSessionCombatStateService` to read/write this state and expose it as arrays for the map context.
- Extend `app/Logic/MapContext/Map/MapContextLiveSession.php::toArray()` to include the persisted killed/obsolete enemies + player positions, so initial page load (and refresh) hydrates from DB.
- Migrations + models + factories; feature tests for the service and for the map-context hydration.

## Dependencies
Independent structurally, but consumed by the events/rendering, incremental ARC, and off-route issues.

## Conventions
Run all commands in Docker. PHPUnit feature tests with factories, `__()` keys in `lang/en_US`, `sprintf` over concatenation, stage new files with `git add`, finish with `composer run fix` + `composer run analyse`. No DB foreign keys.

### LINKED ISSUE #3279: Standalone Node combat-log streamer (skeleton + filter system)
Part of #3275 (live combat-log streaming for live sessions). **Plan: Issue 1 of 8.**

New standalone Node app (new top-level dir, e.g. `combatlog-streamer/`, own `package.json`; separate from the Laravel Mix build) that reads a combat-log file and streams filtered raw lines to keystone.guru.

> **Why standalone / portability is a hard requirement:** this app is intended to be (near) copy-pasted into Raider.IO's existing **desktop client** (which runs on the player's machine and is Node-based). It must be self-contained, depend on as few/standard npm packages as possible, have no coupling to the Laravel repo or its build tooling, and be trivially liftable into their codebase. Design every part with "drop into someone else's Node project" friction in mind.

## Scope
- Tail/read a combat-log `.txt` file line by line (WoW format; lines start with `M/D HH:MM:SS.mmm`).
- A pluggable **filter pipeline**: an ordered list of filter functions, each `(line) => bool`, easy to extend. Filters are undefined for now — ship 1–2 trivial ones (e.g. drop blank lines) plus a clear "add a filter here" extension point.
- Buffer passing lines and POST them in micro-batches (size/interval configurable) to the ingestion endpoint (#3277). Until that endpoint is live, target a stub/log output.
- Config via env/CLI: target URL, API token, live-session key, batch size, flush interval.
- Basic README on running it.

## Dependencies
None to start. Pairs with the ingestion endpoint #3277 for the real target.

### LINKED ISSUE #3280: Node streamer: accelerated playback mode (x2/x5/x10)
Part of #3275 (live combat-log streaming for live sessions). **Plan: Issue 2 of 8.**

Add a mode to the Node streamer (#3279) that replays a *completed* combat log at an accelerated, configurable rate using the per-line timestamps, so the whole pipeline can be debugged repeatedly without generating a live log in-game.

## Scope
- Parse the `M/D HH:MM:SS.mmm` timestamp per line; sleep `delta / speed` between emissions to simulate real timing compressed by the speed factor.
- `--speed` flag (x2, x5, x10, …) and an "instant/no-throttle" option for fast smoke tests.
- Reuses the same filter pipeline + batching from #3279.

## Dependencies
Depends on #3279 (Node streamer skeleton).

### LINKED ISSUE #3281: Live sessions: incremental ARC via accumulated-buffer re-run + kill diffing
Part of #3275 (live combat-log streaming for live sessions). **Plan: Issue 6 of 8.**

Turn the incoming raw-line buffer into detected kills/route state by re-running the **existing one-shot ARC** over the whole accumulated buffer each batch and diffing. No ARC refactor up front — true incremental ARC is a separate, deferred optimization.

## Scope
- A job/service that, per ingestion batch (or on a short debounce), assembles the session's accumulated raw lines (from #3277) into the inputs the existing pipeline expects and runs `CombatLogRouteDungeonRouteService` (or `CombatLogRouteDungeonRouteBuilder` directly) to produce the current set of killed enemies / kill zones.
- **On-disk vs in-memory:** the current ARC/parse path requires the combat log to **exist as a file on disk**. The accumulated buffer (#3277) lives in the DB (gzipped). Two paths, prefer the first:
  - *Low-change interim:* decompress the buffer to a file on a **RAM disk** (tmpfs) per re-run and feed the existing on-disk path. Minimal change, gets us running.
  - *Cleaner follow-up (can fold in here or split out):* teach the parse entry points to accept an **in-memory stream/string** in addition to a file path, removing the temp-file round-trip. Note `CombatLogService::parseCombatLogStreaming()` already parses line by line.
- Reuse the existing parse path (`CombatLogStringParser`, result events under `app/Service/CombatLog/ResultEvents/`) — do **not** add a second parser.
- Diff the new ARC result against the persisted state (#3278): newly-killed enemies, changed obsolete-enemy set, latest player positions (from advanced-event coordinates via `CoordinatesService`).
- Persist deltas (#3278) and fire the broadcast events (events/rendering issue).
- Run as a queued job (`ShouldQueue`, Horizon) so ingestion POSTs stay fast; debounce/coalesce so we don't re-run per single line.

## Tests
Feature tests using a real sample log (`tests/CombatLogs/...`) replayed in chunks, asserting kills accumulate correctly and the final state matches a one-shot run of the full log.

## Known cost
Re-running over a growing buffer is O(n) per batch; acceptable for MVP, flagged for the true-incremental optimization issue.

## Dependencies
Depends on #3277 (ingestion) and #3278 (state schema). Hardest issue — keep tightly scoped to "detect kills + positions"; defer correction polish to the off-route issue.

## Conventions
Run all commands in Docker. PHPUnit feature tests with factories. Finish with `composer run fix` + `composer run analyse`.

### LINKED ISSUE #3282: Live sessions: broadcast events + browser handlers for live combat updates
Part of #3275 (live combat-log streaming for live sessions). **Plan: Issue 5 of 8.**

Push state changes to the browser and have the map react.

## Scope
- New `ContextEvent` subclasses broadcasting on the existing `{app.type}-live-session.{liveSession}` presence channel (see `routes/channels.php`), e.g.:
  - `EnemyKilledEvent` (enemy hidden), `PlayerMovedEvent` (position update), `RouteCorrectionEvent` (obsolete-enemy set changed / off-route suggestion).
  - Model them on `app/Events/OverpulledEnemy/OverpulledEnemyChangedEvent.php`.
- Front-end handlers under `resources/assets/js/custom/echo/messagehandler/listen/...` registered in `echohandler.js`, mirroring the overpulled-enemy handlers: hide killed enemies on the map, move/draw player markers, apply route corrections.
- Player-marker rendering on the Leaflet map (new lightweight layer; reuse existing map/icon utilities under `resources/assets/js/custom/`).
- After changes, rebuild JS (`npm run dev` / `composer run dev`).

## Dependencies
Depends on #3278 (state to broadcast). Can be developed in parallel with #3281 (incremental ARC) using fake/manually-triggered events. Ensure the `reverb` service is up for browser delivery.

## Conventions
Run all commands in Docker. PHPUnit feature tests for the events. Finish with `composer run fix` + `composer run analyse`.

### LINKED ISSUE #3283: Live sessions: off-route detection & automatic correction suggestions
Part of #3275 (live combat-log streaming for live sessions). **Plan: Issue 7 of 8.**

Use the live ARC state + player positions to detect when players deviate from the planned route and suggest corrections, reusing the existing forces-balancing logic.

## Scope
- Build on `OverpulledEnemyService::getRouteCorrection()` / `app/Service/LiveSession/DungeonRouteCorrection.php` to recompute obsolete enemies from the *live-detected* kills (#3281) instead of manual overpull selections.
- Heuristic for "off-route": killed enemies / positions diverging from the planned kill-zone order; emit a `RouteCorrectionEvent` (#3282) with the suggested adjustment.
- Tests for representative deviation scenarios.

## Dependencies
Depends on #3282 (events/rendering) and #3281 (incremental ARC).

## Conventions
Run all commands in Docker. PHPUnit feature tests with factories. Finish with `composer run fix` + `composer run analyse`.

### LINKED ISSUE #3284: Live sessions: optimize ARC to be truly incremental (deferred)
Part of #3275 (live combat-log streaming for live sessions). **Plan: Issue 8 of 8 — optional / deferred.**

If accumulated-buffer re-runs (#3281) strain the system at ~100 concurrent sessions, refactor ARC to maintain `ActivePullCollection` / matched-enemy state between batches so each batch only processes new lines.

## Scope
- Persist/restore in-progress pull + matched-enemy state across batches.
- Guarantee parity with the one-shot path (same fixtures, same final route) to avoid behavioral drift.
- Benchmark before/after under concurrent-session load.

## Dependencies
Depends on #3281 (incremental ARC) being in production with real load data. Explicitly deferred — only tackle if load demands it.

Wotuu:


Wotuu:
:robot:

## `Enemy::factory()->create()` leaks garbage rows into the shared dev DB

`tests/Feature/App/Service/LiveSession/LiveSessionCombatStateServiceTest.php` calls `Enemy::factory()->create()` in five places (currently around lines 239, 264, 268, 301, 509) but none of the surrounding `finally` blocks delete the created `Enemy` row(s) — only the `LiveSession*` join rows and the session itself. Since this project has no `RefreshDatabase` (persistent shared dev DB, cleanup is manual try/finally per `CLAUDE.md`), every run of this test file permanently leaks 1-2 `Enemy` rows straight into the real `enemies` table.

`database/factories/EnemyFactory.php` (new on this branch) fills every column with `$this->faker->randomNumber()`, including `mapping_version_id => randomNumber()` — a value with no relation to any real dungeon. Most leaked rows land on a nonexistent `mapping_version_id` (harmless garbage), but a random collision with a **real, current** mapping version is possible, and I found one: 5 of 28 leaked rows (accumulated from prior test runs) had landed on real mapping versions, including the *current* mapping version for The Eye of Eternity. Any time that dungeon's mapping version gets cloned forward (cutting/saving a new version), the phantom enemy comes along for the ride — this is what caused "saving the mapping version randomly adds enemies to dungeons," reported and root-caused in a separate session today. I've already deleted the 28 accumulated leaked rows from the shared dev DB, but the test still leaks on every run until fixed.

### Fix
Capture the created enemy id(s) and delete them in each `finally`, e.g.:

```php
$enemies = Enemy::factory()->count(2)->create();
try {
    ...
} finally {
    LiveSessionInCombatEnemy::query()->where('live_session_id', $liveSession->id)->delete();
    Enemy::whereIn('id', $enemies->pluck('id'))->delete();
    $liveSession->delete();
    $liveSession->dungeonRoute?->delete();
}
```

Same pattern needed at the other 4 call sites, including the standalone `$oldEnemy = Enemy::factory()->create();` and the two inline `Enemy::factory()->create()` calls passed directly into `collect([...])`.

Separately worth considering (not a blocker, just flagging): even with cleanup fixed, `mapping_version_id => $this->faker->randomNumber()` in `EnemyFactory` still has a nonzero chance of colliding with a real mapping version if a leak happens some other way (e.g. a failed migration mid-suite). Pinning it to a value that structurally can't collide with a real FK (e.g. a large negative number, if the column allows it) would remove that risk entirely.

Wotuu:
:robot: Fixed — captured the created `Enemy` row(s) at each of the 5 call sites and delete them in the surrounding `finally` (moved 2 that were declared inside the `try` back out, since PHPStan correctly flagged them as possibly-undefined in `finally`). Also pinned `EnemyFactory::mapping_version_id` to a negative number (the column is a signed int with no FK constraint), so a future leak from this or any other factory-based test can never land on a real, positive `MappingVersion` id. Verified: all 19 tests in the file still pass, and confirmed 0 negative-mapping-version `Enemy` rows remain after a full run (no leak).

database/migrations/2026_06_12_000004_rename_overpulled_enemies_to_live_session_overpulled_enemies_table.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration {
+    public function up(): void
+    {
+        Schema::rename('overpulled_enemies', 'live_session_overpulled_enemies');
```
Wotuu: :robot: **Migration safety — destructive rename in the same release that removes the last consumer (expand/contract violation).**

`Schema::rename('overpulled_enemies', 'live_session_overpulled_enemies')` ships in the same release that repoints all code to the new table name. Deploys here are **not atomic**: the `migrate` cron runs independently of the ECS web rollout, so during the rollout window the *previous* release's containers are still serving traffic against the new schema.

Old (merge-base) code queries the old table name on user-reachable paths:
- `app/Models/LiveSession.php::getEnemies()` — `->join('overpulled_enemies', ...)` (reached via the live-session map view → `MapContextLiveSession`)
- `app/Service/LiveSession/OverpulledEnemyService::getOverpulledEnemyForces()` — raw SQL `left join overpulled_enemies ...`
- `AjaxOverpulledEnemyController` write path

Once the cron renames the table, every one of those queries hits a now-missing table → `SQLSTATE base table not found` → HTTP 500 on the live-session view (and overpull writes) until the new containers finish rolling out. This is the exact #3497 failure class.

Split expand/contract across two releases: in release N keep the old table (dual-name/view or copy), remove the last old consumer; in release N+1 drop/rename once nothing references the old name.

database/migrations/2026_06_16_000000_drop_enemy_id_from_live_session_overpulled_enemies_table.php
```diff
@@ -0,0 +1,32 @@
...
+return new class extends Migration {
+    public function up(): void
+    {
+        Schema::table('live_session_overpulled_enemies', static function (Blueprint $table) {
+            // Drop indexes that reference enemy_id before dropping the column.
+            // Indexes kept their original prefix from before the table rename.
+            if (Schema::hasColumn('live_session_overpulled_enemies', 'enemy_id')) {
+                $table->dropIndex('overpulled_enemies_enemy_id_index');
+                $table->dropIndex('overpulled_enemies_live_session_id_kill_zone_id_enemy_id_index');
+                $table->dropColumn('enemy_id');
```
Wotuu: :robot: **Migration safety — destructive column drop in the same release as its last consumer.**

`dropColumn('enemy_id')` is destructive and lands in the same release that removes the code using it. During the non-atomic rolling deploy the previous release is still running `OverpulledEnemyService::getOverpulledEnemyForces()`, whose raw SQL joins on the dropped column:

```
left join `enemies` on `enemies`.`id` = `overpulled_enemies`.`enemy_id`
```

(merge-base `app/Service/LiveSession/OverpulledEnemyService.php:121`). Once `enemy_id` is dropped, that still-running old container's query references an unknown column → SQL error → 500 on the live-session correction path — independent of the table rename.

Defer the drop to release N+1, after the last consumer is gone (expand/contract).

app/Service/LiveSession/LiveSessionCombatStateService.php
```diff
@@ -0,0 +1,158 @@
...
+        ?int        $specializationId = null,
+    ): LiveSessionPlayerPosition {
+        $model = LiveSessionPlayerPosition::query()->updateOrCreate(
+            [
+                'live_session_id' => $liveSession->id,
+                'player_guid'     => $playerGuid,
+            ],
+            // Only overwrite the class/spec when we actually resolved them from a COMBATANT_INFO,
+            // so a later position update without one never wipes a previously-known specialization.
+            array_filter([
```
Wotuu: :robot: **Correctness — `array_filter()` silently drops falsy positional values (`lat`/`lng`/`character_name`).**

```php
array_filter([
    'class_id'          => $classId,
    'specialization_id' => $specializationId,
    'character_name'    => $characterName,
    'lat'               => $lat,
    'lng'               => $lng,
    'floor_id'          => $floorId,
    'updated_at'        => now(),
]),
```

`array_filter()` with no callback removes **every** falsy value — `0`, `0.0`, `''` — not just the `null` class/spec you intend to skip. A player whose resolved map coordinate is `lat == 0.0` or `lng == 0.0` has that field stripped:
- on an **existing** row the stale previous coordinate is kept (the dot freezes / never reaches the 0 axis);
- on the **first insert** the field is omitted entirely, and `live_session_player_positions.lat`/`lng` are `double NOT NULL` with no default, so the INSERT fails under strict mode and the job throws.

Only `class_id`/`specialization_id` should be conditionally omitted. Filter just those two keys (e.g. `array_filter([...], fn($v) => $v !== null)`) and merge the always-present positional fields unconditionally.

app/Http/Controllers/Api/V1/InternalTeam/Combatlog/APILiveSessionCombatLogController.php
```diff
@@ -0,0 +1,74 @@
...
+
+        $existingLines = [];
+        if ($buffer->buffer !== null) {
+            $decoded = gzdecode($buffer->buffer);
+            if ($decoded !== false && $decoded !== '') {
+                $existingLines = explode("\n", $decoded);
+            }
+        }
+
+        $allLines   = array_merge($existingLines, $lines);
```
Wotuu: :robot: **Correctness — lost-update race on the shared buffer row (unsynchronized read-modify-write).**

```php
$buffer = LiveSessionCombatLogBuffer::query()->firstOrNew(['live_session_id' => $liveSession->id]);
...
$existingLines = ... gzdecode($buffer->buffer) ...;
$allLines   = array_merge($existingLines, $lines);
$compressed = gzencode(implode("\n", $allLines), 6);
$buffer->buffer = ...;
$buffer->save();
```

Two batches for the same live session arriving close together both `firstOrNew` the same row, both `gzdecode` the identical existing content, each append only their own lines, and each `gzencode` + `save`. The second save clobbers the first → one batch's combat-log lines are permanently lost. There is no row lock, atomic append, or overlap guard on this endpoint (unlike the job, which has `WithoutOverlapping`). The `batch_sequence` check only rejects *older/equal* sequences; it does not serialize two concurrent in-flight appends. Consider a `lockForUpdate()` inside a transaction, or an atomic append strategy.

app/Service/LiveSession/LiveSessionBufferProcessingService.php
```diff
@@ -0,0 +1,492 @@
...
+        $reducedLines = $this->liveSessionCombatLogService->reduceCombatLogForBuffer($tmpFile, $validNpcIds);
+        $compressed   = gzencode(implode("\n", $reducedLines), 6);
+        if ($compressed === false) {
+            $this->log->reduceBufferUnableToCompress();
+
+            return;
+        }
+
+        $query = LiveSessionCombatLogBuffer::query()->where('live_session_id', $liveSession->id);
+        if ($lastSequenceAtRead === null) {
```
Wotuu, line 231: :robot: **Correctness — `reduceBuffer()` can clobber freshly ingested lines when batches use a null `batch_sequence`.**

The conditional-write guard only protects against `last_sequence` changing:

```php
if ($lastSequenceAtRead === null) {
    $query->whereNull('last_sequence');
} else {
    $query->where('last_sequence', $lastSequenceAtRead);
}
$query->update(['buffer' => $compressed]);
```

But `batch_sequence` is nullable in `APILiveSessionCombatLogRequest`, and the ingest endpoint only updates `last_sequence` when a sequence is supplied. A batch ingested with a null sequence appends new lines **without** changing `last_sequence`. If it lands while this job is running, the job's `whereNull('last_sequence')->update()` still matches and overwrites the buffer with the reduced set derived from the pre-append snapshot — silently discarding the just-ingested lines. Either make `batch_sequence` required/monotonic, or guard the reduce write on buffer content/`updated_at` as well.

app/Service/LiveSession/LiveSessionBufferProcessingService.php
```diff
@@ -0,0 +1,492 @@
...
+    private function resolveEnemy(
+        int            $npcId,
+        float          $ingameX,
+        float          $ingameY,
+        int            $uiMapId,
+        Collection     $availableEnemies,
+        MappingVersion $mappingVersion,
+    ): ?Enemy {
+        /** @var Collection<int, Enemy> $candidates */
+        $candidates = $availableEnemies->filter(static fn(Enemy $e) => $e->npc_id === $npcId);
```
Wotuu, line 467: :robot: **Correctness — `resolveEnemy()` compares distances across floors, so a same-`npc_id` enemy on the wrong floor can win.**

Candidates are filtered by `npc_id` only:

```php
$candidates = $availableEnemies->filter(static fn(Enemy $e) => $e->npc_id === $npcId);
```

The sighting is then projected onto the single floor picked from `uiMapId` (or the default floor), and min squared-distance is taken over **all** same-`npc_id` candidates regardless of each candidate's own `floor_id`. In a multi-floor dungeon where the same NPC appears on more than one floor, per-floor lat/lng share the same coordinate space, so a candidate on a different floor can be numerically closer and the kill / in-combat state is attributed to the wrong `Enemy` row (wrong-floor marker lights up). Filter `$candidates` to the resolved floor before the nearest-point search.

app/Models/LiveSession/LiveSessionCombatLogBuffer.php
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * @property int         $id
+ * @property int         $live_session_id
+ * @property string|null $buffer
+ * @property int|null    $last_sequence
+ *
+ * @property LiveSession $liveSession
+ */
+class LiveSessionCombatLogBuffer extends Model
```
Wotuu, line 18: :robot: **Convention — new models are missing their required repositories.**

`.claude/CLAUDE.md` (Model Creation): *"Every new model must also have a repository. Create the interface at `app/Repositories/Interfaces/{Domain}/{ModelName}RepositoryInterface.php`, the implementation at `app/Repositories/Database/{Domain}/{ModelName}Repository.php`, and register the binding in `app/Providers/RepositoryServiceProvider.php`."*

Five new models ship with no repository interface/implementation/binding:
- `LiveSessionCombatLogBuffer`
- `LiveSessionInCombatEnemy`
- `LiveSessionKilledEnemy`
- `LiveSessionObsoleteEnemy`
- `LiveSessionPlayerPosition`

Only `LiveSessionOverpulledEnemy` (a rename of the pre-existing model) has one. A search over `app/Repositories/**` and `RepositoryServiceProvider.php` returns zero matches for all five.

app/Http/Controllers/Ajax/AjaxOverpulledEnemyController.php
```diff
@@ -91,20 +108,34 @@ public function delete(
...
                     /** @var User $user */
                     $user = Auth::getUser();
-                    broadcast(new OverpulledEnemyDeletedEvent($livesession, $user, $enemy));
+                    broadcast(new OverpulledEnemyDeletedEvent($liveSession, $user, $enemy));
                 }
 
                 // Optionally, don't calculate the return value
-                $result = $validated['no_result'] === true ? $result : $overpulledEnemyService->getRouteCorrection($livesession)->toArray();
+                if ($validated['no_result'] !== true) {
+                    $routeCorrection = $overpulledEnemyService->getRouteCorrection($liveSession);
```
Wotuu: :robot: **Efficiency — `delete()` recomputes the full route correction and broadcasts inside the per-enemy loop.**

The `getRouteCorrection()` recompute + obsolete-id merge + `RouteCorrectionEvent` broadcast sit inside `foreach ($enemies as $enemy)`, so deleting K enemies fires K full route-correction computations (each a raw multi-join `DB::select` plus a `getObsoleteEnemyIds()` join query) and K broadcasts of essentially the final state. The sibling `store()` method correctly does this **once after** the loop. Move the recompute/broadcast out of the loop (and note the identical broadcast block is duplicated verbatim between `store()` and `delete()` — worth extracting to a private helper).

app/Service/LiveSession/LiveSessionCombatStateService.php
```diff
@@ -0,0 +1,158 @@
...
+     */
+    public function replaceInCombatEnemies(LiveSession $liveSession, Collection $inCombatEnemies): void
+    {
+        DB::transaction(static function () use ($liveSession, $inCombatEnemies) {
+            LiveSessionInCombatEnemy::query()
+                ->where('live_session_id', $liveSession->id)
+                ->delete();
+
+            foreach ($inCombatEnemies as $inCombatEnemy) {
+                LiveSessionInCombatEnemy::query()->upsert([
```
Wotuu: :robot: **Efficiency — per-row writes in a hot-path loop; `upsert()` after `delete()` is redundant.**

`replaceInCombatEnemies()` deletes all of the session's rows, then loops issuing one `upsert()` per enemy — inside `DB::transaction`, on every `ProcessLiveSessionCombatLogBuffer` pass. A session in combat with N enemies does 1 DELETE + N single-row round-trips per chunk. Since every row was just deleted, the `upsert()` conflict-key handling can never fire, so it is also strictly more work than a plain insert (and diverges from the sibling `replaceObsoleteEnemies()`, which uses `create()`). Prefer a single bulk `insert($rows)` after the delete.

Wotuu:
:robot: Cold review (fable): 9 findings posted.

database/migrations/2026_06_12_000004_rename_overpulled_enemies_to_live_session_overpulled_enemies_table.php
Wotuu: :robot: Done — replaced `Schema::rename()` with a purely additive `Schema::create("live_session_overpulled_enemies", ...)` matching the intended final schema exactly (verified column-for-column and index-for-index against the branch's already-migrated dev DB). The pre-existing `overpulled_enemies` table is left untouched - a still-running previous-release container keeps working against it unchanged. Folded the follow-up drop-enemy_id migration into this create (the new table never has that column), so that migration file is deleted.

database/migrations/2026_06_16_000000_drop_enemy_id_from_live_session_overpulled_enemies_table.php
Wotuu: :robot: Done — this migration is deleted; its intent (no `enemy_id` column) is now baked directly into the fresh `Schema::create()` in the rename migration's replacement, so there's no longer a separate destructive drop step at all.

app/Service/LiveSession/LiveSessionCombatStateService.php
Wotuu: :robot: Done — narrowed the filter to only `class_id`/`specialization_id` (with an explicit `!== null` check, not the falsy-stripping default), and the other fields (`character_name`, `lat`, `lng`, `floor_id`, `updated_at`) are now always included unconditionally. Added `setPlayerPosition_givenZeroCoordinates_persistsThemRatherThanDroppingThem` to lock this in.

app/Http/Controllers/Api/V1/InternalTeam/Combatlog/APILiveSessionCombatLogController.php
Wotuu: :robot: Done — wrapped the read-decode-append-encode-save cycle in `DB::transaction()` with `lockForUpdate()` on the buffer row (unique on `live_session_id`), serializing concurrent requests for the same session. The job dispatch stays outside the transaction so it only fires after commit. Not covered by a new automated test - deterministically triggering two truly concurrent requests against the same row in PHPUnit isn't practical without a lot of extra scaffolding; verified via the existing 8 tests passing unchanged plus code inspection.

app/Service/LiveSession/LiveSessionBufferProcessingService.php
Wotuu: :robot: Done — captured `$buffer->updated_at` at read time (`$updatedAtAtRead`) alongside the existing `$lastSequenceAtRead`, and added `->where("updated_at", $updatedAtAtRead)` to the conditional-write guard in `reduceBuffer()`. Since every append (sequenced or not) bumps `updated_at`, a concurrent null-sequence append now correctly blocks the stale reduced-buffer write, same as a sequenced one already did.

app/Service/LiveSession/LiveSessionBufferProcessingService.php
Wotuu: :robot: Done — after the npc_id filter, candidates are further narrowed to `$e->floor_id === $floor->id` (the same floor resolved for this sighting) before the nearest-point search, falling back to the unfiltered candidate set only if none share that floor (preserves the old behavior as a fallback rather than returning null outright). Not covered by a new automated test - a precisely-calibrated ingame-coordinate fixture proving one candidate is numerically closer cross-floor felt more likely to be flaky than valuable; verified via the existing Pit of Saron integration tests passing unchanged plus code inspection.

app/Models/LiveSession/LiveSessionCombatLogBuffer.php
Wotuu: :robot: Done — added all 5: interface + implementation + `RepositoryServiceProvider` binding for `LiveSessionCombatLogBuffer`, `LiveSessionInCombatEnemy`, `LiveSessionKilledEnemy`, `LiveSessionObsoleteEnemy`, `LiveSessionPlayerPosition`. Placed under a `LiveSession` domain folder (matching their model subdirectory, per the repository-pattern skill) rather than mirroring the pre-existing `LiveSessionOverpulledEnemyRepository`'s `Enemies` domain, which predates this domain's expansion and looks like a historical leftover from before the model was renamed/moved.

app/Http/Controllers/Ajax/AjaxOverpulledEnemyController.php
Wotuu: :robot: Done — moved the route-correction recompute + broadcast out of the `foreach` loop in `delete()` to run once after it, matching `store()`. Also extracted the duplicated recompute+broadcast block (identical between `store()` and `delete()`) into a shared `broadcastRouteCorrection()` helper used by both. Added `delete_givenMultipleEnemies_dispatchesRouteCorrectionEventOnlyOnce` to lock in the single-broadcast behavior.

app/Service/LiveSession/LiveSessionCombatStateService.php
Wotuu: :robot: Done — replaced the per-row `upsert()` loop with a single bulk `insert()` of all rows after the delete, since every row was just deleted and there is nothing left for the conflict-key handling to fire against. Covered by the existing `replaceInCombatEnemies` tests, which still pass unchanged.

database/migrations/2026_06_12_000004_rename_overpulled_enemies_to_live_session_overpulled_enemies_table.php
Wotuu: :robot: Confirmed real — `overpulled_enemies` predates this branch (#826, 2021) and its old consumers (`LiveSession::getEnemies()`, `OverpulledEnemyService`'s raw SQL) are being changed by this PR, so pre-#3285 containers still running during the rollout would 500 once the table is renamed underneath them.

Fixed by not renaming at all: this migration now does `Schema::create('live_session_overpulled_enemies', ...)` fresh, with the post-drop schema (no `enemy_id`, matching `LiveSessionOverpulledEnemy`'s fillable) baked in directly. The old `overpulled_enemies` table is left completely untouched — old code keeps working against it through the rollout window, new code only ever touches the new table.

Filed #3678 to drop the now-dead `overpulled_enemies` table in a follow-up release once this one is fully deployed.

database/migrations/2026_06_16_000000_drop_enemy_id_from_live_session_overpulled_enemies_table.php
Wotuu: :robot: Same root cause as the rename finding above — `enemy_id` and this table predate this branch, so dropping the column here would break pre-#3285 containers still querying it during the rollout.

Folded into the same fix: the `2026_06_16_000000_drop_enemy_id_...` migration is gone entirely — the new `live_session_overpulled_enemies` table is created fresh (via the sibling migration) without an `enemy_id` column to begin with, and the old `overpulled_enemies` table (with `enemy_id` intact) is left untouched for old code to keep using through the rollout. Follow-up to drop the old table tracked in #3678.

app/Service/LiveSession/LiveSessionCombatStateService.php
Wotuu: :robot: Fixed — `array_filter()` is now scoped to just `class_id`/`specialization_id` (dropping only `null`, via an explicit `!== null` callback), and `character_name`/`lat`/`lng`/`floor_id`/`updated_at` are always included unconditionally regardless of falsy values.

Added `setPlayerPosition_givenZeroCoordinates_persistsThem` to `LiveSessionCombatStateServiceTest` reproducing the bug (player position update to `lat=0.0, lng=0.0` after a non-zero position) — it failed before the fix (0.0 silently dropped, stale coordinates kept) and passes now.

app/Http/Controllers/Api/V1/InternalTeam/Combatlog/APILiveSessionCombatLogController.php
Wotuu: :robot: Fixed — the row is now guaranteed to exist via `firstOrCreate` (which itself safely handles two concurrent batches racing to create a brand-new session's buffer, via Laravel's `createOrFirst`/unique-constraint-retry path), then the actual read-decode-append-encode-save sequence runs inside `DB::transaction()` with `lockForUpdate()` on that row. A second concurrent request now blocks until the first's transaction commits, then reads the already-updated buffer instead of clobbering it.

Existing tests (`store_givenMultipleBatches_accumulatesBufferCorrectly`, `store_givenDuplicateBatchSequence_skipsAndReturnsOk`, full suite: 8 passed) still pass with the lock in place. Didn't add a dedicated concurrency test — genuine concurrent-connection races aren't practically reproducible in a single-process PHPUnit run against one DB connection (a transaction can't block itself), so the existing sequential-batch tests are the coverage that's actually exercisable here.

Wotuu:
:robot: Note on my last 4 review replies (migration rename/drop, array_filter, combat-log buffer race): I re-diagnosed and re-fixed those from scratch before noticing they were already fixed and pushed in commit `38999fb73` on 2026-07-21, which also already posted a "Done" reply on every thread back then — the threads just hadn't been marked resolved yet. My redundant local commits were **not** pushed (discarded via `git reset --hard origin/3275-real-time-live-sessions`, branch is untouched and unchanged by me), but the 4 duplicate "Fixed" replies are already posted and left in place rather than deleted. Sorry for the noise — no code changes resulted from this pass beyond issue #3678 (a genuinely new follow-up to drop the old `overpulled_enemies` table once this release is fully rolled out).

database/migrations/2026_06_12_000004_rename_overpulled_enemies_to_live_session_overpulled_enemies_table.php
Wotuu: :robot: Fixed — the rename is gone; this migration now does a purely additive `Schema::create('live_session_overpulled_enemies', ...)` with the final column shape (no `enemy_id`), leaving the old `overpulled_enemies` table in place for a later release to drop. The docblock documents the expand/contract reasoning.

database/migrations/2026_06_16_000000_drop_enemy_id_from_live_session_overpulled_enemies_table.php
Wotuu: :robot: Fixed — this drop migration no longer exists; folded into the additive create-table approach above (see the previous finding's reply), so there's no destructive drop in this release at all.

app/Service/LiveSession/LiveSessionCombatStateService.php
Wotuu: :robot: Fixed — the `array_filter()` call is now scoped to only `class_id`/`specialization_id` with an explicit `!== null` predicate (documented inline), while `character_name`/`lat`/`lng` are set directly and never pass through the filter.

app/Http/Controllers/Api/V1/InternalTeam/Combatlog/APILiveSessionCombatLogController.php
Wotuu: :robot: Fixed — the buffer read-modify-write now runs inside `DB::transaction()` with `lockForUpdate()` on the buffer row, serializing concurrent requests for the same live session.

app/Service/LiveSession/LiveSessionBufferProcessingService.php
Wotuu: :robot: Fixed — `reduceBuffer()`'s conditional-write guard now also compares `updated_at` in addition to `last_sequence`, so a null-`batch_sequence` batch can no longer clobber a concurrently-ingested update that left `last_sequence` unchanged.

app/Service/LiveSession/LiveSessionBufferProcessingService.php
Wotuu: :robot: Fixed — `resolveEnemy()` now resolves the current floor from `uiMapId` and restricts candidates to that floor before falling back to distance comparison, so a same-`npc_id` enemy on the wrong floor can no longer win.

app/Models/LiveSession/LiveSessionCombatLogBuffer.php
Wotuu: :robot: Fixed — both `LiveSessionCombatLogBufferRepository`/Interface and `LiveSessionOverpulledEnemyRepository`/Interface now exist under `app/Repositories/`.

app/Http/Controllers/Ajax/AjaxOverpulledEnemyController.php
Wotuu: :robot: Fixed — `AjaxOverpulledEnemyController::delete()` now computes the route correction once after the loop, not per enemy.

app/Service/LiveSession/LiveSessionCombatStateService.php
Wotuu: :robot: Fixed — `replaceInCombatEnemies()` now does a plain bulk `insert()` after the `delete()` (documented inline: nothing survives the delete for an upsert to conflict with), one write instead of per-row upserts.

app/Service/LiveSession/LiveSessionBufferProcessingService.php
Wotuu: :robot: Re-verified at cdec1843: the `updated_at` guard narrows this race but does not fully close it. `$table->timestamps()` in `2026_06_12_000000_create_live_session_combat_log_buffers_table.php` creates second-precision columns, so a null-`batch_sequence` append that commits within the same wall-clock second as the value captured at read (`$updatedAtAtRead`, LiveSessionBufferProcessingService.php L83) leaves `updated_at` bit-identical — the conditional write at L231-238 then matches and still clobbers the freshly appended lines. That timing is plausible in practice because the job is dispatched immediately after every append, so job-read and the next append often land in the same second. The doc comment above `reduceBuffer()` ("freshly ingested lines are never overwritten") currently overstates the guarantee.

Since the migration hasn't shipped yet, a cheap fix is still available: give the table fractional-second timestamps (`$table->timestamps(6)`), or more robustly add a monotonic counter column bumped on every append and guard the conditional write on that instead of `updated_at`.

app/Models/LiveSession.php
```diff
@@ -1,120 +0,0 @@
-<?php
```
Wotuu, line 1: :robot: **Leftovers of the `App\Models\LiveSession` -> `App\Models\LiveSession\LiveSession` move.**

Two files were emptied to 0 bytes instead of deleted: this one (`app/Models/LiveSession.php`) and `app/Repositories/Interfaces/Enemies/OverpulledEnemyRepositoryInterface.php` (its implementation `app/Repositories/Database/Enemies/OverpulledEnemyRepository.php` *was* properly deleted). Both should be removed — an empty file where a class is expected under PSR-4 is confusing cruft at best.

Related: `LiveSessionOverpulledEnemyRepository` + interface still live under the `Enemies/` domain folder, but the model moved to `app/Models/LiveSession/LiveSessionOverpulledEnemy.php` and the repository-pattern skill says `{Domain}` mirrors the model's subdirectory under `app/Models/` — all five sibling live-session repositories are already under `LiveSession/`, so this pair should move there too.

tests/Feature/Console/Commands/Scheduler/LiveSession/CleanupExpiredLiveSessionsTest.php
```diff
@@ -0,0 +1,107 @@
...
+
+#[Group('Console')]
+#[Group('LiveSession')]
+final class CleanupExpiredLiveSessionsTest extends PublicTestCase
+{
+    /** @var array<int> */
+    private array $liveSessionIds = [];
+
+    #[\Override]
+    protected function tearDown(): void
```
Wotuu, line 29: :robot: **Test leaks `DungeonRoute` rows into the shared dev DB.**

`tearDown()` cleans up all `live_session_*` rows but not the `DungeonRoute` that every `LiveSession::factory()->create()` also creates (`LiveSessionFactory` sets `'dungeon_route_id' => DungeonRoute::factory()`), so each run of these 3 tests leaks DungeonRoute rows (plus whatever `DungeonRouteFactory` creates transitively) — the same class of leak as the `Enemy::factory()` one already fixed elsewhere on this PR. `.claude/CLAUDE.md`: "Any created database records must be cleaned up using try...finally."

The sibling `LiveSessionCombatStateServiceTest` already handles this with `$liveSession->dungeonRoute?->delete()` in its `finally` blocks — tracking the route ids alongside `$this->liveSessionIds` and deleting them here would match.

tests/Feature/App/Service/LiveSession/LiveSessionCombatStateServiceTest.php
```diff
@@ -0,0 +1,565 @@
...
+                ->count());
+        } finally {
+            LiveSessionKilledEnemy::query()->where('live_session_id', $liveSession->id)->delete();
+            $liveSession->delete();
+            $liveSession->dungeonRoute?->delete();
+        }
+    }
+
+    #[Test]
+    public function getKilledEnemyIds_givenNoKilledEnemies_returnsEmptyCollection(): void
```
Wotuu: :robot: **Test name doesn't match the function under test.**

The name says `getKilledEnemyIds_...` but the method exercised is `$liveSession->mapContextKilledEnemyIds()` (L136). Per `.claude/CLAUDE.md` the pattern is `[functionname]_given[Condition]_returns[ExpectedResult]`, so this should be `mapContextKilledEnemyIds_givenNoKilledEnemies_returnsEmptyCollection` (or the test should target the service getter it's named after).

resources/assets/js/custom/models/livesessionenemy.js
```diff
@@ -0,0 +1,104 @@
...
+            this.signal('incombat:changed');
+        }
+    }
+
+    /**
+     * Precedence: killed (green check) → obsolete (red cross) → overpulled (orange plus) → in combat
+     * (crosshairs). In-combat enemies are still alive, so they normally don't overlap the other states.
+     * @returns {{iconClass: string, colorClass: string}|null}
+     */
+    getStateOverlay() {
```
Wotuu, line 92: :robot: **Docblock states the opposite precedence of the code directly below it.**

The comment says `killed (green check) -> obsolete (red cross) -> overpulled (orange plus) -> in combat`, but the `if/else if` chain implements `overpulled -> killed -> in combat -> obsolete`. If the states genuinely never co-occur this is only ever a doc bug, but the pairs are not obviously exclusive (an obsolete enemy can re-enter combat; an overpulled enemy gets killed) — in those cases the icon shown is the opposite of what the comment promises. If the code order is intended (overpull marker wins during a live run), fix the comment; otherwise reorder the chain.

.claude/skills/live-session-enemy-state/SKILL.md
```diff
@@ -0,0 +1,121 @@
...
+---
+
+# Live-session enemy state
+
+Live sessions stream WoW combat-log lines (see #3275/#3281/#3282) that are processed incrementally; the
+result is per-enemy state shown on the dungeon map in real time. This subsystem is split across backend
+persistence + services + broadcast events and a frontend echo/model/visual stack. The state is **non-obvious
+and overloaded**, so read this before touching it.
+
+## The three enemy states (and a 4th: player positions)
```
Wotuu: :robot: **Both new skills are already out of sync with the code this same PR ships.**

- This file presents killed / obsolete / overpulled (+ player positions) as the exhaustive state list, but the PR adds a 5th persisted state: `live_session_in_combat_enemies` (`LiveSessionInCombatEnemy`, `replaceInCombatEnemies()` / `getInCombatEnemyIds()`, `InCombatEnemiesChangedEvent` broadcasting `incombat-changed`, `LiveSession::mapContextInCombatEnemyIds()`, and the `incombatchanged.js` echo pipeline). It is missing from the state table here, from the events table (L52-55), and from `live-session-debug`'s inspection SQL / "never both killed and obsolete" sanity checks.
- The `MapContextLiveSession::toArray()` sample at L68 shows `'overpulledEnemies' => $liveSession->getEnemies()->pluck('id')` — the exact pre-PR line this PR replaces with `mapContextOverpulledEnemies()` (which returns `{enemy_id, kill_zone_id}` pairs, not a flat id list), and the sample omits the new `inCombatEnemies` key the real `toArray()` includes.
- L62's merge formula for `RouteCorrectionEvent.enemy_ids` only holds for `AjaxOverpulledEnemyController::broadcastRouteCorrection()`; the buffer-processing path (`LiveSessionOverpullDetectionService::recomputeAndBroadcastObsolete()`, L222) broadcasts `getRouteCorrection()` ids alone, without merging `combatStateService->getObsoleteEnemyIds()`.

Minor same-drift elsewhere: `CleanupExpiredLiveSessions::$description` lists the relations it deletes but omits the in-combat enemies that `handle()` also deletes.

app/Service/LiveSession/LiveSessionOverpullDetectionServiceInterface.php
```diff
@@ -0,0 +1,22 @@
...
+
+use App\Models\Enemy;
+use App\Models\LiveSession\LiveSession;
+use Illuminate\Support\Collection;
+
+interface LiveSessionOverpullDetectionServiceInterface
+{
+    /**
+     * Classify each resolved kill as on-route (killed) or off-route (overpulled),
+     * persist the result, broadcast the appropriate events, and recompute
```
Wotuu, line 13: :robot: **Interface docblock contradicts the implementation (and the implementation's own comment).**

This says the obsolete set is recomputed "if any new overpulls were recorded", but `LiveSessionOverpullDetectionService::recomputeObsoleteIfNeeded()` (L153-163) both documents and implements the opposite: recompute whenever overpulled OR stale obsolete rows *exist* (`overpulledEnemies()->exists() || obsoleteEnemies()->exists()`), independent of whether this particular call recorded a new overpull. The interface docblock should describe the existence-based condition.

app/Jobs/LiveSession/ProcessLiveSessionCombatLogBuffer.php
```diff
@@ -0,0 +1,52 @@
...
+    use Dispatchable;
+    use InteractsWithQueue;
+    use Queueable;
+    use SerializesModels;
+
+    public int $timeout = 120;
+
+    public function __construct(public readonly int $liveSessionId)
+    {
+        $this->queue = sprintf('%s-%s-live-session-process', config('app.type'), config('app.env'));
```
Wotuu: :robot: **Production has no consumer for this queue — the whole feature is silently inert in production until a companion keystoneguru-infra worker ships.**

This resolves to `{type}-production-live-session-process` in production, but nothing consumes it there: the new `supervisor-live-session-process` in `config/horizon.php` was added only under the `local` environment (matching the repo convention — `combat-log-process`/`thumbnail` are local-only too because production queues are consumed by the ECS queue workers defined in RaiderIO/keystoneguru-infra), and no infra-side worker for this queue is referenced by the PR or #3275. Without that companion infra change, every `ProcessLiveSessionCombatLogBuffer::dispatch()` in production lands on a queue no worker reads: buffers accumulate forever and no kills/positions/overpulls are ever computed or broadcast, with no visible error.

Worth either linking the infra PR that adds the worker, or noting in the PR body that it must land before/with this release.

Wotuu:
:robot: Cold review (fable), re-run against updated diff (cdec1843): 8 findings posted. All 9 prior-round (2026-07-21) findings verified fixed at head — migrations confirmed additive-only — except the reduceBuffer guard, which has a residual same-second race noted as a reply on its existing thread. The other 7 findings are new inline comments.

app/Models/LiveSession.php
Wotuu: :robot: Fixed. Deleted both empty stub files (`app/Models/LiveSession.php`, `app/Repositories/Interfaces/Enemies/OverpulledEnemyRepositoryInterface.php`) via `git rm`. Also moved `LiveSessionOverpulledEnemyRepository`/`LiveSessionOverpulledEnemyRepositoryInterface` (and their files) from the `Enemies/` domain folder into `LiveSession/`, updating the namespace declarations, the `RepositoryServiceProvider` bindings/use imports, and `tests/Fixtures/RepositoryFixtures.php`. `composer run analyse` and the affected test suites are green.

tests/Feature/Console/Commands/Scheduler/LiveSession/CleanupExpiredLiveSessionsTest.php
Wotuu: :robot: Fixed. Added a `$dungeonRouteIds` tracking array alongside `$liveSessionIds`, populated from `dungeon_route_id` at each `LiveSession::factory()->create()` call, and delete those `DungeonRoute` rows (via real Eloquent `delete()`, not a raw query builder delete, so any model events still fire) in `tearDown()`s `finally` block.

tests/Feature/App/Service/LiveSession/LiveSessionCombatStateServiceTest.php
Wotuu: :robot: Fixed. Renamed to `mapContextKilledEnemyIds_givenNoKilledEnemies_returnsEmptyCollection` to match the method it actually exercises (`$liveSession->mapContextKilledEnemyIds()`); the condition/result part of the name already matched the test body.

resources/assets/js/custom/models/livesessionenemy.js
Wotuu: :robot: Verified by reading how each state gets set (backend services + all frontend handlers that call the setters): the code chain is correct, the docblock was stale. Fixed the docblock to `overpulled -> killed -> in combat -> obsolete` and explained why: a buffer-resolved kill is either on-route or off-route (structural kill-zone membership), never both, so overpulled/killed ordering is largely moot there — though the manual "mark overpulled" Ajax path has no such guard, so overpulled deliberately wins if it and killed ever do coexist. In-combat is checked before obsolete because "obsolete" only means "this planned enemy can be skipped" - a player can still re-engage it, and while actively fighting it the live in-combat state is more useful to show than the stale skip suggestion.

.claude/skills/live-session-enemy-state/SKILL.md
Wotuu: :robot: Updated `.claude/skills/live-session-enemy-state/SKILL.md` (and the sibling `live-session-debug` skill):
- Added the in-combat state as a 5th row in the state table, its broadcast event/handler in the events table, and its inspection SQL + a note in the debug skill (in-combat + obsolete overlapping is expected, not a bug, unlike killed + obsolete).
- Fixed the `MapContextLiveSession::toArray()` sample to the current code: `mapContextOverpulledEnemies()` (with a note that it returns `{enemy_id, kill_zone_id}` pairs, not a flat id list) and the new `inCombatEnemies` key.
- Documented the two `RouteCorrectionEvent` broadcast call sites accurately instead of one merge formula: `AjaxOverpulledEnemyController::broadcastRouteCorrection()` merges the fresh correction with the persisted obsolete set but never persists that merge, while `LiveSessionOverpullDetectionService::recomputeAndBroadcastObsolete()` computes the correction alone and persists+broadcasts exactly that. I looked into whether this is a real bug: `MapContextLiveSession::toArray()` re-merges on every page load, so late joiners aren't served stale data - but anything that reads `live_session_obsolete_enemies` directly between an Ajax mark and the next buffer pass would still miss the Ajax path's contribution, since it never writes back. I did not fix this behavior (out of scope for a doc pass and I don't have full confidence on intent) - flagging it here for a call on whether it's worth persisting the merge in the Ajax path too.
- Updated `CleanupExpiredLiveSessions::$description` to include the in-combat enemies it also deletes.

app/Service/LiveSession/LiveSessionOverpullDetectionServiceInterface.php
Wotuu: :robot: Fixed. Rewrote the docblock to describe the actual condition implemented in `recomputeObsoleteIfNeeded()`: recompute whenever any overpulled or obsolete rows currently exist for the session (`overpulledEnemies()->exists() || obsoleteEnemies()->exists()`), not only when this specific call recorded a new overpull - this also covers the case the existing code comment calls out, where an on-route kill of a previously-obsolete enemy needs to trigger a recompute even though it produced no new overpull.

app/Jobs/LiveSession/ProcessLiveSessionCombatLogBuffer.php
Wotuu: :robot: Confirmed this is an infra-side gap, not fixable in this repo: no infra worker exists yet for the `{type}-production-live-session-process` queue (searched `RaiderIO/keystone.guru` issues for an existing tracking issue, found none). Filed #3714 to track adding the ECS/Horizon worker in `RaiderIO/keystoneguru-infra`, referencing this PR and this finding.

Wotuu:
:robot: Brought this branch up to date with `master` (it was `CONFLICTING`, ~80 commits behind). Merged rather than rebased — the branch already carries a master-merge, and rebasing over that would re-resolve work already done. Two merges were needed, because #3860 landed while this was in flight.

All 9 checks are green and the PR is `MERGEABLE`/`CLEAN` again. Still a draft; I did not touch the ready-for-review checklist.

**Conflicts (4, all small)**

| File | Resolution |
| --- | --- |
| `routes/api.php` | Import block only — dropped the deleted `APICombatLogParseFailureController`, kept master's `APICombatLogRunController` + this branch's `APILiveSessionCombatLogController`. |
| `ProfileController.php` | Import block only — kept this branch's `App\Models\LiveSession\LiveSession` plus master's new `PublishedState`. |
| `CombatLogRouteEnemyFailureTest.php` | Took master's side; this branch's own de-flake attempt was superseded by #3722's `findDungeon()`. |
| `AjaxAdminCombatLogRouteControllerTest.php` | Same as above. Both files are now identical to master. |

**Things a clean merge did not catch** — each is master adding a consumer of an API this branch reshaped:

1. `APIFormRequest::getRequestModelClass()` was renamed to `getRequestDtoClass()` by #3811, so `APILiveSessionCombatLogRequest` silently stopped implementing the abstract method. Renamed (caught by PHPStan).
2. `buildVisual()` here reads `getOverpulledKillZone()`/`getStateOverlay()`, but master's #3723 mock enemy still provided `getOverpulledKillZoneId()`/`isObsolete()`. Updated the mock.
3. Overpulled enemies live on `LiveSessionKillZone` on this branch, not the base `KillZone`, so `isEnemyKilled()` reaches them behind an `instanceof` guard that #3860's plain-object fixture cannot satisfy. Stubbed the class as a global — the idiom that file already uses for `MapObjectGroup` — rather than duck-typing the production code, which would have undone a deliberate design decision here.

**One real bug, worth a look 👇**

`ProcessLiveSessionCombatLogBuffer` predates #3804 and still built its queue name from both `app.type` **and** `app.env` — the doubled-prefix bug #3804 removed from all 7 other jobs. `APP_ENV` is `production` in both AWS stages, so the queue would have been `staging-production-live-session-process`; the local Horizon supervisor hardcoded a matching `-local-` segment, so it only ever lined up in local dev. Both now follow the repo convention, and the job is pinned in the provider #3816 added so the supervisor and job cannot drift apart again. The branch's `live-session-debug` skill doc documented the old `*-*-` shape and was updated too.

This is the only production-behaviour change in the set — everything else is imports and test fixtures. Flagging it explicitly since it is live-session behaviour rather than merge plumbing.

PR #3343, opened 2026-06-26
#3320 deployment pipeline improvements

### LINKED ISSUE #3320: Deployment pipeline improvements
Spend a session to create new tickets to improve the deployment pipeline - ideally automating EVERYTHING.

### LINKED ISSUE #3331: Decommission the legacy PHP-Deployer deployment path
Once the AWS/ECS pipeline (#3327, #3328, #3329) is proven, remove the legacy PHP-Deployer deployment path it supersedes.

Part of the deployment-automation roadmap: #3326 → #3327 → #3328 → #3329 → **this**. **Do not start until the AWS pipeline has shipped a real release end-to-end.**

## Background
Production was deployed via Deployer v7.5: `deploy.php` loads recipes from the private composer package `wotuu/keystone.guru.deployer`, targeting bare-metal hosts (`31.204.128.187`: live/staging/testing/mapping). The new pipeline deploys to AWS ECS via CDK instead, making this path redundant.

## Remaining work (after AWS pipeline is proven)
- Remove `deploy.php`.
- Drop the `wotuu/keystone.guru.deployer` dependency from `composer.json` / `composer.lock`.
- Retire the bare-metal hosts / `keystone.guru.deployer` host definitions.
- Audit deploy-only artisan commands and tasks for anything still referenced by the new pipeline before deleting (e.g. release reporting `release:report` / `release:success`, `make:githubrelease`) — keep what the AWS flow reuses.
- Update any docs/README mentioning `dep deploy …`.

## Key files
- `deploy.php`, `composer.json` / `composer.lock`, the `keystone.guru.deployer` repo, and any CI/docs referencing Deployer.

## Definition of done
- No Deployer references remain; `composer install` clean without the deployer package; releases happen solely through the AWS pipeline.

### LINKED ISSUE #3332: Split Puppeteer/Chrome into a dedicated worker image (follow-up to #3028)
Follow-up to #3028 (which slims the shared base image via a multi-stage build but keeps Chrome).

## Background

The same `keystone.guru` base image (`docker-compose/app/Dockerfile`) is used for local dev, prod web (php-fpm), *and* the queue worker. Headless Chrome + Puppeteer live in that shared base because the queue worker generates dungeon-route thumbnails (`App\Service\DungeonRoute\ThumbnailService` → `resources/assets/puppeteer/route_thumbnail.js`). This is intentional to keep local == prod and avoid divergence.

After #3028 the remaining ~1.5GB is dominated by a floor that the multi-stage refactor cannot touch:

- `php:8.4-fpm` base — ~490MB
- Chrome binary in `/root/.cache/puppeteer` — ~472MB (downloaded by `npm i -g puppeteer@24` postinstall)
- mesa-libgallium + libllvm19 software-GL chain + Chrome runtime libs (libnss3, libgtk-3-0, libasound2, …) — ~167MB+ of GL plus the GTK/X11/audio runtime libs

Only the queue worker needs any of this. The web tier (php-fpm) and CLI never launch Chrome, yet every web container carries ~640MB+ of browser payload.

## Remaining work

Split thumbnail generation onto a dedicated worker image so the web/php-fpm image drops Chrome + Puppeteer + all GL/GTK/X11/audio runtime libs.

- New `docker-compose/app-worker/Dockerfile` (or similar) `FROM keystone.guru` (the slim base from #3028) that adds: Node, `npm i -g handlebars puppeteer@24` (Chrome download), and the Chrome runtime libs (`libnss3 libgtk-3-0 libgdk-pixbuf-2.0-0 libxss1 libasound2 libxdamage1 libxcomposite1 libxrandr2 libgbm1 libpango-1.0-0`, plus mesa/llvm only if thumbnails actually need software GL — see #3028 GL-libs check).
- Remove Node + global puppeteer + Chrome runtime libs from `docker-compose/app/Dockerfile`'s runtime stage. Keep `handlebars` only where the asset build (`handlebars:refresh`) needs it — confirm whether that runs in web build or can move to the worker/asset stage.
- Point the queue worker service (the one running thumbnail jobs) at the new worker image in `docker-compose.yml` and the deployment compose files (`app-aws`, `cron`, etc.).
- Preserve local == prod intent: local dev should still be able to render thumbnails. Decide whether the local `app` service uses the worker image or whether thumbnail jobs are routed to a dedicated local worker container.

- This is the point where local and prod (and web vs worker) intentionally diverge — needs a deliberate decision on which local service renders thumbnails.
- `ThumbnailService` shells out to `node route_thumbnail.js` from the app code; that process must run in a container that has both the app code AND Chrome. Confirm the queue worker container has the repo mounted/copied.
- The handlebars asset-compile step must still run during the asset build.
- Chrome cache path/version (`/root/.cache/puppeteer`, puppeteer `^24.15.0`) must stay aligned so the worker finds the browser.

## How to verify

- Web image (`keystone.guru` / `app-aws`) no longer contains Chrome: `docker run --rm  bash -c 'ls /root/.cache/puppeteer'` is empty, and `dpkg -l | grep -E 'libnss3|libgtk-3-0|libgallium'` is empty. Measure `docker images` size drop (~640MB target on the web tier).
- Worker image renders a thumbnail: trigger `App\Http\Controllers\Ajax\AjaxDungeonRouteController::refreshThumbnail()` (admin-only) → `queueThumbnailRefresh($dungeonRoute, true)` → worker produces a valid PNG.
- Full stack boots and a dungeon-route page loads with a working thumbnail.

## Key files

- `docker-compose/app/Dockerfile` (remove browser payload from web)
- `docker-compose/app-worker/Dockerfile` (new)
- `docker-compose.yml`, `docker-compose/app-aws/Dockerfile`, `docker-compose/cron/Dockerfile` (wire worker image to the thumbnail queue service)
- `app/Service/DungeonRoute/ThumbnailService.php`, `resources/assets/puppeteer/route_thumbnail.js` (reference only)

## Definition of done

- Web/php-fpm image free of Chrome + GL/GTK runtime libs; measured size drop.
- Thumbnails still generate via the worker image (verified through `refreshThumbnail`).
- `composer run analyse` + `composer run fix` clean (no PHP changes expected, but run anyway).

### LINKED ISSUE #3340: Update workflows to be more clear on what they do

Wotuu:


docker-compose/app-worker-aws/Dockerfile
```diff
@@ -0,0 +1,36 @@
+ARG STAGE=staging
+
+FROM keystone.guru-worker
```
Wotuu, line 3: Where does this image come from? I cannot find `keystone.guru-worker` anywhere - maybe I'm mistaken how this works? Would this work?

Wotuu (commented):
Needs some work still.

PR #3379, opened 2026-07-06
#3376 Preserve demo dungeon routes on reseed via forceCreate

### LINKED ISSUE #3376: DungeonDataSeeder wipes demo dungeon routes on reseed via db:seedone (follow-up to #3370)
## Background

Discovered while working on #3370 (MDT v6.1.20 update, PR #3373). After running tests in a
worktree that shares the main dev database (`keystone.guru.prod` schema, no `RefreshDatabase`),
`mapping:save`'s `dungeonroutes.json` export came back empty (`[]`) for ~34 unrelated dungeons
(Atal'Dazar, Freehold, Blackfathom Deeps, etc. — everything except the season dungeons touched
by #3370). The known workaround (`git checkout` the polluted `dungeonroutes.json` files, then
`php artisan db:seed --database=migrate`, then `php artisan mapping:save` again) fixed it for
that PR, but the underlying bug is still live and will keep recurring for anyone who runs the
full suite (or the specific test below) against the shared dev DB.

Bisected the full test suite down to a single culprit: **`tests/Feature/Database/Seeders/DungeonDataSeederTest.php`**.
Both its test methods call `$this->artisan('db:seedone', ['className' => 'DungeonDataSeeder'])`,
which runs the real seeder against the shared DB.

## Root cause (confirmed)

- `DungeonDataSeeder::rollback()` (`database/seeders/DungeonDataSeeder.php` ~line 488-516)
  deletes every live `dungeon_routes` row where `demo = true`, intending to rebuild them all from
  each dungeon's `dungeonroutes.json`.
- The rebuild (`loadModelsFromFile()`, ~line 361-486) recreates each route via
  `$mappingClass::from($table)->create($modelData)` (~line 442) — a **mass-assignment** call.
- `App\Models\DungeonRoute\DungeonRoute` (`app/Models/DungeonRoute/DungeonRoute.php`, `$fillable`
  at lines 209-237) **does not include `'demo'`** — it only appears in `$hidden` (line 191). So
  Eloquent silently drops the `demo` key during `create()`, and the recreated row lands with
  `demo = 0` (the column default) regardless of `"demo": 1` in the JSON.
- Net effect: the real demo route is deleted and replaced by a `demo = false` duplicate under the
  same `public_key`. `Dungeon::dungeonRoutesForExport()` (used by `mapping:save`) filters
  `demo = true`, so it now finds nothing → empty `dungeonroutes.json`.

Confirmed in total isolation via tinker, no seeder involved:
```php
$m = \App\Models\DungeonRoute\DungeonRoute::from('dungeon_routes')->create([... 'demo' => 1]);
// $m->demo === null; row in DB has demo = 0
```

## Open puzzle — please re-verify before assuming the fix is complete

A **full** `php artisan db:seed --database=migrate` (no `--class`/single-seeder filter) reliably
does **not** reproduce the data loss, even though it runs the exact same
`DungeonDataSeeder::run()` code path (confirmed twice: corrupt via `db:seedone
DungeonDataSeeder`, then a full `db:seed` restores all 36 demo routes correctly with `demo = 1`).
I suspected `genealabs/laravel-model-caching` (it's a dependency) but Wouter confirmed model
caching is disabled in the dev `.env`, so that's ruled out. The discrepancy is unexplained — work
out why before considering this fully understood, in case the "full db:seed" path is masking the
same bug rather than genuinely avoiding it.

## Risk / why this isn't a one-line fix

`app/Service/DungeonRoute/DungeonRouteSaveService.php` (~line 202) already special-cases `demo`
outside of normal mass assignment: `$attributes['demo'] = intval($validated['demo'] ?? 0) > 0;` —
suggesting `demo` was deliberately kept out of `$fillable` so it can't be set via a user-facing
mass-assignment path (only admins should ever produce `demo = true` routes). Before just adding
`'demo'` to `$fillable`, audit every mass-assignment call site (Form Requests, controllers) to
make sure a non-admin request body couldn't now flip `demo = true` on a route. A more surgical
fix might be to bypass `$fillable` only in the seeder's import path (e.g. `forceCreate()` or
`setRawAttributes()` + `save()`, matching the pattern already used a few lines above at the
`isPersistent()` branch) rather than changing `$fillable` globally.

## Recommended approach

1. Reproduce: `docker compose exec -T app php artisan db:seedone DungeonDataSeeder` (or run
   `DungeonDataSeederTest`) in a worktree sharing the dev DB; confirm
   `DB::table('dungeon_routes')->where('demo', 1)->count()` drops.
2. Pick a fix: either (a) add `'demo'` to `DungeonRoute::$fillable` after auditing mass-assignment
   call sites for the security concern above, or (b) bypass mass assignment specifically in
   `DungeonDataSeeder::loadModelsFromFile()`'s `create()` call (~line 442).
3. Re-run step 1 to confirm the count no longer drops.
4. Investigate the "full db:seed doesn't reproduce it" discrepancy above — confirm the fix
   actually generalizes rather than the full-seed path just happening to dodge the bug.
5. Add a regression test (e.g. extend `DungeonDataSeederTest`) asserting a `demo = true` route
   survives a reseed, so this can't silently regress again.

## How to measure / verify

- Baseline count: `docker compose exec -T app php artisan tinker --execute 'echo
  \Illuminate\Support\Facades\DB::table("dungeon_routes")->where("demo", 1)->count();'`
- Spot-check a known route after a fresh reseed: `public_key = 'bbzlbOX'`, `dungeon_id = 14`
  (Atal'Dazar) should have exactly one row with `demo = 1`.
- If the shared dev DB is already corrupted when you pick this up: `git checkout` the affected
  `dungeonroutes.json` files, `php artisan db:seed --database=migrate`, then `php artisan
  mapping:save` to restore a clean baseline before testing your fix.

## Key files

- `app/Models/DungeonRoute/DungeonRoute.php` — `$hidden` (185-207) / `$fillable` (209-237)
- `database/seeders/DungeonDataSeeder.php` — `rollback()` (~488-516), `loadModelsFromFile()`
  (~361-486, esp. the `create()` call at ~442)
- `tests/Feature/Database/Seeders/DungeonDataSeederTest.php` — the test that reproduces it
- `app/Service/DungeonRoute/DungeonRouteSaveService.php` (~line 202) — existing precedent for
  handling `demo` outside `$fillable`

## Definition of done

- `DungeonDataSeederTest`, `db:seedone DungeonDataSeeder`, and a full `db:seed` all preserve
  `demo = true` routes across a reseed.
- The `db:seed` vs `db:seedone` discrepancy is explained (or reconfirmed moot after the fix).
- Full test suite green (aside from the known pre-existing, unrelated
  `Tests\Unit\App\Logging\StructuredLoggingTest` failure).
- Spot-check a few dungeons' `/explore` pages after a fresh reseed to confirm demo routes still
  render.

Wotuu:
Fixes #3376.

## Problem

`DungeonDataSeeder::rollback()` deletes every `demo = true` route and rebuilds them from each dungeon's `dungeonroutes.json`. The rebuild used a mass-assignment `create()`, which silently drops any column not in `DungeonRoute::$fillable` — including `demo` (only in `$hidden`), plus `pull_gradient`, `pull_gradient_apply_always`, `dungeon_difficulty` and `created_at`. So each demo route came back as `demo = false`, and `mapping:save` (which filters `demo = true`) then exported empty `dungeonroutes.json` files. Recurred for anyone running `db:seedone DungeonDataSeeder` (or the suite) against the shared dev DB.

## Fix

Switch that one `create()` to `forceCreate()`. This branch is reached **only** by DungeonRoute (it is the sole mapping with post-save relation parsers, and it is non-persistent so it never hits the `setRawAttributes()` branch), so the change is fully scoped to the seeder's trusted import path. It mirrors the `setRawAttributes()` bypass already used a few lines above in the `isPersistent()` branch, and imports every exported column verbatim.

`demo` stays out of `$fillable`, so no user-facing mass-assignment path can flip `demo = true` — the security concern raised in the issue is avoided (no `$fillable` change).

## The "full db:seed didn't reproduce it" puzzle

Confirmed the DungeonRoute code path is byte-identical for `db:seedone` and full `db:seed` (both go through `DatabaseSeeder::run()` → the same transaction → the same `create()` branch). There is no second code path by which a full seed could set `demo = 1` while `db:seedone` sets `demo = 0`; the pre-fix discrepancy was an environment/measurement artifact. It is moot after this fix.

## Verification (against the shared dev DB)

- `db:seedone DungeonDataSeeder` → `demo = 1` count holds at **36**; `bbzlbOX` has exactly one `demo = 1` row with its original `created_at` preserved.
- Full `db:seed` → `demo = 1` count holds at **36**.
- Reverting to `create()` drops the count to 0 (regression test fails), re-applying restores it.
- New regression test `run_givenDemoDungeonRoutes_preservesDemoFlagAcrossReseed` + existing `DungeonDataSeederTest` — 3 passed.
- `composer run fix` — 0 changes. `composer run analyse` — no errors.

### DIFF
--- a/database/seeders/DungeonDataSeeder.php
+++ b/database/seeders/DungeonDataSeeder.php
@@ -439,7 +439,10 @@ class DungeonDataSeeder extends Seeder implements TableSeederInterface
             elseif ($mapping->getPostSaveRelationParsers()->isNotEmpty()) {
                 /** @var class-string<Model> $mappingClass */
                 $mappingClass = $mapping->getClass();
-                $createdModel = $mappingClass::from(DatabaseSeeder::getTempTableName($mappingClass))->create($modelData);
+                // forceCreate (not create) so non-$fillable columns exported by mapping:save are imported
+                // verbatim. DungeonRoute keeps demo, pull_gradient etc. out of $fillable to block user-facing
+                // mass assignment; dropping them here would silently reset demo routes to demo = false.
+                $createdModel = $mappingClass::from(DatabaseSeeder::getTempTableName($mappingClass))->forceCreate($modelData);
                 $updatedModels++;
             } // We don't need to do post-processing, add it to the list to be saved
             else {
--- a/tests/Feature/Database/Seeders/DungeonDataSeederTest.php
+++ b/tests/Feature/Database/Seeders/DungeonDataSeederTest.php
@@ -2,6 +2,7 @@
 
 namespace Tests\Feature\Database\Seeders;
 
+use App\Models\DungeonRoute\DungeonRoute;
 use App\Models\Npc\NpcCharacteristic;
 use App\Models\Npc\NpcSpell;
 use App\Models\Spell\Spell;
@@ -104,5 +105,24 @@ final class DungeonDataSeederTest extends PublicTestCase
             NpcCharacteristic::query()->where('npc_id', self::SENTINEL_NPC_ID)->delete();
             SpellDungeon::query()->where('spell_id', self::SENTINEL_SPELL_ID)->delete();
         }
+    }
+
+    #[Test]
+    public function run_givenDemoDungeonRoutes_preservesDemoFlagAcrossReseed(): void
+    {
+        // Arrange - rollback() deletes every demo = true route and rebuilds them from the JSON files,
+        // which do contain demo routes, so a healthy re-seed must leave demo routes behind.
+
+        // Act
+        $this->artisan('db:seedone', ['className' => 'DungeonDataSeeder'])->assertSuccessful();
+
+        // Assert - the rebuild uses forceCreate() so the demo column (deliberately kept out of
+        // $fillable) is imported verbatim. With a plain create() mass assignment would drop demo and
+        // every rebuilt route would land as demo = false, dropping this count to zero (#3376).
+        $this->assertGreaterThan(
+            0,
+            DungeonRoute::query()->where('demo', true)->count(),
+            'Demo dungeon routes must survive a re-seed with demo = true',
+        );
     }
 }


Wotuu (commented):
Ok

PR #3382, opened 2026-07-06
#2947 Fix heatmap IndexSizeError when map container has zero width

### LINKED ISSUE #2947: Heatmap sometimes doesn't load on the weekly route page
```
lib-3daa310b9c7793cb…072bce9b6ff6ff.js:1 Uncaught IndexSizeError: Failed to execute 'getImageData' on 'CanvasRenderingContext2D': The source width is 0.
    at e.draw (lib-3daa310b9c7793cb333795f520072bce9b6ff6ff.js:1:3696)
    at e._redraw (lib-3daa310b9c7793cb333795f520072bce9b6ff6ff.js:1:6792)
    at e._reset (lib-3daa310b9c7793cb333795f520072bce9b6ff6ff.js:1:5947)
    at e.onAdd (lib-3daa310b9c7793cb333795f520072bce9b6ff6ff.js:1:4777)
    at e._layerAdd (app-3daa310b9c7793cb333795f520072bce9b6ff6ff.js:2:892716)
    at e.whenReady (app-3daa310b9c7793cb333795f520072bce9b6ff6ff.js:2:868715)
    at e.addLayer (app-3daa310b9c7793cb333795f520072bce9b6ff6ff.js:2:892989)
    at e.addTo (lib-3daa310b9c7793cb333795f520072bce9b6ff6ff.js:1:5046)
    at n.value (custom-3daa310b9c7793cb333795f520072bce9b6ff6ff.js:1:520117)
    at n.value (custom-3daa310b9c7793cb333795f520072bce9b6ff6ff.js:1:637058)
```

Image

Wotuu:
Fixes #2947.

## Problem
The whole map bootstraps **synchronously on `DOMContentLoaded`** (`inline.blade` → `CommonMapsMap.activate` → `_initDungeonMap` → `new DungeonMap` → `refreshLeafletMap`), i.e. before the browser lays out the percentage/flex-sized `#map` container. Leaflet caches a stale size of `0`, and `HeatPlugin.addToMap()` (called in the plugin loop of `refreshLeafletMap`) creates an `L.heatLayer` whose canvas is sized to `0` → `IndexSizeError: getImageData ... source width is 0`. This aborted map loading intermittently on the explore and weekly-route/heatmap pages.

## Fix — a map-level size gate around the plugin loop
Rather than guarding each plugin individually, the deferral now lives once in `DungeonMap`:

- **`DungeonMap._whenMapSized(callback)`** wraps the plugin remove/add loop in `refreshLeafletMap()`. It runs the callback synchronously when the map already has a non-zero size (floor changes after first load); otherwise it polls the container's `clientWidth` via `requestAnimationFrame` (Leaflet caches a stale `0` from `getSize()`), calls `invalidateSize()` to refresh the cache, then runs the callback. A pending gate is cancelled before a new one is scheduled, so repeated refreshes can't stack. This means future plugins never need their own deferral code.
- **`HeatPlugin`** is reverted to adding its layer directly. It re-applies the current floor's data right after creating the layer, so data that arrived during the (~1 frame) gate window renders instead of silently waiting. The `heatLayer === null` guards in `_applyLatLngsForFloor()`/`setOptions()` stay, since the layer is briefly null during the gate.

## Tests
- `dungeonmap.test.js` (new): exercises `_whenMapSized` on an `Object.create(prototype)` instance (sync path, zero-size-until-laid-out + `invalidateSize`, pending-gate cancellation).
- `heatplugin.test.js`: layer add, floor-data re-apply, and the null-layer guards.

Full JS suite: 105 passed. This is a unit test of the gate; the `refreshLeafletMap` wiring is not integration-tested (impractical in jsdom). No PHP changed.

### DIFF
--- a/resources/assets/js/custom/dungeonmap.js
+++ b/resources/assets/js/custom/dungeonmap.js
@@ -45,6 +45,8 @@ class DungeonMap extends Signalable {
 
         /** @type Boolean */
         this._refreshingMap = false;
+        /** @type {?Number} requestAnimationFrame handle used while waiting for the map container to gain a non-zero size */
+        this._mapSizedRafId = null;
 
         this.options = options;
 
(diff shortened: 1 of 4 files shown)


Wotuu (commented):
Needs manual review in the browser.

Issue #1430, opened 2022-11-04
Handle timed-out CSRF tokens


Wotuu:
https://stackoverflow.com/questions/31449434/handling-expired-token-in-laravel

Wotuu:
I disabled the entire CSRF tokens since it was simply not listening to what I was telling it. It should be re-enabled later on though.

PR #3392, opened 2026-07-06
#3391 Harden request verification: re-enable CSRF, convert admin deletes to DELETE

### LINKED ISSUE #3391: Harden request verification: re-enable CSRF protection and convert admin delete links to DELETE
CSRF token validation is currently disabled for all routes (`bootstrap/app.php`, `validateCsrfTokens(except: ['*'])`). The front-end is already fully wired to send tokens (the `<meta name="csrf-token">` tag in `layouts/app.blade.php` plus `$.ajaxSetup` setting `X-CSRF-TOKEN` in `app.js`, and every traditional POST form carries `@csrf`), so the global disable is legacy and can be removed as a hardening step.

## Work

1. **Publish `config/session.php`** and pin `same_site` explicitly (default `lax`) so the cookie's cross-site behavior is not left to an implicit framework default.
2. **Re-enable CSRF**: change `except: ['*']` to `except: ['webhook/*']` in `bootstrap/app.php`. Only the external webhook endpoints (`webhook/github`, `webhook/wowhead/*`) legitimately need the exception; everything else already sends a token.
3. **Convert the two admin state-changing GET routes to `DELETE`** (defense in depth — CSRF does not protect GET):
   - `admin.mappingversion.delete` (`routes/web.php`) + its link in `resources/views/admin/dungeon/mappingversions.blade.php`.
   - `admin.dungeonspeedrunrequirednpc.delete` (`routes/web.php`) + its link in `resources/views/admin/floor/speedrunrequirednpcs.blade.php`.
   Use the existing DELETE-form pattern from `resources/views/admin/dungeonroute/edit.blade.php` (`<form method="POST"> @csrf @method('DELETE')`).

## Notes / risks

- `gameversion.update` is also a state-changing GET, but it only flips a UI preference (low impact) and its callers are styled nav-dropdown links; converting it is higher churn for little gain — tracked as a minor follow-up, not part of this change.
- Feature tests bypass CSRF (`runningUnitTests()`), so the suite passing does **not** exercise the token flow — verify by clicking through the map editor and an admin form in a browser after the change.

## Definition of done

- `config/session.php` present with `same_site` pinned; `except` narrowed to `webhook/*`.
- The two admin deletes use DELETE forms; links removed.
- Full suite green (ignore the known pre-existing `StructuredLoggingTest` failure); `composer run analyse` + `composer run fix` clean.
- Manual browser check: map editor still saves, admin delete still works, login/register/profile forms still submit.

Wotuu:
Closes #3391.

CSRF token validation was disabled for all routes (`bootstrap/app.php`, `validateCsrfTokens(except: ['*'])`). Because the front-end already sends tokens everywhere — the `<meta name="csrf-token">` tag in `layouts/app.blade.php`, `$.ajaxSetup` injecting `X-CSRF-TOKEN` in `app.js`, and every traditional POST form using `@csrf` — the global disable was legacy and safe to remove.

## Changes
- **`config/session.php`** published (was unpublished, relying on the framework default) with `same_site` pinned to `lax` via env, so cross-site cookie behavior is explicit and reviewable rather than an implicit default.
- **`bootstrap/app.php`** — CSRF exception narrowed from `['*']` to `['webhook/*']`. Only the external GitHub/Wowhead webhooks legitimately POST without a session token; verified there are no other machine-to-machine POST endpoints in `routes/web.php`.
- **Two admin GET deletes → DELETE** (`admin.mappingversion.delete`, `admin.dungeonspeedrunrequirednpc.delete`), with `@csrf @method('DELETE')` forms replacing the anchor links — CSRF does not protect GET, so state-changing GETs are converted as defense in depth. Mirrors the existing DELETE-form pattern in `admin/dungeonroute/edit.blade.php`.

## Verification
- `composer run fix` clean; `composer run analyse` shows only pre-existing errors on `development` (27, all in files this PR does not touch — inline-ignore drift; this PR introduces none).
- Tests green: Breadcrumbs (54), View + Admin controllers (60). Routing and blade rendering unaffected.
- **Note:** feature tests bypass CSRF (`runningUnitTests()`), so the suite does not exercise the token flow. A browser click-through is recommended before merge — confirm the map editor still saves, the two admin deletes work, and login/register/profile forms still submit.

## Follow-up (not in this PR)
`gameversion.update` is also a state-changing GET but only flips a UI preference (low impact) and its callers are styled nav-dropdown links; converting it is higher churn for little gain and can be a separate change.

---

## Update 2026-07-08

- **Session lifetime default raised to 7 days** (`config/session.php`, was 120 min): with CSRF back on, an idle-expired session turns the next POST from an open tab into a 419; 7 days makes that practically unreachable (#1430). Production has no `SESSION_LIFETIME` override, so the default applies.
- **Browser-level CSRF click-through done in a worktree of this branch** (the suite bypasses CSRF, so this was verified over HTTP):
  - ajax POST without token → **419**; with `X-CSRF-TOKEN` header + session cookie → passes CSRF (200/400 from the controller as appropriate).
  - `POST /login` without `_token` → **419**; with token → 302 (normal validation flow).
  - Both converted admin delete routes are DELETE-only in `route:list`; GET no longer reaches the controller.
- Follow-ups filed: #3452 (ajax 419 auto-recovery via XSRF cookie/retry), #3453 (pre-existing: wrong-method requests render 500 instead of 405 — reproduced on production, not a regression of this MR).
- Post-deploy watch: `TokenMismatchException` is (correctly) in `Handler::$dontReport`, so Sentry will not see 419s — monitor **Cloudflare analytics for HTTP 419 counts** instead; baseline is zero.


Wotuu (commented):
Ok, but later

PR #3446, opened 2026-07-07
#3058 Add dungeon landing page

### LINKED ISSUE #3058: Rework view routes for dungeon page
3 columns to break up the 4 column monotone

HEADER
<-- Weekly route PUG route --> <-- Expert route  --> <--  PLTP Route -->

Collapsible? Non-new people don't need to see this really
<-- Dungeon image --> <-- Fluff text "this is Halls of Atonement, you go there for the Revendreth faction to defeat Blargblargh the Destroyer" -->

When logged in - routes that you have made?
<-- Route 1 --> <-- Route 2 --> <-- Route 3 -->
Show link to show more (redirect to what is the current home page with filters set for this dungeon)

List of popular routes as I have it now (WITHOUT WEEKLY ROUTES)
<-- Route 1 --> <-- Route 2 --> <-- Route 3 --> <-- Route 4 -->

NOTHING ELSE - get rid of this week's affix, next week's affix, new routes

Wotuu:


Wotuu:
:robot: Pushed some improvements to the dungeon landing page:

**Bug fix — Bootstrap 4 classes under a BS5 project**
`landing.blade.php` used `mr-2` and `font-weight-bold`, which are no-ops under Bootstrap 5. These have been changed to `me-2` / `fw-bold`, so the expansion icon now gets its right margin and the "Know the dungeon" card titles render bold. (This file was the only content view still using BS4 utilities.)

**Moved queries out of the controller**
Per the service/repository conventions, the two query helpers that lived on the controller now live where they belong:
- `getDungeonStats()` → `DungeonService::getDungeonOverviewStats()` (interface + Stub updated)
- `getUserRoutesForDungeon()` → `DungeonRouteRepository::getRoutesForUserAndDungeon(User, Dungeon, limit)`; the controller keeps only the guest/null-user branch.

While moving the user-routes query I also switched its eager-load from `author` to `author.iconfile` — the route cards render the author avatar, so the old list would have lazy-loaded `iconfile` on render.

**Visual polish**
The featured-routes row now uses `justify-content-center`, so when a dungeon has only 1–2 weekly routes they sit centered instead of left-aligned with dead space on the right.

**Tests**
Added `discoverDungeon_givenUserHasRouteForDungeon_returnsYourRoutesSection` covering the "Your routes" section. All 3 tests in the file pass; `composer run analyse` and `composer run fix` are clean.

Verified in a headless browser (logged in as admin, feature on) on Pit of Saron (no weekly routes) and Magister's Terrace (2 weekly routes) — no page errors, sections render as expected.

app/Http/Controllers/DungeonRoute/DungeonRouteDiscoverController.php
```diff
@@ -294,6 +315,31 @@ public function discoverDungeon(
             $currentAffixGroup = $expansionService->getCurrentAffixGroup($gameVersion->expansion, $userRegion);
         }
 
+        if (Feature::active(DungeonOverview::class)) {
+            EloquentCollection::make(
+                $dungeonWeeklyRoutes
+                    ->map(fn(WeeklyRoute $weeklyRoute) => $weeklyRoute->dungeonRoute)
+                    ->filter()
+                    ->values(),
+            )->load(self::ROUTE_CARD_RELATIONS);
```
Wotuu: :robot: **This `->load()` block is redundant and actively regresses eager loading — I think it can just be deleted (along with `ROUTE_CARD_RELATIONS`).**

`$dungeonWeeklyRoutes` has exactly one source: `$dungeonRouteRepository->getWeeklyRoutes($dungeon)` on line 284. That method already eager-loads **all seven** relations listed in `ROUTE_CARD_RELATIONS`, *plus* `tags` and `withCount('favorites')` — and critically it loads `author.iconfile`, not just `author`.

Two consequences, both verified in tinker against the dev DB:

1. `Collection::load()` re-queries unconditionally (unlike `loadMissing()`), so this runs **12 extra queries per request** that were already run a few lines above.
2. Because the constant lists `'author'` rather than `'author.iconfile'`, the load *replaces* the already-hydrated author models with fresh ones that have no `iconfile`:

```
before: author loaded=1  iconfile loaded=1
queries run by load(): 12
after:  author loaded=1  iconfile loaded=0
```

`common/dungeonroute/cardvertical.blade.php` renders `common.user.name` -> `common.user.userimage`, which does `@isset($user->iconfile)`. `Model::offsetExists()` routes through `getAttribute()`, so that lazy-loads the relation back: one extra query per distinct author, and a logged lazy-loading violation (`AppServiceProvider::boot`) whenever more than one author is hydrated — `Builder::hydrate()` only sets `preventsLazyLoading` when `count($items) > 1`, which is why it stays silent with a single Raider.IO author but not in general.

Nothing here fires in CI because the seeded test DB has no Raider.IO weekly routes for the dungeons under test, so the collection is empty and `load()` short-circuits.

app/Repositories/Database/DungeonRoute/DungeonRouteRepository.php
```diff
@@ -139,6 +140,30 @@ public function getWeeklyRoutes(?Dungeon $dungeon = null, ?Season $season = null
             });
     }
 
+    /**
+     * @return Collection<int, DungeonRoute>
+     */
+    public function getRoutesForUserAndDungeon(User $user, Dungeon $dungeon, int $limit): Collection
+    {
+        return $user->dungeonRoutes()
+            ->where('dungeon_id', $dungeon->id)
```
Wotuu, line 150: :robot: **`getRoutesForUserAndDungeon()` is not scoped to the game version, unlike every other route collection on this page.**

`getWeeklyRoutes()` scopes via the season, and `DiscoverService::popularBuilder()` explicitly filters `mapping_versions.game_version_id`. This query filters only on `dungeon_id`, so "Your routes" is the one panel on the landing page that ignores the `$gameVersion` route parameter.

That is not hypothetical — 15 dungeons currently have mapping versions in more than one game version (Legion dungeons exist under both `retail` and `legion-remix`):

```
dungeons with mapping versions in >1 game version: 15
arcway, blackrookhold, cathedralofeternalnight, courtofstars,
darkheartthicket, eyeofazshara, hallsofvalor, lowerkarazhan, ...
```

So on `/routes/legion-remix/courtofstars` a user sees their **retail** Court of Stars routes. Worse, `cardvertical.blade.php` then evaluates `!$dungeonroute->mappingVersion->isLatestForDungeon()` on those cross-version routes and shows the "outdated mapping version" warning triangle on routes that are perfectly current for their own game version.

Suggest joining/`whereHas`-ing `mappingVersion` on `game_version_id` (which means the method needs the `GameVersion` passed in).

app/Service/Dungeon/DungeonService.php
```diff
@@ -122,4 +125,39 @@ public function getDungeonsForGameVersion(?GameVersion $gameVersion = null): Col
 
         return ($nextSeason ?? $currentSeason)?->dungeons ?? $gameVersion->expansion->dungeons; // @phpstan-ignore nullsafe.neverNull
     }
+
+    public function getDungeonOverviewStats(Dungeon $dungeon, GameVersion $gameVersion): array
+    {
+        $mappingVersion = $dungeon->getCurrentMappingVersion($gameVersion);
+
+        return Cache::remember(
```
Wotuu: :robot: **Direct `Cache::` use bypasses `CacheServiceInterface` — this is the only such call in all of `app/Service/`.**

Every other service (`BaseDiscoverService`, `MapContextService`, `MDTMappingVersionService`, `UserService`, `ViewService`, ...) goes through `CacheServiceInterface`. `DungeonService` already has a DI constructor, so this is a one-parameter change.

Going through the service matters for two reasons here:

- `CacheService::remember()` wraps the closure in `app('model-cache')->runDisabled(...)` so the queries inside do not also populate the model cache, and it honours `isCacheEnabled()` / the bypass flag. `Cache::remember` honours neither.
- Tests share the main redis. A 1-hour entry written under `dungeon.overview.stats.<id>.<id>` during a feature test outlives the test and bleeds into the dev browser and later runs. Latent today (nothing asserts on the stat values), but it is exactly the kind of cross-run leak the cache service exists to gate.

Separately: `$dungeon->getCurrentMappingVersion($gameVersion)` on line 131 sits *outside* the closure, so its query runs on every request even on a cache hit. Moving it inside would make the cache hit free.

app/Service/Dungeon/DungeonService.php
```diff
@@ -122,4 +125,39 @@ public function getDungeonsForGameVersion(?GameVersion $gameVersion = null): Col
...
+                                ->orWhere('enemies.seasonal_type', '!=', Enemy::SEASONAL_TYPE_MDT_PLACEHOLDER);
+                        })
+                        ->count();
+                }
+
+                return [
+                    'npc'                  => $dungeon->npcs()->count(),
+                    'spell'                => $dungeon->spells()->count(),
+                    'pull_count'           => $pullCount,
+                    'avg_enemies_per_pull' => $pullCount > 0 ? round($enemyCount / $pullCount, 1) : 0.0,
```
Wotuu, line 171: :robot: **`avg_enemies_per_pull` divides *all* enemies by the pack count, including enemies that belong to no pack — so the number is inflated relative to the "Avg. enemies per group" label it renders under.**

`$dungeon->enemies()` is not filtered on `enemy_pack_id`, while `$pullCount` counts `enemy_packs` rows. Measured on the dev DB:

```
blackrookhold:   packs=35  allEnemies=119  packedEnemies=119  avgAll=3.4  avgPacked=3.4
courtofstars:    packs=35  allEnemies=152  packedEnemies=134  avgAll=4.3  avgPacked=3.8
darkheartthicket: packs=42 allEnemies=136  packedEnemies=136  avgAll=3.2  avgPacked=3.2
```

Court of Stars advertises 4.3 enemies per group where the packed-only figure is 3.8 — the 18 unpacked enemies (bosses and lone enemies) are counted into an average of group sizes. Adding `->whereNotNull('enemies.enemy_pack_id')` fixes it.

Also worth noting: this is a new public service method with a non-trivial null-mapping-version branch and no direct test — the test suite only asserts the page renders, so neither this number nor the `$pullCount === 0` fallback is pinned anywhere.

Wotuu:

Verified clean: the rebase reconciliation is correct — `$gameVersionDungeons` is computed ahead of both branches and passed by both; reusing `$dungeonWeeklyRoutes` unfiltered is behaviourally identical to the sibling branches' `->filter(...)->values()` because `getWeeklyRoutes()` only ever pushes a `WeeklyRoute` with a non-null `dungeonRoute`; `$currentAffixGroup` is computed before the `DungeonOverview` check so it is populated in that branch. The dynamic `featured.%s` translation key matches `keystoneguru.raider_io.weekly_route.tags` exactly (`pug_friendly`/`expert`/`title`), and the new feature class is auto-discovered by `AdminToolsFeaturesController` so it needs no registration.

One non-finding for your call: the landing page carries no ad units, where the legacy overview has three and the `DungeonRouteListRework` branch has one.

app/Http/Controllers/DungeonRoute/DungeonRouteDiscoverController.php
Wotuu: :robot: Fixed in 306984a39 (pushed to this branch). Confirmed by reading `getWeeklyRoutes()`: it already eager-loads every relation `ROUTE_CARD_RELATIONS` listed, plus `tags`, `withCount(favorites)`, and critically `author.iconfile` (not just `author`). Deleted the `EloquentCollection::make(...)->load(...)` call and the now-unused `ROUTE_CARD_RELATIONS` constant entirely - nothing was left un-eager-loaded for the landing view.

app/Repositories/Database/DungeonRoute/DungeonRouteRepository.php
Wotuu: :robot: Fixed in 306984a39. `getRoutesForUserAndDungeon()` now takes a `GameVersion` parameter and filters via `whereRelation('mappingVersion', 'game_version_id', $gameVersion->id)`, mirroring how `getWeeklyRoutes()` scopes to the season. Updated the interface, the Database and Stub implementations, and the controller call site (`getUserRoutesForDungeon()` now threads `$gameVersion` through). Added a regression test (`DungeonRouteRepositoryTest::getRoutesForUserAndDungeon_givenRoutesOnDifferentGameVersions_returnsOnlyTheMatchingGameVersion`) that creates two isolated mapping versions on different game versions with a route each, and asserts only the matching-game-version route comes back.

Note: `dungeon_routes.mapping_version_id` is nullable at the schema level, and `whereRelation` uses `whereHas` semantics, so a route with a null mapping version would now be excluded from "Your routes". This matches how `getWeeklyRoutes()` and `DiscoverService` already assume a populated mapping version via inner joins, and the model's own PHPDoc declares the property as non-nullable `int`, so this is consistent with the rest of the discover feature rather than a new narrowing.

app/Service/Dungeon/DungeonService.php
Wotuu: :robot: Fixed in 306984a39. `DungeonService` now takes `CacheServiceInterface` via constructor injection and `getDungeonOverviewStats()` calls `$this->cacheService->remember(...)` instead of the `Cache` facade, so it goes through the `model-cache` `runDisabled()` wrapper and `isCacheEnabled()`/bypass checks like every other service. Also moved `$dungeon->getCurrentMappingVersion($gameVersion)` inside the cached closure, since it's purely derived from `(dungeon, gameVersion)` which are already the cache key, so it's fine to bake in - a cache hit no longer runs that extra query.

On the cross-run leak specifically: I verified this is genuinely fixed, not just half-fixed. `KeystoneGuruServiceProvider` binds `CacheServiceInterface` to `DevCacheService` (which sets `bypassCache = true`) whenever `config('app.env')` is `local` or `testing`, and `phpunit.xml` sets `APP_ENV=testing`. `CacheService::remember()` only writes to redis when `!$this->isBypassCache()`. So under the test suite this now never touches real redis at all (previously `Cache::remember()` via the facade always wrote a real 1-hour entry, which is exactly the leak described). The ttl is passed as `'1 hour'` (string), matching the convention in `BannedIpAddressService`.

app/Service/Dungeon/DungeonService.php
Wotuu: :robot: Fixed in 306984a39. Added `->whereNotNull('enemies.enemy_pack_id')` to the enemy count query so only enemies that belong to an actual `EnemyPack` contribute to `avg_enemies_per_pull` - unpacked bosses/lone enemies no longer inflate it.

Also added `tests/Feature/App/Service/Dungeon/DungeonService/GetDungeonOverviewStatsTest.php`, which was missing entirely: two tests build a real isolated mapping version (packs + packed/unpacked enemies), mock `CacheServiceInterface` to just call through the closure (so it's a pure computation test, no real caching involved), and assert `pull_count`/`avg_enemies_per_pull` - one with a mix of packed and unpacked enemies (proves unpacked enemies are excluded), one with zero packs (proves the `pull_count === 0` fallback returns `0.0`).

PR #3447, opened 2026-07-07
#3349 Route poster card

### LINKED ISSUE #3349: Refresh route presentation on dungeon page: poster card orientation (behind feature flag)
## Background

The dungeon routes page (e.g. `/routes/retail/pit-of-saron`) presents routes as a grid of vertical Bootstrap cards (`common.dungeonroute.cardvertical`). That card has existed for years and is space-hungry: a thumbnail strip, a fixed 6rem description box (often empty — it renders `&nbsp;`), an enemy-forces row, and a footer. For a *long* list of routes this is repetitive and scroll-heavy, and it foregrounds the wrong thing.

Two findings drove the new design:

1. **The map thumbnail is not a decision driver.** ~99% of routes have a thumbnail, but people don't choose routes by looking at the map. It should be *de-prioritized* vs. today.
2. **What drives route choice is social proof + author trust.** Rating / views / favorites all exist on the model but are currently commented out of the card. A future feature will let authors link Raider.IO so we can show a trust badge ("this author can clear +18") — not built now, but the layout must reserve room for it.

**Affixes are intentionally dropped** from the new card — they are no longer a deciding factor when choosing a route.

Goal: a fresher, denser, gallery-style presentation where the **map is a scrimmed background texture** and **social proof + author are the foreground hero**. Delivered as a new `poster` card orientation, leaving the existing orientations untouched. (A separate effort is reworking the overall page into a full "dungeon overview" — this issue only touches the route-list presentation component and its wiring.)

The entire rework ships **behind a Pennant feature flag** so it can be deployed merged but toggled on/off in production to gather feedback and refine it.

## Design

A uniform-aspect tile, many per row (masonry/gallery feel):

```
+-------------------------+
|::: map (single image,   |  <- background, heavy gradient scrim for legibility
|::: scrimmed) ::::::::::: |     + outdated-mapping icon / actions menu top-right
|                         |
| Route Title         96% |  <- title (prominent) + EF% check badge
| +18-20                  |  <- level range, secondary
| ----------------------- |
| Author [badge slot]     |  <- HERO strip: author (+ reserved Raider.IO badge slot)
| *4.8 (12)  1.2k  fav 30 |  <- HERO strip: social proof (rating/views/favorites)
+-------------------------+
```

- **Map**: single background image (combined facade / primary floor) via CSS `background-image` with a gradient scrim (reuse the pattern already in `cardvertical` `showDungeonImage` mode). No carousel here — multi-floor browsing stays on the route page and in the legacy `vertical` orientation.
- **Foreground hero**: author name + avatar (`common.user.name`) with an empty, styled slot beside it reserved for the future Raider.IO trust badge; below it a social-proof row showing rating (star + count), views, and favorites count.
- **Secondary**: EF% check/warning badge (reuse `getEnemyForcesPercentage()` + `enemyForcesWarning` logic), level range (`common.dungeonroute.level`), outdated-mapping warning icon, and the existing actions menu (report / admin refresh).
- **Affixes dropped** — no seasonal affix, no tier affix; the affix-popover script block from `cardvertical` is not carried over.
- Description dropped from the tile (low value, often empty).

## Feature flag (Pennant)

- New `app/Features/DungeonRouteListRework.php` — copy the shape of `app/Features/SearchPageRework.php`: a `resolve(?User $user): bool` that returns `false` unless `Feature::getAdminValue(self::class)` (admin master switch) is set, then restricts to `Role::ROLE_ADMIN` / `Role::ROLE_INTERNAL_TEAM` while gathering feedback.
- Toggle point: in `overview.blade.php`, choose the orientation per `Feature::active(DungeonRouteListRework::class)` — `'poster'` when active, `'vertical'` (current behavior) otherwise. Mirrors existing `Feature::active(FrontPageRework::class)` / `Feature::active(SearchPageRework::class)` usage.

## Key files

- **New** `resources/views/common/dungeonroute/cardposter.blade.php` — model on `cardvertical.blade.php`: same `@inject` cache service, `$cacheFn` + `cacheService->remember(...)` wrapper, `DungeonRoute::getCardCacheKey(...)` with orientation `'poster'`. Reuse EF% logic; drop affix-popover `@section('scripts')` and all affix logic.
- **New** `app/Features/DungeonRouteListRework.php`
- `resources/views/common/dungeonroute/cardlist.blade.php` — add `'poster' => 'common.dungeonroute.cardposter'` to the `match($orientation)`.
- `resources/views/dungeonroute/discover/panel.blade.php` — accept `$orientation` (default `'vertical'`) and pass into `cardlist`.
- `resources/views/dungeonroute/discover/dungeon/overview.blade.php` — pass flag-gated orientation into the `panel` includes (coordinate with the parallel page-rework).
- `app/Http/Controllers/DungeonRoute/DungeonRouteDiscoverController.php` — add `->withCount('favorites')` to the queries feeding these panels (poster card needs a favorites count; avoid lazy-load violation which throws in dev).
- `resources/assets/css/custom.css` — `.card_dungeonroute.poster` styles: fixed aspect ratio, `background-size: cover`, `::before` gradient scrim, foreground z-index, hover lift; use existing theme vars.
- `lang/en_US/view_common.php` — reuse `view_common.dungeonroute.card.*` keys; add social-proof labels as needed.

## Definition of done

- With flag **on** (admin master value set + qualifying role), `/routes/retail/pit-of-saron` route sections render poster tiles: map as scrimmed background, author + rating/views/favorites foreground, EF%/level secondary (no affixes), actions menu + outdated-mapping icon working. A route with 0 favorites shows `0` with no lazy-load exception (lazy-loading throws in dev).
- With flag **off**, the page falls back to the existing `vertical` cards unchanged.
- Feature test for the discover dungeon overview: flag active asserts poster card (title, author, rating); flag inactive asserts legacy vertical card. Factory + cleanup in `try...finally`.
- `composer run fix` and `composer run analyse` clean. New files staged.

## Deferred (not in this issue)

- Raider.IO author trust badge — only the layout slot is reserved now.
- True CSS-grid masonry / changing column counts — reuse the existing Bootstrap `cols` mechanism for now.

Wotuu:
Closes #3349

## What this does

Reworks the per-dungeon discovery page (e.g. `/routes/retail/pit-of-saron`) from a grid of vertical image cards into a **hero band + ranked community leaderboard**, entirely behind the `DungeonRouteListRework` Pennant flag (admin / internal-team only). Flag **off** — i.e. everyone else — still renders the existing vertical-card page unchanged.

### Hero band
- **Weekly routes present:** the Raider.IO weekly routes render as labelled *archetype* cards (PUG-friendly / Expert / Title), each with a one-line description. The map is the one cinematic element (wide 1600×640 hero render).
- **No weekly routes:** the **top 3 community routes** fill the hero band instead (`#1 / #2 / #3 community route`), so the band is never a lonely single card.

### Ranked community leaderboard
Remaining routes as a dense, centred, ranked list (max-width 70rem): rank numeral · 72px "route fingerprint" thumbnail · title + author · a right-aligned stat cluster (EF warning · rating · favorites · key-level chip · pull graph · views) · actions menu. Recently-published routes get a small "New" badge.

**Pull-size "fingerprint" graph** — one bar per pull, height ∝ that pull's enemy forces, so "few big pulls vs many small" is eyeballable; boss pulls render as full-height amber bars. Single aggregate query (`DungeonRoute::getEnemyForcesPerKillZone()`) mirroring `getEnemyForces()` accounting.

### Thumbnail pipeline
A larger `hero` render variant (1600×640, zoom 2.1, thin pull lines) alongside the standard render; standard renders draw thicker pull lines so the small fingerprints stay legible. `getHeroThumbnailUrl()` falls back to the standard thumbnail. Hero variants are **not** fanned out on every refresh — a new hourly `thumbnail:ensureheroes` command keeps them fresh only for the routes shown as heroes (weekly routes + top-3 community per dungeon).

## Merge note — reconciled with #3548

This branch predated #3548 ("Fix local dungeon-route thumbnail generation & loading"), which independently rewrote the same `ThumbnailService`. `origin/master` is merged in and the overlap reconciled: #3548's disk-safety guard (`isRemoteDiskUnsafeForLocalGeneration`), configurable `getPreviewUrl()` (`preview_base_url`), and `$disk` threading are kept, with this branch's `$variant`/hero logic layered on top (`thicklines` folded into `getPreviewUrl`).

## Mobile

Reworked at 390px: hero cards stack full-width; leaderboard rank sits top-left, centred against the thumbnail; the stat cluster collapses to one right-aligned row (rating + favorites hidden) so it never wraps or clips.

## Tests

47 affected tests green (`ThumbnailService`, `CardHero`/`CardRow`/`CardPoster`/`PullGraph`, `EnsureHeroThumbnails`, `GetHeroThumbnailUrl`, plus #3548's `MigrateThumbnailDisk`). PhpStan clean, php-cs-fixer applied.

## Follow-up

The "Show more" / category pages the overview links to still use the old cards — tracked in #3561 to receive the same leaderboard treatment.

## Screenshots

**Desktop**

desktop

**Mobile (390px)**

mobile


Wotuu:
:robot: **Handover — dungeon route discovery page redesign (#3349)**

Picking this up tomorrow. Everything below is pushed to `3349-route-poster-card`. Nothing is merged; work continues on this branch.

## What this PR now does

Behind the `DungeonRouteListRework` Pennant flag (admin / internal-team only), the per-dungeon discovery page (`/routes/retail/{dungeon}`) is reworked from a grid of image posters into:

- **Hero band** — the Raider.IO weekly routes as labelled *archetype* cards (PUG-friendly / Expert / Title, each with a one-line description). When a dungeon has no weekly routes, the #1 community route is promoted into the hero slot ("Top community route"). The map is the one cinematic element here (wide 1600×640 render).
- **Ranked community leaderboard** — remaining routes by popularity: rank numeral, 72px "route fingerprint" thumbnail, title + author, and a right-aligned stat cluster.
- The old "New routes" section is gone; recently-published routes get a small "New" badge in the list instead.
- Flag **off** (i.e. guests / everyone else) still renders exactly the old vertical-card page.

Stat cluster order (rarest → always-present, left→right): enemy-forces warning · rating · favorites · key-level chip · pull graph · views.

**Pull-size "fingerprint" graph** — one bar per pull, height ∝ that pull's enemy forces, so you can eyeball "few big pulls vs many small ones". Boss pulls render as full-height amber bars (trash profile + boss milestones). Zero-forces / no-boss pulls are omitted; the tooltip still shows the true pull count. Single aggregate query (`DungeonRoute::getEnemyForcesPerKillZone()`) mirroring `getEnemyForces()` accounting.

**Thumbnail pipeline** — a larger `hero` variant (1600×640, zoom 2, thin pull lines) alongside the standard render; standard renders now draw pull lines 3× thicker so the small fingerprints stay legible. `DungeonRoute::getHeroThumbnailUrl()` falls back to the standard thumbnail. Hero variants are **not** fanned out on every refresh — a new hourly `thumbnail:ensureheroes` command keeps them fresh only for the routes shown as heroes (weekly routes + top-3 community per dungeon). Two real bugs were fixed en route: invisible pull lines in cookie-less headless renders (NaN weight), and debugbar bleeding into screenshots.

## Commits this session (newest first)

- `906640ed1` Refine: stat order, boss pull bars, hero render tuning (thin lines + zoom 2), hourly hero policy command
- `7d2771f31` Fix invisible pull lines in cookie-less renders + debugbar bleed
- `16ab8ae30` Favorites count, pull-size fingerprint graph, hero thumbnail variant
- `dff2d583a` Rework overview into archetype heroes + ranked leaderboard
- `8cabce0b9` Migrate poster card markup BS4 → BS5
- `c155b355b` Poster card transparently replaces vertical card under the flag

Status: 37 affected tests green, PhpStan clean, php-cs-fixer applied.

## How to review locally

- Worktree stack is on **http://localhost:8112**. The flag is admin-only, so log in as an admin to see the new layout (guests get the old page — that's intended).
- Best pages to look at: `/routes/retail/seat-of-the-triumvirate` (weekly-route hero) and `/routes/retail/pit-of-saron` (fallback "top community route" hero).

## Still open (tomorrow)

1. **Mobile polish** — deliberately deferred until desktop is locked. Rows/hero need a pass at 390px (long titles + ⋮ menu wrap awkwardly; hero looks bare when a route lacks a thumbnail).
2. **PR body** — still empty. Should start with `Closes #3349` + a summary.
3. **Before/after screenshots** on the PR via `post-screenshot.sh`.
4. **Cleanup** of my verification scaffolding: throwaway admin user `claude-design-verify@example.com` (id **75191**), the `.chrome-tmp/` dir, and the `chrome` compose service.
5. Only **2 hero thumbnails** were regenerated locally for review (`t13kZXu` PoS, `Tf03QDk` Seat). Other heroes on :8112 still show the older thick-line/zoom-1 render until regenerated — in production the hourly command handles this. The ~30 standard thumbnails from an earlier round already have the thick fingerprint lines.

## Dev-env gotchas (so tomorrow-you doesn't relearn them)

- **Local asset serving**: after any JS/CSS change, rebuild in the worktree app container with `npm run dev` passing `-e npm_config_output_version=$(git rev-parse HEAD)` (the in-container git-revision plugin otherwise fails), then write that same rev into the repo-root `version` file so the page serves the fresh bundle. `version` is gitignored / dev-only.
- **No card cache locally**: dev binds `DevCacheService`, so card fragments aren't cached — edits show immediately (prod caches per-card via `getCardCacheKey`, which includes orientation + the hero archetype, so no cross-variant bleed).
- **Regenerating hero thumbnails locally** (verification only): temporarily bypass the `app()->environment('local')` guard in `ThumbnailService::doCreateThumbnail`, then run the render with `-e APP_URL=http://nginx -e FILESYSTEM_DISK=local -e PUPPETEER_EXECUTABLE_PATH=/var/www/.chrome-tmp/chrome-headless-shell-linux64/chrome-headless-shell` and eager-load `dungeon.floors` (dev lazy-loading prevention throws otherwise; do **not** use `APP_ENV=production` — it force-schemes the preview URL to https and the internal nginx only serves http). Always revert the guard bypass before committing. The `mkdir permission denied` warning during render is benign.
- Default `FILESYSTEM_DISK` in this dev env is the real S3 bucket, which is why the guard exists — never generate thumbnails without the `FILESYSTEM_DISK=local` override.

## Design decisions already settled with Wotuu (don't re-litigate)

- The dungeon **is** the filter (it's in the URL); user metadata (key levels, ratings) is too sparse to filter on; popularity is the only reliable auto signal and is the sort. **No filter controls** on this page — `/search` covers deep filtering.
- Rankings resonate with the WoW community → the leaderboard makes the popularity order visible.
- Thumbnails kept but demoted: cinematic only in the hero, tiny fingerprints in rows.

.claude/skills/seed-dev-routes/generate_thumbnails.php
```diff
@@ -0,0 +1,62 @@
...
+                $route->public_key,
+                $floor->index,
+                $variant,
+                $result !== null ? sprintf('OK -> %s', $result->file?->getURL()) : 'FAILED',
+                microtime(true) - $start,
+                PHP_EOL,
+            );
+        }
+    }
+}
```
Wotuu: Is this file more useful to you, rather than just firing a `ProcessRouteFloorThumbnailCustom` job and letting Horizon handle it? I think doing it yourself should only be done when you think your local changes should have a different thumbnail and you want to see that change?

app/Console/Commands/Scheduler/Thumbnail/EnsureHeroThumbnails.php
```diff
@@ -0,0 +1,100 @@
...
+     * The routes that appear as heroes on the discovery pages: every Raider.IO weekly route plus the top
+     * community routes of each current-season dungeon, deduplicated by id.
+     *
+     * @return Collection<int, DungeonRoute>
+     */
+    private function getHeroRoutes(
+        SeasonServiceInterface          $seasonService,
+        DiscoverServiceInterface        $discoverService,
+        DungeonRouteRepositoryInterface $dungeonRouteRepository,
+    ): Collection {
```
Wotuu: Q: should this maybe move to the `DiscoverService`? That service generates lists of routes. Not saying you should but it's an idea to make it re-usable. Argue with me.

resources/views/dungeonroute/preview.blade.php
```diff
@@ -8,8 +8,11 @@
  * @var Floor                $floor
  * @var float                $defaultZoom
  * @var string               $mapFacadeStyle
+ * @var bool                 $thickLines
  * @var array<string, mixed> $parameters
  */
+
+$thickLines ??= true;
```
Wotuu: `Thicklines` are passed in various places and then end up in `map.blade.php` to be converted to `$thumbnail`. This is all really vague stuff. Can you just pass along a `$pullConnectionsLineThickness` (maybe think of a shorter name), that when it is `null` it does nothing and assumes default, and otherwise takes that default? Would make things a whole bunch clearer.

app/Models/DungeonRoute/DungeonRouteThumbnail.php
```diff
@@ -11,11 +11,12 @@
...
- * @property int  $dungeon_route_id
- * @property int  $floor_id
- * @property int  $file_id
- * @property bool $custom           True if this thumbnail was requested through the API with custom parameters
+ * @property int    $id
+ * @property int    $dungeon_route_id
+ * @property int    $floor_id
+ * @property int    $file_id
+ * @property bool   $custom           True if this thumbnail was requested through the API with custom parameters
+ * @property string $variant          Which render variant this thumbnail is (standard|hero)
```
Wotuu: I propose dropping the `$custom` value and adding a 3rd variant. Now you can be `$custom` AND `$variant` which is an invalid state imo.

app/Models/DungeonRoute/DungeonRoute.php
```diff
@@ -1372,6 +1490,12 @@ public static function dropCaches(int $dungeonRouteId): void
                 'vertical',
                 'horizontal',
                 'horizontal_row',
+                'poster',
```
Wotuu: Please create a follow-up ticket to optimize this function. Right now it will go through hundreds of possible permutations and try to drop them from Redis. There must be a better way.

app/Service/DungeonRoute/ThumbnailService.php
```diff
@@ -42,14 +42,26 @@ public function createThumbnail(
...
                 $dungeonRoute,
                 $floorIndex,
                 self::THUMBNAIL_FOLDER_PATH,
+                $isHero ? config('keystoneguru.api.dungeon_route.thumbnail.hero_viewport_width') : null,
+                $isHero ? config('keystoneguru.api.dungeon_route.thumbnail.hero_viewport_height') : null,
+                $isHero ? config('keystoneguru.api.dungeon_route.thumbnail.hero_image_width') : null,
+                $isHero ? config('keystoneguru.api.dungeon_route.thumbnail.hero_image_height') : null,
+                $isHero ? config('keystoneguru.api.dungeon_route.thumbnail.hero_zoom_level') : null,
+                $isHero ? config('keystoneguru.api.dungeon_route.thumbnail.hero_quality') : null,
+                $variant,
```
Wotuu: I feel like this scales poorly. Add in another variant and how will this code look like? There should be a better way.

app/Service/DungeonRoute/ThumbnailService.php
```diff
@@ -272,6 +289,33 @@ public function queueThumbnailRefresh(DungeonRoute $dungeonRoute, bool $force =
         return $result;
     }
 
+    public function queueHeroThumbnailRefresh(DungeonRoute $dungeonRoute, bool $force = false): bool
```
Wotuu: Can this function be made generic based on the variant? `public function queueThumbnailRefresh(.... $thumbnailVariant)`.

Same then goes for `hasFreshHeroThumbnail()`, can be renamed `hasFreshThumbnailForVariant()` instead.

config/keystoneguru.php
```diff
@@ -464,6 +464,19 @@
...
+                 * render looks stretched. Rendered (and stored) at full size - no downscale. Uses a higher
+                 * zoom so the route fills the wide frame instead of being letterboxed, and keeps the normal
+                 * (thin) pull-line width since the render is large.
+                 */
+                'hero_viewport_width'  => 1600,
+                'hero_viewport_height' => 640,
+                'hero_image_width'     => 1600,
+                'hero_image_height'    => 640,
+                'hero_zoom_level'      => 2.1,
+                'hero_quality'         => 90,
```
Wotuu: Put this in a `variants` => [`hero` => [...]] wrapper instead.

resources/assets/js/custom/constants.js
```diff
@@ -529,6 +529,9 @@ let c = {
                 color: '#F00',
                 colorAnimated: '#F00',
                 weight: 5,
+                // When rendering a thumbnail the pull-connection lines are multiplied by this so the
+                // miniature still reads as a route shape (they'd otherwise be too thin to see).
+                thumbnailWeightMultiplier: 3,
```
Wotuu: I finally have to find the width of the thumbnail pull connections line. `thumbnailWeightMultiplier` is a poor name and I think it should be passed from the PHP to Javascript.

Wotuu (commented):
Not quite there yet.

.claude/skills/seed-dev-routes/generate_thumbnails.php
Wotuu: :robot: Kept as-is. The point of this dev script is to render **locally and synchronously** so you immediately see local map/JS changes reflected, iterating per-variant — it works together with the temporary local-env guard bypass documented in SKILL.md. Dispatching a `ProcessRouteFloorThumbnailCustom` job would hit the same local-env / S3-disk guard and give no synchronous feedback, so it does not serve the "see my local change rendered" use case. Happy to drop the script entirely instead if you would prefer that.

app/Console/Commands/Scheduler/Thumbnail/EnsureHeroThumbnails.php
Wotuu: :robot: Agreed — moved. `heroRoutes(Season $season, int $topPerDungeon)` now lives on `BaseDiscoverService` (so both `DiscoverService` and `DevDiscoverService` inherit it) and the command just delegates to it. Added a `DiscoverServiceTest`.

resources/views/dungeonroute/preview.blade.php
Wotuu: :robot: Done. The vague `thicklines` -> `$thickLines` -> `thumbnail` boolean chain is replaced by a single nullable numeric that flows config -> preview URL (`killzonepathweight`) -> controller -> blade -> map option `killZonePathWeightMultiplier`. Null assumes the default line width.

app/Models/DungeonRoute/DungeonRouteThumbnail.php
Wotuu: :robot: Reworked. `custom` is now the `custom` variant value (new `VARIANT_CUSTOM`), and the write path + a backfill migration make `variant` the source of truth. Since `custom` is a pre-existing prod column, per the expand/contract migration rule I kept it this release (dual-written, reads still on it during the deploy window) and filed #3567 to drop the column and switch the remaining reads to `variant` once this ships.

app/Models/DungeonRoute/DungeonRoute.php
Wotuu: :robot: Filed #3568 to replace the permutation loop with a key-prefix / cache-tag based invalidation.

app/Service/DungeonRoute/ThumbnailService.php
Wotuu: :robot: Done — `createThumbnail` now resolves the render settings by variant (`config("...thumbnail.variants.$variant")`) instead of the `$isHero ? ... : null` ladder, so adding a variant is just a config entry.

app/Service/DungeonRoute/ThumbnailService.php
Wotuu: :robot: Done — `queueHeroThumbnailRefresh` folded into `queueThumbnailRefresh($dr, $force, $variant)`, preserving each variant`s side-effects (standard records `thumbnail_refresh_queued_at` and honours `$force`; hero/custom skip-if-fresh and always force). `hasFreshHeroThumbnail` -> `hasFreshThumbnailForVariant`.

config/keystoneguru.php
Wotuu: :robot: Done — the `hero_*` keys now live under `thumbnail.variants.hero`, alongside a `variants.standard` block; each variant is a self-contained set of render settings.

resources/assets/js/custom/constants.js
Wotuu: :robot: Done — the multiplier now comes from PHP config (`variants.{variant}.kill_zone_path_weight_multiplier`) and is passed to the map as `killZonePathWeightMultiplier`; the hard-coded `thumbnailWeightMultiplier` is removed from constants.js and the JS reads the value off the map options.

Wotuu:
:robot: Addressed this review round in 998bf1515.

**Thumbnail-variant system generalised:**
- Config: per-variant render settings under `thumbnail.variants.{standard,hero}`; `createThumbnail` resolves settings by variant instead of an `$isHero ? … : null` ladder. (#6, #8)
- `custom` is now the `custom` variant value (`VARIANT_CUSTOM`) + backfill migration; kept the legacy column dual-written this release for deploy safety and filed #3567 to drop it. (#4)
- `queueHeroThumbnailRefresh` folded into `queueThumbnailRefresh($dr, $force, $variant)`; `hasFreshHeroThumbnail` → `hasFreshThumbnailForVariant`. (#7)
- Pull-line thickening now flows a single `killZonePathWeightMultiplier` from PHP config through the preview URL/blade to the map option (replaced `thicklines`/hard-coded `thumbnailWeightMultiplier`). (#3, #9)
- Hero-route selection moved to `DiscoverService::heroRoutes()`. (#2)

**Housekeeping:** dev thumbnail script kept (rationale in-thread, #1); `dropCaches` optimisation tracked in #3568 (#5).

50 affected tests green, PhpStan clean, php-cs-fixer applied. Verified the config → URL → blade → map-option weight chain end-to-end against a live preview render (`killzonepathweight=3` → `killZonePathWeightMultiplier":3`; omitted → `null`).

Wotuu:
:robot: Fixed the flaky `php-tests (feature-other)` failure.

The one failing test — `ThumbnailServiceTest::queueThumbnailRefresh_givenHeroVariantAndStaleHeroThumbnail_queuesAHeroJob` — was intermittent (it passed on the prior run; the `#3561` merge just re-rolled the RNG, it doesn't touch thumbnails).

**Root cause:** `queueThumbnailRefresh(..., VARIANT_HERO)` dispatches jobs by iterating `Dungeon::floorsForMapFacade($mappingVersion, useFacade: true)`, which returns **only facade floors** when the mapping version is `facade_enabled`. The `getDungeonWithNonFacadeFloor()` fixture helper guaranteed a dungeon with *no facade floor* but never constrained `facade_enabled`. Two Midnight dungeons (`the_blinding_vale`, `voidscar_arena`) are in an in-between seed state — `facade_enabled = true` yet no facade floor row — so a ~2% random pick made the floor loop empty, `$result` stayed `false`, and `assertTrue($result)` failed.

**Fix (`tests/Feature/Traits/ProvidesDungeon.php`):** require the current mapping version to have facade rendering **disabled** in `getDungeonWithNonFacadeFloor()` (and, symmetrically, **enabled** in `getDungeonWithFacadeFloor()`), so `floorsForMapFacade(..., true)` always returns the floors the fixture promises. Verified across all seeded dungeons that every selectable dungeon now dispatches ≥1 floor; `ThumbnailServiceTest` green 3×; PHPStan clean.

Wotuu:
:robot: Merged `origin/master` in to resolve the conflict with #3599 (this-week/next-week affix discovery removed for good) and #3594 (jquery-visible removal).

Notable resolution: this branch still had `discoverDungeonThisWeek`/`discoverDungeonNextWeek` redirect stubs (and the `nextAffixGroup`/`thisweek`/`nextweek` view data feeding the legacy flag-off overview) from an earlier "keep a redirect" deprecation approach. Master's #3599 went further and removed the routes entirely, so I matched that: dropped the dead controller methods/view data, and deleted the 2 now-unrunnable tests for those routes (`DungeonRouteDiscoverCategoryTest::discoverDungeonThisWeek_givenAnyState_redirectsToPopular` / `...NextWeek...`) — confirmed with Wotuu before removing them per the no-test-removal-without-approval rule.

`griddiscover.blade.php` took this branch's version wholesale: master's post-#3599 version (no `DungeonRouteListRework` branching) was already a strict subset of this branch's `@else` fallback content, so no logic was lost.

Verified: full PHP + JS test suites pass, phpstan clean, php-cs-fixer clean.

app/Models/DungeonRoute/DungeonRoute.php
```diff
@@ -827,6 +856,95 @@ public function getEnemyForces(): int
...
 
+    /**
+     * Gets the summed enemy forces for each kill zone (pull) in this route, ordered by the kill zone's
+     * index, along with whether that pull contains a boss. Used to render a compact "route fingerprint"
+     * bar graph. This is a single aggregate query that mirrors the enemy-forces accounting of
+     * getEnemyForces() but groups per pull instead of per route.
+     *
+     * @return Collection<int, stdClass> Each row exposes an int `enemy_forces` and a bool `has_boss`.
+     */
+    public function getEnemyForcesPerKillZone(): Collection
```
Wotuu: I'm trying to shrink the DungeonRoute.php class - not add to it. Please move this to either the DungeonRouteService.php or create a new service such as a DungeonRouteKillZoneService (to put methods in that need logic for any DR's pulls).

You can then also offload these queries into a fitting repository.

And use a Dto when you return the data.

app/Models/DungeonRoute/DungeonRouteThumbnail.php
```diff
@@ -28,11 +29,18 @@
  */
 class DungeonRouteThumbnail extends Model
 {
+    public const string VARIANT_STANDARD = 'standard';
+
+    public const string VARIANT_HERO = 'hero';
+
+    public const string VARIANT_CUSTOM = 'custom';
```
Wotuu: Please convert this in a proper `enum`.

app/Service/DungeonRoute/BaseDiscoverService.php
```diff
@@ -121,4 +130,35 @@ public function excludeTeam(?Team $team): DiscoverServiceInterface
...
+    /**
+     * {@inheritDoc}
+     */
+    public function heroRoutes(Season $season, int $topPerDungeon = self::HERO_TOP_ROUTES_PER_DUNGEON): Collection
+    {
+        /** @var Collection<int, DungeonRoute> $heroRoutes */
+        $heroRoutes = collect();
+
+        // The Raider.IO weekly routes (grouped by dungeon key) are always shown as heroes.
+        App::make(DungeonRouteRepositoryInterface::class)->getWeeklyRoutes()
```
Wotuu: Put the `App::make()` in the constructor instead to follow the pattern and make it easily visible what dependencies this service has.

app/Service/DungeonRoute/ThumbnailService.php
```diff
@@ -429,18 +466,40 @@ public function hasThumbnailsGenerated(DungeonRoute $dungeonRoute): bool
...
+    private function hasFreshThumbnailForVariant(DungeonRoute $dungeonRoute, string $variant): bool
+    {
+        $latestRenderedAt = $dungeonRoute->dungeonRouteThumbnails()
+            ->where('custom', false)
+            ->where('variant', $variant)
+            ->max('updated_at');
+
+        return $latestRenderedAt !== null
+            && Carbon::parse($latestRenderedAt)->greaterThanOrEqualTo($dungeonRoute->updated_at);
+    }
```
Wotuu: I think this would be better served as a Repository function.

app/Service/DungeonRoute/ThumbnailService.php
```diff
@@ -523,11 +589,33 @@ private function getPreviewUrl(DungeonRoute $dungeonRoute, int $floorIndex, ?flo
...
+    /**
+     * The factor the killzone-path (pull-connection) line weight is multiplied by when rendering this
+     * variant, or null for no thickening. Custom API renders reuse the standard variant's multiplier;
+     * only the standard and hero variants define their own value.
+     */
+    private function getKillZonePathWeightMultiplier(string $variant): int|float|null
+    {
+        $settingsVariant = $variant === DungeonRouteThumbnail::VARIANT_CUSTOM
+            ? DungeonRouteThumbnail::VARIANT_STANDARD
+            : $variant;
```
Wotuu: Can't you just include config for `custom`? I understand `custom` won't have with width height etc, but it can have the `kill_zone_path_weight_multiplier`? Saves this confusing ternary/function from existing.

resources/views/common/dungeonroute/cardhero.blade.php
```diff
@@ -0,0 +1,189 @@
...
+                            ? sprintf('+%d', $dungeonroute->level_min)
+                            : sprintf('+%d – +%d', $dungeonroute->level_min, $dungeonroute->level_max) }}
+                    </span>
+                @endif
+                @include('common.dungeonroute.pullgraph', [
+                    'pullForces'  => $pullForces,
+                    'chartHeight' => 26,
+                    'fill'        => 'rgba(255, 255, 255, 0.6)',
+                    'bossFill'    => 'rgba(240, 180, 60, 0.95)',
+                    'graphClass'  => 'hero_pull_graph me-3',
```
Wotuu, line 158: I think these are great defaults that _can_ be overwritten. Not should be passed always.

resources/views/common/dungeonroute/pullgraph.blade.php
```diff
@@ -0,0 +1,66 @@
...
+    <?php
+    $barWidth = 3;
+    $barGap   = 1;
+    $minBar   = 2;
+    // Zero-forces pulls without a boss carry no information - drop them before capping so they
+    // neither render as noise nor eat into the 30-bar budget
+    $bars = $pullForces
+        ->filter(static fn(stdClass $pull) => $pull->enemy_forces > 0 || $pull->has_boss)
+        // Routes rarely exceed 30 pulls; cap the bar count to keep the graph tidy while the tooltip keeps the true total
+        ->take(30)
```
Wotuu: Can you pull this up to a `const` in the top of this file? It's pretty hidden in there right now.

Wotuu (commented):
A few more issues.

app/Models/DungeonRoute/DungeonRoute.php
Wotuu: :robot: Done — moved off the model into `DungeonRouteRepository::getEnemyForcesPerKillZone()` plus a new `DungeonRouteKillZoneService` (`app/Service/DungeonRoute/DungeonRouteKillZoneService.php`), returning a `KillZoneEnemyForces` DTO (camelCase `enemyForces`/`hasBoss`) instead of the previous snake_case stdClass rows.

app/Models/DungeonRoute/DungeonRouteThumbnail.php
Wotuu: :robot: Done — `VARIANT_STANDARD`/`VARIANT_HERO`/`VARIANT_CUSTOM` are now a native `DungeonRouteThumbnailVariant: string` backed enum (`Standard`/`Hero`/`Custom`), cast on the model via `casts()`.

app/Service/DungeonRoute/BaseDiscoverService.php
Wotuu: :robot: Done — both `App::make()` calls replaced with constructor injection; promoted all three params (`CacheServiceInterface`/`ExpansionServiceInterface` as `protected`, `DungeonRouteRepositoryInterface` as `private readonly`) so the constructor is consistent rather than mixing promoted and non-promoted params.

app/Service/DungeonRoute/ThumbnailService.php
Wotuu: :robot: Done — moved `hasFreshThumbnailForVariant()` off `ThumbnailService` into `DungeonRouteThumbnailRepository::hasFreshThumbnailForVariant()`.

app/Service/DungeonRoute/ThumbnailService.php
Wotuu: :robot: Done — the `custom` variant now has its own `kill_zone_path_weight_multiplier` entry in `config/keystoneguru.php`, so the ternary/`getKillZonePathWeightMultiplier()` helper is gone entirely — `getPreviewUrl()` just inlines a single `config(sprintf(...))` call now. One caveat: it is a static literal, not a live mirror of `standard`'s value like the old ternary was, so it needs to be kept in sync manually if `standard`'s multiplier ever changes — left a comment on the config entry noting that.

resources/views/common/dungeonroute/cardhero.blade.php
Wotuu: :robot: Done — `pullgraph.blade.php` now sets `??=` defaults for `chartHeight`/`fill`/`bossFill`, so `cardhero.blade.php` no longer passes them explicitly. `cardrow.blade.php` still overrides them since its values differ.

resources/views/common/dungeonroute/pullgraph.blade.php
Wotuu: :robot: Pulled the `30`-pull bar cap into a named `$maxBars` variable at the top of the file — not a PHP `const` though: confirmed via a test that a top-level `const` in a Blade partial that's `@include`d more than once per request throws "Constant already defined" on the second include.

app/Repositories/Database/DungeonRoute/DungeonRouteRepository.php
```diff
@@ -211,6 +216,83 @@ public function findSimilarRoutes(DungeonRoute $dungeonRoute, int $limit = 5): C
...
+                     left join `dungeons` on `dungeons`.`id` = `dungeon_routes`.`dungeon_id`
+                     left join `kill_zones` on `kill_zones`.`dungeon_route_id` = `dungeon_routes`.`id`
+                     left join `kill_zone_enemies` on `kill_zone_enemies`.`kill_zone_id` = `kill_zones`.`id`
+                     left join `enemies` on coalesce(`enemies`.`mdt_npc_id`, `enemies`.`npc_id`) = `kill_zone_enemies`.`npc_id`
+                        AND `enemies`.`mdt_id` = `kill_zone_enemies`.`mdt_id`
+                        AND `enemies`.`mapping_version_id` = `dungeon_routes`.`mapping_version_id`
+                     left join `npcs` on `npcs`.`id` = `kill_zone_enemies`.`npc_id`
+                     left join `npc_enemy_forces` on `npcs`.`id` = `npc_enemy_forces`.`npc_id` AND `dungeon_routes`.`mapping_version_id` = `npc_enemy_forces`.`mapping_version_id`
+                     {$ifIsShroudedJoins}
+            where `dungeon_routes`.`id` = :id
```
Wotuu: :robot: **Correctness — this per-kill-zone aggregate diverges from `DungeonRoute::getEnemyForces()`, which the interface doc says it mirrors.** Two guards from the original query are missing here:

1. The `WHERE` clause lacks `AND enemies.mapping_version_id = dungeon_routes.mapping_version_id`. In `getEnemyForces()` that predicate turns the `enemies` LEFT JOIN into an effective INNER JOIN, dropping any `kill_zone_enemies` row that has no matching `enemies` row for the route's mapping version. Here such orphaned rows survive and contribute the `npc_enemy_forces.enemy_forces` fallback branch, inflating that pull's total.
2. `getEnemyForces()` divides the SUM by `COUNT(concat(kill_zone_enemies.npc_id, kill_zone_enemies.mdt_id))` and groups per-enemy to cancel join fan-out / duplicate enemies. This query groups only by `kill_zones.id/index` with no such normalization, so any fan-out (an enemy matching >1 `npc_enemy_forces` or `enemies` row) is counted N times.

Net effect: the sum of the per-pull "fingerprint" bars can exceed the route's real `getEnemyForces()` total, contradicting the documented "mirrors the enemy-forces accounting" contract. Please reconcile the WHERE filter and the fan-out normalization with `getEnemyForces()`.

resources/views/common/dungeonroute/cardrow.blade.php
```diff
@@ -0,0 +1,157 @@
...
+    // The favorites count is only present when the route was loaded with withCount('favorites'); never trigger a query
+    $favoritesCount = $dungeonroute->favorites_count ?? null;
+    // Enemy forces per pull, ordered by pull index - drives the "route fingerprint" bar graph. Resolved
+    // lazily here (not via top-level @inject) so a cache hit never pays the container-resolution cost.
+    $pullForces = app(DungeonRouteKillZoneServiceInterface::class)->getEnemyForcesPerKillZone($dungeonroute);
+    // The key-level chip is only meaningful when the route deviates from the season's catch-all range
+    $showLevel = $dungeonroute->level_min !== $dungeonroute->season?->key_level_min
+        || $dungeonroute->level_max !== $dungeonroute->season?->key_level_max;
+    // A route counts as "new" while it is within its first two weeks of being published
+    $isNew = $dungeonroute->published_at->greaterThan(Carbon::now()->subDays(14));
```
Wotuu, line 43: :robot: **Correctness — unguarded null deref can 500 the whole leaderboard.** `$dungeonroute->published_at->greaterThan(...)` dereferences `published_at` with no null-safe. Although `@property Carbon` is declared, `popularBuilder()` only filters on `published_state = WORLD` and `expires_at IS NULL` — it never requires `published_at IS NOT NULL`. A single WORLD route with a null `published_at` (legacy/imported/seeded rows, or a route promoted to WORLD without the timestamp being stamped) throws `Call to a member function greaterThan() on null` and 500s the entire reworked dungeon page. This is the only card template that dereferences `published_at` unguarded.

```suggestion
    $isNew = $dungeonroute->published_at?->greaterThan(Carbon::now()->subDays(14)) ?? false;
```

app/Jobs/ProcessRouteFloorThumbnail.php
```diff
@@ -55,13 +57,13 @@ public function handle(): void
                 // Give some additional space since we're refreshing ALL floors - the first floor may get processed,
                 // but the floors after that will otherwise think "oh the thumbnail is up-to-date" and not refresh.
                 if ($this->dungeonRoute->thumbnail_updated_at->isBefore($this->dungeonRoute->updated_at->addHour()) || $this->force) {
```
Wotuu, line 59: :robot: **Correctness — NPE on force-queued hero renders.** The left operand `$this->dungeonRoute->thumbnail_updated_at->isBefore(...)` is evaluated before the `|| $this->force` short-circuit. `thumbnail_updated_at` is null for any route that has never had a standard thumbnail generated. The new `EnsureHeroThumbnails` command force-queues hero renders (`force = true`) for the top community + weekly routes, some of which may not yet have a standard thumbnail — so `null->isBefore()` throws before `|| $this->force` can rescue the render, and the hero thumbnail is never produced. Reorder so `$this->force` is checked first (or null-guard `thumbnail_updated_at`):

```suggestion
                if ($this->force || $this->dungeonRoute->thumbnail_updated_at?->isBefore($this->dungeonRoute->updated_at->addHour())) {
```

app/Service/DungeonRoute/DevDiscoverService.php
```diff
@@ -134,6 +135,14 @@ public function popularByDungeon(Dungeon $dungeon): Collection
         return $this->popularBuilder()->get();
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    public function popularByDungeonPaginated(Dungeon $dungeon, int $perPage): LengthAwarePaginator
+    {
+        return $this->popularBuilder()->paginate($perPage);
```
Wotuu: :robot: **Correctness (dev environment) — the `$dungeon` argument is ignored.** The real `DiscoverService::popularByDungeonPaginated()` filters `->where('dungeon_routes.dungeon_id', $dungeon->id)` before `paginate()`, but this dev override drops that filter and returns globally-popular routes. In a local/dev environment (where `DevDiscoverService` is bound), the reworked `DungeonRouteListRework` leaderboard renders the same all-dungeon list on every dungeon's overview page. Add the missing `dungeon_id` filter to match the production implementation.

resources/views/common/dungeonroute/cardrow.blade.php
```diff
@@ -0,0 +1,157 @@
...
+    $enemyForcesWarning    = $dungeonroute->enemy_forces < $dungeonroute->mappingVersion->enemy_forces_required || $enemyForcesPercentage >= 105;
+    $backgroundUrl = $dungeonroute->has_thumbnail
+        ? $dungeonroute->thumbnails->first()->getURL()
+        : $dungeonroute->dungeon->getImageTransparentUrl();
+    $ratingCount = $dungeonroute->rating_count;
+    // The favorites count is only present when the route was loaded with withCount('favorites'); never trigger a query
+    $favoritesCount = $dungeonroute->favorites_count ?? null;
+    // Enemy forces per pull, ordered by pull index - drives the "route fingerprint" bar graph. Resolved
+    // lazily here (not via top-level @inject) so a cache hit never pays the container-resolution cost.
+    $pullForces = app(DungeonRouteKillZoneServiceInterface::class)->getEnemyForcesPerKillZone($dungeonroute);
```
Wotuu, line 38: :robot: **Efficiency — per-card N+1.** `getEnemyForcesPerKillZone()` issues one raw `DB::select` (a multi-join aggregate over `dungeon_routes`/`kill_zones`/`enemies`/`npcs`/`npc_enemy_forces`) per rendered card, and `popularBuilder()` does not eager-load `killZones`. Only the rendered HTML is cached, not the query result, so on a cold card cache one leaderboard page (up to 18 rows + up to 3 hero cards, same pattern in `cardhero.blade.php`) fires ~21 separate aggregate queries, scaling linearly with page size and repeated per cache variant (locale/admin). Consider batching the per-pull forces for the whole page in one grouped query, or deriving from an eager-loaded `killZones` relation.

tests/Feature/Console/Commands/Scheduler/Thumbnail/EnsureHeroThumbnailsTest.php
```diff
@@ -0,0 +1,25 @@
...
+{
+    #[Test]
+    public function handle_givenCurrentSeasonRoutes_runsSuccessfullyAndQueuesOnlyHeroJobs(): void
+    {
+        // Arrange - the seeded test DB has a current season with dungeons and popular routes; faking the
+        // queue keeps the command from actually rendering thumbnails
+        Queue::fake();
+
+        // Act & Assert - the command resolves its hero routes and completes without error
+        $this->artisan(EnsureHeroThumbnails::class)->assertSuccessful();
```
Wotuu: :robot: **Test coverage — the assertion doesn't match the test's promise.** `handle_givenCurrentSeasonRoutes_runsSuccessfullyAndQueuesOnlyHeroJobs` only asserts `->assertSuccessful()`. It never asserts that any job was pushed, nor that the pushed `ProcessRouteFloorThumbnail` jobs carry `variant = DungeonRouteThumbnailVariant::Hero`. If `heroRoutes()` returned empty, or the command queued standard-variant jobs by mistake, this test would still pass — the command's core behavior is unverified. Add a `Queue::assertPushed(ProcessRouteFloorThumbnail::class, fn ($job) => $job->variant === DungeonRouteThumbnailVariant::Hero)` (and assert the expected count), so the "queues only hero jobs" claim is actually pinned.

app/Repositories/Database/DungeonRoute/DungeonRouteThumbnailRepository.php
```diff
@@ -2,14 +2,28 @@
...
 {
     public function __construct()
     {
         parent::__construct(DungeonRouteThumbnail::class);
     }
+
+    public function hasFreshThumbnailForVariant(DungeonRoute $dungeonRoute, DungeonRouteThumbnailVariant $variant): bool
+    {
+        $latestRenderedAt = $dungeonRoute->dungeonRouteThumbnails()
+            ->where('custom', false)
```
Wotuu: :robot: **Correctness / expand-contract hazard.** Two issues on this freshness query:

1. `->where('custom', false)` is now redundant with `->where('variant', $variant)` (the `custom` boolean is dual-written as a function of `variant`), and it will actively break once the follow-up migration drops the `custom` column — this method still references it. Gate on `variant` alone. (It also means `hasFreshThumbnailForVariant(..., Custom)` can never return true: `custom=false AND variant='custom'` matches nothing.)
2. `max('updated_at')` is taken across *all* of a route's hero thumbnails. For a non-facade multi-floor route, `queueThumbnailRefresh` dispatches one job per floor; if one floor's hero render succeeds recently but another floor's is missing/failed, `max()` is still `>= route.updated_at`, so the hourly `EnsureHeroThumbnails` treats the route as fresh and never re-queues the missing floor — it stays absent until the route is edited again.

database/migrations/2026_07_14_000000_add_variant_to_dungeon_route_thumbnails_table.php
```diff
@@ -0,0 +1,29 @@
...
+return new class extends Migration {
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::table('dungeon_route_thumbnails', function (Blueprint $table) {
+            // Distinguishes the standard (small) thumbnail from the larger hero-band variant.
+            // Nullable-safe default keeps every pre-existing row a 'standard' thumbnail.
+            $table->string('variant', 16)->default('standard')->after('custom')->index();
```
Wotuu: :robot: **Migration — combining ADD COLUMN and ADD INDEX in one ALTER forfeits `ALGORITHM=INSTANT`.** Adding a defaulted column on its own is an instant metadata-only change in MySQL 8, but adding an index in the same statement forces the whole `ALTER` to run `INPLACE`, rebuilding `dungeon_route_thumbnails` (one row per route/floor — potentially large). Beyond the I/O cost, the longer-running ALTER widens the window during the rolling deploy where new code (which now reads `variant` on the render-every-card hot path) coexists with the pre-migration schema. Splitting into two statements — instant `ADD COLUMN`, then a separate online `ADD INDEX` — keeps the column add instant and shrinks that window.

app/Helpers/CustomHelper.php
```diff
@@ -167,3 +167,19 @@ function array_insert_after(array $array, string $afterKey, array $insert): arra
...
+ * Formats a number in a compact human-readable form (6800 -> 6.8K) without requiring ext-intl.
+ */
+function abbreviateNumber(int $number): string
+{
+    if ($number >= 1000000) {
+        return sprintf('%sM', rtrim(rtrim(number_format($number / 1000000, 1), '0'), '.'));
+    }
+
+    if ($number >= 1000) {
+        return sprintf('%sK', rtrim(rtrim(number_format($number / 1000, 1), '0'), '.'));
```
Wotuu: :robot: **Minor correctness — stray thousands separator at the bucket boundary.** `number_format($number / 1000, 1)` uses the default `,` thousands separator, so any value that rounds up to `>= 1000` within its bucket renders a comma: e.g. `999_950` → `number_format(999.95, 1)` = `"1,000.0"` → after `rtrim` → `"1,000"` → output `"1,000K"` instead of `"1M"` (same for the millions boundary). Pass an empty thousands separator: `number_format($number / 1000, 1, '.', '')`.

resources/views/common/dungeonroute/cardposter.blade.php
```diff
@@ -0,0 +1,160 @@
...
+                </div>
+                <div class="poster_social small text-muted">
+                    @if( $ratingCount > 0 )
+                        <span class="poster_rating">
+                            @include('common.dungeonroute.rating', ['count' => $ratingCount, 'rating' => (int) round($dungeonroute->rating)])
+                        </span>
+                    @endif
+                    <span class="poster_views ms-1" data-bs-toggle="tooltip"
+                          title="{{ sprintf(__('view_common.dungeonroute.poster.views'), $dungeonroute->views) }}">
+                        <i class="fas fa-eye"></i> {{ $dungeonroute->views }}
```
Wotuu: :robot: **Consistency — same metric formatted two ways across the card family.** `cardposter` prints the raw `{{ $dungeonroute->views }}` (and raw `$favoritesCount`), while the sibling `cardhero`/`cardrow` cards added in this same PR render the identical fields through `abbreviateNumber()`. A high-view route will show `40000` on a poster card but `40K` on a hero/row card of the same route on the same discovery surface, and the long raw number can overflow the compact poster footer. Pick one formatting for the view/favorite counts across the card family.

Wotuu:
:robot: Cold review (fable): 10 findings posted.

app/Repositories/Database/DungeonRoute/DungeonRouteRepository.php
Wotuu: :robot: Done — restructured the query as a nested aggregation mirroring `getEnemyForces()`' exact accounting: an inner subquery groups by `(kill_zone, enemy identity)` to apply the `/ COUNT(...)` fan-out normalization per identity, and the outer query sums those normalized per-identity values per kill zone. Also added the missing `enemies.mapping_version_id = dungeon_routes.mapping_version_id` WHERE filter so orphaned `kill_zone_enemies` rows (no matching `enemies` row for this mapping version) are excluded instead of leaking their `npc_enemy_forces` fallback into the sum. Verified against the existing 2 tests (both pass) plus manual SQL/paren-balance verification. Note: I did not add a data-driven regression test for the specific fan-out/orphaned-row scenarios — there's no existing factory for `Enemy`/`Npc`/`NpcEnemyForce`, and the sibling `getEnemyForces()` this mirrors has no test either, so building one from scratch felt like a separate, larger follow-up rather than something to rush here. Flagging that gap rather than hiding it.

resources/views/common/dungeonroute/cardrow.blade.php
Wotuu: :robot: Investigated and reverted — `published_at` is `NOT NULL` with an epoch-sentinel default (`1970-01-01 00:00:00`) in the live schema (confirmed via `SHOW COLUMNS`), and the model's `@property Carbon $published_at` docblock agrees. It can never actually be null, so the suggested `?->` introduced a real PHPStan error (`nullsafe.neverNull`) rather than fixing a real bug. Left the original code as-is.

app/Jobs/ProcessRouteFloorThumbnail.php
Wotuu: :robot: Investigated and reverted — same as the `published_at` finding below: `thumbnail_updated_at` is `NOT NULL` with an epoch-sentinel default in the live schema (confirmed via `SHOW COLUMNS`) and the model's docblock agrees. It can never be null, so `?->` was a real PHPStan error (`nullsafe.neverNull`), not a fix. Left the original code (including operand order) as-is.

app/Service/DungeonRoute/DevDiscoverService.php
Wotuu: :robot: Done — added the missing `->where("dungeon_routes.dungeon_id", $dungeon->id)` filter to `popularByDungeonPaginated()`, matching the production `DiscoverService` implementation. Left the two pre-existing sibling methods (`popularByDungeon`, `popularByDungeonAndAffixGroup`) alone since they're unchanged by this PR and have the same gap — flagging as a separate pre-existing issue rather than expanding this fix's scope.

resources/views/common/dungeonroute/cardrow.blade.php
Wotuu: :robot: Acknowledged, deferring rather than fixing here — batching per-pull forces across a whole leaderboard page requires restructuring how routes flow from the controller/service layer into these cards (today each card independently resolves its own forces), and interacts with the "only rendered HTML is cached, not the query result" caching note in this same file. That's a bigger, cache-sensitive change I don't want to rush into an already-large fix batch. Filing this as a good candidate for its own follow-up rather than attempting it alongside the other 9 findings in this pass.

tests/Feature/Console/Commands/Scheduler/Thumbnail/EnsureHeroThumbnailsTest.php
Wotuu: :robot: Done — the test now asserts `Queue::assertPushed(ProcessRouteFloorThumbnail::class)` (at least one job queued) and, via a bound closure reading the protected `variant` property (matching the existing pattern in `RepairBrokenThumbnailsTest::isThumbnailJobForRoute`), that every queued job is Hero-variant and none is Standard-variant.

app/Repositories/Database/DungeonRoute/DungeonRouteThumbnailRepository.php
Wotuu: :robot: Partially done. (1) Fixed — dropped the redundant `where("custom", false)`; `variant` is now the sole gate (confirmed via the backfill migration's own doc comment that `variant` is the source of truth), so `Custom`-variant freshness checks work again. (2) Partially fixed — switched `max("updated_at")` to `min("updated_at")`, so a stale-but-present row on one floor no longer gets masked by a fresh row on another (added a test for this). This does NOT catch the fully-missing-floor case (a floor with zero thumbnail rows for the variant) — that needs comparing the row count against the expected floor count from `Dungeon::floorsForMapFacade()`, which I attempted but reverted: the existing single-thumbnail test fixtures use `getDungeonWithNonFacadeFloor()`, and I confirmed several seeded dungeons in that pool have multiple non-facade floors (Dire Maul East: 2, Karazhan: 17), so an expected-floor-count check would need those fixtures reworked to seed a thumbnail per floor - a larger test-fixture change I didn't want to rush. Added 2 new tests (Custom-variant freshness, stale-vs-fresh multi-floor) alongside the existing 3.

database/migrations/2026_07_14_000000_add_variant_to_dungeon_route_thumbnails_table.php
Wotuu: :robot: Done — split into two `Schema::table()` calls (two ALTERs): the defaulted `ADD COLUMN` stays on its own (instant, metadata-only), and `ADD INDEX` runs as a separate statement. Not able to verify this against a live migrate-down/migrate-up round-trip in this worktree — this migration's batch already has 2 later, unrelated migrations from master stacked on top of it in the shared dev DB, so a targeted rollback would also roll back those. Verified via `php -l` and by reasoning through the Schema Builder semantics instead; the change has no custom logic, just splitting one chained call into two sequential ones.

app/Helpers/CustomHelper.php
Wotuu: :robot: Done — added an empty thousands separator (`number_format($number / 1000, 1, ".", "")`) to both the K and M branches (the same root cause applies to the M branch at the billions boundary, just less likely to occur). Note the fix removes the stray comma but the boundary-rounding itself still displays e.g. `999_950` as `"1000K"` rather than `"1M"` - that's a separate quirk (the branch threshold check happens on the raw number before rounding), out of scope for this comma fix. Added a `CustomHelperTest` with a data provider covering this exact boundary case (verified the real output via tinker first, not by assumption).

resources/views/common/dungeonroute/cardposter.blade.php
Wotuu: :robot: Done — both `views` and `favoritesCount` on the poster card now render through `abbreviateNumber()`, matching `cardhero`/`cardrow`.

app/Repositories/Database/DungeonRoute/DungeonRouteThumbnailRepository.php
```diff
@@ -2,14 +2,32 @@
...
         parent::__construct(DungeonRouteThumbnail::class);
     }
+
+    public function hasFreshThumbnailForVariant(DungeonRoute $dungeonRoute, DungeonRouteThumbnailVariant $variant): bool
+    {
+        // Gate on the OLDEST render among this route's thumbnails for the variant, not the newest: for a
+        // multi-floor route, one freshly-rendered floor must not mask another floor whose thumbnail row
+        // still exists but is stale (e.g. a failed re-render left the old row in place). Note this still
+        // can't detect a floor whose thumbnail row doesn't exist at all (a wholly failed render) - that
+        // requires knowing the expected floor count, which is a larger follow-up.
```
Wotuu: Is this follow-up documented somewhere? Follow-up = new ticket, otherwise it will be forgotten and never done.

resources/views/common/dungeonroute/cardrow.blade.php
Wotuu: Ok, file the follow-up or it will be forgotten.

Wotuu (commented):
Some follow ups required

Wotuu:
:robot: Follow-up issues filed for the two deferred findings on this PR:
- #3650 (batch per-pull enemy-forces queries / N+1 on leaderboard cards)
- #3651 (EnsureHeroThumbnails missing-floor freshness gap)

(Noticed there's an open pending review on this PR — leaving this as a top-level comment rather than inline replies so as not to touch it.)

app/Jobs/ProcessRouteFloorThumbnail.php
Wotuu: :robot: Heads up — this got reintroduced by the "Apply suggestions from code review" batch commit (95c79bfef) and broke PHPStan again (nullsafe.neverNull, same reason as before: `thumbnail_updated_at` is `NOT NULL` with a sentinel default). Reverted back to the original operand order in a9899e53d.

resources/views/common/dungeonroute/cardrow.blade.php
Wotuu: :robot: Heads up — this got reintroduced by the "Apply suggestions from code review" batch commit (95c79bfef) and broke PHPStan again (nullsafe.neverNull, same reason as before: `published_at` is `NOT NULL` with a sentinel default). Reverted back in a9899e53d.

app/Repositories/Database/DungeonRoute/DungeonRouteThumbnailRepository.php
Wotuu: :robot: Yes — filed as #3651 (`EnsureHeroThumbnails never re-queues a fully-missing floor's hero thumbnail`), and it's already been fixed and merged: #3654 (commit ebb44d2fd) adds the expected-floor-count check via `Dungeon::floorsForMapFacade()`, and this branch has since pulled that commit in.

Wotuu (commented):
Ok

PR #3472, opened 2026-07-09
#3445 Make discord log channel a safe no-op when webhook is unset

### LINKED ISSUE #3445: scheduler log stack throws when APP_LOG_DISCORD_WEBHOOK is empty (swallowed as 404 in arrow delete)
## Background

Discovered while landing the Rector change set (#3434 / PR #3440). The Rector PR is **not** the cause — this reproduces on clean `development` and is unrelated to those changes.

`tests/Feature/Controller/Ajax/AjaxArrowControllerTest::delete_givenExistingArrow_shouldDeleteArrow` fails with **404 instead of 204**. The 404 is a red herring — a real exception is being swallowed.

## Root cause

The `scheduler` log channel in `config/logging.php` is a `stack` that **unconditionally** includes the `discord` channel:

```php
'scheduler' => [
    'driver'   => 'stack',
    'channels' => ['scheduler_file', 'discord'],
],
```

But `discord` is defined conditionally:

```php
'discord' => empty(env('APP_LOG_DISCORD_WEBHOOK')) ? [] : [
    'driver' => 'custom',
    'url'    => env('APP_LOG_DISCORD_WEBHOOK'),
    'via'    => MarvinLabs\DiscordLogger\Logger::class,
    'level'  => 'error',
],
```

When `APP_LOG_DISCORD_WEBHOOK` is empty (the dev/test default — it's empty in `.env` on both machines), `discord` resolves to `[]`. As soon as anything logs to the `scheduler` channel, Laravel's `LogManager` tries to read `$config['driver']` on that empty array and throws:

```
ErrorException: Undefined array key "driver" at vendor/laravel/framework/src/Illuminate/Log/LogManager.php:225
```

## How it surfaces as a 404

1. Arrow delete → `AjaxArrowController::delete()` → `dungeonRouteChanged()` dispatches the `RefreshEnemyForces` job, which runs synchronously under `QUEUE_CONNECTION=sync` in tests.
2. `App\Jobs\RefreshEnemyForces::handle()` (line 35) calls `Log::channel('scheduler')->info(...)`.
3. That resolves the broken `discord` channel → `Undefined array key "driver"`.
4. `AjaxArrowController::delete()` wraps its body in `catch (Exception) { return response(..., Http::NOT_FOUND); }`, so the real error is swallowed and the client gets a 404.

## Why we haven't hit it before

- In production `APP_LOG_DISCORD_WEBHOOK` is set, so `discord` is a valid channel and never throws.
- It only fires in dev/test where the webhook is empty **and** a code path logs to `scheduler` synchronously (the arrow-delete → `RefreshEnemyForces` path does).

## Remaining work

Make the `scheduler` stack tolerate a disabled discord channel. Options:
- Only include `'discord'` in the `scheduler` stack's `channels` when `APP_LOG_DISCORD_WEBHOOK` is set, or
- Resolve `discord` to a valid `'null'`/monolog channel (`['driver' => 'monolog', 'handler' => NullHandler::class]`) instead of `[]` when the webhook is empty.

Secondary (optional): `AjaxArrowController::delete()`'s blanket `catch (Exception) => 404` masks unrelated failures — consider narrowing it or at least logging the swallowed exception so this class of problem isn't invisible.

## Key files

- `config/logging.php` — `scheduler` stack + conditional `discord` channel
- `app/Jobs/RefreshEnemyForces.php:35` — `Log::channel('scheduler')`
- `app/Http/Controllers/Ajax/AjaxArrowController.php` — `delete()` `catch (Exception)` → 404

## How to verify

- With `APP_LOG_DISCORD_WEBHOOK=` empty: `docker compose exec -T app php artisan test --compact tests/Feature/Controller/Ajax/AjaxArrowControllerTest.php` — the `delete_givenExistingArrow` test should pass (204) after the fix.
- Confirm `scheduler`-channel logging no longer throws when the webhook is unset.

## Definition of done

- `AjaxArrowControllerTest` fully green with an empty discord webhook.
- Logging to the `scheduler` channel is a no-op (not an exception) when `APP_LOG_DISCORD_WEBHOOK` is empty.

Wotuu:
Closes #3445

## Problem

The `scheduler` log stack unconditionally includes the `discord` channel, but `discord` resolved to an empty array (`[]`) whenever `APP_LOG_DISCORD_WEBHOOK` is empty — the local/testing default. Building the stack then read `$config['driver']` on that empty array, emitting `Undefined array key "driver"` warnings.

## Fix

`discord` now resolves to a valid monolog `NullHandler` channel when no webhook is configured, so any stack that includes it (like `scheduler`) builds cleanly and discord logging becomes a genuine no-op. When a webhook is set, behaviour is unchanged.

## Note on the reported 404

The literal 404 in the issue **no longer reproduces on Laravel 12**: framework `LogManager::get()` now wraps channel resolution in `try/catch (Throwable)` and swallows the failure into the *emergency logger*, so the arrow-delete test was already green before this change — it does not discriminate the bug on v12.

The underlying defect is still real and matches the issue's Definition of Done, and this fix has a positive side effect: because the broken `discord` channel previously fell back to the emergency logger (a `StreamHandler` → `laravel.log`), **every** scheduler log line was leaking into `laravel.log` plus emitting an "Unable to create configured logger" line on every call. With the `NullHandler`, scheduler logs now go only to `scheduler.log`.

## Tests

New `tests/Unit/App/Logging/SchedulerLogChannelTest.php`:
- `discordChannel_givenEmptyWebhook_resolvesToValidChannel` — the resolved `discord` channel config has a `driver` key.
- `schedulerChannel_givenEmptyWebhook_logsWithoutEmittingPhpErrors` — logging to the `scheduler` channel emits no PHP warnings.

Both were verified red before the change and green after. `AjaxArrowControllerTest` and the full `Logging` group pass; `composer run analyse` remains at its pre-existing baseline and `composer run fix` is clean.

## Deferred (optional, follow-up)

The issue's secondary suggestion — narrowing `AjaxArrowController::delete()`'s blanket `catch (Exception) => 404` — was intentionally left out to keep this PR focused. Note `store()` has the identical pattern, so any change there should cover both.

### DIFF
--- a/config/logging.php
+++ b/config/logging.php
@@ -2,6 +2,7 @@
 
 use App\Logging\Handlers\ColoredLineFormatter;
 use App\Logging\Handlers\DeduplicateHandlers;
+use Monolog\Handler\NullHandler;
 use Monolog\Handler\StreamHandler;
 use Monolog\Processor\PsrLogMessageProcessor;
 
@@ -67,7 +68,13 @@ return [
             'stream'     => 'php://stderr',
         ],
 
-        'discord' => empty(env('APP_LOG_DISCORD_WEBHOOK')) ? [] : [
+        // When no webhook is configured (e.g. local/testing) discord must still resolve to a
+        // valid channel, otherwise any stack that includes it (like 'scheduler') fails to build
+        // and falls back to the emergency logger. A NullHandler makes discord logging a safe no-op.
+        'discord' => empty(env('APP_LOG_DISCORD_WEBHOOK')) ? [
+            'driver'  => 'monolog',
+            'handler' => NullHandler::class,
+        ] : [
             'driver' => 'custom',
             'url'    => env('APP_LOG_DISCORD_WEBHOOK'),
             'via'    => MarvinLabs\DiscordLogger\Logger::class,
--- a/tests/Unit/App/Logging/SchedulerLogChannelTest.php
+++ b/tests/Unit/App/Logging/SchedulerLogChannelTest.php
@@ -1,0 +1,60 @@
+<?php
+
+namespace Tests\Unit\App\Logging;
+
+use Illuminate\Support\Facades\Log;
+use PHPUnit\Framework\Attributes\Group;
+use PHPUnit\Framework\Attributes\Test;
+use Tests\TestCases\PublicTestCase;
+
+/**
+ * Regression coverage for #3445: when APP_LOG_DISCORD_WEBHOOK is empty (the local/testing default),
+ * the 'discord' channel must still resolve to a valid channel so the 'scheduler' stack that includes
+ * it can be built without emitting "Undefined array key \"driver\"" warnings or falling back to the
+ * emergency logger.
+ */
+#[Group('Logging')]
+final class SchedulerLogChannelTest extends PublicTestCase
+{
+    #[Test]
+    public function discordChannel_givenEmptyWebhook_resolvesToValidChannel(): void
+    {
+        // Arrange - a configured 'url' means a webhook is set, which is not the case this test guards
+        $discordConfig = config('logging.channels.discord');
+        if (array_key_exists('url', $discordConfig)) {
+            self::markTestSkipped('A discord webhook is configured; this test covers the empty-webhook default.');
+        }
+
+        // Act & Assert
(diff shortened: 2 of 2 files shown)


Wotuu (commented):
Ok

PR #3476, opened 2026-07-09
#3452 Gracefully recover from 419 CSRF token mismatches in ajax requests

### LINKED ISSUE #3452: Gracefully recover from 419 CSRF token mismatches in ajax requests (follow-up to #3391)
## Background

#3391 / MR #3392 re-enables CSRF validation (disabled since #1430 because of "page expired" friction). Two mitigations are already in place:

- The CSRF token is stable for the whole session, and MR #3392 raises the default session idle timeout to 7 days (`config/session.php`), so organic 419s should be near zero.
- `$.ajaxSetup` in `resources/assets/js/app.js` sends `X-CSRF-TOKEN` from the `<meta name="csrf-token">` tag on every ajax request.

The residual gap: a tab whose session has been rotated (re-login in another tab, or >7 days idle) still holds a stale meta-tag token, and its next ajax write fails with 419 until the user F5s. This is exactly the scenario described in #1430.

## Remaining work

Make 419 invisible to users, either of:

1. **Cookie-based token (preferred):** have the global ajax setup read the `XSRF-TOKEN` cookie (Laravel refreshes it on every response) and send it as `X-XSRF-TOKEN` instead of the static meta tag value. A long-open tab that has done *any* request then always has a current token. This is what axios does out of the box.
2. **Retry-on-419:** a global ajax error handler that, on 419, fetches a fresh token (any GET returns a new `XSRF-TOKEN` cookie), updates the header, and retries the original request once.

## Verification

- Load a page, invalidate the session server-side (e.g. log out in a second tab), fire an ajax POST from the first tab: it should succeed (option 1) or transparently retry (option 2) instead of surfacing a 419.
- `TokenMismatchException` stays in `Handler::$dontReport` — this is UX recovery, not alerting.

Refs #1430, follow-up to #3391.

Wotuu:
Closes #3452

Follow-up to #3391. Makes a stale CSRF token invisible to users by switching ajax requests to Laravel's cookie-based token (issue's **preferred option 1**).

## What changed
- **`resources/assets/js/csrf.js`** (new) — `getCsrfToken()` reads the always-fresh, encrypted `XSRF-TOKEN` cookie (via js-cookie, which url-decodes it exactly as axios does) and returns it as `X-XSRF-TOKEN`. Falls back to the `<meta name="csrf-token">` value as `X-CSRF-TOKEN` only when the cookie is absent.
- **`resources/assets/js/app.js`** — replaced the one-time `$.ajaxSetup({headers})` with a global `$(document).ajaxSend(...)` handler that re-reads the token on every request.
- **`resources/assets/js/csrf.test.js`** (new) — 5 vitest cases covering cookie/meta/empty/null branches.

## Three non-obvious design points
1. **Global `ajaxSend`, not `ajaxSetup.beforeSend`** — ~10 callers already define their own per-request `beforeSend`, which would clobber an `ajaxSetup` one. `ajaxSend` fires additionally and can't be overridden. (Confirmed no writes use `global: false`, which would bypass it.)
2. **Removed the static `X-CSRF-TOKEN`** — `VerifyCsrfToken::getTokenFromRequest()` checks `X-CSRF-TOKEN` first and *ignores* `X-XSRF-TOKEN` when present, so a stale meta header would defeat the fix. The two are now mutually exclusive.
3. **`crossDomain` guard** — the token is now sent same-origin only (the old code sent it to every host).

## Security
Standard Laravel/axios double-submit-cookie pattern, verified safe:
- The `XSRF-TOKEN` cookie is **encrypted** (not in any `encryptCookies(except:)` list), so it can't be forged without `APP_KEY` nor read cross-origin (same-origin policy).
- Defense is layered: `SameSite=lax` + POST-only writes already block the session cookie cross-site; the token is the second layer.
- XSS still defeats CSRF, but that is unchanged — a script could already read the meta token today. No regression.

## Verification
- Unit tests pass (5).
- **Real round-trip against the running app:** control `POST /login` with no token → **419**; the same POST with `X-XSRF-TOKEN` set to the url-decoded cookie value (exactly what js-cookie produces) → **302**. Laravel's `decrypt()` accepts it.

## Scope
Covers the common re-login-in-another-tab case elegantly (session + XSRF cookies are browser-global, so a long-open tab reads the fresh cookie with no prior request). It does **not** cover the >7-day idle case on the very first request (one 419, F5 fixes) — but that user is logged out anyway. Per the issue's "either of / preferred: option 1", this is a conscious call; option 2 (retry-on-419) can close that last case if desired.

### DIFF
--- a/resources/assets/js/app.js
+++ b/resources/assets/js/app.js
@@ -6,6 +6,8 @@
 
 require('./bootstrap');
 
+const {getCsrfToken} = require('./csrf');
+
 /**
  * Next, we will create a fresh Vue application instance and attach it to
  * the page. Then, you may begin adding components to this application
@@ -18,9 +20,19 @@ require('./bootstrap');
 //     el: '#app'
 // });
 
-// Prepare ajax to always use the CSRF token
-$.ajaxSetup({
-    headers: {
-        'X-CSRF-TOKEN': $('meta[name="csrf-token"]').attr('content')
+// Attach a fresh CSRF token to every same-origin ajax request. We use the global `ajaxSend`
+// event rather than `$.ajaxSetup` because the token is re-read from the cookie on every request
+// (staying valid after the session is rotated in another tab) and because `ajaxSetup`'s
+// `headers`/`beforeSend` are clobbered by the per-request `beforeSend` handlers several callers
+// already define. See resources/assets/js/csrf.js and issue #3452.
+$(document).ajaxSend(function (event, jqXHR, settings) {
+    // Never leak the CSRF token to a third-party host.
+    if (settings.crossDomain) {
+        return;
+    }
+
+    const token = getCsrfToken();
+    if (token !== null) {
+        jqXHR.setRequestHeader(token.header, token.value);
     }
 });
--- a/resources/assets/js/csrf.js
+++ b/resources/assets/js/csrf.js
@@ -1,0 +1,42 @@
+/**
+ * CSRF token resolution for jQuery ajax requests.
+ *
+ * Laravel refreshes the encrypted `XSRF-TOKEN` cookie on every response, so reading it
+ * per-request always yields a token valid for the *current* session - even after the session
+ * was rotated in another tab (re-login) or expired. This replaces the static
+ * `<meta name="csrf-token">` value, which goes stale on long-open tabs and surfaced as 419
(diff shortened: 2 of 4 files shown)


Wotuu (commented):
Ok

Wotuu (commented):
Ok

PR #3482, opened 2026-07-09
#3453 Render wrong-HTTP-method requests as 405 instead of 500

### LINKED ISSUE #3453: Wrong-HTTP-method requests render as 500 instead of 405
## Background

Sending a GET to a DELETE-only route (e.g. `GET /ajax/profile/adfree/{key}`, or the two admin delete routes converted in MR #3392) returns **HTTP 500** with the custom error page, instead of **405 Method Not Allowed**. Reproduced on production (pre-#3392 code), so this is long-standing behavior of the error rendering, not a regression.

The page body even renders the 404-titled error view while the status line says 500 — the custom `errors` blade / `Handler::render` path apparently does not map `MethodNotAllowedHttpException` to its proper status.

## Remaining work

- Make `MethodNotAllowedHttpException` render as a 405 with a sensible error page (the generic error blade is probably fine; it's the status code and title that are wrong).
- Check whether other `HttpException` subclasses suffer the same mis-mapping in `app/Exceptions/Handler.php` / `resources/views/errors`.

## Verification

`curl -s -o /dev/null -w '%{http_code}' -X GET https://<env>/ajax/profile/adfree/xyz -H 'X-Requested-With: XMLHttpRequest'` should print 405.

Low priority — cosmetic/API-hygiene; no functional impact (the request is still rejected).

Wotuu:
Closes #3453

## Problem

A GET to a DELETE-only route (e.g. `GET /ajax/profile/adfree/{key}`, or the admin delete routes from #3392) returned **HTTP 500** with a 404-titled error page instead of **405 Method Not Allowed**.

Root cause: the `Route::fallback` closure rendered `errors.404` with a **hardcoded 500** status. Because the fallback is a GET/HEAD catch-all (`.*`), it also *swallowed every GET method mismatch* before the router could raise a 405 — so those requests never reached the exception handler as a 405 at all.

A secondary mis-mapping lived in `Handler::render`: on JSON/API requests, every `HttpException` other than `NotFoundHttpException`/`ThrottleRequestsException` was collapsed to a generic 500, losing the real status code.

## Changes

- **`routes/web.php`** — the fallback now inspects the route table. If the URI is served under other verbs it throws `MethodNotAllowedHttpException` (which advertises the allowed verbs via the `Allow` header); otherwise it `abort(404)`s. Aborting (instead of rendering a view inline) routes the response through `App\Exceptions\Handler`, while the `web` middleware keeps the session active so the rendered HTML error page still shows a logged-in nav.
- **`app/Exceptions/Handler.php`** — JSON/API responses now preserve the real HTTP status code and headers for any `HttpException` (405, 403, 400, …) instead of returning 500.
- Added the missing **`errors/405`** view + `view_errors.405` translations.

## Verification

```
$ curl -s -o /dev/null -w '%{http_code}' -X GET http://…/ajax/profile/adfree/xyz -H 'X-Requested-With: XMLHttpRequest'
405
```

| Request | Before | After |
|---|---|---|
| GET a DELETE-only route (browser) | 500, 404-titled page | **405** HTML page, `Allow` header |
| GET a DELETE-only route (ajax/JSON) | 500 | **405** JSON, `Allow` header |
| GET a genuinely unmatched URI | 500 | **404** |
| POST a GET-only route (JSON) | 500 | **405** JSON, `Allow: GET, HEAD` |

## Tests

Added `tests/Feature/Controller/ErrorResponseTest.php` covering the 405/404 HTML and JSON paths (both the fallback-thrown and router-thrown 405s). Existing Ajax/API/Site controller suites (144 tests) still pass.

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -16,8 +16,10 @@ use Illuminate\Validation\ValidationException;
 use MarvinLabs\DiscordLogger\Discord\Exceptions\MessageCouldNotBeSent;
 use Override;
 use Symfony\Component\HttpFoundation\Exception\BadRequestException;
+use Symfony\Component\HttpFoundation\Response as SymfonyResponse;
 use Symfony\Component\HttpKernel\Exception\AccessDeniedHttpException;
 use Symfony\Component\HttpKernel\Exception\HttpException;
+use Symfony\Component\HttpKernel\Exception\HttpExceptionInterface;
 use Symfony\Component\HttpKernel\Exception\MethodNotAllowedHttpException;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 use Symfony\Component\HttpKernel\Exception\TooManyRequestsHttpException;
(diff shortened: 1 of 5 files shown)


Wotuu (commented):
Ok

PR #3419, opened 2026-07-07
#3397 Upgrade Bootstrap from v4 to v5

Wotuu:
Closes #3397

Migrates the frontend from Bootstrap 4.6.2 to 5.3.8, one stage per commit:

1. **Dependencies** — bootstrap 5.3.8, @popperjs/core 2 (replaces popper.js v1), bootswatch 5.3.8 (was mismatched at 5.2.2 against the v4 core), bootstrap5-toggle (replaces bootstrap4-toggle, markup-compatible), bootstrap-select pinned to 1.14.0-beta3 (BS5 support, same jQuery API — follow-up issue tracks a maintained replacement). `window.bootstrap` is exposed for the concatenated custom-*.js scripts; load order vs bootstrap-select is documented in a comment. The orphaned superhero theme (absent from `User::THEME_ALL`) is removed.
2. **data attributes** — `data-toggle/target/dismiss/parent/backdrop/keyboard/placement/html/container` and popover `data-content` renamed to `data-bs-*` across blade, custom JS, and handlebars templates. Deliberately unprefixed: `data-toggle="toggle"` (bootstrap5-toggle), `data-toggle="navbar-shrink"` (custom hook), and `data-content` on `<option>` (read by bootstrap-select).
3. **jQuery plugin calls** — all `.modal()/.tooltip()/.popover()/.collapse()/.tab()/.carousel()/.button()` call sites converted to `bootstrap.*.getOrCreateInstance()`. The removed private `.tooltip('_fixTitle')` is replaced by dispose-and-recreate in `refreshTooltips`.
4. **bootstrap-4-autocomplete** — removed (hard-depends on the BS4 jQuery dropdown API); its single consumer (route table tag input) now uses a small in-repo dropdown type-ahead with the same behavior.
5. **Utility class sweep** — `ml/mr/pl/pr → ms/me/ps/pe`, `float/text/border/rounded-left|right → start|end`, `no-gutters → g-0`, `font-weight-* → fw-*`, `sr-only → visually-hidden`, `badge-pill → rounded-pill`, `dropright → dropend`, plus BS3-era `pull-*` leftovers.
[...]

PR #3484, opened 2026-07-09
#3420 Replace bootstrap-select 1.14.0-beta3 with Tom Select 2.6

### LINKED ISSUE #3420: Evaluate maintained replacement for bootstrap-select 1.14.0-beta3 (follow-up to #3397)
## Background

As part of the Bootstrap 4 → 5 migration (#3397, MR #3419), `bootstrap-select` was bumped from 1.13.18 to **1.14.0-beta3** — the only release with Bootstrap 5 support. It kept the migration small because the markup (`selectpicker` class, `data-live-search`, `data-content` HTML options) and the jQuery API (`$().selectpicker('refresh'/'render')`, `changed.bs.select`) are unchanged.

However, the package is effectively **unmaintained** (the beta is from 2022, no stable release since), so it should not be a long-term dependency.

## Remaining work

Evaluate and (if a good candidate exists) migrate to a maintained select/multiselect library. Candidates:

- **Tom Select** (https://tom-select.js.org) — actively maintained, jQuery-free, has a Bootstrap 5 theme, supports option templates (needed for the image/icon selects) and live search.
- **Choices.js** — maintained, lighter, but weaker HTML-option-template support.

## Scope / blast radius

- ~70 `selectpicker` occurrences across ~35 blade files + ~12 handlebars templates.
- Image/icon selects rely on `data-content` HTML on `<option>`: `common/forms/select/imageselect.blade.php`, `imageselectcategories.blade.php`, `common/npc/select.blade.php`, `admin/npc/edit.blade.php`, `layouts/sitepage.blade.php`, `resources/assets/js/handlebars/map_popup_type_select_template.handlebars`.
- Live search (`data-live-search`) used in ~18 places.
- JS integration: `resources/assets/js/custom/inline/layouts/app.js` (global init + `refreshSelectPickers()`), `changed.bs.select` binding in `searchhandlercombatlogrouteenemyfailures.js`.
- `resources/assets/sass/theme/theme.scss` has a large `.bootstrap-select` override block that would need a rewrite for the new library's classes.
- `resources/assets/js/bootstrap.js` load-order note: bootstrap-select requires `window.bootstrap` to be assigned before its `require()` — remove that constraint when the package goes.

## Local patches to remove with the migration

- `resources/assets/js/bootstrapselectrefreshfix.js` (applied from `resources/assets/js/bootstrap.js`): works around a 1.14.0-beta3 regression where `buildData()` appends to `selectpicker.main.data` without clearing it, so every `.selectpicker('refresh')` duplicated all options in the menu and in the selected-count label.
- `resources/assets/js/bootstrapselectrefreshfix.test.js`: covers the patch. Its first test asserts the upstream bug still exists, so it doubles as a canary — if a newer bootstrap-select fixes `refresh()`, that test fails to signal the patch can go. Delete both files together with the package.

## Definition of done

- bootstrap-select removed from `package.json`, `resources/assets/js/bootstrap.js`, and `resources/assets/sass/app.scss` (line importing `bootstrap-select/dist/css/bootstrap-select.min.css`).
- `resources/assets/js/bootstrapselectrefreshfix.js` and its test removed (see local patches above).
- All selects (plain, multiple, live-search, image/HTML options) work in all 3 themes.
- `npm run prod` clean; manual walk of admin edit forms, map sidebar selects, and the heatmap player-spells filter.

Wotuu:
Closes #3420

**Stacked on #3419** (`3397-bootstrap-5-migration`) — this MR only makes sense on top of the Bootstrap 5 migration and should be merged into that branch (or retargeted to `master` once #3419 lands).

## Evaluation

Tom Select over Choices.js, as the issue anticipated: actively maintained, jQuery-free, ships a Bootstrap 5 theme, and natively covers everything we used from bootstrap-select — live search, HTML option templates (`data-content`), optgroups, multi selects, disabled options/selects. Its DOM import even copies `data-*` attributes into option data, so the existing `data-content` markup works untouched.

## Approach: keep the markup convention, swap the engine

Blades/handlebars keep `select.selectpicker` + data attributes; a new adapter (`resources/assets/js/selectpicker.js`) interprets them:

- `data-live-search="true"` → `dropdown_input` plugin (search box inside the dropdown, same UX as before)
- `data-content` → HTML option/item rendering
- `title` / `data-none-selected-text` → placeholder
- `data-width` → explicit wrapper width
- `multiple` → tags UI with per-tag remove buttons (**UX change**: replaces the "N selected" count label of `data-selected-text-format`, which is now inert)

The global `refreshSelectPickers()` keeps its contract (init new pickers, sync existing ones after `$select.val()` / repopulation / `disabled` changes). It fingerprints the option set and only rebuilds when it actually changed — Tom Select moves selected `<option>`s to the end of the underlying select on every selection, so a blind re-import would scramble dropdown order (and this doubles as a perf win for the map's frequent refresh calls).

Covered by a 12-test vitest suite (`selectpicker.test.js`), including the no-duplication-on-refresh case that previously needed the local patch.

## Removed with the package

- `bootstrapselectrefreshfix.js` + its canary test (the beta3 `refresh()` duplication patch)
- the `window.bootstrap`-before-require load-order constraint in `bootstrap.js`
- the `shown.bs.select` layout hacks in the affix week/group selects + their `custom.css` workarounds (`span.text` wrappers don't exist anymore); the `readonly` → disable-options behavior is preserved
- jsdom back to master's `^29.1.1` (the `^26.1.0` pin existed only for bootstrap-select)
- `theme.scss` / `map.scss` `.bootstrap-select` overrides rewritten for `.ts-wrapper`/`.ts-control`/`.ts-dropdown`, using theme CSS vars; single selects get a `currentColor` caret so it stays visible in all themes

Also fixed en passant: the sitepage model-switcher options had no `value` attribute and empty text when showing an icon — they'd resolve to empty-value options and be treated as placeholders; they now carry the model key as value.

## Verified

- `npx vitest run`: 118/118 (including 12 new adapter tests), on jsdom 29
- `npm run prod` compiles clean
- PHPStan + PhpCsFixer clean; `tests/Feature/Controller/{DungeonRoute,Admin}` green
- Headless Chrome walk (worktree stack): `/new` dungeon select — dropdown opens, live search filters, **ArrowDown+Enter selects and closes the menu** (the beta3 regression from the issue comment is gone), underlying `<select>` synced; `/search` affix image selects render `data-content` HTML with correct sizing (no width-0 dead space), multi select produces removable tags and syncs values; route map page loads without JS errors; all of the above in darkly + lux + Xal'atath themes with no page errors. (Affix icon images 404 in the dev env — `localhost:8009` asset host — pre-existing, unrelated.)


Wotuu:
The select dropdown needs to show more rows. It shows like 6 now, it should be at least double.

Wotuu (commented):
Ok now!

PR #3485, opened 2026-07-09
#3414 Run local Docker app containers as host UID so they stop creating root-owned files in the bind mount

### LINKED ISSUE #3414: Run local Docker app containers as host UID so they stop creating root-owned files in the bind mount
## Background

On local dev (WSL2, host user UID 1000), every container that bind-mounts the repo (`./:/var/www`) runs its PHP process as **root**, so any file the app writes into the mount — `vendor/`, `node_modules/`, `storage/**`, `bootstrap/cache/**`, built assets in `public/js/**` — is created as `root:root` on the host. Because WSL2 preserves the container UID on the bind mount, the host user can't cleanly manage those files.

Concrete symptoms this causes:
- `git status` intermittently fails with `warning: could not open directory 'storage/framework/testing/disks/': Permission denied` — git tries to scan a root-owned `0700` runtime dir the test suite created. (Partially mitigated already by adding the missing `storage/framework/testing/.gitignore`, matching the other `storage/framework/*` dirs, so git no longer descends into it. That's a band-aid, not the root cause.)
- `git pull` / `git reset` abort mid-operation with `unable to unlink old '...'/unable to create file '...': Permission denied` on root-owned tracked placeholder files (e.g. `docker-compose/mysql/.gitkeep`, `storage/app/private/.gitignore`, `storage/logs/.gitignore`), leaving the tree half-updated.
- Recurring manual `chown` from inside the container (`docker compose exec -u 0 app chown -R 1000:1000 ...`) needed to undo the drift.

Root cause: the containers run as root. Fix: run the app process as the host UID (1000) so writes are owned by the host user.

## Remaining work

Make the local dev containers run as UID/GID 1000 (or a configurable `PUID`/`PGID`) instead of root, for every service that bind-mounts the repo:

- `app` (docker-compose.yml ~L2)
- `app-test` (~L21)
- `app-swoole` / Octane (~L49, `octane:start`)
- `cron` (~L71)
- `horizon` (~L90)
- `reverb` (~L110)

`docker-compose/app/Dockerfile` currently sets `USER root` (~L66). PHP-FPM pool and the `fixownership` cron (`docker-compose/cron/etc/cron.d/fixownership`, which `chown -R www-data:www-data /var/www/storage/logs` every minute — www-data = UID 33) also participate in the ownership story and must be reconciled, otherwise logs keep flipping to UID 33.

- **Not just one `user:` line.** Multiple services share the mount; all must agree on the UID or the drift just moves to whichever container still runs as root.
- **php-fpm master vs workers.** The FPM master may need root to bind/setuid; workers run as the pool user. The pool user (`docker-compose/app/usr/local/etc/php-fpm.d/`) must be the target UID.
- **The `fixownership` cron** actively chowns `storage/logs` to `www-data` (UID 33). Either retarget it to the same UID or drop it once the process no longer runs as root.
- **UID mismatch across machines.** This is machine 2 (host UID 1000). A hardcoded `1000` breaks on a host with a different UID — prefer build/runtime `PUID`/`PGID` args defaulting to 1000, wired through `.env`.
- **Image-baked files.** `vendor/` baked at image build time (composer install as root in the Dockerfile) will still be root-owned inside the image layer; only bind-mount writes are affected by the runtime user. A one-time host `chown` clears existing drift.
- **This is a local-dev concern only.** The AWS/production images (`docker-compose/app-aws`, `cron-aws`) already `chown www-data:www-data` and should be left as-is.

## Recommended approach

1. Add `PUID`/`PGID` (default `1000`) to `.env.example` / `.env.docker.example` and the compose env.
2. Give the repo-mounting services a `user: "${PUID}:${PGID}"` (or an entrypoint that `usermod`s the app user to match), one service at a time, testing after each.
3. Align the php-fpm pool user and remove/retarget the `fixownership` cron.
4. One-time host cleanup: `docker compose exec -u 0 app chown -R 1000:1000 /var/www/vendor /var/www/node_modules /var/www/storage /var/www/bootstrap/cache`.
5. Keep the `*-aws` production Dockerfiles unchanged.

## How to verify

After the change, from a clean state:
```
docker compose exec -T app id                       # expect uid=1000, not 0
docker compose exec -T app php artisan cache:clear   # writes to storage as 1000
find storage bootstrap/cache -not -user $(whoami)    # expect empty
git status                                            # no permission-denied warnings
```
Run the test suite (`docker compose exec -T app php artisan test --compact`) and confirm storage/cache writes still work and no container fails to boot.

## Key files

- `docker-compose.yml` (services: app, app-test, app-swoole, cron, horizon, reverb)
- `docker-compose/app/Dockerfile` (`USER root`)
- `docker-compose/app/usr/local/etc/php-fpm.d/` (pool user)
- `docker-compose/cron/etc/cron.d/fixownership`
- `.env.example`, `.env.docker.example` (new `PUID`/`PGID`)
- Leave alone: `docker-compose/app-aws/Dockerfile`, `docker-compose/cron-aws/*`

## Definition of done

- App, cron, horizon, swoole, reverb, and test containers boot and function with the app process running as UID 1000.
- No new root-owned files appear in the bind mount after normal use (artisan, tests, asset build).
- `git status` is free of permission-denied warnings without needing the `storage/framework/testing/.gitignore` band-aid to hide it.
- Works with a configurable UID (not hardcoded) so it doesn't break on a host with a different UID.
- Production `*-aws` images untouched.

Wotuu:
Closes #3414

## What changed

**Base app image** (`docker-compose/app/Dockerfile`): new `PUID`/`PGID` build args (default 1000) bake a `ksg` user whose home mirrors root's composer/bash setup. The image itself still ends as `USER root`, so the production images that build `FROM` it (app-aws, worker/worker-aws, cron) are byte-for-byte unaffected at runtime.

**Local compose** (`docker-compose.yml`): `app`, `app-swoole`, `horizon` and `reverb` now run with `user: "${PUID:-1000}:${PGID:-1000}"`, and `~/.ssh` mounts to `/home/ksg/.ssh`. php-fpm's master runs unprivileged (pool `user` directive is ignored with a benign NOTICE; workers inherit the UID). `cron` keeps a root daemon but the scheduler cron.d entry runs artisan as `ksg`, and the `fixownership` cron (chown storage/logs to www-data every minute) is deleted — nothing writes as www-data locally anymore.

**Worker image**: the puppeteer Chrome cache moves from `/root/.cache/puppeteer` (unreadable for non-root) to `PUPPETEER_CACHE_DIR=/opt/puppeteer-cache`, world-readable. The env persists into `app-worker-aws`, whose `npm install` resolves the same cache — verified Chrome 148 launches as UID 1000.

**Scheduler output** (`routes/console.php`): scheduled commands appended output to `/proc/1/fd/1`, which only root can open — as `ksg` every scheduled command failed with exit 2 (dash redirection failure). Now falls back to `/dev/stdout` when `/proc/1/fd/1` isn't writable; production cron (root, cron-aws) keeps the exact old behavior.

**Worktrees**: `docker-compose.worktree.yml` runs the app as the host UID too, and `sh/worktree.sh` no longer chowns `storage`/`bootstrap/cache` to www-data at init (unneeded and impossible from an unprivileged exec). The remove-time chown-back stays for worktrees created on older images.

**Incidental test fix**: `SiteControllerTest::index_givenAuthenticatedUser_returnsHomeLayout` was failing on master — `cardvertical.blade.php` read `Auth::user()->locale` directly (null for factory users → TypeError in `getCardCacheKey()`). It now uses `app()->getLocale()` like the horizontal card variants.

## Verification (worktree stack on the rebuilt image)

- `id` in app container → `uid=1000(ksg)`, HOME=/home/ksg; fpm boots with only the expected 'user directive ignored' NOTICEs
- Front page 200 through nginx; artisan cache:clear; `composer run fix`/`analyse` all as UID 1000
- `sudo cli_weakauras_parser` works via the new sudoers entries (MDT decode tests pass)
- Cron container: scheduler tick runs every command as `ksg` with 0 failures after the `/dev/stdout` fallback; storage/logs files created as 1000
- Reverb and octane boot cleanly as UID 1000
- Full test suite: 574 passed, 1 failed → the pre-existing SiteControllerTest failure fixed above (fails identically on master); file passes after the fix
- `find . ! -user $(whoami)` over the checkout after all of the above → empty
- PHPStan: same 35 pre-existing local-env errors as master (CI PHPStan green today on other branches)

## Rolling this out on a dev machine (not done automatically)

1. `docker compose build app app-swoole cron horizon` and `docker compose up -d` to recreate on the new images
2. One-time cleanup of existing drift: `docker compose exec -u 0 -T app chown -R 1000:1000 /var/www/storage /var/www/bootstrap/cache /var/www/vendor /var/www/node_modules`
3. Optionally set `PUID`/`PGID` in `.env` on hosts where `id -u` ≠ 1000


.env.example
```diff
@@ -25,6 +25,11 @@ FILESYSTEM_DISK=s3_user_uploads
 
 OCTANE_SERVER=swoole
 
+# Host UID/GID the local Docker app containers run as (and bake their `ksg` user with), so files
+# written into the bind mount stay owned by you instead of root. Match your `id -u` / `id -g`.
+PUID=1000
+PGID=1000
+
```
Wotuu: Is this required here? The idea is that the .env.example will be copied to .env on staging/production-like environments for init. Otherwise, for local dev, .env.docker.example should be used. This sounds like a local dev change?

Wotuu (commented):
1 comment

.env.example
Wotuu: Good catch — you're right, this is local-dev only. Removed the PUID/PGID block from `.env.example` in 670d3e679; it stays in `.env.docker.example` where local Docker dev picks it up.

Issue #3388, opened 2026-07-06
Reduce global $with on DungeonRoute, Npc and User (follow-up to #3302)


Wotuu:
Follow-up to #3302 (which handled `Dungeon` and `Season`). Those two closed the header-path levers; this issue covers the three remaining heavy global `$with` declarations that #3302 explicitly did not touch, all on hot paths.

## Background

`app/Providers/AppServiceProvider.php` enables `Model::preventLazyLoading(!isProduction())` **and** `Model::automaticallyEagerLoadRelationships()`. The second one is the key enabler here: on Laravel 12, relation access on a *collection* is auto-batched, so a global `$with` is no longer needed to avoid N+1 when iterating a collection — it only saves queries for genuinely-always-needed relations, and otherwise over-fetches on every query of the model, including single-row and count paths.

Three models still eager-load a large relation set globally:

- **`DungeonRoute`** (`app/Models/DungeonRoute/DungeonRoute.php:239`) — 9 relations: `mappingVersion`, `dungeon`, `season`, `faction`, `specializations`, `classes`, `races`, `affixes`, `thumbnails`. This is the model behind the route list / discover / profile pages, which paginate dozens of routes — every one drags all 9 relation loads whether the current view renders them or not.
- **`Npc`** (`app/Models/Npc/Npc.php:86`) — 6 relations: `type`, `class`, `npcbolsteringwhitelists`, `npcHealths`, `characteristics`, `spells`. Npcs are loaded via enemy relations on every map render and across the compendium, so this multiplies badly.
- **`User`** (`app/Models/User.php:155`) — 5 relations: `iconfile`, `patreonUserLink`, `dungeon`, `gameVersion`, `roles`. Loaded on nearly every authenticated request and wherever authors are listed.

## Remaining work

For each model, remove the global `$with` (or trim it to the relations that are *genuinely* needed on every single load — likely none for DungeonRoute) and add explicit `->with([...])` at the query sites that actually render those relations.

## Risks / why it's non-trivial

- These relations are read app-wide (Blade views, resources, services), so a naive removal will surface a lot of `preventLazyLoading` exceptions in dev. That is the safety net, not the enemy — work through them.
- `DungeonRoute::$season` / `->dungeon` are read in policies and services, not just views; check those paths.
- Do **one model per PR**, DungeonRoute last (biggest blast radius).

## Recommended approach

1. Pick one model. Remove/trim its `$with`.
[...]

PR #3486, opened 2026-07-09
#3388 Remove global $with from Npc, eager-load at consumers

Wotuu:
Part of #3388 (PR 1 of 3 — Npc; User and DungeonRoute follow in separate PRs).

## What

Removes the global `$with = [type, class, npcbolsteringwhitelists, npcHealths, characteristics, spells]` from `Npc` and adds explicit eager-loads at the query sites that actually serialize or render those relations. With `automaticallyEagerLoadRelationships()` enabled, collection-iteration paths need no changes — the work here is preserving **serialized payload shapes**, since unloaded relations silently vanish from JSON.

## Consumer changes

- **MapContextDungeonData** (`dungeonNpcs` payload — the canonical npc source for all map JS): explicitly loads `type`, `class`, `npcbolsteringwhitelists`, `npcHealths` alongside the existing `spells:id`. Payload key set verified identical to master; the nested `whitelistnpc` no longer drags its own 5 cascaded relations along (JS only reads its `name`).
- **MappingExportService::serializeNpcs**: adds `npcHealths` to the explicit loads (was implicit via `$with`); relation order chosen so `npcs.json` stays **byte-identical** (verified with `mapping:save` + git diff). Dead `without()`/`makeHidden(type/class)` removed.
- **NpcController::store**: the relation reload now happens *before* the `NpcChangedEvent` broadcasts (the front-end re-assigns `enemy.npc` from that payload and reads type/class/npc_healths/whitelists in visuals/tooltips), and includes the full set.
- **AjaxEnemyController::store**: response npc keeps the relation set the admin map re-assigns on `save:success`.
- **NpcCompendiumController**: `get()` loads `spells` (the datatable renders that column from the serialized relation); `show()` loads what the view reads.
- **MDTMappingExportService**: explicit `type/characteristics/spells/npcHealths` for the MDT mapping export.
- **MDTDungeon / SpellCompendiumController**: removed now-dead `makeHidden`/`without` of npc relations.

## Incidental fix

`tests/Feature/Controller/SiteControllerTest` failed on master: `cardvertical.blade.php` passed a null authenticated-user locale into `DungeonRoute::getCardCacheKey()`. It now uses `app()->getLocale()`, consistent with `cardhorizontal` / `cardhorizontalrow`.

## Measured

- `Npc::all()` (5953 rows): **13 queries → 1** — and every `Enemy` load (`Enemy::$with = ['npc']`, i.e. every map render/combat-log path) stops cascading the 6 npc relation loads.
- `dungeonNpcs` map-context build: 9 → 8 queries, and the cached payload shrinks (no more nested whitelistnpc relation trees).

## Verified

- Full test suite green in the worktree stack.
- `mapping:save` → `npcs.json` byte-identical.
- Headless Chrome: route map page (96 enemies rendered, npc payload complete, 0 JS errors), NPC compendium table (spells column renders), NPC detail page (health/characteristics/spells sections) — all 200, no page errors.
- `composer run analyse`: same 35 pre-existing errors as master (local env skew), none in changed files.

### DIFF
--- a/app/Http/Controllers/Ajax/AjaxEnemyController.php
+++ b/app/Http/Controllers/Ajax/AjaxEnemyController.php
@@ -81,6 +81,12 @@ class AjaxEnemyController extends AjaxMappingModelBaseController
             $enemy->load([
                 'npc',
                 'npc.enemyForces',
+                // The admin map re-assigns the enemy's npc from this response - include what the visuals/tooltips read
+                'npc.type',
+                'npc.class',
+                'npc.npcbolsteringwhitelists',
+                'npc.npcHealths',
+                'npc.spells',
                 'floor',
             ])->makeHidden(['floor']);
             // Perform floor change and move enemy to the correct location on the new floor
--- a/app/Http/Controllers/Compendium/NpcCompendiumController.php
+++ b/app/Http/Controllers/Compendium/NpcCompendiumController.php
@@ -36,7 +36,7 @@ class NpcCompendiumController extends Controller
             return redirect(route('npc.compendium.show', $npc), 301);
         }
 
-        $npc->load(['classification', 'dungeons.expansion', 'npcSpells']);
+        $npc->load(['classification', 'type', 'dungeons.expansion', 'npcSpells', 'npcHealths', 'characteristics', 'spells']);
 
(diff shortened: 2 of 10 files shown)


Wotuu (commented):
Ok! Checked visually as well.

Issue #3492, opened 2026-07-10
Build migration stage 2: bundle app.js with esbuild, compile sass directly, delete laravel-mix/webpack, add MR CI build job (part of #3449)


Wotuu:
:robot: Part of #3449. Depends on the stage 1 issue (extraction of handlebars/lang/concat/copies to `scripts/build/`).

## Background

Stage 2 of the laravel-mix → esbuild migration: replace the last two things mix still does after stage 1 — bundling `resources/assets/js/app.js` and compiling the 4 sass entries — then **delete laravel-mix and webpack entirely**, and add a build job to MR CI so bundler breakage can never again hide behind green checks (the original trigger for #3449).

Core constraint: `resources/assets/js/bootstrap.js` must remain **byte-unchanged**. Its in-place `require()` order is load-bearing (`window.$ = window.jQuery = require('jquery')` at line 9 must execute before the ~15 jQuery-plugin requires at line 88+). esbuild bundles mixed CJS/ESM with in-place execution order preserved and emits a plain script — this is exactly why esbuild was chosen over Vite/Rollup, whose ESM import hoisting breaks the plugin registration order (the failure that sank the Laravel Shift attempt, #2232).

## Remaining work

Create in `scripts/build/`:

- `app.mjs` — esbuild `build()`: entry `resources/assets/js/app.js`, `bundle: true`, plain-script output (IIFE) → `public/js/app-{v}.js`, alias `handlebars` → `handlebars/dist/handlebars.min.js` (same workaround webpack.mix.js has today), minify in prod, sourcemap in dev. No `define` needed — bootstrap.js references no `process.env`/`MIX_*`.
- `sass.mjs` — dart-sass JS API (`sass` is already a dependency) for the 4 entries (`app.scss`, `theme/theme.scss`, `custom/custom.scss`, `custom/assets/assets.scss`) → `public/css/{app,theme,custom-compiled,assets-compiled}-{v}.css`. Prepend `$asset-url: "${ASSETS_BASE_URL}";` (replaces sass-loader `additionalData`). **Custom importer required**: the scss uses webpack tilde imports (`@import "~bootstrap/scss/bootstrap"`, and ~15 plain-`.css` imports in `app.scss` like `~leaflet/dist/leaflet.css`) — map `~x` → `node_modules/x` and load `.css` files so they inline the way css-loader does today. `processCssUrls: false` parity is automatic (plain sass never rewrites urls). Minify with esbuild CSS transform in prod.
[...]

PR #3477, opened 2026-07-09
#3469 Remove old Mobirise front-page, FrontPageRework flag, and jarallax

Wotuu:
Closes #3469

The reworked front page has fully rolled out on production, so the `FrontPageRework` Pennant flag is now the only path. This makes the new `home.layout` the landing page for **everyone** (guests and logged-in users — the per-user route overview stays reachable at `/profile/routes`), removes all the now-dead old-page machinery, and drops `jarallax` (which was only used by the retired Mobirise page — declines Dependabot #3458).

## Changes
- **Controller**: `SiteController::index` now always renders `home.layout`; the flag / auth branches (and their unused imports + `$coverageService` param) are gone.
- **Feature flag**: deleted `app/Features/FrontPageRework.php` (the admin feature list is class-discovered, so it drops off automatically).
- **Old page**: deleted `resources/views/home.blade.php`, `HomeComposer` + its `KeystoneGuruServiceProvider` wiring, and `inline/home/home.js`.
- **jarallax**: removed from `bootstrap.js`, the `layouts/app.js` theme-switcher callsite, `custom/home.js`, `package.json` + `package-lock.json`.
- **SCSS**: deleted the `.theme .home`-scoped `home.scss` / `home/main.scss` / `home/additional.scss` and their `webpack.mix.js` JS + SASS entries.
- **Tests**: added `SiteControllerTest` (guest + authenticated `/` both render `home.layout`); removed the now-invalid `HomeComposer` case from `ViewComposerTest` and repointed the `GlobalComposer` case at `home.layout`.

The shared footer's `mbr-*` classes are unaffected — the deleted SCSS was scoped to `.theme .home` (the old page's rootClass), so it never styled the footer on any other page.

## Verification
- `composer run analyse` — no new errors (remaining ones are pre-existing in unrelated MDT/geometry files).
- `composer run fix` — clean (unrelated reformats discarded).
- `npm run dev` builds clean; the active bundles contain **0** `jarallax` / `mbr-parallax` references, the manifest has no `home-*.css`, old `HomeHome` inline gone, new `HomeLayout` present.
- Affected feature tests pass (43 in the view/controller group).
- HTTP smoke: `/` (new page, no `mbr-parallax` in HTML), `/dungeonroutes`, `/credits`, `/about` all `200`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #3449, opened 2026-07-07
Migrate build tooling off laravel-mix (webpack pinned to ~5.99.9) — likely to Vite (follow-up to #2225)


Wotuu:
## Background

While triaging the Dependabot npm security group (#3413, merged 2026-07-07), the proposed webpack bump 5.99.9 → 5.108.4 turned out to **silently break `npm run production`** — and CI can't catch this because the js-tests workflow only runs `npm ci` + vitest, never the webpack build.

laravel-mix 6 is incompatible with newer webpack 5.x in two independent ways:

1. **webpack ≥5.107 removed `webpack/lib/SizeFormatHelpers`**, which mix's `BuildOutputPlugin` requires → `MODULE_NOT_FOUND` at build start. (Bisected via npm tarballs: the module exists through 5.106.2, gone in 5.107.0.)
2. Even at 5.106.x, webpack's **strict `ProgressPlugin` option validation** rejects the options (`name`, `color`, `reporters`, `reporter`) passed by mix's progress bar dependency `webpackbar@5` → `Invalid options object` at startup.

laravel-mix is unmaintained — **6.0.49 is the final release** (confirmed via `npm view laravel-mix dist-tags`, 2026-07-07) — so no upstream fix will ever come.

## Current mitigation (already in place)

- `package.json` pins `"webpack": "~5.99.9"` so neither Dependabot nor `npm update` can float onto a broken version (added in #3413).
- `.github/dependabot.yml` has an `ignore` rule for `webpack` versions `>5.99` with an explanatory comment.
- Accepted residual: webpack's two open security alerts stay open (both **low** severity, both only affect `experiments.buildHttp` — a build-time HTTP module import feature this project does not use; fixes landed in 5.104 which we can't reach).

## Remaining work

Replace laravel-mix/webpack with a maintained bundler — Vite being the natural choice in the Laravel ecosystem (`laravel-vite-plugin`, `@vitejs` is already transitively present via vitest, and the JS test suite already runs on Vitest per #3292, which shares Vite's transform pipeline).

[...]

PR #3494, opened 2026-07-10
#3491 Extract handlebars/lang/concat/copies/versioning from webpack.mix.js into node build scripts

### LINKED ISSUE #3491: Build migration stage 1: extract handlebars/lang/concat/copies/versioning from webpack.mix.js into node build scripts (part of #3449)
:robot: Part of #3449.

## Background

#3449 migrates the build off laravel-mix/webpack. The chosen target is **esbuild + small node scripts** (not Vite — see the plan comment on #3449 for the full rationale). This is stage 1 of 3: move everything that isn't real bundling out of `webpack.mix.js` into standalone node scripts, **while mix keeps building `app.js` and the 4 sass entries**. That keeps this MR low-risk and independently shippable: `npm run production` must emit the identical file set before and after.

Key architectural fact (why this is possible without blade/PHP changes): the PHP side never uses `mix()`/`mix-manifest.json`. Blades load plain version-suffixed files via `ksgCompiledAsset()` (`app/Helpers/CustomHelper.php`), keyed on the repo-root `version` file. The build only has to keep producing the same filenames in `public/`.

## Task 0: baseline capture (before touching anything)

- On master: `ASSETS_BASE_URL=https://assets.keystone.guru npm --output_version=baseline run production` (~13–18 min). Record sorted file list + sizes of `public/js|css|webfonts|vendor`.
- Extract the messages object from each of the 23 `public/js/lang-{locale}-baseline.js` bundles (run in a node `vm` with a stub `window`) and save as JSON — this is the diff oracle for the new lang generator.
- Headless-Chrome smoke pass (map view, route editor draw/killzone, admin mapping page, a page using Inline codes) with screenshots; snapshot `Object.getOwnPropertyNames(window)`.

## Remaining work

Create `scripts/build/` (new directory):

- `build.mjs` — orchestrator with `--production` flag; runs the node steps and spawns `npx mix --production` (or dev) in the middle, passing the resolved version via an env var.
- `version.mjs` — `process.env.npm_config_output_version` if set, else `git rev-list HEAD -1`; write the repo-root `version` file **only** in the git-derived case (exact current behavior — the release map-context job writes its own).
- `handlebars.mjs` — spawn the existing `handlebars` CLI (`-m` in prod) over `resources/assets/js/handlebars/` (96 templates) → `resources/assets/js/handlebars.js` (replaces `webpack-shell-plugin-next`).
- `custom-scripts-order.mjs` — the ~200-entry ordered scripts array, moved **verbatim** from `webpack.mix.js` (lines ~162–361), including the trailing `custom/inline/*/**/*.js` glob.
- `concat.mjs` — builds `public/js/custom-{v}.js` and `public/js/lib-{v}.js`: ordered concat with `concat-with-sourcemaps` (dev), plus esbuild **transform API** minify in prod. Critical: the 288 `custom/**` files are global-scope scripts (`class Foo extends Bar {}` at top level, `eval(className)` in `inlinemanager.js`) — the output must stay a plain script, never ESM, and top-level names must survive minification (esbuild's transform API never renames top-level symbols by design).
- `lang.mjs` — replaces the webpack-only `require.context` + `laravel-localization-loader` mechanism: parse `lang/{locale}/*.php` with **`php-array-loader`** (the exact parser laravel-localization-loader wraps internally, so parsing semantics are identical), emit `public/js/lang-{locale}-{v}.js` containing the same `window.lang.setMessages(messages)` wrapper. Keep the local-env "en_US only" shortcut. Delete the generator block and `resources/assets/js/generated/lang/` handling from `webpack.mix.js`.
- `css-concat.mjs` — `public/css/custom-{v}.css` from `resources/assets/css/**/*.css` (match mix's glob order); esbuild CSS minify in prod.
- `copy.mjs` — `node_modules/@fortawesome/fontawesome-free/webfonts` + `resources/assets/webfonts` → `public/webfonts`; `resources/assets/vendor` → `public/vendor`.

Modify:

- `webpack.mix.js` — delete the moved sections (shell plugin, lang generator, scripts array, styles concat, copies, git-revision/version logic); read the version from the env var `build.mjs` sets. Mix keeps only `mix.js(app.js)` + the 4 `mix.sass()` calls + sourcemaps.
- `package.json` — scripts `dev`/`watch`/`production` point at `build.mjs` (keep the names: the release pipeline's `npm --output_version=<tag> run production` contract must be untouched; `npm_config_output_version` propagates into scripts automatically). Remove `laravel-localization-loader`, `webpack-shell-plugin-next`, `git-revision-webpack-plugin`; add `esbuild`, `php-array-loader`, `concat-with-sourcemaps`, `fast-glob`.

## Verification / definition of done

- [ ] Prod build file list identical to baseline (names incl. `-{version}` suffixes).
- [ ] Automated deep-equal diff of all 23 locales' messages objects vs the baseline JSON oracle.
- [ ] Dev-mode `custom-{v}.js` ~byte-identical to mix's `mix.scripts` output.
- [ ] Prod `custom-{v}.js`: post-build assertion that every top-level class name from the order list still appears (guards `eval(className)` and cross-file globals).
- [ ] `npm test` (vitest) green — it's already build-independent and must stay so.
- [ ] Headless smoke incl. an Inline-code page; A/B against master stack.
- [ ] `npm --output_version=vTEST run production` dry-run: filenames carry the tag, `version` file not clobbered.
- [ ] Sourcemap parity: dev builds emit maps, prod does not (current `mix.sourceMaps()` behavior).

## Key files

`webpack.mix.js`, `package.json`, new `scripts/build/*.mjs`, `resources/assets/js/custom/inline/inlinemanager.js` (read-only — the `eval(className)` constraint), `.github/workflows/release-deploy.yml` (read-only — the `--output_version` contract).

## Gotchas

- `npm run production` can OOM WSL — run on the host, avoid parallel heavy work.
- The `version` file is read by PHP (`ViewService`, `config/laravel-model-caching.php`, `MapContext` `SavesToFile`) — do not change its semantics.

Wotuu:
:robot: Closes #3491

Part of #3449 (stage 1 of 3). Everything in `webpack.mix.js` that wasn't real bundling moves to plain node scripts under `scripts/build/`; laravel-mix keeps building only `app.js` and the four sass entries until #3492 removes it entirely. `npm run dev` / `npm run production` / the release pipeline's `npm --output_version=<tag> run production` all keep working with unchanged semantics and unchanged output filenames.

## What moved

| Output | Was | Now |
|---|---|---|
| `js/custom-{v}.js` | `mix.scripts` (dev) / `mix.babel` (prod) over a ~200-entry ordered list | `scripts/build/concat.mjs` — ordered concat + esbuild **transform** minify in prod (transform never renames top-level symbols, so `eval(className)` in InlineManager and all cross-file globals keep working) |
| `js/lib-{v}.js` | `mix.babel` over `resources/assets/lib/**` | same concat script |
| `js/lang-{locale}-{v}.js` ×23 | generated `require.context` entries + `laravel-localization-loader` | `scripts/build/lang.mjs` using `php-array-parser` (the exact parser that loader wrapped) emitting the `window.lang.setMessages()` wrapper directly |
| handlebars precompile | `webpack-shell-plugin-next` onBuildStart | `scripts/build/handlebars.mjs` spawning the same CLI |
| `css/custom-{v}.css` | `mix.styles` glob concat | `scripts/build/css-concat.mjs` |
| webfonts/vendor copies | `mix.copy` | `scripts/build/copy.mjs` |
| version resolution + `version` file | `git-revision-webpack-plugin` | `scripts/build/version.mjs` (same rules: `npm_config_output_version` wins and leaves the `version` file alone; git-derived writes it) |

Dependencies: − `git-revision-webpack-plugin`, `laravel-localization-loader`, `webpack-shell-plugin-next`; + `esbuild`, `php-array-parser`, `fast-glob` (the latter two were already in the tree transitively).

## Verification

- **All 23 locales deep-equal**: messages objects extracted (node `vm`) from the old webpack lang bundles vs the new generator — identical for every locale (64 files each).
- **File-set/order parity for the concat**: all 280 top-level class names from the ordered list survive prod minification, and every one of them is present in the old mix production bundle (same input file set); glob expansion matches `find` (115 inline files).
- **Release dry-run**: `ASSETS_BASE_URL=… npm --output_version=v99.99.99 run production` → filenames carry the tag, repo-root `version` file untouched, `$asset-url` (CDN base) present in compiled CSS.
- **Headless Chrome A/B vs master** (front page, `/search`, explore map, route view incl. an opened bootstrap-select): zero page errors on the branch, error profile identical to master, `window` globals diff fully explained (babel ES5 helpers gone by design; `jarallax`/`initSwitchArrowPlugin` were stale master assets removed in #3477). Route view (heaviest JS page — map, 24 pulls, enemy visuals, handlebars sidebar) is pixel-equivalent:

| master | this branch |
|---|---|
| master | branch |

- **Vitest**: 116 tests green, including 5 new ones covering the PHP-lang parser and the ordered glob expansion (`scripts/build/*.test.mjs`, run in node environment).

## Behavior changes to be aware of

1. **`custom-{v}.js` / `lib-{v}.js` ship modern JS in production instead of babel ES5.** This matches what dev mode has always shipped (`mix.scripts` never transpiled), so the semantics are battle-tested; the change drops support for pre-ES2015 browsers in these two bundles. esbuild `target` downleveling (to ~es2017) can be added later if wanted — full ES5 is impossible with esbuild (no `let`/`const` lowering).
2. **Found & fixed by the A/B**: the vendored `lib/rollbar/rollbar.js` used top-level `let _rollbarConfig`, which babel used to lower to `var`. Untranspiled it collided with the `var _rollbarConfig` from `rollbar.blade.php` and aborted the entire lib bundle (also killing `L.heatLayer`). Changed to `var` in our vendored copy.
3. `lang-{locale}-{v}.js.map` files are no longer emitted (the bundles are generated one-liner wrappers now; the old maps were webpack-runtime noise). Everything else keeps its old sourcemap behavior (dev maps for app/sass via mix; the concat bundles never had maps).
4. npm scripts `watch-poll` and `hot` were removed: `hot` needed webpack-dev-server which isn't in the dependencies (it was already broken), and mix's webpack config already sets `watchOptions.poll`. `npm run watch` behaves as before: chokidar (polling, same 2s cadence as the webpack watchOptions) rebuilds the node-built bundles (custom/lib/css/lang/handlebars) on change, while the spawned `mix watch` keeps covering app.js/sass — verified end-to-end (touched `custom/util.js` → `custom-{v}.js` rebuilt in ~15 ms).
5. Building all 23 locales takes a few minutes of PHP-array parsing (release pipeline only — local/MR builds do en_US per `APP_ENV=local`). Same cost as the old webpack loader; can be parallelized in #3492 if CI time matters.

No PHP was touched.

https://claude.ai/code/session_018JA6zr55whFohw8SnzDjKa


Wotuu (commented):
Ok

Issue #3302, opened 2026-06-18
Reduce global $with N+1 on Dungeon and Season (follow-up to #3300)


Wotuu:
Follow-up to #3300. The view-data split + `Expansion::$with` cleanup landed there; this issue covers the two remaining (riskier) `$with` levers that still drive most of the per-request query count.

## Background — what #3300 already did

Two commits on the `3300-split-viewservice-into-modules` branch:

- `f0738e158` — Split the monolithic `ViewService::getGameServerRegionViewVariables()` into granular, independently cached getters (`getCurrentSeasonForRegion`, `getNextSeasonForRegion`, `getExpansionsData`, `getAllAffixGroupsForRegion`, `getAllAffixGroupsByActiveExpansion`, etc.). Repointed all 12 view composers to the slice each renders so text pages no longer build the affix/create-route graph. Moved `getDemoRoutes()` off the global `*` composer into `HomeComposer`.
- `dad1b390f` — Removed `timewalkingEvent` from `Expansion::$with` (stopped a per-hydration N+1).

Real MySQL `SELECT`s per request dropped: `/credits` 1510 → 86, `/` (home) 2874 → 391.

## Remaining work (this issue)

Two global `$with` declarations still cascade across the many season/expansion/dungeon loads in the header path:

1. **`app/Models/Dungeon.php`** — `public $with = ['expansion', 'floors'];` → remove `floors`.
2. **`app/Models/Season.php`** — `public $with = ['expansion', 'affixGroups', 'dungeons'];` → trim (likely all three).

### Measured potential (model cache off)
Removing `floors` from `Dungeon::$with` alone measured roughly: `/credits` −7, `/` −30, `/affixes` −32 selects. `Season::$with` was not individually measured but is accessed even more widely.

## Why this is the riskier slice

- **`floors` feeds the map renderer** — the core of the app. Every dungeon/route map view needs floors. Removing it from `$with` requires auditing each map/route-view page to add explicit eager-loads.
- **`Season`'s relations (`expansion`, `affixGroups`, `dungeons`) are accessed app-wide** — policies (`SeasonPolicy`), models (`DungeonRoute`), and many services read `$season->expansion` / `->affixGroups` / `->dungeons` directly.
- `Model::preventLazyLoading(!isProduction())` is ON (`app/Providers/AppServiceProvider.php`), so a missed eager-load **throws in local/dev** but **silently lazy-loads (N+1) in production**. The dev exception is the safety net — use it.
[...]

PR #3496, opened 2026-07-10
#3388 Remove global $with from DungeonRoute, eager-load at consumers

### LINKED ISSUE #3388: Reduce global $with on DungeonRoute, Npc and User (follow-up to #3302)

Wotuu:
:robot: Closes #3388 — third and final part (Npc #3486 → User #3495 → **DungeonRoute**). Stacked on #3495 (base branch), so this diff shows only the DungeonRoute work; GitHub will retarget to `master` when #3495 merges.

## What changed

- **`DungeonRoute`** no longer globally eager loads its 9 relations (`mappingVersion`, `dungeon`, `season`, `faction`, `specializations`, `classes`, `races`, `affixes`, `thumbnails`).
- `getSetupAttribute()` switched from `load()` to `loadMissing()` — the old code **re-queried faction/specs/classes/races on every serialization even when eager loaded**; now it only queries when missing.
- `getHasThumbnailAttribute()` now `loadMissing('thumbnails')` so the appended attribute is safe on collection-hydrated routes.
- Card-rendering query sites got the full card relation set (`author.iconfile`, `dungeon`, `affixes`, `mappingVersion`, `season.expansion`, `thumbnails`, `ratings`): weekly routes + route search in `DungeonRouteRepository`, `htmlsearch` in `AjaxDungeonRouteController`.
- `DungeonRouteRepository::getDungeonRoutesWithExpiredThumbnails()` now eager loads `dungeon`+`mappingVersion` — `ThumbnailService::queueThumbnailRefresh()` reads both on every returned route (this method re-queries bare models, so callers' eager loads never helped).
- `AdminToolsThumbnailsController` same fix; `AdminToolsEnemyForcesController` builder reads only `id` — no relations at all now.
- `map.blade.php` explicitly `loadMissing(['affixes'])` on the route — the inline map JS reads `affixes`/`setup` off the serialized route options.
- Dead code removed: every `->without([...])` on DungeonRoute (`AjaxDungeonRouteController`, `DiscoverService`, `DevDiscoverService`, `ViewService` demo routes, both AdminTools controllers) plus the `dungeon`/`season` sub-`without()` closures in the discover services, which referred to `Dungeon`/`Season` `$with` entries that #3302 already removed.

## Why the rest needs no changes

- `mapping:save` demo-route export calls `unsetRelations()` + `setAppends([])` + explicit `load()` — seeder JSON is byte-shape independent of `$with`.
- `DungeonRoutePolicy`, map context, MDT/SimulationCraft exports, and all single-route pages operate on route-bound/`find()` models, which may lazy-load freely (`preventLazyLoading` only arms on >1-row hydrations).
- The datatables JS reads `dungeon.name`/`dungeon.expansion.shortname`/`affixes`/`setup`/`author.name` — all still in that endpoint's explicit `with`; `mappingVersion`/`season` objects were serialized but never read, so the rows just got smaller.
- The delete cascade explicitly loads everything it touches.

## Measured (authed, per-request dev query log — deterministic, no cross-traffic)

| Page | master | after User #3495 | after this PR |
|---|---|---|---|
| `/` | 220 | 212 | 212 |
| `/dungeonroutes` | 342 | 313 | 313 |
| discover dungeon | 308 | 298 | 298 |
| **route map page** | 386 | 361 | **340** |
| **`/ajax/search` (20 cards)** | 467 | 450 | **437** |
| **single `DungeonRoute` hydration** | 14 | 14 | **1** |

The last row is the big one: every route-model binding (all `/ajax/route/{route}` PATCH/PUT/DELETE ops, API `show`, ratings, favorites, thumbnail jobs, ...) used to drag 13 relation queries.

## Verified

- Full suite: 987 passed; only failure is `MapTilesExistenceTest` (env-only, fails identically on master on this machine).
- Headless-Chrome vs master: home, discover, search, route map page (172 markers) — 0 pageErrors; `/ajax/search` rendered card HTML byte-identical (class histogram + route set).
- Serialized map options A/B: `dungeonroute.affixes` and `.setup` present and equal on both sides.

### DIFF
--- a/app/Http/Controllers/AdminTools/AdminToolsEnemyForcesController.php
+++ b/app/Http/Controllers/AdminTools/AdminToolsEnemyForcesController.php
@@ -58,13 +58,7 @@ class AdminToolsEnemyForcesController extends Controller
     {
         $dungeonId = (int)$request->get('dungeon_id');
 
-        $builder = DungeonRoute::without([
-            'faction',
-            'specializations',
-            'classes',
-            'races',
-            'affixes',
-        ])
+        $builder = DungeonRoute::query()
             ->select('id')
             ->when($dungeonId !== -1, static fn(Builder $builder) => $builder->where('dungeon_id', $dungeonId));
 
--- a/app/Http/Controllers/AdminTools/AdminToolsThumbnailsController.php
+++ b/app/Http/Controllers/AdminTools/AdminToolsThumbnailsController.php
@@ -24,14 +24,8 @@ class AdminToolsThumbnailsController extends Controller
         $dungeonId   = (int)$request->get('dungeon_id');
         $onlyMissing = (int)$request->get('only_missing');
 
-        $builder = DungeonRoute::without([
-            'faction',
-            'specializations',
-            'classes',
-            'races',
-            'affixes',
-        ])
-            ->with('dungeon')
+        // ThumbnailService::queueThumbnailRefresh() reads dungeon and mappingVersion on every route
+        $builder = DungeonRoute::with(['dungeon', 'mappingVersion'])
             ->when($dungeonId !== -1, static fn(Builder $builder) => $builder->where('dungeon_id', $dungeonId))
             ->orderByDesc('created_at');
 
--- a/app/Http/Controllers/Ajax/AjaxDungeonRouteController.php
+++ b/app/Http/Controllers/Ajax/AjaxDungeonRouteController.php
@@ -100,7 +100,6 @@ class AjaxDungeonRouteController extends Controller
         }
 
         $routes = DungeonRoute::with($withRelations)
-            ->without(['season'])
             // Specific selection of dungeon columns; if we don't do it somehow the Affixes and Attributes of the result is cleared.
             // Probably selecting similar named columns leading Laravel to believe the relation is already satisfied.
             ->selectRaw('dungeon_routes.*, mapping_versions.enemy_forces_required_teeming, mapping_versions.enemy_forces_required, MAX(mapping_versions.id) as dungeon_latest_mapping_version_id')
@@ -269,11 +268,8 @@ class AjaxDungeonRouteController extends Controller
             $season = Season::find($request->get('season'));
         }
 
+        // Everything the rendered route cards read - DungeonRoute no longer eager loads relations globally
         $query = DungeonRoute::with([
-            'faction',
-            'specializations',
-            'classes',
-            'races',
             // The route cards render the author's avatar - User no longer eager loads iconfile globally
             'author.iconfile',
             'affixes',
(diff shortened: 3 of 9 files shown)


Wotuu (commented):
Ok

PR #3502, opened 2026-07-11
#3462 Migrate .left_checkbox to BS5 .form-check-input (shim 3)

### LINKED ISSUE #3462: Migrate Bootstrap-4-compat shims to native Bootstrap 5 patterns (follow-up to #3397)
## Background

The Bootstrap 4 → 5 migration (#3397, MR #3419) deliberately kept a number of **BS4-compat shims**: places where we restore Bootstrap 4's look or behavior globally instead of migrating each call site to the Bootstrap 5-native pattern. That was the right call to keep the MR reviewable, but every shim is a global override fighting the framework — future Bootstrap upgrades, new pages, or third-party markup can interact with them in surprising ways. This issue is the definitive inventory: **when something styling-related looks "off" after #3419 merges, check this list first.**

Each item below lists: where the shim lives, what it does, why it exists, the risk, and the BS5-native fix that would let us delete it.

---

### 1. Global column shim (biggest blast radius)

**Where:** `resources/assets/css/custom.css` (top of file, `BOOTSTRAP 5 COMPAT` block)

```css
.col, [class^="col-"], [class*=" col-"] {
    position: relative;
    padding-right: calc(var(--bs-gutter-x, 1.5rem) * .5);
    padding-left: calc(var(--bs-gutter-x, 1.5rem) * .5);
}
```

**Why:** BS4 gave *every* column `position: relative` + gutter padding. BS5 removed `position: relative` entirely and only pads columns that are **direct children of a `.row`** (via `.row > *`). Without the shim, absolutely-positioned overlays anchored to a column break (e.g. dungeon-card titles over their images) and standalone columns lose their padding (e.g. the affix list).

**Risk:** the selector matches every element whose class *starts with or contains* `col-` — including any future component or third-party widget. A column outside a row now always gets `1.5rem` gutters even where BS5 would give none. Anyone adding `.g-0`/`.gx-*` rows must rely on the CSS variable cascade doing the right thing.

**Native fix:** grep for the two breakage patterns and fix call sites, then delete the shim:
- overlays anchored to a col → add Bootstrap's `position-relative` utility to that col;
- standalone cols used as padded containers → replace with `px-*` utilities or wrap in a `.row`.

### 2. `select.form-control` native arrow restore

**Where:** `resources/assets/css/general.css` (commit f2794e9c4)

```css
select.form-control { appearance: auto !important; /* + -webkit/-moz */ }
```

**Why:** BS5 sets `appearance: none` on `.form-control`; BS4-idiom `<select class="form-control">` (profile Region/Timezone, etc.) lost its native dropdown arrow and looked like a dead text box. Needs `!important` because compiled theme rules (`.darkly .form-control`) out-specify it.

**Risk:** `!important` in a global file; if we ever *want* a custom-styled select via `.form-control`, this fights it.

**Native fix:** change every native (non-selectpicker) `<select class="form-control">` to BS5's `form-select` class, then delete the shim. Find them with:
`grep -rn '<select[^>]*form-control' resources/views/ | grep -v selectpicker` (also check `html()->select(...)` builder calls). The selectpicker-managed selects don't matter — they're hidden and replaced.

### 3. `.left_checkbox` BS4 checkbox look

**Where:** `resources/assets/sass/theme/theme.scss` (`.theme .left_checkbox`)

```scss
.left_checkbox { appearance: auto; width: 40px; height: calc(1.6em + 0.75rem + 2px); }
```

**Why:** same `appearance: none` problem for checkboxes rendered as `.form-control` (login "remember me", register "legal agreed", profile "show as anonymous", …), plus BS5 dropped the fixed height that made them big.

**Native fix:** migrate these to BS5's `.form-check` / `.form-check-input` markup (or `.form-switch`), then delete `.left_checkbox` entirely.

### 4. `.dismiss-close` — project-owned replacement for BS4's removed `.close`

**Where:** `resources/assets/sass/theme/theme.scss` (button styling) + `resources/assets/css/general.css` (modal positioning, `color: var(--theme-text-contrast)`) + `resources/views/common/general/modal.blade.php` (markup: `<button class="dismiss-close" data-bs-dismiss="modal"><i class="fas fa-times"></i></button>`)

**Why:** BS5 removed `.close` in favor of `.btn-close` (a background-image × with its own light/dark variants). We kept the FontAwesome × design, renamed to `.dismiss-close` so BS5 can't collide with it, and colored it with `--theme-text-contrast` so it's visible on lux (white ×  was invisible on white modals).

**Risk:** low — it's fully project-owned now. But it duplicates what `.btn-close` provides (including focus/hover states and `data-bs-theme` awareness).

**Native fix:** replace with `.btn-close` (BS5 handles dark themes via `data-bs-theme="dark"` on the modal or `.btn-close-white`); delete `.dismiss-close` styling from both files.

### 5. `$navbar-padding-x: 1rem` restore (darkly)

**Where:** `resources/assets/sass/theme/darkly/darkly.scss` (commit 83d066e2a)

**Why:** BS5 changed the `$navbar-padding-x` default from `1rem` to `null` (0). On mobile the darkly navbar's dark box touched the screen edges; desktop hides it because the inner `.container` is max-width'd. lux and vapor were unaffected only because they already override `.navbar { padding: 1rem 1rem }` directly (lux without, vapor with `!important`).

**Risk:** low, but the *same value lives in three different places in three different forms* (SCSS var for darkly, plain CSS for lux, `!important` CSS for vapor).

**Native fix:** pick one mechanism — either set `$navbar-padding-x`/`$navbar-padding-y` per theme in SCSS and delete the CSS overrides, or drop the variable and put a spacing utility on the navbar markup itself.

### 6. Per-theme `$body-bg` — and vapor pinned to a historical accident

**Where:** `resources/assets/sass/_variables.scss` (global removed, with explanatory comment), `theme/darkly/darkly.scss` (`$body-bg: #2B3E50`), `theme/vapor/vapor.scss` (`$body-bg: #2B3E50`)

**Why:** BS5 derives `--bs-body-bg` from `$body-bg` and feeds it into cards, modals, dropdowns and list-groups; the old global navy leaked into all three themes (lux got dark cards on a light page). Now each theme sets its own. **Note:** vapor's navy is deliberately pinned to the *pre-BS5 rendering, which was itself the pollution artifact* — bootswatch vapor's real body-bg is deep purple (#1a0933). We chose visual continuity over palette purity.

**Risk:** if anyone "cleans up" vapor to bootswatch defaults, inputs/modals will change color site-wide and it will look like a regression (or vice versa: this pin is why vapor never looks like stock bootswatch vapor).

**Native fix:** none needed technically — but make an explicit design decision for vapor (keep navy or adopt bootswatch purple) and note it in the theme file (partially done via comments).

### 7. Link underlines disabled per theme

**Where:** all three theme scss files: `$link-decoration: none; $link-hover-decoration: underline;`

**Why:** BS5 underlines links by default; BS4 didn't. Restored BS4 behavior in each theme.

**Risk:** minimal — but it's repeated in three files; a fourth theme would silently get underlines.

**Native fix:** accept as a design choice; optionally centralize (it's one of the few vars that's safe to set globally before the theme imports, since no bootswatch theme sets it).

### 8. bootstrap-select 1.14.0-beta3 — a BS4-era library running against BS5

**Where:** `package.json` (`bootstrap-select: 1.14.0-beta3`), plus a render workaround in `resources/assets/css/custom.css`:

```css
.affixselect.bootstrap-select .dropdown-menu li a span.text { display: block; }
```

and `resources/assets/js/bootstrapselectrefreshfix.js`.

**Why:** there is no stable bootstrap-select release for BS5; beta3 mostly works. Known beta3-internal issues are tracked in #3420 (notably: **Enter-key selection leaves the menu open**, mouse click closes fine; the `span.text` shim fixes affix-icon rows collapsing to width 0 inside the shrink-to-fit `data-content` span on first render).

**Risk:** the highest long-term risk on this list — an unmaintained beta pinned against a major version it doesn't officially support. Every odd dropdown behavior should be suspected of coming from here first.

**Native fix:** when a BS5-compatible bootstrap-select ships, upgrade and delete the CSS shim + refresh fix; if it never ships, evaluate replacing with a maintained picker (e.g. tom-select). Tracked in #3420.

### 9. `scopedThemeRootFix` postcss plugin (migration infrastructure)

**Where:** `webpack.mix.js`

**Why:** our themes wrap all of Bootstrap inside a theme class (`.darkly { @import bootstrap }`). BS5 emits its design tokens as CSS variables on `:root`, which compiles to the never-matching selector `.darkly :root`. The plugin rewrites it to `:root.darkly`.

**Risk:** the regex only rewrites selectors of the exact shape `^.theme :root$`. Any future rule where `:root` appears in a different nested shape silently stays broken (symptom: `--bs-*` variables undefined in one theme). Also: nobody expects a postcss plugin in `webpack.mix.js` to be load-bearing for theming — hence this paragraph.

**Native fix:** the real solution is restructuring theming to BS5's `data-bs-theme` / CSS-variable overrides instead of compiling Bootstrap three times into scoped classes — a much bigger refactor, likely its own issue. Until then the plugin stays.

### 10. Vendored bootswatch file hand-patched

**Where:** `resources/assets/sass/theme/vapor/dist/_bootswatch.scss` — `.arrow` → `.tooltip-arrow` (BS5 rename)

**Why:** vapor imports a *vendored copy* of bootswatch's scss (unlike darkly/lux which import from node_modules), so BS5 renames must be patched by hand in our copy.

**Risk:** the vendored copy silently drifts from upstream bootswatch; future BS5.x renames won't arrive automatically.

**Native fix:** figure out why vapor needs the vendored copy and switch back to the node_modules import if possible.

---

## Not shims (checked during #3419 verification, listed to prevent re-investigation)

- Modals not closing on Escape → pre-existing `data-keyboard="false"` in `common/general/modal.blade.php`, faithfully converted to `data-bs-keyboard="false"`. Intended behavior.
- `data-toggle="navbar-shrink"` (2 occurrences) → custom site hook (`navbarshrink.js`), not a missed BS5 conversion. Do not rename to `data-bs-toggle`.
- Embed-header fixes (e2f796dd3, 7e8ff360a) → proper BS5 fixes (removed nested `.row` as direct row child), no compat debt.

## Recommended approach

One shim per PR, in roughly this order (payoff ÷ risk): **2 → 3 → 4 → 5/7 → 1 → 10**, with **8** whenever upstream moves and **9** as its own refactor issue. For each: fix the call sites, delete the shim, run the visual sweep below.

## How to verify

Headless-Chrome sweep of both themes and mobile, as done for #3419 (see MR wrap-up comment: https://github.com/RaiderIO/keystone.guru/pull/3419#issuecomment-4914789308): home, search, route view/edit, embed (regular+compact), login/register modals, profile, admin dungeon/npc edit — darkly/lux/vapor, 1600×1000 + 390×844. Zero `pageerror`s; compare against pre-change screenshots.

## Definition of done

- Every numbered shim above is either deleted (call sites migrated to BS5-native) or explicitly re-documented as a permanent design choice in the file itself.
- `npm run production` builds clean; visual sweep matches; PHPUnit green (ignore the known MapTilesExistenceTest env failure); `composer run fix` / `analyse` clean (35 pre-existing PhpStan errors on master are not part of this issue).

Wotuu:
:robot: Closes #3462 (shim 3 of the inventory — one shim per PR)

## What

Removes the **`.left_checkbox`** BS4-compat shim (issue #3462, item 3). Bootstrap 5's `.form-control` sets `appearance: none` on checkboxes (hiding the native box) and dropped the fixed height that made them large; the shim restored the BS4 look globally via `.theme .left_checkbox { appearance: auto; width: 40px; height: … }`.

**Native fix:** every checkbox call site moves from `.form-control left_checkbox` to BS5's native **`.form-check-input`** (the `.form-control` must go too — on a checkbox it's what strips the checkmark), then the shim is deleted from `theme.scss`.

- 24 blade forms + the `map_popup_type_bool_template.handlebars` template (compiled `handlebars.js` is a gitignored build artifact, regenerated).
- `.left_checkbox` block deleted from `resources/assets/sass/theme/theme.scss`.
- Also fixes a pre-existing PHPStan error on master (`panel.blade.php` `$linkOptions` iterable value type) so this branch's CI is green.

## Verification

- `npm run production` builds clean; compiled `theme`/`custom` CSS + `handlebars.js` confirmed free of `left_checkbox`, `.form-check-input` present.
- `composer run analyse` clean, `composer run fix` clean, `ViewComposerTest` green (36 passed).
- Headless-Chrome A/B (login + register modals, 1600px): master renders `.form-control left_checkbox` @ **40px**, this branch renders `.form-check-input` @ **15px** (BS5-native), alignment clean. Screenshots in a comment below.
- Admin edit pages (npc/dungeon/floor/release/spell) use the identical mechanical class swap; not individually screenshotted (auth-gated) but covered by the compiled-CSS + measurement check.

### DIFF
--- a/resources/assets/js/handlebars/map_popup_type_bool_template.handlebars
+++ b/resources/assets/js/handlebars/map_popup_type_bool_template.handlebars
@@ -5,7 +5,7 @@
             {{ description }}
         </div>
     {{/if}}
-    <input type="checkbox" class="form-control left_checkbox" value="{{value}}"
+    <input type="checkbox" class="form-check-input" value="{{value}}"
            id="map_{{map_object_name}}_edit_popup_{{property}}_{{id}}"
            name="map_{{map_object_name}}_edit_popup_{{property}}_{{id}}"
         {{#ifCond value '===' 1 }}
--- a/resources/assets/sass/theme/theme.scss
+++ b/resources/assets/sass/theme/theme.scss
@@ -348,6 +348,15 @@ body {
     -webkit-user-select: none;
   }
 
+  /*
+   * Bootstrap 5's `.form-check-input` checkboxes default to 1em, which is tiny in our standalone
+   * (label-less) checkbox layouts. Restore the larger size the old `.left_checkbox` shim gave them.
+   */
+  .form-check-input[type="checkbox"] {
+    width: 40px;
+    height: calc(1.6em + 0.75rem + 2px);
+  }
+
   .modal-backdrop {
     z-index: 100010 !important;
   }
--- a/resources/assets/sass/theme/theme.scss
+++ b/resources/assets/sass/theme/theme.scss
@@ -348,16 +348,6 @@ body {
     -webkit-user-select: none;
   }
 
-  /*
-   * Restores the Bootstrap 4 look of `.form-control` checkboxes: Bootstrap 5's .form-control sets
-   * `appearance: none` (hiding the native checkbox) and no longer has a fixed height (which made it big).
-   */
-  .left_checkbox {
-    appearance: auto;
-    width: 40px;
-    height: calc(1.6em + 0.75rem + 2px);
-  }
-
   .modal-backdrop {
     z-index: 100010 !important;
   }
--- a/resources/views/admin/dungeon/edit.blade.php
+++ b/resources/views/admin/dungeon/edit.blade.php
@@ -32,25 +32,25 @@ use Illuminate\Support\Collection;
         <div class="row mb-3">
             <div class="col {{ $errors->has('active') ? ' has-error' : '' }}">
                 {{ html()->label(__('view_admin.dungeon.edit.active'), 'active') }}
-                {{ html()->checkbox('active', $dungeon?->active ?? 1, 1)->class('form-control left_checkbox') }}
+                {{ html()->checkbox('active', $dungeon?->active ?? 1, 1)->class('form-check-input') }}
                 @include('common.forms.form-error', ['key' => 'active'])
             </div>
 
             <div class="col {{ $errors->has('raid') ? ' has-error' : '' }}">
                 {{ html()->label(__('view_admin.dungeon.edit.raid'), 'raid') }}
-                {{ html()->checkbox('raid', $dungeon?->raid ?? 0, 1)->class('form-control left_checkbox') }}
+                {{ html()->checkbox('raid', $dungeon?->raid ?? 0, 1)->class('form-check-input') }}
                 @include('common.forms.form-error', ['key' => 'raid'])
             </div>
 
             <div class="col {{ $errors->has('heatmap_enabled') ? ' has-error' : '' }}">
                 {{ html()->label(__('view_admin.dungeon.edit.heatmap_enabled'), 'heatmap_enabled') }}
-                {{ html()->checkbox('heatmap_enabled', $dungeon?->heatmap_enabled ?? 0, 1)->class('form-control left_checkbox') }}
+                {{ html()->checkbox('heatmap_enabled', $dungeon?->heatmap_enabled ?? 0, 1)->class('form-check-input') }}
                 @include('common.forms.form-error', ['key' => 'heatmap_enabled'])
             </div>
 
             <div class="col {{ $errors->has('has_wallpaper') ? ' has-error' : '' }}">
                 {{ html()->label(__('view_admin.dungeon.edit.has_wallpaper'), 'has_wallpaper') }}
-                {{ html()->checkbox('has_wallpaper', $dungeon?->has_wallpaper ?? 0, 1)->class('form-control left_checkbox') }}
+                {{ html()->checkbox('has_wallpaper', $dungeon?->has_wallpaper ?? 0, 1)->class('form-check-input') }}
                 @include('common.forms.form-error', ['key' => 'has_wallpaper'])
             </div>
 
@@ -58,7 +58,7 @@ use Illuminate\Support\Collection;
                 <div class="row">
                     <div class="col-auto {{ $errors->has('speedrun_enabled') ? ' has-error' : '' }}">
                         {{ html()->label(__('view_admin.dungeon.edit.speedrun_enabled'), 'speedrun_enabled') }}
-                        {{ html()->checkbox('speedrun_enabled', $dungeon?->speedrun_enabled ?? 0, 1)->id('speedrun_enabled')->class('form-control left_checkbox') }}
+                        {{ html()->checkbox('speedrun_enabled', $dungeon?->speedrun_enabled ?? 0, 1)->id('speedrun_enabled')->class('form-check-input') }}
                         @include('common.forms.form-error', ['key' => 'speedrun_enabled'])
                     </div>
 
--- a/resources/views/admin/expansion/edit.blade.php
+++ b/resources/views/admin/expansion/edit.blade.php
@@ -14,7 +14,7 @@
 
     <div class="mb-3{{ $errors->has('active') ? ' has-error' : '' }}">
         {{ html()->label(__('view_admin.expansion.edit.active'), 'active') }}
-        {{ html()->checkbox('active', isset($expansion) ? $expansion->active : 1, 1)->class('form-control left_checkbox') }}
+        {{ html()->checkbox('active', isset($expansion) ? $expansion->active : 1, 1)->class('form-check-input') }}
         @include('common.forms.form-error', ['key' => 'active'])
     </div>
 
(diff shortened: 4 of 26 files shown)


Wotuu:
:robot: **Headless-Chrome A/B verification** (1600px viewport)

The checkbox goes from the BS4 shim's big 40px box to Bootstrap 5's native ~15px checkbox. Alignment stays clean.

**Login modal — before (master, `.form-control left_checkbox` @ 40px):**
before-login

**Login modal — after (this branch, `.form-check-input` @ 15px):**
after-login

**Register modal — after (the "agree" checkbox under the multi-line consent label):**
after-register

Wotuu (commented):
Ok

PR #3519, opened 2026-07-12
#3515 Add sh/release-watch.sh: track and drive a full release from one terminal

### LINKED ISSUE #3515: Add sh/release-watch.sh: track and drive a full release (both repos) from one terminal

## Background

A release currently means watching pipelines across two repos: `release-deploy.yml` here (3 build jobs, ~12 min), then a `repository_dispatch` fire-and-forget to RaiderIO/keystoneguru-infra whose `Deploy` runs (staging ~8-15 min, production ~8-16 min after the manual gate) must be tracked down by hand. Nothing reports status back. Verification (asset 200s, staging/production HTML flipping to `compiled/<tag>/`) is also manual. Measured on v15.3.4/v15.4.1.

## Remaining work

New `sh/release-watch.sh [vX.Y.Z]` (bash + gh + jq, defaults to newest v* tag's run):

1. Find the release-deploy run for the tag; poll **job** conclusions (run status parks on `waiting` at the production gate forever).
2. Track the build jobs with elapsed times; print a failed job's URL immediately.
3. After the `Deploy to Staging` dispatch job succeeds, correlate the infra run (`gh run list --repo RaiderIO/keystoneguru-infra --workflow=Deploy --event=repository_dispatch`, `createdAt >=` dispatch time, version confirmed via the job name `Deploy deploy-staging (vX.Y.Z)`) and poll its steps.
4. Auto-verify staging: HEAD the three asset URLs (`compiled/<tag>/js/app-<tag>.js`, css, `mapcontext/static/en_US.js`) expecting 200; poll `https://staging.keystone.guru/` HTML until it references `compiled/<tag>/`.
5. Production gate: show waiting state; offer approval ONLY on explicit interactive confirmation (type the version string) via `gh api repos/RaiderIO/keystone.guru/actions/runs/<run-id>/pending_deployments` (GET for `environment.id`, POST `state=approved`); fall back to printing the gate URL if the API rejects it. `--watch-only` disables the approval path. Never auto-approve.
6. Track the infra production run the same way, verify production HTML flips, print a final summary.

Re-entrant: safe to start or restart mid-release — discovers current state and resumes.

Also: new `.claude/skills/release-watch/SKILL.md` + pointers from the create-release and deployment-pipeline skills.

## Verification

- `bash -n` + shellcheck clean.
- Read-only runs against the completed v15.3.4 release (run 29123763527 / infra 29125867692 + 29145975046) and the v15.4.1 run parked at the gate (29164420834). Do not approve any gate while testing.

## Definition of done

- [ ] Script tracks build jobs, both infra runs, and verification checks in one view.
- [ ] Gate approval requires typing the version; `--watch-only` supported.
- [ ] Skill added; create-release + deployment-pipeline skills reference it.

Wotuu:
:robot: Closes #3515

## Summary

Adds `sh/release-watch.sh`, a re-entrant polling loop that watches (and can drive) a
full release across both repos from one terminal:

- Polls `release-deploy.yml` **job** conclusions, never the run status (which parks on
  `waiting` at the production environment gate forever).
- Enumerates build jobs dynamically (`gh run view --json jobs`) — no hardcoded job
  names, since two in-flight MRs are changing them (images job becomes a matrix; the
  map-context job splits into a gating `priority` job and a non-gating `rest` job). Only
  `Deploy to Staging` / `Deploy to Production` are special-cased.
- Once `Deploy to Staging` succeeds, correlates the matching `RaiderIO/keystoneguru-infra`
  "Deploy" run by stage + version, preferring the **latest** match so a re-dispatch
  after an earlier infra failure is picked up correctly, and shows its current step.
  The correlation handles **both** displayTitle forms: the current bare event type
  (`deploy-staging`, version then confirmed via the job name
  `Deploy deploy-staging (vX.Y.Z)`) and the `Deploy deploy-staging (vX.Y.Z)` run-name
  form that infra PR RaiderIO/keystoneguru-infra#37 introduces (version confirmed from
  the title directly, skipping the per-run job fetch). A same-stage run for a different
  version never matches, and the two stage names can't cross-match as substrings.
- Runs staging verification once that infra run succeeds: 3 asset HEAD checks + polling
  staging's HTML until it references the tag (own ~5 min timeout, checked fresh every
  cycle rather than blocking the loop).
- Shows `GATE: waiting for approval` when `Deploy to Production`'s job status is
  `waiting` and staging verification passed. Approval (`GET`/`POST`
  `.../actions/runs/<id>/pending_deployments`) only ever happens if **all** of: not
  `--watch-only`, stdin is a real TTY, staging verification passed, **and** the user
  types the exact version string that cycle. Anything else skips for that cycle (re-asks
  next cycle). If the API rejects the approval, it prints the run URL for manual
  approval instead.
- Same correlation + verification for production once the gate opens.
- Final summary once every job/infra-run/verification item has concluded; exit 0 on full
  success, 1 if anything failed (with the failing item(s) called out).
- Fully re-entrant: every cycle re-derives all state from the GitHub API — restarting
  the script mid-release just resumes wherever the release actually is.

Also adds the `.claude/skills/release-watch/SKILL.md` skill, and short pointers from
`create-release` (suggest running it after cutting the tag) and `deployment-pipeline`
(prefer it over the manual verify steps).

## Test plan (read-only — see safety note below)

- [x] `bash -n sh/release-watch.sh` — clean. `shellcheck` isn't installed on the host
      it was built on; skipped per instructions (no install).
- [x] Ran with `--run-id 29123763527 --watch-only < /dev/null` (completed v15.3.4
      release). Correctly walked to a final summary in one cycle: all build jobs
      success, both infra runs correlated correctly (including preferring the *later*
      successful staging infra run over an earlier failed retry — this release had a
      real mid-release infra failure + re-dispatch), all asset checks passed. The
      HTML-flip check for staging/production correctly **failed** (with the timeout
      message, not a hang) — expected, because production has since moved on to
      v15.4.1 in the real environment, so the live site genuinely no longer references
      `compiled/v15.3.4/`. Exit code 1, as it should be for that live state.
- [x] Ran with `--run-id 29164420834 --watch-only < /dev/null` (v15.4.1). Staging is
      still on v15.4.1 live, so staging verification fully passed (all 4 checks PASS).
      Production happened to have a **real, currently in-progress** infra deploy run at
      test time (approved by Wotuu this morning) — the script correctly tracked it
      across cycles, showing the live current step advancing
      (`Deploy services stack` → `Announce release (post-deploy)`). Only GET requests
      were made throughout; nothing was mutated.
- [x] Ran with `--run-id 29162084548 --watch-only < /dev/null` (v15.4.0, a real failed
      run: `Build & Upload Frontend Assets` failed, `Deploy to Staging`/
      `Deploy to Production` skipped). Correctly printed the failed job's URL
      immediately, correctly treated `skipped` as non-failing-but-non-success, exited 1
      in a single cycle.
- [x] Verified argument parsing: no-arg default resolution, explicit `vX.Y.Z` resolution,
      `--run-id`, `--interval 0` rejection, unrecognized-flag rejection, `--help`.
- [x] Dual-form infra correlation: unit-tested `find_infra_run` against a stubbed `gh`
      and a hand-built fixture containing both displayTitle forms — 7/7 checks pass
      (new-form fast path with zero job fetches, same-stage/different-version rejection,
      old-form job-name fallback, no stage cross-matching, correct version selection when
      multiple new-form runs exist). The live v15.3.4 re-run above doubles as a real-data
      test of the old-form fallback path, since the infra repo still uses bare
      displayTitles today.
  - [x] Verified the gate-approval prompt is skipped safely when stdin isn't a TTY
      (`[[ ! -t 0 ]]` guard, confirmed directly) — this repo currently has no run
      sitting at a live gate to test end-to-end, so the `waiting`-state branch itself is
      exercised via GitHub's documented job-status enum (confirmed `waiting` is a valid
      `status` value via the REST API schema) rather than a live observed example.

**Safety note:** every test above used `--watch-only` and/or piped stdin from
`/dev/null`, so the approval POST path was never reachable. No gate was approved and
nothing was mutated during testing — there is a real, separate production gate pending
elsewhere in the pipeline that this work deliberately did not touch.

https://claude.ai/code/session_01ArakgH1TXrqehHM3QECW9w

### DIFF
--- a/.claude/skills/create-release/SKILL.md
+++ b/.claude/skills/create-release/SKILL.md
@@ -232,3 +232,7 @@ issue URL, and any commits without an issue number or any new category you added
 that the tag push triggers the `release-deploy` pipeline (staging first), and that the GitHub
 *Release* itself is cut separately from the infra project once it's going to production — see
 the deployment-pipeline roadmap (issues #3327–#3329).
+
+Suggest running `sh/release-watch.sh <version>` next — it tracks the build jobs, both
+staging/production infra deploys, and verification from one terminal (see the `release-watch`
+skill).
--- a/.claude/skills/deployment-pipeline/SKILL.md
+++ b/.claude/skills/deployment-pipeline/SKILL.md
@@ -79,6 +79,11 @@ There is **no** push-to-branch deploy. Deploys only happen on a tag (and the inf
 
 ## Verify a release end-to-end
 
+Preferred: `sh/release-watch.sh <tag>` (or no argument for the newest release-deploy run) —
+it does all of the below itself in a re-entrant polling loop, correlates the infra runs, and
+can drive the production gate. See the `release-watch` skill. The manual steps below are
+useful for spot-checks or when reasoning about a single piece.
+
 ```bash
 # Assets (the thing that used to 404) — expect 200:
 curl -sI https://assets.keystone.guru/compiled/<tag>/js/app-<tag>.js
--- a/.claude/skills/release-watch/SKILL.md
+++ b/.claude/skills/release-watch/SKILL.md
@@ -1,0 +1,80 @@
+---
+name: release-watch
+description: >
+  Use after a v* tag has been pushed and release-deploy.yml is running (or about to run),
+  when the user wants to watch or drive a release to completion from one terminal instead
+  of tracking release-deploy.yml, the keystoneguru-infra Deploy runs, and the manual
+  production gate by hand. Runs sh/release-watch.sh, a re-entrant polling loop that tracks
+  build jobs, correlates the staging/production infra runs, runs asset and HTML
+  verification, and can approve the production gate. Use --watch-only for a pure observer
+  with no approval prompt. Gate approval always requires the human typing the exact
+  version string; the script (and this skill) must never approve unattended. Not for
+  authoring the release changelog itself (create-release) or a general map of the pipeline
+  (deployment-pipeline).
+---
+
+# Release watch
+
+`sh/release-watch.sh` tracks (and can drive) a full release across both repos from one
+terminal, instead of the manual back-and-forth of watching `release-deploy.yml` here, then
+guessing which `keystoneguru-infra` "Deploy" run corresponds to which stage, then manually
+curling the verification endpoints.
+
+## Usage
+
+```
+sh/release-watch.sh [vX.Y.Z] [--watch-only] [--interval <seconds>] [--run-id <id>]
+```
+
+- No version → newest `release-deploy.yml` run's tag.
+- `--watch-only` → never prompts for or performs production gate approval (pure observer).
+- `--interval <seconds>` → poll interval, default 15.
+- `--run-id <id>` → explicit release-deploy run id (skips the tag → run lookup).
+
+Only depends on bash, `gh`, `jq`, `curl`, `date` — no extra tooling.
+
+## What it does, per polling cycle
+
+1. Fetches the release-deploy run's jobs (dynamically — build job names are not
+   hardcoded, since they change over time) and prints status + elapsed time for each. A
+   failed job's URL is printed immediately.
+2. Once `Deploy to Staging` succeeds, correlates the matching `keystoneguru-infra` "Deploy"
+   run (by stage + version, preferring the latest match if a stage was re-dispatched after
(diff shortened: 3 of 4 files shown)


Wotuu (commented):
Ok

PR #3520, opened 2026-07-12
#3517 Season-first map context: --scope option + split release job

### LINKED ISSUE #3517: Season-first map context: deploy once current-season dungeons are uploaded, stream the rest in

## Background

The `build-map-context` job in `release-deploy.yml` seeds a full DB and regenerates map context for **all ~158 dungeons x locales x mapping versions** before `deploy-staging` may even be dispatched (5.8 min on v15.4.1; becomes the critical path once #3449 collapses the assets job). But ~99% of traffic hits the ~8 current-season dungeons — only those must exist in S3 when ECS flips; the rest can stream in while the infra deploy (8-15 min) runs.

## Remaining work

1. **Scope filter on the generators** (`app/Console/Commands/MapContext/`): add `--scope=current-season|rest|all` (default `all`, existing behavior unchanged) to `make:mapcontextdungeon` and `make:mapcontextmappingversion`. Both currently iterate `Dungeon::all()` / `MappingVersion::all()`. Resolve current-season dungeon IDs via the existing Season relationships/`SeasonServiceInterface` (current season per active expansion, so Classic-era seasons included); `rest` is the complement; mapping versions scope by their dungeon. `make:mapcontextstatic` stays global (always priority).
2. **Split the workflow job**: `build-map-context-priority` (static + `--scope=current-season`; `deploy-staging` needs this) and `build-map-context-rest` (`--scope=rest`; NOT in `deploy-staging`'s needs — it uploads while the infra deploy runs).

Accepted trade-off: a few-minute window right after the staging flip where a non-current-season route's map context can 404; it self-heals when the rest lands, and the production gate is approved long after everything is uploaded.

## Verification

- PHPUnit tests for the scope resolution (current-season vs rest vs all), Arrange-Act-Assert, try/finally cleanup, run in Docker.
- Locally: file tree of `current-season` + `rest` runs unioned must be identical to an `all` run.
- `composer run fix` + `composer run analyse` clean.

## Definition of done

- [ ] `--scope` option with tests; default behavior unchanged.
- [ ] Workflow split so `deploy-staging` only waits on priority map context.
- [ ] Expected: dispatch-blocking map context work ~5.8 → ~2.5-3 min.

Wotuu:
:robot: Closes #3517

## Summary

- Adds a `--scope=priority|rest|all` option (default `all`, existing behavior unchanged) to `make:mapcontextdungeon` and `make:mapcontextmappingversion`. `make:mapcontextstatic` is untouched (stays global/priority).
- `priority` resolves via `GameVersion::active()->get()` filtered to `has_seasons`, then both `SeasonServiceInterface::getCurrentSeason($gameVersion->expansion)` and `getNextSeasonOfExpansion($gameVersion->expansion)` per matching game version — so it naturally covers non-retail game versions too (e.g. Classic-era) the moment they get a `has_seasons` flag and a current/next season, without any hardcoded expansion/dungeon list. Today only `retail` (→ expansion `midnight`) qualifies. The next season is included alongside the current one because it starts receiving traffic (previews, early route creation) before it's the fully active season.
- `rest` is the exact complement of `priority` over `Dungeon::all()`. Mapping versions are scoped by their dungeon's id (`MappingVersion::whereIn('dungeon_id', $dungeonIds)`), so **all** mapping versions of an in-scope dungeon are included, not just the latest.
- The resolution logic is factored into one shared trait, `app/Console/Commands/MapContext/Traits/ResolvesMapContextScope.php`, used by both commands (no duplication). The valid scope values are a native backed enum, `App\Console\Commands\MapContext\Enums\MapContextScope`. Invalid `--scope` values print an error and return a non-zero exit code.
- Splits the `build-map-context` job in `.github/workflows/release-deploy.yml` into `build-map-context-priority` (static + `--scope=priority`; `deploy-staging`'s `needs` now points here) and `build-map-context-rest` (`--scope=rest`; deliberately **not** in `deploy-staging`'s or `deploy-production`'s `needs`, so it uploads while the infra deploy runs). The diff is kept surgical — `build-push-images` (touched by #3516) is untouched.

## How priority resolution works

Investigated `SeasonServiceInterface`/`SeasonService`, the `Season` model's `dungeons()` relation, and `GameVersion`/`Expansion`. The front page (`DungeonRouteDiscoverController::discoverCurrentGameVersion`) already does exactly this for the current season: check `$gameVersion->has_seasons`, then `$seasonService->getCurrentSeason($gameVersion->expansion)`. Reused that pattern instead of `Expansion::active()` directly, because `active` on `Expansion` just means "released/supported" (true for almost every past expansion, including long-retired ones like Legion/BfA/Shadowlands) — it does not mean "currently the active season line". `GameVersion.has_seasons` is the actual signal for "this game version currently has an ongoing season", and correctly yields only `retail` today.

`getNextSeasonOfExpansion($expansion)` is merged in alongside `getCurrentSeason($expansion)` — per review, there's a period where a next season is known (and already receiving some traffic) but not active yet, so both need to land in "priority" mode.

## Verification

- Tests: `tests/Feature/Console/Commands/MapContext/ResolvesMapContextScopeTest.php` — 7 tests, all green (`php artisan test --compact --filter=ResolvesMapContextScopeTest`): priority is a non-empty strict subset of all dungeons, priority merges in the next season's dungeons when known (mocked `SeasonServiceInterface` with TWW S1/S2, which have disjoint dungeon rosters, since Midnight doesn't have a season 2 in the DB yet to exercise this for real), rest is the exact complement (and their union recombines to all), `all` matches `Dungeon::all()`, invalid scope fails.
- End-to-end equivalence check in the worktree container (`--scope=priority` + `--scope=rest` vs. unscoped `all`, both commands):
  - **all**: 2493 files
  - **priority**: 208 files (8 dungeons: Algeth'ar Academy, Magister's Terrace, Maisara Caverns, Nexus-Point: Xenas, Pit of Saron, Seat of the Triumvirate, Skyreach, Windrunner Spire — Midnight Season 1; no Midnight S2 seeded yet, so priority == current season only in practice today)
  - **rest**: 2285 files
  - union(priority, rest) == all exactly (byte-identical sorted file list diff); intersection is empty. 208 + 2285 = 2493. ✓
- `.github/workflows/release-deploy.yml` parses cleanly (`Symfony\Component\Yaml\Yaml::parseFile(...)` in the app container); job graph verified (`build-map-context-priority`/`build-map-context-rest` present, `deploy-staging.needs` = `[build-push-images, build-assets, build-map-context-priority]`). CI does not run this workflow on the MR (tag-triggered only) — **the workflow change will be validated on the next release**.
- `composer run fix` + `composer run analyse` both clean (an unrelated pre-existing formatting drift in `WowheadTranslationService.php` was reverted so the diff stays focused).
- Full `SeasonService` + `Console` test groups (59 tests total, run together) green; no regressions.

## Test plan

- [x] `ResolvesMapContextScopeTest` green (incl. the new next-season-merge test)
- [x] File-tree equivalence check (priority ∪ rest == all, priority ∩ rest == ∅)
- [x] Workflow YAML parses + job graph verified
- [x] `SeasonService` + `Console` test groups green

### DIFF
--- a/app/Console/Commands/MapContext/MakeMapContextDungeon.php
+++ b/app/Console/Commands/MapContext/MakeMapContextDungeon.php
@@ -2,14 +2,17 @@
 
 namespace App\Console\Commands\MapContext;
 
+use App\Console\Commands\MapContext\Traits\ResolvesMapContextScope;
 use App\Console\Commands\MapContext\Traits\SavesToFile;
 use App\Models\Dungeon;
 use App\Service\MapContext\MapContextServiceInterface;
+use App\Service\Season\SeasonServiceInterface;
 use Illuminate\Console\Command;
 use Illuminate\Support\Collection;
 
 class MakeMapContextDungeon extends Command
 {
+    use ResolvesMapContextScope;
     use SavesToFile;
 
     /**
@@ -17,7 +20,9 @@ class MakeMapContextDungeon extends Command
      *
      * @var string
      */
-    protected $signature = 'make:mapcontextdungeon {--output= : The output folder to place the generated map context in}';
+    protected $signature = 'make:mapcontextdungeon
+        {--output= : The output folder to place the generated map context in}
+        {--scope=all : Which dungeons to generate map context for (current-season, rest, all)}';
 
     /**
      * The console command description.
@@ -31,11 +36,17 @@ class MakeMapContextDungeon extends Command
      */
     public function handle(
         MapContextServiceInterface $mapContextService,
+        SeasonServiceInterface     $seasonService,
     ): int {
         $output = $this->option('output') ?? storage_path('mapcontext');
 
+        $dungeonIds = $this->resolveDungeonIdsForScope((string)$this->option('scope'), $seasonService);
+        if ($dungeonIds === null) {
+            return self::FAILURE;
+        }
+
         /** @var Collection<int, Dungeon> $dungeonsById */
-        $dungeonsById     = Dungeon::all()->keyBy('id');
+        $dungeonsById     = Dungeon::whereIn('id', $dungeonIds)->get()->keyBy('id');
         $allowedLanguages = language()->allowed();
 
         $bar = $this->output->createProgressBar($dungeonsById->count() * count($allowedLanguages));
--- a/app/Console/Commands/MapContext/MakeMapContextDungeon.php
+++ b/app/Console/Commands/MapContext/MakeMapContextDungeon.php
@@ -22,7 +22,7 @@ class MakeMapContextDungeon extends Command
      */
     protected $signature = 'make:mapcontextdungeon
         {--output= : The output folder to place the generated map context in}
-        {--scope=all : Which dungeons to generate map context for (current-season, rest, all)}';
+        {--scope=all : Which dungeons to generate map context for (priority, rest, all)}';
 
     /**
      * The console command description.
--- a/app/Console/Commands/MapContext/Traits/ResolvesMapContextScope.php
+++ b/app/Console/Commands/MapContext/Traits/ResolvesMapContextScope.php
@@ -1,0 +1,85 @@
+<?php
+
+namespace App\Console\Commands\MapContext\Traits;
+
+use App\Models\Dungeon;
+use App\Models\GameVersion\GameVersion;
+use App\Service\Season\SeasonServiceInterface;
+use Illuminate\Support\Collection;
+
+/**
+ * Resolves the `--scope` option shared by the map context generator commands into the set of
+ * dungeon IDs that should be processed.
+ */
+trait ResolvesMapContextScope
+{
+    private const string SCOPE_CURRENT_SEASON = 'current-season';
+
+    private const string SCOPE_REST = 'rest';
+
+    private const string SCOPE_ALL = 'all';
+
+    /** @var array<int, string> */
+    private const array VALID_SCOPES = [
+        self::SCOPE_CURRENT_SEASON,
+        self::SCOPE_REST,
+        self::SCOPE_ALL,
+    ];
+
+    /**
+     * Resolves the given `--scope` option value to the set of dungeon IDs it covers.
+     *
+     * Writes an error to the command's output and returns null if the scope value is not recognized.
+     *
+     * @return Collection<int, int>|null
+     */
+    protected function resolveDungeonIdsForScope(string $scope, SeasonServiceInterface $seasonService): ?Collection
+    {
+        if (!in_array($scope, self::VALID_SCOPES, true)) {
+            $this->error(sprintf(
+                'Invalid --scope value "%s", expected one of: %s',
+                $scope,
+                implode(', ', self::VALID_SCOPES),
+            ));
+
+            return null;
+        }
+
+        $allDungeonIds = Dungeon::query()->pluck('id');
+
+        return match ($scope) {
+            self::SCOPE_ALL  => $allDungeonIds,
+            self::SCOPE_REST => $allDungeonIds
+                ->diff($this->currentSeasonDungeonIds($seasonService))
+                ->values(),
+            default => $this->currentSeasonDungeonIds($seasonService),
+        };
+    }
+
+    /**
+     * Dungeon IDs belonging to the current season of every expansion that currently has an active
+     * season - this includes non-retail game versions (e.g. Classic-era) whenever they have one.
+     *
+     * @return Collection<int, int>
+     */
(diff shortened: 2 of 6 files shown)


app/Console/Commands/MapContext/Traits/ResolvesMapContextScope.php
```diff
@@ -0,0 +1,85 @@
...
+    private const string SCOPE_REST = 'rest';
+
+    private const string SCOPE_ALL = 'all';
+
+    /** @var array<int, string> */
+    private const array VALID_SCOPES = [
+        self::SCOPE_CURRENT_SEASON,
+        self::SCOPE_REST,
+        self::SCOPE_ALL,
+    ];
```
Wotuu: Should we convert this to a native Enum instead? I'm trying to do away with these string arrays if possible.

app/Console/Commands/MapContext/Traits/ResolvesMapContextScope.php
```diff
@@ -0,0 +1,85 @@
...
+
+            if ($currentSeason === null) {
+                continue;
+            }
+
+            $dungeonIds = $dungeonIds->merge($currentSeason->dungeons()->pluck('dungeons.id'));
+        }
+
+        return $dungeonIds->unique()->values();
+    }
```
Wotuu: This should also take into account the `$seasonService->getNextSeason()` if it's not null. There's a period of time where a next season is known but not active yet. It will start receiving traffic though. So both current and next season need to go in "priority" mode.

app/Console/Commands/MapContext/MakeMapContextDungeon.php
```diff
@@ -2,22 +2,27 @@
...
 
     /**
      * The name and signature of the console command.
      *
      * @var string
      */
-    protected $signature = 'make:mapcontextdungeon {--output= : The output folder to place the generated map context in}';
+    protected $signature = 'make:mapcontextdungeon
+        {--output= : The output folder to place the generated map context in}
+        {--scope=all : Which dungeons to generate map context for (current-season, rest, all)}';
```
Wotuu: `current-season` needs to be renamed to `priority` or something else to signify it'll be current + next season.

Wotuu (commented):
A few comments

app/Console/Commands/MapContext/Traits/ResolvesMapContextScope.php
Wotuu: :robot: Converted to a native backed enum — `App\Console\Commands\MapContext\Enums\MapContextScope` (`Priority`/`Rest`/`All`). The trait now does `MapContextScope::tryFrom($scope)` and matches on the enum cases instead of the string array. See 886604e5d.

app/Console/Commands/MapContext/Traits/ResolvesMapContextScope.php
Wotuu: :robot: Good catch. `priorityDungeonIds()` now also merges in `$seasonService->getNextSeasonOfExpansion($gameVersion->expansion)` alongside `getCurrentSeason()`, so both current and next season dungeons land in "priority" once a next season is known. Added a test (`resolveDungeonIdsForScope_givenPriorityAndNextSeasonKnown_includesNextSeasonDungeons`) that mocks `SeasonServiceInterface` with TWW S1/S2 (disjoint dungeon rosters) to prove both get merged, since Midnight doesn't have a season 2 in the DB yet to exercise this for real. See 886604e5d.

app/Console/Commands/MapContext/MakeMapContextDungeon.php
Wotuu: :robot: Renamed `current-season` to `priority` everywhere — both command signatures, the enum, the release-deploy.yml invocations (job was already named `build-map-context-priority`, just the `--scope=` value lagged), and the tests. See 886604e5d.

Wotuu:
:robot: Addressed all three review comments in 886604e5d:
1. `--scope` is now a native backed enum (`App\Console\Commands\MapContext\Enums\MapContextScope`) instead of a string const array.
2. `priority` now merges in each active season line's *next* season too (`getNextSeasonOfExpansion`), not just the current one.
3. Renamed `current-season` → `priority` everywhere (signatures, workflow, tests) to match.

Added a test that mocks `SeasonServiceInterface` with TWW S1/S2 (disjoint rosters) to prove the merge, since Midnight doesn't have a season 2 seeded yet to exercise it for real. `ResolvesMapContextScopeTest` (7 tests), `SeasonService`+`Console` groups (59 tests), `composer run fix`/`analyse` all green. PR description updated to match.

Wotuu (commented):
Ok

Issue #3535, opened 2026-07-13
Stale dungeon_difficulty is submitted and persisted when switching from a speedrun raid to a normal dungeon


Wotuu:
:robot: Found while planning #3534 (front-end tests for the create route flow).

## Bug

On the create route form, switching the dungeon select from a speedrun-enabled raid (e.g. Serpentshrine Cavern) to a non-speedrun dungeon leaves the previously built difficulty options — and the user's selection — in the now-hidden `dungeon_difficulty` select:

- `resources/assets/js/custom/inline/common/dungeonroute/create/dungeondifficultyselect.js` only calls `$dungeonDifficultySelectContainer.hide()` in the non-speedrun branch; it never clears the options or the selection.
- A hidden `<select>` is still a successful form control, so the browser POSTs the stale `dungeon_difficulty` (e.g. `1`).
- Validation accepts it: `Rule::in(DungeonConstants::DIFFICULTY_ALL)` is dungeon-agnostic.
- `DungeonRouteSaveService::resolveDungeonDifficulty()` returns the submitted difficulty unchanged when the dungeon is not `speedrun_enabled`, so the stale value is persisted onto a non-speedrun route.

## Repro

1. Switch to Classic, open the create route form.
2. Select Serpentshrine Cavern, pick 10 man.
3. Switch the dungeon select to a normal dungeon (e.g. The Deadmines) and submit.
4. The created route has `dungeon_difficulty = 1` even though the dungeon is not speedrun-enabled.

## Suggested fix

- Front end: in the else-branch of `dungeonSelectionChanged`, also remove the options (`$dungeonDifficultySelect.find('option').remove()` + `refreshSelectPickers()`) so nothing is submitted.
- Back end (defense in depth): `resolveDungeonDifficulty()` should return `null` when `!$dungeon->speedrun_enabled`.

The JS/contract tests from #3534 pin the current (buggy) behavior with a comment referencing this issue; update those pins when fixing.

PR #3533, opened 2026-07-13
#3514 Add in-depth feature tests for the Create Route flow

Wotuu:
:robot: Closes #3514

Extends `tests/Feature/Controller/DungeonRouteControllerTest.php` into a thorough suite for both route-creation endpoints — the most critical write path on the site, which had no creation-POST coverage before the v15.4.1 hotfix.

## What's covered

**`POST /new/temporary` (`saveNewTemporary`, guest-accessible)**
- happy path: guest `author_id = -1`, generated title, `expires_at` set, `season_id` set for the has-seasons (retail) default; logged-in user sets `author_id`
- `dungeon_id`: missing / nonexistent / inactive → validation error
- `dungeon_difficulty`: empty (the regression), each valid `1–4`, invalid (`0`, `99`, `'10_man'` slug); speedrun enabled-difficulty kept vs non-enabled → falls back to first enabled (`resolveDungeonDifficulty()`)
- `dungeon_route_level`: valid `X;Y` sets range, malformed rejected
- `dungeon_start_map_icon_id`: valid start icon set, foreign/bogus icon → resolves to `null` (`resolveDungeonStartMapIconId()`)

**`POST /new` (`saveNew`, authenticated)**
- happy path (`author_id` = user, no `expires_at`); title + description with `<script>` → `HtmlSanitizer` strips it
- `faction_id`: `FactionSelectionRequiredRule` on the faction-required dungeon (thenexus) — unspecified rejected, valid faction accepted; nonexistent faction rejected
- `team_id`: own team accepted, foreign team rejected, `-1` → no team
- inactive / missing `dungeon_id` rejected; admin `demo` flag; speedrun difficulty fallback

**Browser-contract cases** — a dataprovider sends `""` for each optional field a real browser submits empty (empty text inputs / empty Tom Select selects) and asserts creation still succeeds. Fields a browser never submits empty and that lack `nullable` (`dungeon_route_level`, `dungeon_route_sandbox`, `faction_id`, `route_select_affixes`) are documented and excluded, since for those an empty value is a genuine validation error rather than a browser case.

**Throttle** — one test proves the `throttle:create-dungeonroute` limiter/route wiring by driving the limit down and asserting a `429`.

## Notes
- Dungeons/factions/icons are selected by querying the same active criteria the form requests use, not hardcoded ids (the seeded DB's select/validation state differs).
- All created rows cleaned up with `try/finally` (no `RefreshDatabase`).

## Verification
[...]

Issue #3514, opened 2026-07-11
Add in-depth feature tests for the Create Route flow (all creation parameters)


Wotuu:
:robot: Follow-up to the v15.4.1 hotfix: creating any route through the front end failed with "The dungeon difficulty field value is not in the list of allowed values."

## Background

The Tom Select migration (#3420 / MR #3484, shipped in v15.4.0) changed what the browser submits for an optionless single select: Tom Select injects an `<option value="">` (bootstrap-select submitted nothing), so the create-route form started POSTing `dungeon_difficulty=""`. The `Rule::in(...)` validation on both submit form requests was not `nullable`, so every route creation failed. Fixed by hotfix on v15.4.1 (`nullable` added in `DungeonRouteSubmitFormRequest` + `DungeonRouteSubmitTemporaryFormRequest`).

Why no test caught it: **nothing in `tests/` had ever POSTed to `dungeonroute.savenew` or `dungeonroute.temporary.savenew`.** Existing DungeonRoute coverage is authorization (#3389/#3393 policy tests) and the Ajax editor tests, which create routes via factories. The first creation-flow test is the regression test added with the hotfix (`tests/Feature/Controller/DungeonRouteControllerTest.php`, 2 tests). Smoke-level "GET returns 200" coverage exists for other pages, but the creation POST — arguably the most critical write path on the site — had none.

## Remaining work

Extend `tests/Feature/Controller/DungeonRouteControllerTest.php` into a thorough suite for both creation endpoints:

**`POST /new/temporary` (`saveNewTemporary`, guest-accessible):**
- happy path per game version: current-season dungeon, non-season active dungeon, speedrun-enabled (classic raid) dungeon
- `dungeon_id`: missing, inactive dungeon, inactive expansion, nonexistent id → validation error
- `dungeon_difficulty`: absent, `""` (the regression), each valid value 1–4 on a speedrun dungeon, invalid (0, 99, `'10_man'` slug) → rejected; non-enabled difficulty on a speedrun dungeon → falls back to first enabled (`DungeonRouteSaveService::resolveDungeonDifficulty()`)
- `dungeon_route_level`: valid `X;Y`, out-of-season-bounds, malformed (`DungeonRouteLevelRule`)
- `dungeon_start_map_icon_id`: valid start icon, icon of another dungeon / another mapping version / wrong type → resolves to null instead of erroring (`resolveDungeonStartMapIconId()`)
[...]

PR #3536, opened 2026-07-13
#3534 Add front-end tests + FE/BE payload contract for the Create Route flow

### LINKED ISSUE #3534: Add front-end tests + FE/BE payload contract for the Create Route flow (follow-up to #3514)
:robot: Follow-up to #3514. That issue added in-depth backend feature tests for `POST /routes/new` and `POST /routes/new/temporary` (MR #3533), but the v15.4.0 regression was the **front end** producing malformed data: Tom Select injected `<option value="">` into the optionless `dungeon_difficulty` select, so the browser POSTed `dungeon_difficulty=""` and every route creation failed. Nothing tests the exact POST body the browser produces after the JS runs — that blind spot remains.

## Remaining work

**1. Vitest+jsdom tests exercising the real browser code path** — real inline JS classes (`resources/assets/js/custom/inline/...`: `createbase.js`, `dungeondifficultyselect.js`, `dungeonstartselect.js`, `affixes.js`, `composition.js`) + real Tom Select (via `selectpicker.js`) + real jQuery, asserting what `new FormData(form)` would serialize. Per-class unit tests plus one integration "form serialization" suite. DOM built by a shared hand-written builder mirroring the blade partials' exact field names/ids (`resources/assets/js/test/fixtures/createRouteForm.js`).

**2. Shared payload fixtures = the FE/BE contract** — `tests/Fixtures/CreateRoute/payloads/*.json`, each an array of `{name, value}` pairs (preserves duplicate `[]` keys) with `{{placeholder}}` tokens for DB-seeded ids. The JS integration test asserts the serialized form matches the fixture; a new PHP feature test (`tests/Feature/Controller/DungeonRouteCreateContractTest.php`, deliberately separate from MR #3533's file) POSTs the same fixtures to the real endpoints and asserts creation succeeds. If FE serialization or BE validation drifts, one side breaks.

**3. Scenarios (~12)**: retail minimal defaults (regression pin), multiple affixes, with team, without team, full group composition, attributes, multiple start icons, classic non-speedrun, classic speedrun Serpentshrine Cavern 10/25-man, Tempest Keep default difficulty, speedrun-then-switch-away (stale difficulty pin), guest temporary route.

**4. Local Puppeteer E2E script** (`sh/e2e/create-route.e2e.js`, not CI) — reuses the compose `chrome` CDP service: login, create a Retail route via real UI, switch game version to Classic, create a Serpentshrine Cavern speedrun route with difficulty, clean up created routes.

## Gotchas

- Inline classes are concatenated globals, not ES modules — add the guarded `module.exports` block (same as `inlinecode.js`) to the touched sources; no behavior change.
- Pin what the real code path actually emits for the untouched difficulty select (`""` vs absent) by running the integration test — don't guess in the fixture.
- Tom Select's `updateOriginalInput` reorders options, so per-key value comparison must be order-insensitive.
- Keep zero shared files with MR #3533 to stay conflict-free; dedupe overlap after both merge.

## Definition of done

- `npm test` green (new + existing suites), fixture chain proven by mutation check (reverting `nullable` on `dungeon_difficulty` in `DungeonRouteSubmitFormRequest` must fail the PHP contract test).
- `php artisan test --compact tests/Feature/Controller/DungeonRouteCreateContractTest.php` green, `composer run fix` + `composer run analyse` clean.
- E2E script runs green against the local stack for both Retail and Classic flows.

Wotuu:
:robot: Closes #3534

The v15.4.0 regression (#3514) was the front end producing malformed data — Tom Select injected `<option value="">` into the optionless `dungeon_difficulty` select, so the browser POSTed `dungeon_difficulty=""`. Backend tests alone can't catch that class of bug; this MR closes the blind spot from the JS side and ties both sides together with a shared payload contract.

## What's in here

**1. Vitest+jsdom tests running the real browser code path** — real inline classes (`createbase`, `dungeondifficultyselect`, `dungeonstartselect`, `affixes`, `composition`) + real Tom Select + real jQuery, asserting what `new FormData(form)` actually serializes. 23 per-class unit tests + a 13-scenario integration serialization suite (`createroute.serialization.test.js`). Full suite: 173 tests. The inline classes got the established guarded `module.exports` block (no behavior change).

**2. The FE/BE contract: `tests/Fixtures/CreateRoute/payloads/*.json`** — 13 hand-written payload fixtures ({name, value} pairs, `{{placeholder}}` tokens for seeded ids). The JS integration suite proves the front end serializes exactly these payloads; `tests/Feature/Controller/DungeonRouteCreateContractTest.php` (14 tests, 76 assertions) POSTs the same payloads to the real endpoints and proves the backend accepts them. If either side drifts, one suite breaks. **Mutation-verified**: removing `nullable` from `dungeon_difficulty` in `DungeonRouteSubmitFormRequest` fails the retail-minimal-defaults contract test with the exact v15.4.0 error message. A field-name hardening test keeps the hand-written JS DOM fixture honest against the rendered `/new` form.

Scenarios cover: retail defaults (the #3514 pin), multiple affixes, with/without team, full group composition, attributes, multiple start icons, classic non-speedrun, classic speedrun 10/25-man + default-difficulty, speedrun-then-switch-away, and guest temporary routes.

**3. Local E2E script `sh/e2e/create-route.e2e.js`** (not CI) — drives the real form in headless Chrome via the compose `chrome` service: Retail creation, game-version switch to Classic, Serpentshrine Cavern speedrun with 25-man difficulty, route cleanup, game-version restore. Validated live against the dev stack.

## Bugs/quirks found while building this (pinned as current behavior, NOT fixed here)

- **#3535 (filed)**: stale `dungeon_difficulty` is submitted and persisted when switching from a speedrun raid to a normal dungeon. Pinned in `classic-speedrun-then-switch.json`; the E2E probe reproduces it in a real Chrome.
- `dungeon_start_map_icon_id` serializes as `""` for non-speedrun dungeons but is entirely **absent** for speedrun dungeons (an option-stripping interaction between the difficulty and start-select refreshes).
- `seasonal_index` auto-select on affix clicks is dead in create mode (`_automaticSeasonalIndexChange` requires an existing `dungeonroute`).
- `composition.js` class/race cross-filter via `.find("[data-id=...]")` never matches (options don't carry `data-id`); only the faction-based filter path actually works.
- `DungeonRouteSubmitFormRequest` has no rules for `specialization`, so the submitted `specialization[]` is silently dropped by `validated()`; and `race[]` sentinel `"0"` values are persisted verbatim as `character_race_id = 0` rows (unlike `class[]`, which is existence-filtered).

## Coordination with MR #3533

Zero shared files: the PHP contract tests live in a new `DungeonRouteCreateContractTest.php`, not `DungeonRouteControllerTest.php`. After both merge, some overlap can be deduped.

## Verification

- `npm test`: 173/173 (runs in CI via js-tests.yml)
- `php artisan test --compact tests/Feature/Controller/DungeonRouteCreateContractTest.php`: 14 passed (76 assertions)
- `composer run analyse` + `composer run fix`: clean
- E2E script: both Retail and Classic flows green against the local stack, DB cleanup verified


Wotuu (commented):
You added some README.mds. Aren't they better suited to be a CLAUDE.md for that folder instead? They are for you - not for me.

Wotuu:
:robot: You're right — fixed in 72e4dbb.

Both files were written in second person addressed to a future Claude session (gotchas, "if you extend this, prefer `nativeClick()`", cross-references to `.claude/skills/...`), not general project documentation. Renamed `sh/e2e/README.md` → `sh/e2e/CLAUDE.md` and `tests/Fixtures/CreateRoute/README.md` → `tests/Fixtures/CreateRoute/CLAUDE.md`, content unchanged. As a bonus, nested `CLAUDE.md` files are auto-loaded as context by Claude Code when working in that directory tree, which a `README.md` never would be — so this isn't just naming, it makes the notes actually surface next time.

Wotuu (commented):
Ok now!

PR #3550, opened 2026-07-14
#3548 Fix local dungeon-route thumbnail generation & loading

### LINKED ISSUE #3548: Fix local dungeon-route thumbnail generation & loading (public disk + live worktree share)
:robot: _Plan drafted by Claude in a planning session; to be executed in a follow-up session._

## Context

Local dungeon-route thumbnails don't render (e.g. `/routes/retail/pit-of-saron`). URLs come out as
`http://localhost:<port>/storage//thumbnails/FvfZSWD/...` and 403/404. Root-causing found several
compounding problems, all **local-only** — production uses an S3 disk and is unaffected:

1. **Double-slash.** `ThumbnailService::THUMBNAIL_FOLDER_PATH = '/thumbnails'` bakes a leading slash
   into `File.path`. Laravel's `local`-disk URL builder (`getLocalUrl`) prepends `/storage/`
   **without** `ltrim` → `/storage//thumbnails/...`. S3/`public` disks use `concatPathToUrl` which
   ltrims, so they never showed it.
2. **The `local` disk can't serve plain `` URLs.** Its `serve => true` route runs `ServeFile`,
   which **requires a signed URL** unless the disk visibility is `public`. The `local` disk is
   private → a plain `Storage::disk('local')->url()` request returns **403** (confirmed by curl).
   `getURL()` never signs.
3. **Local generation is hard-disabled.** `doCreateThumbnail()` early-returns when
   `app()->environment('local')` — a stale guard from when local dev's `FILESYSTEM_DISK` pointed at
   real S3. The `seed-dev-routes` skill currently works around it by editing source (`false &&`).
4. **Puppeteer nav host is unreachable.** Generation navigates to `route('dungeonroute.preview')`
   = `APP_URL` = `http://localhost:<port>`, not reachable from inside the app container (nginx is a
   separate container). The skill works around it with a per-run `APP_URL=http://nginx` override.
5. **Worktrees ship with no thumbnail files.** A fresh worktree's `storage/app` is empty.

**Decisions made during planning:**
- Use Laravel's **`public` disk** locally (files in `storage/app/public/thumbnails`, served
  statically by nginx via the `storage:link` symlink — no signing, no double-slash). Prod stays s3.
- Worktrees **share the main checkout's thumbnails live** (implemented as a docker bind-mount, since
  a host symlink into the main checkout won't resolve inside the worktree container).
- Keep worktrees lean — **no Horizon/puppeteer**. Generation happens only on the main stack, which
  already has puppeteer + chromium (`/home/ksg/.cache/puppeteer`) + Horizon.

Verified: all existing dev thumbnail `File` rows are `disk=local`, `path=/thumbnails/...`;
pit-of-saron's files **exist** on the main disk (serving bug, not missing files); the framework
`public` disk is available (root `storage/app/public`, `url => APP_URL.'/storage'`, `visibility
public`); main lacks the `public/storage` symlink; main `.env` has `FILESYSTEM_DISK=local`.

---

### A1. Strip the leading slash from the folder constants
`app/Service/DungeonRoute/ThumbnailService.php:28-30`
```php
public const string THUMBNAIL_FOLDER_PATH        = 'thumbnails';
public const string THUMBNAIL_CUSTOM_FOLDER_PATH = 'thumbnails_custom';
```
`getTargetFilePath()` already uses `sprintf('%s/%s/%s', ...)`, so new paths become
`thumbnails/<key>/<file>.jpg`. Harmless on S3/public (they ltrim); fixes the raw local case.
Also drop the now-dead `mkdir($targetFolder ...)` at line 183-185 (it tried to `mkdir('/thumbnails')`
at FS root; `Storage::put` creates dirs itself).

### A2. Point local dev at the `public` disk + create the symlink
- `.env` (both dev machines) and `.env.example`: `FILESYSTEM_DISK=public`.
- Run `php artisan storage:link` on the **main** checkout (creates `public/storage ->
  storage/app/public`). Worktrees already have this symlink from their init.
- Result: `Storage::disk('public')->url('thumbnails/x.jpg')` →
  `http://localhost:<port>/storage/thumbnails/x.jpg`, served statically by nginx, no signing.
- Note: `FILESYSTEM_DISK` is the generic default, so other local `File` uploads also move to the
  `public` disk in dev — acceptable (prod is s3).

### A3. Migrate existing dev rows (`disk=local` → `public`)
Add a **production-guarded** Artisan command
`app/Console/Commands/DungeonRoute/MigrateThumbnailDisk.php` (abort if `app()->isProduction()`):
- For each `File` where `model_class = DungeonRouteThumbnail` and `disk = 'local'`: move the file
  from the `local` disk (`storage/app/private/<path>`) to the `public` disk
  (`storage/app/public/<path>`, stripping any leading slash), then update `disk = 'public'`,
  `path = ltrim(path,'/')`.
- Run once on the main stack. pit-of-saron's existing files then load immediately (no regeneration).

## Part B — Worktrees instantly show main's thumbnails (live share)

- `docker-compose.worktree.yml`: add a bind-mount to **both** `app` and `nginx`:
  `${MAIN_THUMBNAILS_DIR}:/var/www/storage/app/public/thumbnails`.
- `sh/worktree.sh` (`cmd_create`, alongside the existing artifact seeding): export
  `MAIN_THUMBNAILS_DIR="$REPO_ROOT/storage/app/public/thumbnails"` into the compose invocation, and
  `mkdir -p` that dir on main so the mount source always exists.
- Worktree already runs `storage:link` in its init, so `public/storage -> storage/app/public` is
  present; the bind-mount makes the `thumbnails` subdir resolve to main's live files.
- Net: main-stack generation appears in every worktree with no re-seeding.

### C1. Replace the env guard with a disk-safety guard
`app/Service/DungeonRoute/ThumbnailService.php:126-130`. Extract a testable private helper and gate
on the **target disk driver**, not the env — the real danger is a dev machine writing to remote S3:
```php
$disk   = config('filesystems.default', 'public');
$driver = config(sprintf('filesystems.disks.%s.driver', $disk));
if (app()->environment('local') && $driver === 's3') {
    $this->log->doCreateThumbnailSkippedRemoteDiskFromLocal();
    return null;
}
```
local+public/local → allowed; local+s3 → blocked (protects prod S3); prod+s3 → allowed (unchanged).
Compute `$disk` once and reuse it in `attachThumbnailToDungeonRoute` for consistency.

### C2. Config-driven internal preview URL (no global APP_URL override)
- `config/keystoneguru.php` `thumbnail` block: add
  `'preview_base_url' => env('THUMBNAIL_PREVIEW_BASE_URL')` (default null).
- `.env.example` (+ main `.env`): `THUMBNAIL_PREVIEW_BASE_URL=http://nginx`.
- In `ThumbnailService`, extract `getPreviewUrl(DungeonRoute, floorIndex, zoom): string`: when
  `preview_base_url` is set, build from the relative route (`route('dungeonroute.preview', $params,
  false)`) prefixed with that base; otherwise the absolute `route(...)` (prod → `keystone.guru`,
  reachable). Replaces the inline `route()` at line 153.
- This removes the skill's per-run `APP_URL=http://nginx` hack.

### C3. Update the `seed-dev-routes` skill
Rewrite for the new flow: drop the source-editing guard hack (now disk-gated), drop/replace step 3
"make thumbnails serve" (now the `public` disk + `storage:link` handle it — no `public/storage ->
private` symlink, no `chmod 0700` fix), and reference `THUMBNAIL_PREVIEW_BASE_URL` instead of
overriding `APP_URL`. Keep the environmental prereqs (node_modules hardlink, chrome-headless-shell +
apt deps, `ASSETS_BASE_URL_INTERNAL`/`DEBUGBAR_ENABLED` per-run toggles) — those stay main-stack
runbook items.

## Critical files
- `app/Service/DungeonRoute/ThumbnailService.php` — constants, guard→disk-safety, `getPreviewUrl()`.
- `app/Models/File.php` — reference only; `getURL()` needs no change once paths/disk are clean.
- `config/keystoneguru.php` — `thumbnail.preview_base_url`.
- `.env` / `.env.example` — `FILESYSTEM_DISK=public`, `THUMBNAIL_PREVIEW_BASE_URL=http://nginx`.
- `app/Console/Commands/DungeonRoute/MigrateThumbnailDisk.php` — new, dev-only migration.
- `docker-compose.worktree.yml`, `sh/worktree.sh` — live thumbnail bind-mount.
- `.claude/skills/seed-dev-routes/*` — rewrite for the new flow.

## Testing
- Unit `tests/Unit/Service/DungeonRoute/ThumbnailServiceTest.php`:
  - `getTargetFilePath` returns a path without a leading slash.
  - Disk-safety helper: `local`+`s3` → skip; `local`+`public`/`local` → proceed; `production`+`s3` → proceed.
  - `getPreviewUrl` returns the internal-base URL when `preview_base_url` is set, absolute otherwise.
- Feature test for `MigrateThumbnailDisk`: seed a `File` (`disk=local`) + a fake file on the `local`
  disk (`Storage::fake`), run the command, assert the row is `disk=public`, path de-slashed, and the
  file exists on the `public` disk. Clean up in `try/finally`.
- Follow the `writing-tests` skill (base test case, `#[Group]` at class level, AAA, naming).

## End-to-end verification
1. **Serving (worktree, existing data):** run `storage:link` on main, run `MigrateThumbnailDisk` on
   main, then on the worktree
   `curl -s -o /dev/null -w "%{http_code}" http://localhost:<port>/storage/thumbnails/<key>/<file>.jpg`
   → **200**; load `/routes/retail/pit-of-saron` and confirm the card `getURL()` has no `//` and the
   image renders (use the `headless-browser-verify` skill for a screenshot).
2. **Generation (main stack):** with `FILESYSTEM_DISK=public` + `THUMBNAIL_PREVIEW_BASE_URL=http://nginx`
   and the seed-dev-routes prereqs in place, run generation for a route **without editing
   ThumbnailService** — confirm a new file appears under `storage/app/public/thumbnails`, a `File`
   row is created with `disk=public`, and the image serves 200.
3. **Live worktree share:** confirm the thumbnail generated on main in step 2 is visible from the
   worktree URL without re-seeding (bind-mount working).
4. `composer run fix` and `composer run analyse`; stage only intended files.

Wotuu:
:robot: Closes #3548

## Summary

Local dungeon-route thumbnails were 403/404ing (`/storage//thumbnails/...`) and generation
was hard-disabled locally. Implements Parts A, B, C1, C2 from the plan on #3548:

- **A1** — `ThumbnailService::THUMBNAIL_FOLDER_PATH`/`THUMBNAIL_CUSTOM_FOLDER_PATH` drop their
  leading slash (fixes the `/storage//...` double-slash on the `local` disk's URL builder).
  Dropped the dead `mkdir($targetFolder)` call (operated on a disk-relative path, not a real
  filesystem path — `Storage::put` creates directories itself).
- **A2** — `.env.example`: `FILESYSTEM_DISK=public`. Laravel 12 merges the framework's default
  `filesystems.disks` on top of this app's `config/filesystems.php`, so the `public` disk
  (`storage/app/public`, served via the `storage:link` symlink, no signing) already exists —
  no new disk config needed.
- **A3** — new dev-only `dungeonroute:migratethumbnaildisk` command
  (`app/Console/Commands/DungeonRoute/MigrateThumbnailDisk.php`), guarded against production,
  moves existing `disk=local` thumbnail `File` rows + their bytes to the `public` disk and
  strips any leading slash from `path`.
- **C1** — replaced the blanket `app()->environment('local')` generation guard with a
  disk-safety check (`isRemoteDiskUnsafeForLocalGeneration`): only refuses when running locally
  **and** the target disk's driver is `s3` (protects against a local run mutating real S3).
  `local`+`public`/`local` now proceeds; `production`+`s3` is unchanged.
- **C2** — `config('keystoneguru.thumbnail.preview_base_url')` (env
  `THUMBNAIL_PREVIEW_BASE_URL`, e.g. `http://nginx` in dev) lets puppeteer reach the preview
  route from inside the app container without overriding the app's real `APP_URL`.
- **B** — worktrees live-share the main stack's generated thumbnails via a bind-mount
  (`docker-compose.worktree.yml` + `sh/worktree.sh`: `MAIN_THUMBNAILS_DIR` →
  `storage/app/public/thumbnails` on both `app` and `nginx`). Worktrees intentionally still run
  no Horizon/puppeteer of their own — a worktree-created route's thumbnail job lands in the
  shared Redis/Horizon queue and is generated by the **main** stack's app container, which this
  bind-mount then makes visible back in the worktree. Confirmed acceptable per the issue.

**Not included** (see issue discussion): Part C3 (rewriting the `seed-dev-routes` skill) —
that skill only exists on the still-open #3447 branch, not on `master`, so it's left for that
PR / a follow-up once merged. The main-checkout operational steps (editing the real `.env`,
running `storage:link` + the migration command on main) are runtime steps, not part of this
diff — happy to run them or hand off, whichever is preferred.

## Test plan

- [x] `tests/Feature/App/Service/DungeonRoute/ThumbnailServiceTest.php` — disk-safety guard
      (local+s3 skip / local+public proceed / production+s3 proceed), `getTargetFilePath` has
      no leading slash, `getPreviewUrl` (absolute vs. base-url-prefixed), existing
      `attachThumbnailToDungeonRoute` coverage updated for the new `$disk` param.
- [x] `tests/Feature/Console/Commands/DungeonRoute/MigrateThumbnailDiskTest.php` — migrates a
      local-disk file + de-slashes its path, skips a row whose file is missing from disk,
      refuses to run in production.
- [ ] End-to-end verification (serving, generation, live worktree share) — not yet run; needs
      the main-checkout operational steps above first.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -21,7 +21,7 @@ PHP_CLI_SERVER_WORKERS=4
 
 BCRYPT_ROUNDS=10
 DEBUGBAR_ENABLED=false
-FILESYSTEM_DISK=s3_user_uploads
+FILESYSTEM_DISK=public
 
 OCTANE_SERVER=swoole
 
@@ -82,6 +82,9 @@ QUEUE_CONNECTION=redis
 
 # A secret string that must be provided in order to gain access to the preview image
 THUMBNAIL_PREVIEW_SECRET=secret
+# Base URL puppeteer uses to reach the preview route from inside the app container (dev only;
+# the app's own APP_URL may point at a separate nginx container it can't reach)
+THUMBNAIL_PREVIEW_BASE_URL=http://nginx
 
 REVERB_APP_ID=
 REVERB_APP_KEY=
--- a/.env.example
+++ b/.env.example
@@ -21,7 +21,7 @@ PHP_CLI_SERVER_WORKERS=4
 
 BCRYPT_ROUNDS=10
 DEBUGBAR_ENABLED=false
-FILESYSTEM_DISK=public
+FILESYSTEM_DISK=s3_user_uploads
 
 OCTANE_SERVER=swoole
 
@@ -82,9 +82,6 @@ QUEUE_CONNECTION=redis
 
 # A secret string that must be provided in order to gain access to the preview image
 THUMBNAIL_PREVIEW_SECRET=secret
-# Base URL puppeteer uses to reach the preview route from inside the app container (dev only;
-# the app's own APP_URL may point at a separate nginx container it can't reach)
-THUMBNAIL_PREVIEW_BASE_URL=http://nginx
 
 REVERB_APP_ID=
 REVERB_APP_KEY=
--- a/.env.docker.example
+++ b/.env.docker.example
@@ -19,7 +19,7 @@ AUTH_PASSWORD_RESET_TOKEN_TABLE=password_resets
 
 BCRYPT_ROUNDS=10
 DEBUGBAR_ENABLED=true
-FILESYSTEM_DISK=local
+FILESYSTEM_DISK=public
 CACHE_FILE_PATH=/tmp/laravel-cache
 
 OCTANE_SERVER=swoole
@@ -84,6 +84,9 @@ QUEUE_CONNECTION=redis
 
 # A secret string that must be provided in order to gain access to the preview image
 THUMBNAIL_PREVIEW_SECRET=secret
+# Base URL puppeteer uses to reach the preview route from inside the app container (dev only;
+# the app's own APP_URL may point at a separate nginx container it can't reach)
+THUMBNAIL_PREVIEW_BASE_URL=http://nginx
 
 REVERB_APP_ID=
 REVERB_APP_KEY=
--- a/app/Console/Commands/DungeonRoute/MigrateThumbnailDisk.php
+++ b/app/Console/Commands/DungeonRoute/MigrateThumbnailDisk.php
@@ -1,0 +1,67 @@
+<?php
+
+namespace App\Console\Commands\DungeonRoute;
+
+use App\Models\DungeonRoute\DungeonRouteThumbnail;
+use App\Models\File;
+use Illuminate\Console\Command;
+use Illuminate\Support\Collection;
+use Illuminate\Support\Facades\Storage;
+
+class MigrateThumbnailDisk extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'dungeonroute:migratethumbnaildisk';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Dev-only: moves dungeon route thumbnail files (and their File rows) from the local disk to the public disk.';
+
+    public function handle(): int
+    {
+        if (app()->isProduction()) {
+            $this->error('This command must not be run in production.');
+
+            return 1;
+        }
+
+        /** @var Collection<int, File> $files */
+        $files = File::query()
+            ->where('model_class', DungeonRouteThumbnail::class)
+            ->where('disk', 'local')
+            ->get();
+
+        $this->info(sprintf('Found %d thumbnail file(s) on the local disk.', $files->count()));
+
+        $migrated = 0;
+        foreach ($files as $file) {
+            $path = ltrim($file->path, '/');
+
+            if (!Storage::disk('local')->exists($path)) {
+                $this->warn(sprintf('File #%d: %s does not exist on the local disk, skipping.', $file->id, $path));
+
+                continue;
+            }
+
+            Storage::disk('public')->put($path, Storage::disk('local')->get($path));
+
+            $file->update([
+                'disk' => 'public',
+                'path' => $path,
+            ]);
+
+            $migrated++;
+        }
+
+        $this->info(sprintf('Migrated %d file(s) to the public disk.', $migrated));
+
+        return 0;
+    }
+}
--- a/app/Service/DungeonRoute/Logging/ThumbnailServiceLogging.php
+++ b/app/Service/DungeonRoute/Logging/ThumbnailServiceLogging.php
@@ -56,7 +56,7 @@ class ThumbnailServiceLogging extends StructuredLogging implements ThumbnailServ
         $this->info(__METHOD__);
     }
 
-    public function doCreateThumbnailSkippedLocalEnvironment(): void
+    public function doCreateThumbnailSkippedRemoteDiskFromLocal(): void
     {
         $this->info(__METHOD__);
     }
--- a/app/Service/DungeonRoute/Logging/ThumbnailServiceLoggingInterface.php
+++ b/app/Service/DungeonRoute/Logging/ThumbnailServiceLoggingInterface.php
@@ -39,7 +39,7 @@ interface ThumbnailServiceLoggingInterface
 
     public function doCreateThumbnailMaintenanceMode(): void;
 
-    public function doCreateThumbnailSkippedLocalEnvironment(): void;
+    public function doCreateThumbnailSkippedRemoteDiskFromLocal(): void;
 
     public function doCreateThumbnailProcessStart(string $commandLine): void;
 
--- a/app/Service/DungeonRoute/ThumbnailService.php
+++ b/app/Service/DungeonRoute/ThumbnailService.php
@@ -25,9 +25,9 @@ use Throwable;
 
 class ThumbnailService implements ThumbnailServiceInterface
 {
-    public const string THUMBNAIL_FOLDER_PATH = '/thumbnails';
+    public const string THUMBNAIL_FOLDER_PATH = 'thumbnails';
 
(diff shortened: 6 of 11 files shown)


.env.example
```diff
@@ -82,6 +82,9 @@ QUEUE_CONNECTION=redis
 
 # A secret string that must be provided in order to gain access to the preview image
 THUMBNAIL_PREVIEW_SECRET=secret
+# Base URL puppeteer uses to reach the preview route from inside the app container (dev only;
+# the app's own APP_URL may point at a separate nginx container it can't reach)
+THUMBNAIL_PREVIEW_BASE_URL=http://nginx
```
Wotuu: I think this should be empty? `.env.example` is for new environments (such as staging/prod, not dev), `.env.docker.example` is where this should go too. Same for the `FILESYSTEM_DISK` change

Wotuu (commented):
One comment

.env.example
Wotuu: :robot: Good catch — moved both to `.env.docker.example` instead in aa63e649c.

Wotuu (commented):
Ok

PR #3552, opened 2026-07-14
#3535 Clear stale dungeon_difficulty when switching away from a speedrun dungeon

### LINKED ISSUE #3535: Stale dungeon_difficulty is submitted and persisted when switching from a speedrun raid to a normal dungeon

Wotuu:
:robot:

Closes #3535

## Summary
- `dungeondifficultyselect.js` only hid the difficulty select's container when switching away from a speedrun dungeon; it never cleared the select's options/selection, so a hidden `<select>` (still a successful form control) posted the stale difficulty value from the previously-selected speedrun dungeon.
- The else-branch now also clears the select's options (only when there are options to clear, to avoid prematurely triggering a page-wide `refreshSelectPickers()` sync on unrelated `.selectpicker` elements that haven't activated yet).
- `DungeonRouteSaveService::resolveDungeonDifficulty()` now discards any submitted difficulty for a non-speedrun dungeon (defense in depth), instead of passing it through unchanged.

## Test plan
- [x] Updated JS unit test (`dungeondifficultyselect.test.js`) and integration test (`createroute.serialization.test.js`) + shared fixture (`classic-speedrun-then-switch.json`) to assert the fixed behaviour instead of pinning the bug.
- [x] Updated `DungeonRouteSaveServiceSaveTest` and `DungeonRouteCreateContractTest` to assert the difficulty is discarded for non-speedrun dungeons.
- [x] `npx vitest run` — 199 passed.
- [x] `php artisan test --compact` on the affected DungeonRoute/contract test files — 104 passed.

### DIFF
--- a/app/Service/DungeonRoute/DungeonRouteSaveService.php
+++ b/app/Service/DungeonRoute/DungeonRouteSaveService.php
@@ -436,21 +436,27 @@ readonly class DungeonRouteSaveService implements DungeonRouteSaveServiceInterfa
 
     /**
      * Resolves the dungeon difficulty for speedrun-enabled dungeons: keeps the chosen difficulty when it is one of the
-     * dungeon's enabled speedrun difficulties, otherwise falls back to the first enabled difficulty.
+     * dungeon's enabled speedrun difficulties, otherwise falls back to the first enabled difficulty. Non-speedrun
+     * dungeons never have a difficulty, so a submitted value is discarded (defense in depth: see GitHub issue #3535,
+     * where a stale value from a previously-selected speedrun dungeon could otherwise survive a dungeon switch).
      */
     private function resolveDungeonDifficulty(Dungeon $dungeon, ?int $difficulty): ?int
     {
-        if ($difficulty !== null && $dungeon->speedrun_enabled) {
-            $enabledDifficulties = $dungeon->getEnabledSpeedrunDifficulties();
-
-            if (in_array($difficulty, $enabledDifficulties, true)) {
-                return $difficulty;
-            }
-
-            return $enabledDifficulties[0] ?? $difficulty;
+        if (!$dungeon->speedrun_enabled) {
+            return null;
         }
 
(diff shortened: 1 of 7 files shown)


Wotuu (commented):
Ok

PR #3497, opened 2026-07-10
#3480 Retire DB-seeded releases in favour of GitHub Releases

Wotuu:
Closes #3480

Makes GitHub Releases the single source of truth for release notes and retires the DB tables, the 367 JSON seeder files, and the hardcoded-id machinery.

## What changed
- **Models/repositories:** Removed `Release`, `ReleaseChangelog`, `ReleaseChangelogCategory`, `ReleaseChangelogChange` and their repositories/interfaces/provider bindings. `ReleaseReportLog` is kept but re-keyed by `version` (instead of `release_id`) so the announce dedupe survives.
- **Seeders/data:** Deleted all 367 `database/seeders/releases/v*.json` files, `ReleasesSeeder`, `ReleaseChangelogCategorySeeder`, and their `DatabaseSeeder` entries.
- **Migration:** One migration drops `releases`, `release_changelogs`, `release_changelog_changes`, `release_changelog_categories`, backfilling `release_report_logs.version` from `releases` before the drop.
- **Site:** Removed the on-site changelog list, single release page, and the shared render partial. The footer "changelog" link now points at the GitHub Releases page; the footer/Rollbar version string is sourced without the release tables.
- **Admin/commands:** Removed the admin release CRUD, nav link, datadump export, the now-dead `release:{export,save,current,body,success}` and `make:githubrelease*` commands, and the Reddit API service. `release:report` now dedupes by version.
- **Docs/tests:** Updated the `create-release` skill to the new flow, plus composers, breadcrumbs, translations, and tests.

## Verification
- `composer run analyse` — clean (no errors).
- `composer run fix` — clean (one unrelated pre-existing drift file was reverted to keep the diff focused).
- Full test suite green: **991 passed**. The only "failure" is `MapTilesExistenceTest` (`MapTiles` group), which requires the sibling `keystone.guru.assets` tiles repo that the worktree stack doesn't mount — it is **excluded from CI in every shard** (`--exclude-group=MapTiles`) and passes on the main stack. Unrelated to this change.

## Notes for review
- The drop migration is intentionally left **Pending** locally (the worktree shares the main DB; running it would drop the tables from under the `master` stack). CI runs migrations on a fresh DB, so the drop and the `release_report_logs` re-key are exercised there. It applies on the production deploy after merge.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #3480, opened 2026-07-09
Retire DB-seeded releases in favour of GitHub Releases as source of truth


Wotuu:
## Background

The release-notes system is built around per-release JSON seeder files that are re-seeded into every environment's DB on each deploy. Now that the project is trunk-based (the `development` branch is gone), the delivery half was cleaned up (release JSON is committed directly to `master`, see the `create-release` skill Step 8), but the underlying DB-seeded machinery still carries a lot of weight:

- `database/seeders/releases/v*.json` — **367 files**, each with **hardcoded sequential** `id` / `release_changelog_id` / `release_id`. Adding a release means hand-computing the next ids.
- `database/seeders/ReleasesSeeder.php` reads **all** files every run and does a full wipe-and-replace via `*_temp` tables + one atomic `RENAME TABLE` swap. Runs on **every deploy** via `environment:update`.
- A legacy admin CRUD UI (`admin.release.*`) whose only remaining job was the copy-paste JSON textarea, now superseded by the `create-release` skill.

**Proposal:** make **GitHub Releases the single source of truth** for release notes, and retire the DB tables + JSON seeder + hardcoded-id machinery. The site footer links out to the GitHub Releases page. Announcements (Discord/Reddit) and the `silent` flag get reworked.

This issue is a **design/plan for a later effort** — not something to build immediately. The full plan was worked out from a codebase exploration; this issue captures it so it can be picked up cold.

## Why this is worth doing

Removes: hardcoded-id brittleness, 367 JSON files, full-reseed-every-deploy, the temp-table swap complexity, the legacy admin CRUD, and the copy-paste flow.

Cost/tradeoffs to accept: a GitHub runtime/link dependency for release notes, loss of the structured category + `is_public` DB model (it becomes markdown in the Release body), and a one-time history backfill.

## How the release system works today (facts)

- Tag/deploy is already separate: `make:githubrelease vX.X.X --hash=<commit>` creates a GitHub Release + `v*` tag via the API using the **public** body only; that tag triggers `.github/workflows/release-deploy.yml`.
- On production, `environment:update` calls `make:githubrelease` (DB -> GitHub Release, public body) + `git pull`. It does **not** post to Discord/Reddit.
- The GitHub Release body (`Release::getGithubBodyAttribute()`) **already is** the public, category-grouped changelog. Non-public changes live only in the release **issue** (`getGithubFullBodyAttribute()`).
[...]

PR #3554, opened 2026-07-14
#3553 Defer release_* table drops to a guarded follow-up migration

### LINKED ISSUE #3553: Drop release_* tables (expand/contract follow-up to #3480)
:robot: Expand/contract follow-up to #3480.

v15.4.5 removed all code that used the `releases` / `release_changelog*` tables and `release_report_logs.release_id`, but the destructive drop migration (`2026_07_09_000000_drop_release_tables.php`) could not ship in that same release without 500-ing still-running v15.4.4 containers during the non-atomic deploy window (same failure mode as #3497). For v15.4.5 the drops were neutralized in production by hotfixing an **additive-only** version of that migration (kept `release_report_logs.version` add + backfill + index, dropped nothing).

Now that v15.4.5 is fully live and nothing references those tables, ship the real drops as the contract half of expand/contract:

- [ ] Edit `2026_07_09_000000_drop_release_tables.php` to the additive-only body (so master / fresh DBs match what prod actually ran).
- [ ] Add a new **guarded** migration that drops `release_report_logs.release_id` and tables `releases`, `release_changelogs`, `release_changelog_categories`, `release_changelog_changes`. Must be idempotent: staging already ran the original destructive body, prod still has the tables, fresh DBs also have them.
- [ ] Ship in the next release (v15.4.6+). By then the last consumer (v15.4.5) is already live, so the drop is safe.

Wotuu:
:robot: Closes #3553

Expand/contract follow-up to #3480. The original `2026_07_09_000000_drop_release_tables` dropped the `release_*` tables and `release_report_logs.release_id` **in the same release (v15.4.5)** that removed their code consumers — which would have 500'd old still-running v15.4.4 containers during the non-atomic deploy window (same failure mode as #3497). For v15.4.5 the drops were neutralised in production via an **additive-only hotfix** of that migration file.

This lands the proper split on master:

- **`2026_07_09_000000_drop_release_tables`** is reduced to its **additive half** — add + backfill `release_report_logs.version` + index — matching what production actually ran via the hotfix. No drops.
- **`2026_07_14_000000_drop_release_tables_followup`** (new) performs the destructive drops (`release_id` column + `releases`, `release_changelogs`, `release_changelog_categories`, `release_changelog_changes`), **guarded** with `hasColumn` / `dropIfExists` so it is idempotent across all environment states:
  - **staging** — original destructive body already ran, tables/column already gone → no-op;
  - **production** — neutralised via hotfix, tables/column still present → dropped;
  - **fresh DBs / CI** — tables/column present → dropped.

Ships the destructive half one release later (v15.4.6+), by which point the last consumer (v15.4.5) is already live everywhere, so the drop is safe.

## Verification

Ran both migrations against an isolated throwaway SQLite database (no shared DB touched) across both real-world states:

- **Fresh/prod:** additive-only `07_09` adds `version` and drops nothing; follow-up then drops `release_id` + all four tables and retains `version`. ✅
- **Staging:** follow-up runs clean when the tables/column are already absent, and is idempotent on a repeat run. ✅

`composer run fix` and `composer run analyse` (PHPStan) both clean.

### DIFF
--- a/database/migrations/2026_07_09_000000_drop_release_tables.php
+++ b/database/migrations/2026_07_09_000000_drop_release_tables.php
@@ -22,24 +22,20 @@ return new class extends Migration {
             ->update(['release_report_logs.version' => DB::raw('releases.version')]);
 
         Schema::table('release_report_logs', static function (Blueprint $table): void {
-            $table->dropColumn('release_id');
             $table->index('version');
         });
 
-        Schema::dropIfExists('release_changelog_changes');
-        Schema::dropIfExists('release_changelogs');
-        Schema::dropIfExists('release_changelog_categories');
-        Schema::dropIfExists('releases');
+        // The destructive drops (release_id column + release_* tables) are deferred to the follow-up
+        // migration 2026_07_14_000000_drop_release_tables_followup so that old still-running containers
+        // keep working during the non-atomic deploy window (expand/contract, see #3553).
     }
 
     /**
-     * Reverse the migrations. The dropped tables (and their seeded contents) are not restored - release
-     * notes now live on GitHub Releases.
+     * Reverse the migrations.
      */
     public function down(): void
     {
         Schema::table('release_report_logs', static function (Blueprint $table): void {
-            $table->integer('release_id')->after('id')->default(0);
             $table->dropIndex(['version']);
             $table->dropColumn('version');
         });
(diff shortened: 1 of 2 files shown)


Wotuu (commented):
Ok

PR #3562, opened 2026-07-15
#2460 Fix Rollbar logging

### LINKED ISSUE #2460: Fix Rollbar logging
logging.php
AppServiceProvider.php
RollbarStructuredLogging.php

Wotuu:
:robot: Closes #2460

## Summary

Restores the backend Rollbar integration, which went dark during the Laravel 11 upgrade (commit `3adc23373` disabled it, and the automated Laravel 11 config rewrite `eb8762ae2` later stripped the leftover dead channel config).

- `AppServiceProvider::boot()` — uncommented the `Rollbar::init()` call in the `booted()` callback. Fixed up two things that no longer matched the current codebase since the plan in #2460 was written:
  - The `App\Models\Release` model referenced by the old commented-out code was deleted in #3497 (release notes moved to GitHub Releases). `code_version` now comes from `ViewServiceInterface::getAppVersionInfo()['version']` — the same source the frontend Rollbar composer (`RollbarComposer`) already uses, so backend/frontend report the same version string.
  - `optional($user)->id` was modernized to `$user?->id` (nullsafe is already the sole convention used elsewhere in this codebase; `optional()` had zero other usages).
  - Extended `enabled` to also skip `app()->runningUnitTests()`, matching `StructuredLogging::__construct()`'s existing convention, so CI/PHPUnit runs stay quiet.
- `config/logging.php` — stripped the now-meaningless commented-out `/*'discord', 'rollbar'*/` remnants from the `stack_docker`/`stack_docker_local` channel arrays (Rollbar goes through the direct SDK, never a Monolog channel).
- Replaced the abstract `RollbarStructuredLogging` base class with an `App\Logging\Concerns\InteractsWithRollbar` trait, so structured-logging classes opt in without an extra inheritance layer. Migrated all 30 classes that extended it (one more than the 29 estimated in the issue) to `extends StructuredLogging` + `use InteractsWithRollbar;`.
- Added `tests/Unit/App/Logging/Concerns/InteractsWithRollbarTest.php` — asserts `Rollbar::logger()` is included once `Rollbar::init()` has run, and that resolving default loggers doesn't throw when it hasn't.

`.env.example` / `.env.docker.example` / `.env.ci.example` already had the `ROLLBAR_*` keys, so no env changes were needed.

## Test plan

- [x] `composer run fix` / `composer run analyse` — clean (one pre-existing PHPStan `nullsafe.neverNull` false positive on the restored `$user?->id ?? 0` line, suppressed inline the same way this codebase already does elsewhere, e.g. `AjaxKillZoneController.php`)
- [x] `php artisan test --compact --group=Logging` — 16 passed
- [x] `php artisan test --compact --exclude-group=MapTiles` (full suite) — 1074 passed
- [ ] Manually confirm a real log line reaches the Rollbar dashboard once deployed (can't be asserted by an automated test, per the issue plan)


Wotuu (commented):
Ok

PR #3563, opened 2026-07-15
#3380 Assign mapping version by MDT version/timestamp on import

### LINKED ISSUE #3380: Assign proper mapping version based on MDT version/timestamp in the string
This way we most closely resemble when the string was made and with what version it was built. A newly generated route can then be created using an outdated mapping version, but offer an upgrade.

A year or 2 ago MDT started including what version the string was built with in the `addonVersion: 6114,` in the root of the MDT string JSON. Based on this string we can determine when this release was made (through the MDT Github) and when the next release comes, and during that window look what mapping version was active.

We can then import MDT strings using the correct MDT version. We can also assign MDT versions to mapping versions and then import using the correct mapping version. Note that MDT may release v6.2.5, when I'm on v6.2.2, and I skip to v6.2.5, entirely going over 3 and 4. So this also needs to be handled.

Development contains a dump of the production database. It contains thousands of routes, some of which may be imported through MDT. If this is the case, an entry exists in `mdt_imports` table. You can use this data to cross-reference import dates and MDT versions and test any implementation you make. If you use this data for your tests, ensure there's a copy of the data in the tests available.

Wotuu:
:robot: Closes #3380

## What & why

An imported MDT string carries an `addonVersion` (the MDT addon's semver with dots stripped, e.g. `6115` for v6.1.15). Until now the importer ignored it and always attached the **newest** mapping version, so a route imported from an older MDT string was never flagged as outdated.

This attaches the route to the mapping version matching the MDT version the string was built with, so the existing "new mapping version → upgrade" flow (`isLatestForDungeon()` → header prompt + `dungeonroute.upgrade`) offers an upgrade. No UI changes were needed.

## Why date-based, not int-based

`addonVersion` is `tonumber(version:gsub("%.",""))` and is **not orderable** across MDT's historical schemes — verified against the production dump: `40120` (2022) → `5014` (2024) → `6115` (2026), yet `40120 > 5014`. So the int is only a **lookup key** into a committed `addonVersion → GitHub release-date` map; all comparisons are on dates. Digit-stripping the GitHub release tags reproduces every observed `addonVersion` exactly (`v4.0.1.20`→`40120`, `5.0.14`→`5014`, `6.1.15`→`6115`).

## Approach

- **`MDTAddonVersionService`** + committed `database/data/mdt/addon_versions.json` (built from the MDT GitHub releases). `php artisan mdt:syncaddonversions --refresh` rebuilds the map; the plain command backfills existing mapping versions.
- Additive nullable **`mapping_versions.mdt_addon_version`** = the MDT version each mapping version was imported from. Populated going forward in `MappingService`; existing rows backfilled from release dates vs `created_at`.
- **`Dungeon::getMappingVersionForMdtAddonVersion()`** — selects the *oldest* mapping version whose imported-from release date is **≥** the string's release date. So a version released *between* two imports (e.g. v6.1.14 between imports of v5.0.14 and v6.1.15) attaches to the **later** mapping version, and skipped MDT versions are handled. Falls back to the current mapping version when the string has no / unknown / newer `addonVersion`.
- Threads the chosen mapping version through the importer so sub-importers no longer re-resolve the current one; the chosen version is re-fetched as a single model to satisfy the lazy-loading guard.

## Scope

New imports only — existing routes keep their `mapping_version_id` (retroactive re-assignment is a possible follow-up).

## Tests

- Unit-style coverage of the selection algorithm (exact match, in-between → later version, non-orderable-int trap, newer/older-than-all, absent/unknown, same-date tiebreak, NULL-column `created_at` fallback).
- Integration test using a **real committed production import string** (MDT v5.0.7) asserting it resolves to a non-latest mapping version, plus a data-map guard.
- Full existing MDT suite green; `composer run fix` + `analyse` clean.


app/Models/Dungeon.php
```diff
@@ -283,6 +284,78 @@ public function hasMappingVersionForGameVersion(GameVersion $gameVersion): bool
...
+     */
+    public function getMappingVersionForMdtAddonVersion(?int $addonVersion, ?GameVersion $gameVersion = null): ?MappingVersion
+    {
+        $currentMappingVersion = $this->getCurrentMappingVersion($gameVersion);
+
+        if ($addonVersion === null || $addonVersion === 0 || $currentMappingVersion === null) {
+            return $currentMappingVersion;
+        }
+
+        $mdtAddonVersionService = app(MDTAddonVersionServiceInterface::class);
```
Wotuu: Calling upon a Service in a model is tell-tale sign this code should not be here. It should be moved to perhaps the `MappingVersionService` instead. Then, the `MDTAddonVersionServiceInterface` can simply require the service in the constructor and resolve the dependency that way.

app/Service/MDT/MDTAddonVersionService.php
```diff
@@ -0,0 +1,71 @@
...
+                /** @var array<string, string> $decoded */
+                $decoded = json_decode(File::get($path), true) ?? [];
+                foreach ($decoded as $addonVersion => $publishedAt) {
+                    $this->releaseDates[(int)$addonVersion] = Carbon::parse($publishedAt);
+                }
+            }
+        }
+
+        return $this->releaseDates;
+    }
```
Wotuu: This is a pattern that's never done in this project and I don't really agree with it. I think it's better to have the database handle this. Import the `addon_versions.json` into the database, with a model, repository etc. Then this service can dep inject the repository and perform a query to retrieve the data.

Wotuu (commented):
Refactoring required.

Wotuu:
:robot: Pushed a refactor commit (`dde6e627c`) addressing both review comments.

**1. Model calling a service (`Dungeon::getMappingVersionForMdtAddonVersion`)**
Moved the selection logic into `MappingService` (which already constructor-injects `MDTAddonVersionServiceInterface`), taking the dungeon as a parameter. Removed the method + service import from `Dungeon`, and updated the two `MDTImportStringService` call sites, injecting `MappingServiceInterface` there.

**2. Reading `addon_versions.json` at runtime**
Imported the JSON into the database:
- New `MDTAddonVersion` model + `mdt_addon_versions` table (additive migration).
- `MDTAddonVersionRepository` (`findReleaseDate` / `findLatestAddonVersionAtDate`) + binding.
- `MDTAddonVersionSeeder` imports the still-committed JSON (idempotent upsert). It populates the real table directly rather than via the temp-table swap, because the MDT class basename snake-cases to `m_d_t_addon_versions`, which `DatabaseSeeder::getTempTableName()` can't address — same reason `MappingVersionSeeder` manages its own table.
- `MDTAddonVersionService` now queries the repository (no more file read / in-memory map). `mdt:syncaddonversions --refresh` writes the JSON *and* upserts the DB.

The committed JSON stays as the single seed source of truth (mirrors how `DungeonDataSeeder` seeds committed JSON into tables), so there's no second drifting source.

Repository lives at the root `App\Repositories` namespace to match the sibling `MDTImportRepository` rather than a `{Domain}` subfolder — happy to move it if you'd prefer.

**Tests**: relocated the selection test to `tests/Feature/App/Service/Mapping/` (all 9 algorithm cases preserved, retargeted to the service), retargeted the MDT import-string integration test, and added repository + seeder coverage. Full `MDT` + `Mapping` groups green; `composer run fix` + `analyse` clean.

**⚠️ One thing to confirm on your side — production population.** The change is additive/backward-compatible, but the new code needs the table to both *exist* (migration) and be *populated* (seeder): an empty-but-migrated table degrades gracefully to the current mapping version, but nothing gets attached to older versions until it's seeded. CI seeds it, but I can't see from this repo whether prod deploys run the reference-data seeders. If they don't, `php artisan mdt:syncaddonversions --refresh` populates it manually. Can you confirm your deploy seeds reference tables?

Wotuu (commented):
Ok

Issue #3564, opened 2026-07-16
Ban specific IP addresses from accessing the site at the application layer


Wotuu:
:robot: _Filed by Claude on behalf of @Wotuu._

## Background

Rate limiting keys off the real visitor IP (`$request->ip()`), which is resolved from
CloudFlare's `CF-Connecting-IP` header in `app/Http/Middleware/TrustProxies.php` (see the
related fix that makes per-visitor IP resolution reliable). Now that we can trust the
real client IP, we want to act on it directly.

Today, blocking an abusive IP means manually creating a CloudFlare security rule. That is
slow, lives outside the app, and can't be managed by admins from the site itself. During a
recent scriptkiddy probing episode this was the only lever available.

## Goal

Detect a banned IP as early as possible in the request lifecycle and refuse to serve it
anything (return `403`/deny, ideally with no view rendering, DB work, or downstream
middleware). Bans should be managed from within the application.

## Remaining work / proposed scope

1. **Storage** — a `banned_ip_addresses` table + `BannedIpAddress` model (with a repository
   per project convention: interface in `app/Repositories/Interfaces/...`, implementation in
   `app/Repositories/Database/...`, binding in `RepositoryServiceProvider`). Suggested
   columns: `ip_address` (support both single IPs and CIDR ranges), `reason` (nullable),
   `expires_at` (nullable, for temporary bans), `created_by` (admin user id), timestamps.
2. **Enforcement middleware** — a new global middleware registered early in
   `bootstrap/app.php` (after `TrustProxies` so `$request->ip()` is the real visitor, but
   before heavy middleware / routing). It checks the resolved IP against the ban list and
   short-circuits with a `403` when matched. Use a cached lookup (Redis) so the check is a
   single fast in-memory match per request, not a DB query — invalidate the cache on
   ban add/remove. Reuse `Symfony\Component\HttpFoundation\IpUtils::checkIp` so CIDR ranges
   are supported.
3. **Admin management UI** — list / add / remove bans from the admin panel (see the
   `admin-tools-page` skill and existing admin pages for the pattern). Show reason, who
   banned, and expiry. Ideally a one-click "ban this IP" from somewhere the offending IP is
   already visible.
4. **Optional convenience** — an Artisan command to ban/unban an IP from the CLI for
   incident response (`artisan-command-conventions` skill).

## Risks / notes

- The enforcement middleware runs on **every** request, so the lookup must be cache-backed
  and cheap; a per-request DB query is not acceptable.
[...]

PR #3566, opened 2026-07-16
#3565 Resolve real visitor IP from CF-Connecting-IP for rate limiting

### LINKED ISSUE #3565: Rate limiting buckets swaths of users under CloudFlare edge IPs

## Problem

Rate limiting in `app/Providers/AppServiceProvider.php` keys off `$request->ip()` (via
`userKey()`). In production the app sits behind CloudFlare, and
`app/Http/Middleware/TrustProxies.php` trusts CloudFlare's published IP ranges so
`$request->ip()` should resolve to the real visitor.

Instead, many legitimate users are being throttled together, and the throttle log lines
(`Handler::tooManyRequests`, which logs `$request->ip()`) show *CloudFlare edge IPs* as the
offender. Meanwhile an attacker probing the site is correctly resolved to their real public
IP (e.g. `172.41.23.193`).

## Root cause

- The CloudFlare range list loads correctly (the fetch URL 301-redirects and `Curl` follows
  redirects), and includes `172.64.0.0/13`. The trusted-proxy list is not the problem.
- `172.41.23.193` is not in any CloudFlare range, so it is a real client IP correctly
  resolved — proving resolution works when `X-Forwarded-For` carries a usable client entry.
- The bug is Symfony's fallback: when the `X-Forwarded-For` chain does not yield a
  non-trusted client IP, `getClientIps()` falls back to the connecting peer — a trusted
  CloudFlare edge IP. With only a handful of edge IPs, large numbers of distinct users
  collapse into one rate-limit bucket and get throttled together.

## Fix

In `TrustProxies::handle()`, when the connecting peer is a trusted CloudFlare proxy, rewrite
`X-Forwarded-For` to the single `CF-Connecting-IP` value (which CloudFlare sets on every
proxied request and contains exactly the real visitor IP) before `parent::handle()`. Gate
strictly on `IpUtils::checkIp(REMOTE_ADDR, cloudflare_ranges)` so the header is only honored
from a genuine CloudFlare peer and cannot be spoofed by a client reaching the origin
directly.

Follow-up: #3564 (application-layer IP banning, which depends on this real-IP resolution).

## Definition of done

- `CF-Connecting-IP` honored only from a trusted CloudFlare peer.
- Unit tests: trusted peer + usable XFF returns real IP; trusted peer + degenerate XFF
  returns `CF-Connecting-IP` (the bug today); untrusted peer + spoofed header is ignored.

Wotuu:
:robot: _Opened by Claude on behalf of @Wotuu._

Closes #3565

## Problem

Rate limiting (`AppServiceProvider::userKey`) keys off `$request->ip()`. Behind CloudFlare, `TrustProxies` trusts CloudFlare's published IP ranges — but when the `X-Forwarded-For` chain does not yield a non-trusted client entry, Symfony's `getClientIps()` falls back to returning the connecting peer, which is a **trusted CloudFlare edge IP**. With only a handful of edge IPs, swaths of unrelated visitors collapse into one rate-limit bucket and get throttled together, while an attacker whose XFF entry survives (e.g. `172.41.23.193`, not a CloudFlare range) is bucketed correctly.

## Fix

In `TrustProxies::handle()`, when the connecting peer is a trusted CloudFlare proxy, rewrite `X-Forwarded-For` to the single `CF-Connecting-IP` value (which CloudFlare sets on every proxied request and contains exactly the real visitor IP) before delegating to `parent::handle()`. Symfony's normal trusted-proxy walk then resolves `$request->ip()` to the real visitor.

The override is gated strictly on `IpUtils::checkIp(REMOTE_ADDR, cloudflare_ranges)` — the same trust boundary the parent middleware enforces — so a client reaching the origin outside CloudFlare cannot spoof the header to evade rate limiting.

Notes:
- Only affects the exact class of broken requests: a request misresolving to a CloudFlare IP must have `REMOTE_ADDR` in the CF ranges, which means it is a genuine CloudFlare request that always carries `CF-Connecting-IP`.
- No regression for currently-correct traffic: with CF connecting directly to origin (production nginx has no `real_ip` stripping), the rightmost XFF entry already equals `CF-Connecting-IP`, so the rewrite yields the same value.
- Self-healing after deploy: stale `perHour` buckets keyed by CF edge IPs age out while real-IP buckets start fresh — cut-off users unblock with no Redis purge.

## Tests

New `tests/Unit/App/Http/Middleware/TrustProxiesTest.php`:
1. Trusted CF peer + usable XFF + `CF-Connecting-IP` → returns the `CF-Connecting-IP` value (authoritative over XFF).
2. Trusted CF peer + missing XFF + `CF-Connecting-IP` → returns the real visitor IP instead of the CloudFlare edge IP (the bug).
3. Untrusted peer + spoofed `CF-Connecting-IP` → header ignored, real connecting peer used (security case).

`composer run analyse` and `composer run fix` clean.

Follow-up: #3564 (application-layer IP banning, which depends on this real-IP resolution).

### DIFF
--- a/app/Http/Middleware/TrustProxies.php
+++ b/app/Http/Middleware/TrustProxies.php
@@ -7,6 +7,7 @@ use Closure;
 use Illuminate\Http\Middleware\TrustProxies as Middleware;
 use Illuminate\Http\Request;
 use Override;
+use Symfony\Component\HttpFoundation\IpUtils;
 use Symfony\Component\HttpFoundation\Response;
 
 class TrustProxies extends Middleware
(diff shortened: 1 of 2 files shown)


Wotuu (commented):
Ok

PR #3569, opened 2026-07-16
#3564 Ban IP addresses from accessing the site at the application layer

### LINKED ISSUE #3564: Ban specific IP addresses from accessing the site at the application layer

Wotuu:
:robot: Closes #3564

## Summary
- `banned_ip_addresses` table + `BannedIpAddress` model + repository (interface/impl/binding), following project convention.
- `BlockBannedIpAddresses` global middleware, registered right after `TrustProxies` in `bootstrap/app.php` so `$request->ip()` is already the real visitor IP resolved from `CF-Connecting-IP`. Short-circuits with a plain `403` (no view render) before routing.
- `BannedIpAddressService`: write-through cache (`Cache`-backed via the existing `CacheServiceInterface`) refreshed on every ban/add - `isBanned()` reads through a 5-minute TTL cache, checks CIDR ranges via Symfony's `IpUtils::checkIp`, and skips expired bans in-memory so temporary bans self-heal without waiting on the TTL. Defensively falls back to a direct DB read if the cache ever returns something unusable, since this runs on every request and must never crash the site.
- Admin CRUD page under Admin > Tools > Security to list/add/remove bans (`admin.tools.bannedipaddresses.*`).
- `bannedipaddress:ban` / `bannedipaddress:unban` Artisan commands for CLI incident response.
- Validation (`BannedIpRangeRule`): rejects malformed IP/CIDR input, rejects ranges wider than **/24 (IPv4) or /64 (IPv6)**, and blocks an admin from banning their own current IP (self-lockout prevention). **Note:** this cap only applies to the admin UI - the `bannedipaddress:ban` CLI command is intentionally unrestricted, so a wider emergency range ban (e.g. during a scripted probing episode) is still possible via CLI/incident response. The `/24` and `/64` cap constants live at the top of `app/Rules/BannedIpRangeRule.php` if they need tuning.

## Test plan
- [x] `composer run fix` clean (only touched intended files)
- [x] New tests (22): service (cache/CIDR/expiry), middleware (unit + full HTTP short-circuit through the real middleware stack), admin controller (auth gate, render, validation, add/remove), artisan commands
- [x] Full `Controller`, `Middleware`, and `API` test groups (523 tests) pass with the new global middleware in place

### DIFF
--- a/app/Models/BannedIpAddress.php
+++ b/app/Models/BannedIpAddress.php
@@ -1,0 +1,55 @@
+<?php
+
+namespace App\Models;
+
+use Database\Factories\BannedIpAddressFactory;
+use Eloquent;
+use Illuminate\Database\Eloquent\Factories\HasFactory;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\Relations\BelongsTo;
+use Illuminate\Support\Carbon;
+
+/**
+ * @property int         $id
+ * @property string      $ip_address
+ * @property string|null $reason
+ * @property Carbon|null $expires_at
+ * @property int|null    $created_by
+ *
+ * @property Carbon $updated_at
+ * @property Carbon $created_at
+ *
+ * @property User|null $createdBy
+ *
+ * @mixin Eloquent
+ */
+class BannedIpAddress extends Model
+{
+    /** @use HasFactory<BannedIpAddressFactory> */
+    use HasFactory;
+
+    protected $fillable = [
+        'ip_address',
+        'reason',
+        'expires_at',
+        'created_by',
+    ];
+
+    protected function casts(): array
+    {
+        return [
+            'expires_at' => 'datetime',
+        ];
+    }
+
+    /** @return BelongsTo<User, $this> */
+    public function createdBy(): BelongsTo
+    {
+        return $this->belongsTo(User::class, 'created_by');
+    }
+
+    public function isExpired(): bool
+    {
+        return $this->expires_at !== null && $this->expires_at->isPast();
+    }
+}
--- a/app/Models/BannedIpAddress.php
+++ b/app/Models/BannedIpAddress.php
@@ -14,12 +14,12 @@ use Illuminate\Support\Carbon;
  * @property string      $ip_address
  * @property string|null $reason
  * @property Carbon|null $expires_at
- * @property int|null    $created_by
+ * @property int         $created_by
  *
  * @property Carbon $updated_at
  * @property Carbon $created_at
  *
- * @property User|null $createdBy
+ * @property User $createdBy
  *
  * @mixin Eloquent
  */
--- a/app/Console/Commands/BannedIpAddress/Ban.php
+++ b/app/Console/Commands/BannedIpAddress/Ban.php
@@ -1,0 +1,44 @@
+<?php
+
+namespace App\Console\Commands\BannedIpAddress;
+
+use App\Service\BannedIpAddress\BannedIpAddressServiceInterface;
+use Illuminate\Console\Command;
+use Illuminate\Support\Carbon;
+
+class Ban extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'bannedipaddress:ban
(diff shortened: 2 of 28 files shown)


app/Models/BannedIpAddress.php
```diff
@@ -0,0 +1,55 @@
...
+ * @property int         $id
+ * @property string      $ip_address
+ * @property string|null $reason
+ * @property Carbon|null $expires_at
+ * @property int|null    $created_by
+ *
+ * @property Carbon $updated_at
+ * @property Carbon $created_at
+ *
+ * @property User|null $createdBy
```
Wotuu: Why is this optional? It should always be created by _someone_ right?

Wotuu (commented):
Small change

app/Models/BannedIpAddress.php
Wotuu: :robot: Good catch - it wasn't required so the artisan CLI path could create a ban with no attributed admin. Fixed in 153f701: `created_by` is now `NOT NULL`, and `bannedipaddress:ban` takes a required `adminId` argument (validated against the admin role) instead of silently recording no one.

Wotuu (commented):
Ok

PR #3571, opened 2026-07-16
#3570 Fix double slash in compiled asset URLs

### LINKED ISSUE #3570: Double slash in asset URLs breaks sprite images since webpack removal
:robot: ### Problem

The v15.5.0 compiled CSS references sprite images with a double slash, e.g.:

```
background-image: url(https://assets.keystone.guru//images/affixes/affixes-midnight-s1.png)
```

The `//` produces a broken asset path (empty path segment), so affix/class/spec/etc. sprite icons fail to load on production.

### Root cause

Two things combine:

1. **`.github/workflows/release-deploy.yml` sets `ASSETS_BASE_URL: https://assets.keystone.guru/` with a trailing slash** — every other definition (`.env.example`, `js-tests.yml`, prod runtime) has none. The SCSS writes `url('#{$asset-url}/images/...')`, already supplying the slash, so a trailing slash yields `//`.
2. This trailing slash is **not new** (present since #3320). It was silently masked because the old webpack production build ran **cssnano** (`postcss-normalize-url` collapses `//` → `/`). After the webpack removal (#3449), the new `scripts/build/sass.mjs` pipeline minifies with **esbuild**, which does no URL normalization — so the pre-existing double slash now leaks into the shipped CSS.

Reproduced locally: with a trailing slash the esbuild output is `url(https://…guru//images/…)`; without it, the single-slash correct form.

### Fix

- Drop the trailing slash in `release-deploy.yml`.
- Harden `scripts/build/sass.mjs` to normalize `ASSETS_BASE_URL` (strip trailing slashes) so a stray slash can never reintroduce the bug, since esbuild won't catch it.

### Note

v15.5.0's compiled CSS is immutable on S3; the fix takes effect on the next release.

Wotuu:
:robot: Closes #3570

### Problem

v15.5.0's compiled CSS references sprites with a double slash, breaking them on production:

```
background-image: url(https://assets.keystone.guru//images/affixes/affixes-midnight-s1.png)
```

### Root cause

- `.github/workflows/release-deploy.yml` injected `ASSETS_BASE_URL: https://assets.keystone.guru/` **with a trailing slash**, while every other definition (`.env.example`, `js-tests.yml`, prod runtime) has none. The SCSS writes `url('#{$asset-url}/images/...')`, already supplying the slash → `//`.
- The trailing slash is **not new** (present since #3320). It was masked because the old webpack production build ran **cssnano** (`postcss-normalize-url` collapses `//` → `/`). After the webpack removal (#3449), the new `scripts/build/sass.mjs` pipeline minifies with **esbuild**, which does no URL normalization — so the double slash now leaks into shipped CSS.

Reproduced locally: with a trailing slash esbuild emits `guru//images/…`; without it, the correct single-slash form.

### Fix

- Drop the trailing slash in `release-deploy.yml`.
- Normalize `ASSETS_BASE_URL` in `sass.mjs` (`normalizeAssetBaseUrl`, strips trailing slashes) so a stray slash can never reintroduce the bug — esbuild won't catch it.

### Verification

- New unit tests for `normalizeAssetBaseUrl`; full `sass.test.mjs` suite green (19 passed).
- Ran `buildSassBundles` in production mode with the worst-case trailing-slash env: compiled bundle now emits `url(https://assets.keystone.guru/images/...)` with **zero** double slashes.

### Note

v15.5.0's compiled CSS is immutable on S3; this fix takes effect on the **next release**. It cannot be hotfixed (compiled CSS is baked into release assets).

https://claude.ai/code/session_01ArakgH1TXrqehHM3QECW9w

### DIFF
--- a/scripts/build/sass.mjs
+++ b/scripts/build/sass.mjs
@@ -29,6 +29,20 @@ const sassEntries = [
 ];
 
 /**
+ * Normalizes the CDN base injected as `$asset-url`, stripping any trailing slashes. The
+ * stylesheets already write `url('#{$asset-url}/images/...')`, supplying their own slash, so a
+ * base with a trailing slash (as the release workflow env once carried) produced `guru//images`.
+ * The old webpack production build hid this because cssnano's normalize-url collapsed the `//`;
+ * esbuild — the current minifier — does not, so the double slash reached shipped CSS (#3570).
+ *
+ * @param {string|undefined} base
+ * @returns {string}
+ */
+export function normalizeAssetBaseUrl(base) {
+    return (base ?? '').replace(/\/+$/, '');
+}
+
+/**
  * @param {string} rootDir
  * @param {string} version
  * @param {boolean} production
@@ -42,7 +56,7 @@ export function buildSassBundles(rootDir, version, production) {
     for (const [entry, outName] of sassEntries) {
         const entryPath = path.join(rootDir, entry);
         // Same injection sass-loader's additionalData used to do (CDN base for the sprite urls)
-        const source = `$asset-url: "${process.env.ASSETS_BASE_URL}";\n`
+        const source = `$asset-url: "${normalizeAssetBaseUrl(process.env.ASSETS_BASE_URL)}";\n`
             + rewritePlainCssImports(fs.readFileSync(entryPath, 'utf8'));
 
         const result = sass.compileString(source, {
--- a/scripts/build/sass.mjs
+++ b/scripts/build/sass.mjs
@@ -29,17 +29,17 @@ const sassEntries = [
 ];
 
 /**
- * Normalizes the CDN base injected as `$asset-url`, stripping any trailing slashes. The
- * stylesheets already write `url('#{$asset-url}/images/...')`, supplying their own slash, so a
- * base with a trailing slash (as the release workflow env once carried) produced `guru//images`.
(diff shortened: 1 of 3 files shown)


scripts/build/sass.mjs
```diff
@@ -28,6 +28,20 @@ const sassEntries = [
...
+ * Normalizes the CDN base injected as `$asset-url`, stripping any trailing slashes. The
+ * stylesheets already write `url('#{$asset-url}/images/...')`, supplying their own slash, so a
+ * base with a trailing slash (as the release workflow env once carried) produced `guru//images`.
+ * The old webpack production build hid this because cssnano's normalize-url collapsed the `//`;
+ * esbuild — the current minifier — does not, so the double slash reached shipped CSS (#3570).
+ *
+ * @param {string|undefined} base
+ * @returns {string}
+ */
+export function normalizeAssetBaseUrl(base) {
```
Wotuu: `normalizeUrl` no need to make it so specific.

Wotuu (commented):
One comment

scripts/build/sass.mjs
Wotuu: :robot: Done — renamed `normalizeAssetBaseUrl` → `normalizeUrl` (and updated the tests/doc); it just strips trailing slashes now. Pushed as 116bafd08.

PR #3573, opened 2026-07-16
#3568 Optimize DungeonRoute::dropCaches() with a per-route Redis hash

### LINKED ISSUE #3568: Optimize DungeonRoute::dropCaches() — replace the permutation loop with prefix/tag invalidation
:robot: Follow-up raised in review of #3447.

`DungeonRoute::dropCaches()` invalidates a route's card caches by looping over every combination of orientation × locale × showAffixes × showDungeon × isAdmin and issuing a `Cache::delete` per permutation — hundreds of Redis deletes to drop the caches for a single route.

This scales poorly and grows every time a new orientation/dimension is added. Replace it with a better strategy, e.g. a key prefix per route + prefix scan/delete, or Redis cache tags, so dropping a route's caches is a bounded operation regardless of the number of card variants.

Reference: `app/Models/DungeonRoute/DungeonRoute.php` (`dropCaches` / `getCardCacheKey`).

Wotuu:
:robot: Closes #3568

## Problem

`DungeonRoute::dropCaches($routeId)` invalidated a route's card caches by looping over every combination of `orientation × locale × showAffixes × showDungeon × isAdmin` and issuing one `Cache::delete` per permutation — **~200–360 blind Redis deletes** for a single route, on **every** `touch()`/save. It scaled poorly and grew every time a new card dimension was added.

## Fix

Move the permutation out of the Redis **key name** and into Redis **hash fields**. All card variants of a route now live as fields in a single hash keyed `dungeonroute_card:{id}`, so dropping a route's caches is a **single `DEL`** — bounded regardless of how many variants exist or are added later.

- **`CacheService`**: new `rememberInHash()` (HGET read; HSET write + whole-hash `EXPIRE`) and `dropHashCache()` (single `DEL`), reusing the existing `RedisServiceInterface::rawCommand` plumbing so it stays client-agnostic and mockable. The global `dropCaches()` "drop all caches" card regex is updated to the new `dungeonroute_card:\d+` key shape.
- **`DungeonRoute`**: `getCardCacheKey()` now returns just the per-route hash key; the variant tuple moved to a new `getCardCacheField()`. `dropCaches()` delegates one `dropHashCache()`. Signature stays `(int)`, so all existing callers are unaffected.
- **Card blades** (`cardvertical`/`cardhorizontal`/`cardhorizontalrow`): switched from `remember()` to `rememberInHash()`.

## Behaviour note

TTL is now per-route-hash (Redis can only expire a hash as a whole), refreshed on each variant write. Practically harmless with the 1-hour card TTL, and arguably better — a hot route keeps all its variants warm together.

## Tests

- `CacheServiceHashTest` — cache miss writes `HSET`+`EXPIRE` and returns the value; cache hit returns the cached value without running the closure; `dropHashCache` issues exactly one `DEL`.
- `DungeonRouteDropCachesTest` — `dropCaches()` delegates a single `dropHashCache('dungeonroute_card:{id}')`; locks the key/field formats.
- Verified end-to-end against real Redis: two variants → one hash key (`fields=2`, `ttl=3600`); re-read is a cache hit; `dropCaches()` clears the whole route in one op.

`composer run fix` + `composer run analyse` clean.

### DIFF
--- a/app/Models/DungeonRoute/DungeonRoute.php
+++ b/app/Models/DungeonRoute/DungeonRoute.php
@@ -45,6 +45,7 @@ use App\Models\Traits\Reportable;
 use App\Models\Traits\SerializesDates;
 use App\Models\Traits\Taggable;
 use App\Models\User;
+use App\Service\Cache\CacheServiceInterface;
 use App\Service\Coordinates\CoordinatesServiceInterface;
 use App\Service\Expansion\ExpansionServiceInterface;
 use App\Service\Season\SeasonAffixGroupServiceInterface;
@@ -66,11 +67,9 @@ use Illuminate\Support\Carbon;
 use Illuminate\Support\Collection;
 use Illuminate\Support\Facades\App;
 use Illuminate\Support\Facades\Auth;
-use Illuminate\Support\Facades\Cache;
 use Illuminate\Support\Facades\DB;
 use Illuminate\Support\Str;
 use Override;
-use Psr\SimpleCache\InvalidArgumentException;
 
 /**
  * @property int                  $id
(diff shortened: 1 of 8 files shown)


Wotuu (commented):
Ok!

PR #3581, opened 2026-07-16
#3579 Bump app-worker Puppeteer pin to 25.3.0 (bake matching Chrome)

### LINKED ISSUE #3579: app-worker Dockerfile Chrome pin drifts from package.json puppeteer version (breaks thumbnail generation, incl. prod)

## Problem

`docker-compose/app-worker/Dockerfile` bakes Puppeteer's Chrome via a **hardcoded** pin:

```dockerfile
RUN npm i -g puppeteer@24 && chmod -R a+rwX /opt/puppeteer-cache
```

The repo's `puppeteer`/`puppeteer-core` (bind-mounted from `node_modules`) is what actually calls `puppeteer.launch()`, and it launches the **exact** Chrome build its version pins. So when the repo's puppeteer is bumped but this pin isn't, they drift and the baked Chrome is the wrong version:

- **#3556** bumped `puppeteer` **24.43.1 → 25.3.0** (wants Chrome **150.0.7871.24**).
- The worker image still bakes `puppeteer@24` → Chrome **148.0.7778.97**.
- Result at runtime: `Error: Could not find Chrome (ver. 150.0.7871.24)` — every thumbnail render fails fast and exhausts its retries. Locally the thumbnails dir stays empty.

## Production impact

The prod `app-worker-aws` image inherits this same pin, so **production thumbnail generation will break the same way once #3556 releases**. This should ship before/with that release.

## Fix

- Immediate: bump the pin to match the repo (`puppeteer@25.3.0`), rebuild the worker image, recreate the container. (PR incoming.)
- Better: **derive the version from `package.json`** at build time (e.g. copy `package.json` into the build context and `npm i -g puppeteer@$(node -p "require('./package.json').devDependencies.puppeteer")`), so the pin can't drift again. Add a CI check or Dependabot grouping so a puppeteer bump also rebuilds/repins the worker image.

## Notes
- `PUPPETEER_CACHE_DIR=/opt/puppeteer-cache` and the baked-Chrome approach are correct; only the version pin is the issue.
- Discovered while verifying thumbnail regeneration for the #3561 discover rework.

Wotuu:
:robot: _Opened by Claude on behalf of @Wotuu._

Closes #3579

## Problem
`docker-compose/app-worker/Dockerfile` bakes Puppeteer's Chrome from a hardcoded `npm i -g puppeteer@24`, but the repo's `puppeteer-core` (bind-mounted from `node_modules`, and what actually calls `puppeteer.launch()`) was bumped to **25.3.0** in #3556. puppeteer 25.3.0 pins **Chrome 150.0.7871.24**; the image baked **148.0.7778.97**. Result at runtime:

```
Error: Could not find Chrome (ver. 150.0.7871.24).
```

Every thumbnail render fails fast and exhausts its retries — thumbnail generation is fully broken. The prod `app-worker-aws` image inherits the same pin, so **production breaks the same way once #3556 releases**.

## Fix
Bump the pin to `puppeteer@25.3.0` so the baked Chrome matches the repo, and add a comment to keep it in sync with `package.json`. Verified locally: rebuilt the worker image (bakes `chrome/linux-150.0.7871.24`), recreated horizon, and thumbnail jobs now render in ~2s and land on disk (previously failed in ~30ms).

Follow-up idea (noted in #3579): derive the version from `package.json` at build time so it can't drift again.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,6 @@
+# Root build context is used only by docker-compose/app-worker/Dockerfile (see docker-compose.yml
+# and release-deploy.yml), which needs nothing from the repo besides package.json to derive its
+# Chrome/puppeteer version pin. Deny-all + allow-list keeps that build fast and means never having
+# to remember to add some new heavy directory (node_modules, vendor, storage, .git, ...) here.
+*
+!package.json
--- a/.github/workflows/release-deploy.yml
+++ b/.github/workflows/release-deploy.yml
@@ -194,16 +194,21 @@ jobs:
             ecr_repos: "ksg-php-fpm ksg-reverb ksg-swoole"
           # Puppeteer/Chrome live only in this image (docker-compose/app-worker), so the
           # thumbnail-generation queue workers get a browser; the app image above doesn't.
+          # base_context is the repo root (not docker-compose/app-worker) so its Dockerfile can
+          # COPY package.json and derive the baked Chrome version from it instead of a hardcoded,
+          # driftable pin.
           - image: worker
             base_tag: keystone.guru-worker
             needs_own_base_build: true
-            base_context: docker-compose/app-worker
+            base_context: .
+            base_dockerfile: docker-compose/app-worker/Dockerfile
             prod_context: docker-compose/app-worker-aws
             ecr_repos: "ksg-queue-worker"
           - image: cron
             base_tag: keystone.guru-cron
             needs_own_base_build: true
             base_context: docker-compose/cron
+            base_dockerfile: docker-compose/cron/Dockerfile
             prod_context: docker-compose/cron-aws
             ecr_repos: "ksg-cron"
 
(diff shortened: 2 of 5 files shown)


Wotuu:
:robot: Pushed the "better fix" noted in #3579: instead of hardcoding `puppeteer@25.3.0` (which will just drift again on the next bump), the Dockerfile now derives the version from `package.json` at build time.

Required moving the horizon build context from `docker-compose/app-worker` to the repo root (both in `docker-compose.yml` and `release-deploy.yml`) so the Dockerfile can `COPY package.json` in. Added a root `.dockerignore` (deny-all + allow-list `package.json`) so that wider context doesn't balloon the build — verified the context transfer is ~34-446 bytes, not the whole repo.

Verified locally: built the worker image from this branch, confirms it resolves `puppeteer@25.3.0` from `package.json` and bakes the matching `chrome/linux-150.0.7871.24`.

Wotuu (commented):
Ok

PR #3582, opened 2026-07-17
#3580 Repair broken dungeon route thumbnails (delete fileless + re-queue disk-missing) & facade regression test

### LINKED ISSUE #3580: Facade routes create fileless thumbnail rows (file_id null) → 403; ~1292 orphan thumbnails to clean up

## Problem

For **facade** routes, thumbnail generation succeeds but the resulting `DungeonRouteThumbnail` row ends up with **`file_id = null`** (no attached `File`). The route's `has_thumbnail` is still true, so cards render a broken thumbnail URL and the request **403s** (served through Laravel's local-serve/`public` path with nothing behind it).

### Repro
- Route `4daaHko` (Pit of Saron), floor index 2 is a **facade** floor (`facade = true`).
- `ThumbnailService::createThumbnail($route, 2, 0)` returns **`true`** (render + resize succeed), yet afterwards `$route->thumbnails()->first()->file` is `null`.
- Re-dispatching cleanly (no concurrent jobs) reproduces it every time, so it is not a race — it's specific to the facade attach path.

### Scale
```
DungeonRouteThumbnail rows with file_id NULL or a missing File row: 1292
```
These are pre-existing (they predate the Chrome fix / disk change).

## Likely area
`ThumbnailService::attachThumbnailToDungeonRoute()` + the facade branch of the "existing thumbnails to delete" query (`->when(!$floor->facade, …)`), and/or the `getTargetFilePath`/`File` attach for facade floors. The new row is created but its `file_id` isn't persisted (or the file it points to is removed by the same pass).

## Asks
1. Fix the facade path so a successful render always yields a file-backed thumbnail.
2. One-off cleanup of the **1292** fileless thumbnail rows: delete them so `has_thumbnail` reflects reality (cards fall back to the dungeon image) and/or re-queue those routes once (1) lands.

## Notes
- Separate from the Chrome/puppeteer pin issue and from `FILESYSTEM_DISK` — standard (non-facade) routes regenerate and load fine (verified HTTP 200).
- Discovered while verifying thumbnail regeneration for the #3561 discover rework.

Wotuu:
:robot: Closes #3580

## Summary

`dungeonroute:repairbrokenthumbnails` (with `--dry-run`) repairs broken dungeon route thumbnails in two phases, plus a facade regression test.

### Phase 1 — delete fileless rows
Deletes `DungeonRouteThumbnail` rows **not backed by a usable `File`** — `file_id` null **or** pointing at a `File` row that no longer exists (`whereDoesntHave('file')`). Rows are deleted per-model so the `deleting` observer cleans up any lingering `File`/disk object. Afterwards `has_thumbnail` (an accessor over the file-join) correctly reports `false` and the card falls back to the dungeon image.

### Phase 2 — re-queue rows whose disk object is missing
Detects thumbnails whose `File` row still exists but whose **object is gone from disk** (`Storage::disk($file->disk)->exists($file->path) === false`) and re-queues each affected route (`queueThumbnailRefresh(force: true)`). This is the population that actually **403s**: `has_thumbnail` stays true, a URL is generated, but nothing is on disk.

Broken **custom** (API-requested) thumbnails are only reported via a warning, not re-queued — re-queueing regenerates the default per-floor thumbnails and cannot repair a custom one, so it would re-flag them on every run. The scan also shows a progress bar (one `Storage::exists()` per row can be slow against S3).

Confirmed against real data — e.g. `tNA1Irn`'s thumbnail is a legacy `disk=local`, leading-slash-path row (`/thumbnails/…`) whose object is absent, rendering `/storage//thumbnails/…` → 403. `MigrateThumbnailDisk` (#3550) skips it because it only migrates local rows whose object still exists, so it was stuck. A dry-run flags **14** such routes locally.

> Re-queued regeneration runs via the queued thumbnail jobs, so it only clears the 403 once thumbnail generation itself is working (currently blocked by the Chrome/puppeteer breakage in #3579).

### Regression test
`attachThumbnailToDungeonRoute_givenFacadeFloorWithExistingThumbnailsOnMultipleFloors_yieldsSingleFileBackedFacadeThumbnail` seeds a route with thumbnails on every floor, runs a facade attach, and asserts exactly one **file-backed** thumbnail remains (+ `has_thumbnail == true`). Adds a `getDungeonWithFacadeFloor` test helper.

## Investigation note

I could **not** reproduce the issue's "successful render → `file_id` null" on current code: `attachThumbnailToDungeonRoute()` is a single atomic `DB::transaction`, `file_id` is always set via `update()`, and the delete pass is fetched before the new row exists. Verified empirically for single- and multi-floor facade routes — every path yields a file-backed thumbnail. The reported orphans read as historical (the issue notes they predate the disk/Chrome changes); hence a guard test rather than an attach-path code change.

## Review fixes (17e7be87d)

- Phase 2 skips `custom` thumbnails when re-queueing (warned instead, see above) and no longer hydrates the unused `floor` relation during the scan.
- Tests: added `Queue::fake()` to the delete test — it previously pushed real jobs to the shared redis queue, because under `Storage::fake()` the seeded test DB's 31 file-backed thumbnails all read as "missing from disk". For the same reason the queue assertions are now **scoped per-route** (closure-bound into the job's protected `$dungeonRoute`). New test covers the custom-thumbnail path.

## Flagged, not fixed (scope)

The facade branch of the delete query has no `custom = false` filter, so a non-custom facade regeneration also deletes any **custom** API thumbnails for the route. Not the orphan cause (row + file are deleted together); left as-is to keep scope tight — happy to add `->where('custom', false)` if wanted.


Wotuu (commented):
Ok

PR #3584, opened 2026-07-17
Standardize agent thumbnail generation into two canonical paths (skill + commands)

Wotuu:
:robot: Standardizes how agents generate dungeon-route thumbnails so we stop re-deriving it every time.

## Why

Agents kept landing on a different ad-hoc thumbnail method each time (guard bypasses, `.chrome-tmp` apt-installs, env swaps, `public/storage` symlink hacks, running `createThumbnail` in a container with no Chrome). Root cause: Chrome is baked into the `keystone.guru-worker`/Horizon image **only**, and the **DB is shared** across the main stack and all worktrees — two facts that no single place documented.

## The two — and only two — supported paths

**Path B — real thumbnails (shared).** `dungeonroute:queuethumbnail <key> --force` dispatches to the shared Horizon container (the only one with Chrome), which renders to the `public` disk. Bind-mounted + shared DB means the main stack and every worktree see it. Works from anywhere (dispatch needs only Redis). Uses master's code.

**Path A — verify MY branch's render (isolated).** `docker compose --profile render run --rm render dungeonroute:renderthumbnail <key> --floor=<n>` renders *this checkout's* code inside an on-demand container built from the existing `keystone.guru-worker` image (no app-image bloat, Chrome already version-matched), writes to the `local` disk (`storage/app/private`), and prints the path so the agent **Reads the JPG**. It runs inside a **transaction that is always rolled back**, so the shared DB is never mutated — critical, because otherwise a worktree render replaces the real public-disk row with one pointing at worktree-only storage and 403s the route everywhere else.

Grid check (why only two): need = *(whose code renders)* × *(isolated vs shared output)*. Path A = my-branch + isolated; Path B = master + shared. The other cells are pointless or the deliberately-avoided per-worktree Chrome worker.

## Changes

- `app/Console/Commands/DungeonRoute/QueueThumbnail.php` — Path B command (thin wrapper over `queueThumbnailRefresh`).
- `app/Console/Commands/DungeonRoute/RenderThumbnail.php` — Path A command; forces the `local` disk and wraps the render in a rolled-back transaction (no DB writes).
- `docker-compose.worktree.yml` — profiled `render` service (worker image; `NODE_PATH=/usr/lib/node_modules` so `require('puppeteer')` resolves to the image's global, Chrome-matched puppeteer since the `./` mount shadows any worktree node_modules).
- `sh/worktree.sh` — persist `MAIN_THUMBNAILS_DIR` into the worktree `.env` so plain `docker compose` (the `render` service) resolves the bind-mount source (also hardens all manual compose calls).
- `.claude/skills/generating-thumbnails/SKILL.md` — the single source of truth; cross-referenced from `worktree-docker` and `docs/developer/scheduler.md`.
- Tests: `RenderThumbnailTest` (prod guard, unknown key, forces `local` disk, **and proves the render is rolled back / never persists**), `QueueThumbnailTest` (queues a job; unknown key fails).

## Verification (done end-to-end on this machine)

- **Path A:** `render` container rendered `Ruv4Rtj` to `storage/app/private/...jpg`; read the JPG (valid image); the shared DB row stayed `disk=public`, count=1 — untouched.
- **Path B:** `dungeonroute:queuethumbnail` → Horizon rendered to the `public` disk in ~5s; `curl http://nginx/storage/<path>` → `200`.
- **Fresh worktree:** created a throwaway worktree off this branch and confirmed `sh/worktree.sh` writes `MAIN_THUMBNAILS_DIR` into its `.env` and `docker compose --profile render config` resolves with no empty-mount error — so the `render` path works in a new worktree without any manual `.env` patching.
- Both renders use the same `route_thumbnail.js` pipeline; local renders show the DebugBar + black tiles because local `app-assets` has no tiles — identical for both paths, not a Path A bug. (Path A uses the worker image's global puppeteer, Path B the repo's local puppeteer.)
- `RenderThumbnail|QueueThumbnail` tests pass (6, incl. asserting the Path A render is rolled back in DB **and** the file survives on disk); `composer run fix` + `analyse` clean.

## Note

The stale `seed-dev-routes` thumbnail runbook (guard bypass + `public/storage → private` symlink) lives in the unmerged #3349 worktree, not master — the new skill documents that those steps are obsolete; the actual edit lands if/when #3349 does.

### DIFF
--- a/.claude/skills/generating-thumbnails/SKILL.md
+++ b/.claude/skills/generating-thumbnails/SKILL.md
@@ -52,6 +52,15 @@ docker compose --profile render run --rm render dungeonroute:renderthumbnail <pu
 - `--floor` is optional (defaults to the map-facade floor set). `--disk` defaults to `local`; keep
   it — writing to `public` from a worktree would land in the shared, bind-mounted dir.
 
+**Path A is inspection-only — it does NOT give the route a thumbnail.** The rollback undoes the whole
+attach (including its delete-of-existing pass), so the route is left exactly as it was: if it had a
+real (public) thumbnail it keeps it, if it had none it still falls back to the dungeon image. Path A
+only adds a standalone JPG on the local disk for you to Read. To actually give a route a working,
+site-rendered thumbnail, use **Path B** — you can't make the route card show your branch's render in
+the worktree UI, because the `public` file is the shared bind-mount (overwriting it, or re-pointing
+the shared DB row, would bleed into the main stack and every worktree). Your branch's rendering
+change reaches real thumbnails via Path B **after merge**, when Horizon runs your merged code.
+
 **Local renders show the DebugBar and black (missing) map tiles** — the local `app-assets` container
 has no tiles and dev has the DebugBar on. This is the same for Path A and Path B (same render
 script/pipeline), so it's fine for verifying layout/enemies/positioning. If you need a tile-accurate,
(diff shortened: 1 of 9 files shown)


Wotuu (commented):
Ok

PR #3607, opened 2026-07-18
#3605 Harden worktree.sh for parallel agents + .gitignore junk coverage

### LINKED ISSUE #3605: Harden sh/worktree.sh for parallel agents (create lock, repair command, hardlinked vendor seed) + .gitignore junk coverage
:robot: Filed by Claude after a productivity audit of the parallel-agent worktree setup.

## Background

Multiple Claude agents each work in their own git worktree + Docker stack via `sh/worktree.sh`. An audit found three recurring/latent problems in the script and one repo-hygiene gap:

1. **Port-pick race, no locking.** `cmd_create` picks a host port by scanning `ss -ltnH` (`find_free_port`, `sh/worktree.sh:64-73`) and only binds it later at `docker compose up -d`. Two concurrent `create`s can both see e.g. 8100 free and collide at bind time. There is no lockfile/mutex anywhere in the script.
2. **No recovery command for detached shared-service aliases.** Restarting the main stack detaches the shared containers (`db`, `redis`, `app-swoole`, ... — `SHARED_SERVICES`, `sh/worktree.sh:23-32`) from every worktree network. Each worktree's nginx then fails until someone manually re-runs `docker network connect` with the right aliases + restarts that nginx. This is a known recurring hand-fix.
3. **`cp -a vendor/` per worktree is the dominant create cost.** Step 2c (`sh/worktree.sh:157-172`) copies the full `vendor/` tree (hundreds of MB, tens of thousands of files) into every worktree. Pure duplication of identical files.
4. **`.gitignore` gaps.** `__pycache__/`, `tbc/`, `tbc_ssc/`, `tbc_the_eye/`, `tmp/`, `tmp_combatlogs/`, `docker-compose/mysql-prod*/` (real DB dumps, UID-999-owned, emit permission-denied noise in `git status`), and `storage/api-docs/` are untracked but not ignored. `git add *`-style commands are allowlisted for agents, so these are one blanket add away from being staged.

## Remaining work

- Wrap the racy section of `cmd_create` (port pick through `docker compose up -d`) in `flock` on a lockfile (e.g. `/tmp/ksg-worktree-create.lock`) so parallel creates serialize.
- Add a `repair [slug]` subcommand (no slug = all live worktree stacks): idempotently re-`docker network connect --alias` every `SHARED_SERVICES` container into each worktree network (skip when already attached), then `docker compose restart nginx` in that worktree. Safe to run blindly. Mention it in the `worktree-docker` skill so agents self-heal.
- Hybrid hardlink seed for `vendor/`: `cp -al` the tree, then **real-copy** `vendor/composer/` and `vendor/autoload.php` on top. Composer replaces changed packages by delete+re-extract (new inodes → hardlinks detach safely) but regenerates the autoload files via in-place truncate (`file_put_contents`) which would leak through hardlinks into the main checkout. `public/*` compiled assets must **stay** `cp -a`: webpack writes output by truncating in place, so hardlinks there would let a worktree `npm run production` clobber the main checkout's assets.
- Add the `.gitignore` entries listed above (ignore only — delete nothing; `docker-compose/data/*.sql` stays visible on purpose).

## How to verify

- `sh/worktree.sh create` a throwaway worktree: create is fast; site responds on its port; vendor package files have link count > 1 (`stat -c %h`) while `vendor/composer/*`, `vendor/autoload.php` are private copies (link count 1).
- In the worktree, run `composer dump-autoload` + a frontend build, then checksum-compare the main checkout's `vendor/autoload.php` and `public/js` — must be unchanged.
- Two concurrent `create`s → distinct ports, both stacks healthy.
- `docker network disconnect` a shared container from a worktree network → `sh/worktree.sh repair` restores it and the worktree serves pages again.
- `git status` in the main checkout no longer lists the junk dirs.

## Key files

- `sh/worktree.sh`
- `.gitignore`
- `.claude/skills/worktree-docker/SKILL.md`

## Definition of done

- All verification steps above pass; `worktree.sh remove` still tears down cleanly (hardlinked vendor must not block removal).
- No PHP changes, so no test/PHPStan impact expected.

Wotuu:
Closes #3605

:robot: Hardens the parallel-agent worktree tooling and closes `.gitignore` gaps.

## Changes

- **`sh/worktree.sh` — create lock**: the port-pick → `docker compose up -d` window now holds an exclusive `flock` (`/tmp/ksg-worktree-create.lock`). `find_free_port` only scans current listeners, so two concurrent creates could previously pick the same port and collide at bind time.
- **`sh/worktree.sh` — `repair [slug]` subcommand**: restarting the main stack detaches the shared containers from every worktree network, breaking each worktree's nginx. `repair` idempotently reattaches all shared-service aliases (per stack or all `ksg-*` stacks) and restarts nginx. The attach loop is factored into `attach_shared_services()` shared with `create`. Documented in the `worktree-docker` skill.
- **`sh/worktree.sh` — hardlinked vendor seed**: `vendor/` is now seeded with `cp -al` (near-instant, no extra disk) instead of a full copy. `vendor/composer/` and `vendor/autoload.php` are real-copied on top because composer rewrites them **in place** (would leak through hardlinks into the main checkout); package files are safe since composer replaces packages by delete + re-extract. `public/*` stays a real copy — webpack truncates output in place, so hardlinks there would let a worktree build clobber the main checkout's assets. Falls back to a full copy if hardlinking fails (different filesystem).
- **`.gitignore`**: covers `docker-compose/mysql-prod*/` (local prod dump restores, UID-999-owned noise in `git status`), `tbc/`, `tbc_ssc/`, `tbc_the_eye/`, `tmp/`, `tmp_combatlogs/`, `__pycache__/`, `storage/api-docs/` — `git add *` is allowlisted for agents, so these were one blanket add away from being staged. `docker-compose/data/*.sql` intentionally stays visible.

## Verification (all done with the modified script)

- Throwaway create completed in **11 s**; page served 200; vendor package files had link count 2 sharing the source checkout's inodes, while `vendor/autoload.php`, `vendor/composer/*`, and `public/js/*` were private copies (link count 1).
- Leak test: `composer dump-autoload` + a simulated in-place asset write in the throwaway left the source checkout's `vendor/autoload.php`, `autoload_classmap.php`, and `public/js` checksums unchanged.
- Two concurrent `create`s → distinct ports (8103/8104), both stacks healthy.
- Detached `redis` + `app-swoole` from a worktree network → `repair <slug>` reattached exactly the missing ones (`+`) and skipped the rest (`=`); a blind `repair` across all six running stacks left every worktree serving 200.
- `remove` of hardlink-seeded worktrees tore down cleanly; main checkout vendor checksums intact afterwards.
- `git check-ignore` confirms every new pattern matches (including `mysql-prod.bak/`) and `docker-compose/data/*.sql` stays visible.

No PHP changes — no test/PHPStan/style impact.

### DIFF
--- a/.claude/skills/worktree-docker/SKILL.md
+++ b/.claude/skills/worktree-docker/SKILL.md
@@ -127,6 +127,20 @@ real thumbnails, or render this branch's code to the local disk for inspection
 (`docker compose --profile render run --rm render dungeonroute:renderthumbnail <key>`). Never call
 `ThumbnailService::createThumbnail()` synchronously in the `app` container — it has no Chrome.
 
+## Broken worktree after a main-stack restart? Run `repair`
+
+Restarting the **main** stack detaches its shared containers from every worktree network — the
+worktree's nginx then 502s/fails on every request because its upstreams (`db`, `app-swoole`,
+`reverb`, ...) no longer resolve. Don't debug this by hand; run:
+
+```bash
+sh/worktree.sh repair                  # fix ALL running worktree stacks
+sh/worktree.sh repair <issue>-<slug>   # fix just one
+```
+
+It reattaches the shared-service aliases (idempotent) and restarts each worktree's nginx. Safe to
+run blindly whenever a worktree suddenly stops serving pages.
+
 ## Tear down
 
 ```bash
--- a/.gitignore
+++ b/.gitignore
@@ -41,5 +41,16 @@ docker-compose/mysql
 docker-compose/mysql-combatlog
 !docker-compose/mysql-combatlog/.gitkeep
 docker-compose/redis-data
+# Local restores of production database dumps
+docker-compose/mysql-prod*/
 
 .ladebugermcp/
+
+# Local scratch data that must never be staged (git add * is allowlisted for agents)
+__pycache__/
+/tbc/
+/tbc_ssc/
+/tbc_the_eye/
+/tmp/
+/tmp_combatlogs/
+/storage/api-docs/
--- a/sh/worktree.sh
+++ b/sh/worktree.sh
@@ -9,6 +9,7 @@
 #   sh/worktree.sh create <issue>-<slug> [base-ref]   # default base-ref: origin/master
 #   sh/worktree.sh down   <issue>-<slug>              # stop the stack, keep the checkout
 #   sh/worktree.sh remove <issue>-<slug>              # stop the stack and remove the worktree
+#   sh/worktree.sh repair [<issue>-<slug>]            # reattach shared services (all stacks if omitted)
 #   sh/worktree.sh list                               # list worktrees and their stacks
 #
 set -euo pipefail
@@ -33,6 +34,10 @@ influxdb:keystone.guru-influxdb
 
 PORT_RANGE_START=8100
 PORT_RANGE_END=8199
+
+# Serializes the port-pick → compose-up window of concurrent `create`s: find_free_port only scans
+# current listeners, so without a lock two creates can pick the same port before either binds it.
+CREATE_LOCK="/tmp/ksg-worktree-create.lock"
 
 # Scoped, passphraseless deploy key (write access, RaiderIO/keystone.guru only) for non-interactive
 # pushes of worktree branches. `-F /dev/null` bypasses ~/.ssh/config, which maps github.com to a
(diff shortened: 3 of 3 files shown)


Wotuu (commented):
Ok

PR #3615, opened 2026-07-19
#3614 Add generate-floor-unions skill: OpenCV facade registration

Wotuu:
Closes #3614

:robot: New Claude skill `generate-floor-unions`: generates FloorUnions + FloorUnionAreas for a dungeon facade from the per-floor images and the combined facade image, replacing the manual GIMP overlay workflow (wiki: *Adding support for dungeon facades*).

## What's in the box

- `scripts/register_floors.py` — SIFT+RANSAC multi-instance registration (`match`), art-based full-coverage area partition (`areas`), stripe-blend verification overlays, CDN tile stitcher fallback (`stitch`). Runs in a throwaway pinned Docker image (`scripts/Dockerfile` + `scripts/run.sh`) — no host installs, no app-image changes.
- `scripts/insert_floor_unions.php` — tinker script that inserts the result into a **fresh bare mapping version** (quiet `insertGetId`, dodging the clone-on-create boot hook) with validation (game-version-scoped current version, floor ownership, no empty areas).
- `SKILL.md` — the runbook, including all calibration facts (px→latlng math, rotation sign convention, phantom-fit thresholds, rejected approaches).

## FloorUnionArea partition

One hard rule, enforced mechanically: **art stays with the floor it depicts; only empty parchment is negotiable** (it goes to the nearest art, like hand-drawn areas). Content masks come from diffing each floor image against the per-pixel median of all floor images (>= 3 images) or a pairwise diff gated by self-artness (exactly 2). Boundary smoothing only moves boundaries over parchment, small-region cleanup never reassigns art, and the polygon overlap margin exceeds the simplification epsilon. The `areas` step prints a per-placement **exclusive own-art coverage** figure as an objective gate.

## Validation (against real hand-made ground truth, approved by Wotuu)

| Dungeon | Placements | Result |
|---|---|---|
| Skyreach (2 floors) | 2/2 | max Δ 0.52 map units vs mv 788, 100% own-art coverage, 102/102 enemies round-trip |
| Magister's Terrace Midnight (7 images, rotations up to 149.5°, duplicated-art void floors) | 8/8 recoverable | max Δ 0.72 map units vs mv 800, rotation ±0.3°, 100% own-art coverage, 264/264 enemies round-trip |

The 9th MT union (the repeated void-backtrack corridor of `grand_magister_asylum`) is not derivable from the 7 provided images — its floor's own image (different orientation + the corridor variant) wasn't in the set. SKILL.md requires one image per real floor for exactly this reason.

[...]

Issue #3614, opened 2026-07-19
Skill: generate FloorUnions/FloorUnionAreas from facade images via image registration


Wotuu:
:robot: **Goal**: automate the manual GIMP workflow described in the wiki page *Adding support for dungeon facades (MDT combined view)* with a new Claude skill, `generate-floor-unions`.

## Background

Adding facade support to a dungeon currently means overlaying each floor image on the combined facade image at ~60% opacity in GIMP, eyeballing the center/scale, measuring rotation by hand (typically 5-7°), and then hand-drawing full-coverage FloorUnionArea polygons in the map editor. Since the facade image is composed from the exact same per-floor images, this is a classic image-registration problem: OpenCV feature matching can recover each floor's placement (translation + uniform scale + rotation) precisely.

## Scope

- New skill `.claude/skills/generate-floor-unions/` with a Python/OpenCV helper (`register_floors.py`) run in a throwaway Docker container (no app-image or host changes).
- `match`: per-floor feature matching against the facade image → `lat/lng/size/rotation` in map units, with iterative multi-instance support (Nokhud-style zoomed sub-regions, `size > 256`).
- `areas`: full-coverage facade partition → simplified FloorUnionArea polygons (no gaps, every point owned by exactly one union).
- `overlay`: verification renders (warped floors + partition) for vision review.
- Insertion into the dev DB via tinker against a user-named mapping version; review/tuning stays in the admin map editor; `mapping:save` exports as usual.
- No app code changes.

## Validation

Calibrated against two already-mapped dungeons using the actual source images used to derive their existing FloorUnions (ground truth = latest mapping version):
- Skyreach (simple, 2 floors)
- Magister's Terrace (Midnight; complex, 7 floors)

Generated output goes into a fresh bare mapping version per dungeon so the generated FloorUnions can be compared side-by-side with the hand-made ones.

PR #3619, opened 2026-07-20
#3616 Harden generate-floor-unions for blind first runs

### LINKED ISSUE #3616: Harden generate-floor-unions for blind first runs (follow-up to #3614)
:robot: Follow-up to #3614 / MR #3615. Wotuu wants to explore this in a dedicated session.

## Background

The `generate-floor-unions` skill was built and validated in **reconstruction mode**: both calibration dungeons (Skyreach, Magister's Terrace Midnight) were already mapped, so every strong check compared against ground truth - hand-made unions (max Δ0.72 map units), hand-drawn areas, and full enemy round-trips (0 failures). See MR #3615 for the validation record.

## The gap

On a genuinely NEW dungeon none of that ground truth exists, and the strongest verifier - the enemy round-trip through `CoordinatesService::convertMapLocationToFacadeMapLocation` / `convertFacadeMapLocationToMapLocation` - is impossible at insert time: enemies only appear after the MDT import, which itself consumes the unions being verified. The remaining net (edge-correlation, exclusive own-art coverage, stripe overlays + vision) cannot catch the one invisible failure mode:

**Wrong floor identity on duplicated-art placements.** When one image places twice (normal + void variant), deciding which instance is which floor is pure game knowledge. Every overlay looks perfect either way; a swapped assignment ships a mapping where MDT-imported enemies silently land on the wrong (visually identical) floor. Midnight leans into void-variant floors, so this will come up.

## Remaining work (the plan)

1. **Post-import verification step** (highest value): extend the skill (SKILL.md workflow + optionally a small tinker script next to `insert_floor_unions.php`) with a "close the loop after MDT import" step - once enemies exist on the new mapping version, run the round-trip check and report per-floor enemy counts for a human sanity pass. Treat inserted unions as provisional until this has run.
2. **Identity-assignment guardrails**: make the skill surface duplicated-art ambiguity explicitly - when one image yields multiple placements (or two floors share art), require the user's instance→floor decision interactively and record the reasoning in the run output. Never let the model guess silently.
3. **Threshold tuning on more dungeons** (Wotuu will supply image sets): exercise `--min-edge-corr 0.2` (current margin: phantoms ≤0.17 vs genuine ≥0.25 on two dungeons of one art lineage), `--content-diff-threshold 20` and `--content-close-px 35` against new art styles before the first blind run, and record per-dungeon deviations in SKILL.md.
4. Optional hardening ideas from review, still open: image-count-vs-real-floor-count warning at match time; promoting bare-mapping-version creation from the tinker script into `MappingService` (offered in MR #3615, deliberately deferred to keep that MR app-code-free).

## Key files

- `.claude/skills/generate-floor-unions/SKILL.md` (workflow + calibration facts)
- `.claude/skills/generate-floor-unions/scripts/register_floors.py`
- `.claude/skills/generate-floor-unions/scripts/insert_floor_unions.php`
- `app/Service/Coordinates/CoordinatesService.php` (round-trip primitives)

## Definition of done

- The skill's runbook contains a mandatory post-MDT-import verification step with exact commands.
- Duplicated-art instance→floor assignment cannot happen without explicit user confirmation.
- Thresholds validated (or re-tuned, documented) on at least one additional dungeon image set.
- First blind production run of the skill completes with the new loop and its mapping verified.

Wotuu:
:robot: Closes #3616

**Stacks on #3615** (targets its branch `3614-generate-floor-unions-skill`, not `master`, so the diff is only the hardening; retarget to `master` once #3615 merges).

## Problem

#3614's skill was validated in *reconstruction* mode against hand-made ground truth. Running it blind on 8 Midnight/BfA dungeons surfaced a systematic failure: **the dungeon-title banner**. Every floor cut carries the banner (and frame), the facade shows it too, so SIFT matches floor→facade banner-to-banner with an `edge_corr` *higher* than a small genuine placement (up to 0.61) — which `--min-edge-corr` cannot reject. Left unchecked it adds phantoms (Temple of Sethraliss, Voidscar) or, when a floor's own art is faint, outscores and masks the real placement (Altar of Fangs, Murder Row).

## Fix (`register_floors.py`)

1. **Own-art keypoint masking** — restrict each floor's SIFT keypoints to its own art (reusing the `areas`-step median backdrop) so the shared banner/frame is never matched. Removes banner phantoms *and* frees the real placements they were masking. Needs ≥2 floor images; single-floor dungeons match unmasked.
2. **Off-image center guard** — reject a fit whose center lands >15% outside the facade.
3. **Review band** (`--review-edge-corr`) — a floor with no fit above `--min-edge-corr` but a geometrically sound sub-threshold fit is surfaced as `[NEEDS REVIEW]` instead of silently dropped. Recovers genuinely faint floors (Mutation Chambers 0.11, Illicit Rain 0.19) for a human call.

Threshold guidance corrected: the old "genuine ≥0.25 / phantom ≤0.17" margin doesn't hold on this art; **don't lower the global threshold** — confirm the review candidate on the overlay.

## Regression tests (`test.sh`)

Golden `match` comparison over both calibration dungeons (Skyreach, Magister's Terrace) + all 8 new dungeons, plus coordinate math, in the pinned Docker image. **Magister's Terrace still recovers all 8 unions** (duplicated-art `cut_1` still places twice) — no regression. Fixture **images are committed in-repo** under `scripts/test_fixtures/` as **256-colour indexed PNGs** (the same trick MDT uses — ~40% smaller than source RGB, 37 MB → 22 MB, verified no measurable match change). Comparison is order-independent within a floor image, so a duplicated-art instance relabel is not a false failure. Verified the guard actually fails on drift.

## Post-insert verification (`verify_floor_unions.php`)

Round-trip self-consistency check (needs no MDT data): projects every enemy on a reference version onto the facade through the new unions and back, reporting per-floor same-floor / deviation counts. Calibration runs: **Kings' Rest 108/108, Temple of Sethraliss 123/123, Ruby Life Pools 162/164** within 1 map unit.

## Verified

- 8 dungeons generated + inserted into bare mapping versions (804–811); every multi-floor partition hit 100% exclusive own-art coverage; overlays reviewed.
- **Den of Nalorakk** exercised the identity guardrail: the cut→floor order I was told contradicted the images' own "Heart of Rage" / "The Foraging" labels; the skill surfaced it, the flip was confirmed, and the corrected mapping was inserted — no silent mis-assignment.
- `composer run fix` (0 changed) and `composer run analyse` (no errors) pass; `code-review` self-run applied.


Wotuu (commented):
Ok

Issue #3593, opened 2026-07-17
Migrate away from unmaintained jquery-bar-rating (follow-up to #3590)


Wotuu:
:robot: ### Background

#3590 audited the bundled jQuery plugins for jQuery-4 removed-API usage and, as a
bonus, checked which of them are still maintained upstream. `jquery-bar-rating` is
not: last npm release **2017**, last GitHub push **2021** ([antennaio/jquery-bar-rating](https://github.com/antennaio/jquery-bar-rating)).

This is also the plugin that already broke production once: it called the
jQuery-4-removed `$.isNumeric`, which took down the pull lists on v15.6.0 (#3588,
fixed in #3589 with a shim). An unmaintained dependency means the next jQuery
major bump (or any other breaking upstream change) has no one to fix it upstream —
we'd be shimming around it again.

### Usage

Star-rating widget for pull lists, wired up in `resources/assets/js/custom/inline/common/maps/map.js`
(guarded behind a `MapContextLiveSession` check, see #3589).

### Task

Research an actively-maintained star-rating alternative (or a small custom
implementation, given the usage surface is narrow) and migrate the pull-list
rating widget to it. Update/replace `resources/assets/js/jquery-isnumeric-shim.test.js`
and the `jquery4-plugin-audit.test.js` coverage from #3590 accordingly once the
plugin is swapped out.

### Related

- #3590 (audit that surfaced this)
- #3588 / #3589 (the incident this plugin already caused)

PR #3621, opened 2026-07-20
#3597 Replace unmaintained password-strength-meter with self-owned widget

### LINKED ISSUE #3597: Migrate away from unmaintained password-strength-meter (follow-up to #3590)
:robot: ### Background

bonus, checked which of them are still maintained upstream. `password-strength-meter`
is not: last npm release **2020**, last GitHub push **2022** ([elboletaire/password-strength-meter](https://github.com/elboletaire/password-strength-meter)).

Worth noting: this plugin initially looked unused during the #3590 audit because it
registers as `$.fn.password`, not `$.fn.pwstrength` (the bootstrap.js variable name
is `window.pwstrengthmeter`) - a first-pass grep for the wrong method name missed
the real call site.

### Usage

Registration password strength meter, wired up in
`resources/assets/js/custom/inline/layouts/app.js` (`_newPassword`, called for
`#register_password` and `#modal-register_password`) via `.password({...})`.

### Task

Research an actively-maintained password-strength-meter alternative (e.g. zxcvbn
paired with a small custom UI, since the current usage is a single self-contained
widget) and migrate `_newPassword` to it. Update/remove the
`password-strength-meter` block in `resources/assets/js/jquery4-plugin-audit.test.js`
(added in #3590) once swapped.

### Related

Wotuu:
:robot: Closes #3597

## Summary
- `password-strength-meter` is unmaintained (last npm release 2020, last GitHub push 2022). Replaces it with a self-owned `$.fn.passwordStrength` widget (same pattern as the #3593 star-rating migration).
- Renders Bootstrap's own `.progress`/`.progress-bar` component (role=progressbar, aria-valuenow/min/max) instead of hand-rolled bar CSS, matching every other progress bar in the app (e.g. `misc/mapping.blade.php`).
- Scoring is a small, **dependency-free** character-class/repetition heuristic - base score from length, a penalty for short repeating patterns ("aaaa", "abab", "abcabc", ...), and bonuses for case mixing / digits / symbols. This is the same *kind* of check the old plugin did, just self-owned instead of pulled from an abandoned package.
- `_newPassword()` (used by `#register_password`, `#modal-register_password`, and `#new_password` on the profile "change password" tab) keeps the same `shortPass`/`badPass`/`goodPass`/`strongPass`/`minimumLength` options.

## A dictionary-backed library was tried first and reverted
An earlier version of this PR used [zxcvbn-ts](https://github.com/zxcvbn-ts/zxcvbn) (the maintained continuation of Dropbox's zxcvbn) for proper dictionary-aware scoring. It worked, but its merged English dictionaries added **+1.6MB minified / +834KB gzip (+180%)** to the site's single global JS bundle (confirmed via esbuild's production build + metafile) - and since this app intentionally builds one monolithic IIFE bundle (see `scripts/build/app.mjs`, chosen to preserve CommonJS plugin-registration order, #2232/#3449), that's downloaded by **every page**, not just the two register fields and the profile password tab that use it. Lazily constructing the zxcvbn factory only deferred the CPU cost, not the download.

That's too much weight for a login-form nicety, so this PR drops the dependency entirely for the heuristic described above. Trade-off: it's not dictionary-aware (e.g. `P@ssw0rd` scores well despite being a well-known bad password) - the same limitation the original plugin had, and no worse than before this migration.

## Known limitation (intentionally not addressed)
The old plugin's `containsUsername` check (flagging a password containing the account's username) was already dead code before this migration - the app.js call site passed the wrong option name and never set the selector the plugin needed to enable the check at all, so it never fired in production. I removed the now-fully-orphaned `js.contains_username` lang key rather than re-wire the check, to keep this migration behavior-equivalent. Happy to file a follow-up issue if that check is wanted.

## Test plan
- [x] `npx vitest run` - 228/228 passing (`password-strength.test.js` covers short/weak/repeating/medium/strong tiers + init guards)
- [x] `composer run fix` / `composer run analyse` - clean (no PHP changed besides the lang file)
- [x] `npm run development` build succeeds; production bundle size confirmed back at baseline (1.55MB minified / 474KB gzip, vs 3.07MB / 1.27MB with zxcvbn-ts)
- [x] Verified in headless Chrome on all three call sites: register page, register modal, and the profile "change password" tab (`#new_password`) - short/weak/medium/strong each render the correct color/text, updates instantly (no debounce needed), no console or page errors


Wotuu:
:robot: Visual verification (register page, password set to a common weak word so the meter has something to show):

**Before (master, old plugin):**
before

**After (this branch, Bootstrap progress bar + zxcvbn-ts):**
after

Also manually verified the register modal and the profile "change password" tab (`#new_password`) - all three call sites render correctly (weak/medium/strong tiers, correct color/text) with no console or page errors.

Wotuu:
:robot: Update: dropped zxcvbn-ts after Wotuu flagged the bundle-size cost during review (it added +1.6MB minified/+834KB gzip to the site's single global JS bundle, downloaded on every page - lazy construction only deferred the CPU cost, not the download, since esbuild bundles every require() into that one file regardless of call-site). Replaced with a small dependency-free character-class/repetition heuristic instead - same idea the old plugin used, just self-owned. Bundle size confirmed back at baseline (1.55MB minified / 474KB gzip). PR description updated with details.

Wotuu (commented):
Ok

Issue #3590, opened 2026-07-17
Audit bundled jQuery plugins for other jQuery 4.0 removals (follow-up to #3588)


Wotuu:
:robot: ### Background

#3560 bumped jQuery **3.6.1 → 4.0.0**. jQuery 4.0 removed a number of long-deprecated global helpers. This already caused a production incident: `$.isNumeric` was removed, which the bundled `jquery-bar-rating` plugin still calls, breaking the pull lists on v15.6.0 (fixed in #3589 / #3588 with a targeted shim + guard).

That fix only addresses `$.isNumeric`. Other bundled legacy jQuery plugins may call other removed APIs that simply haven't been exercised yet.

### Task

Audit the bundled jQuery plugins (and our own code) for other jQuery-4 removals and either shim, patch, or replace as needed. Removed/changed APIs to grep for include:

- `$.isNumeric`, `$.isFunction`, `$.isArray`, `$.type`, `$.isWindow`
- `$.trim`, `$.nodeName`, `$.parseJSON`, `$.holdReady`, `$.camelCase`, `$.now`
- `$.fn.bind/unbind/delegate/undelegate` removals, `$.fn.hover`, `$.cssNumber` changes
- `jQuery.Deferred` / `.andSelf()` / other 3.x-deprecated methods

Plugins to check (from `resources/assets/js/bootstrap.js`): `jquery-bar-rating`, `jquery-mousewheel`, `jquery-visible`, `jquery-touchswipe`, `bootstrap5-toggle`, `lightslider`, `intro.js`, `password-strength-meter`, plus any inline usage in `resources/assets/js/**`.

### Notes

- The `$.isNumeric` shim lives in `resources/assets/js/bootstrap.js`; additional shims should sit alongside it, before any plugin `require`.
- Consider whether staying on jQuery 4 is worth it vs. pinning back to 3.x, given the number of legacy plugins in the bundle.

PR #3622, opened 2026-07-20
#3596 Migrate off unmaintained ion-rangeslider to a self-owned noUiSlider adapter

### LINKED ISSUE #3596: Migrate away from unmaintained ion-rangeslider (follow-up to #3590)
:robot: ### Background

bonus, checked which of them are still maintained upstream. `ion-rangeslider` is
not: last npm release **2019**, last GitHub push **2023** ([IonDen/ion.rangeSlider](https://github.com/IonDen/ion.rangeSlider)).

### Usage

**By far the largest usage surface of any plugin in this audit** - roughly 15 call
sites across:

- Heat-option controls: `heatoptionmaxhandler.js`, `heatoptionminopacityhandler.js`,
  `heatoptionradiushandler.js`, `heatoptionblurhandler.js`, `heatoptionmaxzoomhandler.js`
- Dungeon-route handlers: `minsamplesrequiredhandler.js`, `keylevelhandler.js`,
  `itemlevelhandler.js`, `playerdeathshandler.js`
- Search filters: `filterinputrating.js`, `filterinputkeylevel.js`,
  `filterinputitemlevel.js`, `filterinputplayerdeaths.js`, `filterinputduration.js`,
  `filterinputminsamplesrequired.js`
- `resources/assets/js/custom/inline/common/dungeonroute/simulate.js`
- `resources/assets/js/custom/inline/admin/npchealth/edit.js`

All under `resources/assets/js/custom/inline/`. Both `.ionRangeSlider({...})`
initialisation and the `.data('ionRangeSlider').update({...})` runtime-update API
are used throughout.

### Task

Given the size of this surface, this is the highest-risk/highest-effort migration
of the five follow-ups from #3590. Research an actively-maintained range-slider
alternative (e.g. noUiSlider) and plan a phased migration - likely worth its own
design pass given the number of call sites and the shared `init` + `update` usage
pattern across handler classes. Update/remove the `ion-rangeslider` block in
`resources/assets/js/jquery4-plugin-audit.test.js` (added in #3590) once swapped.

### Related

Wotuu:
:robot: Closes #3596

## Summary

Replaces the unmaintained `ion-rangeslider` npm package (last release 2019, last GitHub push 2023 — flagged by #3590's plugin audit as the heaviest of five follow-up migrations, ~15 call sites) with a self-owned `$.fn.rangeSlider` adapter (`resources/assets/js/range-slider.js`) around the actively-maintained `nouislider` package, mirroring the #3593 `star-rating.js` precedent.

The adapter preserves the exact call shape every handler/filter class already used — `$(selector).rangeSlider(opts).data('rangeSlider')` on init, `.update({min,max,from,to})` afterward — so the 10 handler classes + 6 filter classes needed only a mechanical rename (`.ionRangeSlider(` → `.rangeSlider(`, `.data('ionRangeSlider')` → `.data('rangeSlider')`). Only the adapter itself, `bootstrap.js`, the 3 direct-init call sites (`filterinputrating.js`, `simulate.js`, `admin/npchealth/edit.js`), and CSS needed real design work.

Key behavioral details preserved (verified live, not just assumed):
- `onFinish` fires on noUiSlider's `change` event only, never on programmatic `.update()` — matches ion-rangeslider's `is_update`-suppressed `onFinish` semantics (search filters, heat-option redraws).
- The `change` handler also triggers a native DOM `change` event on the input — `simulate.js` persists slider state purely via `$('#simulate_x').on('change', ...)` with no `onFinish` at all.
- When a caller doesn't pass `from`/`to` at init, the adapter falls back to the input's existing value (admin/npchealth and simulate.js pre-fill from a cookie, then init with no `from`).
- Tooltips, step-derived value formatting (so e.g. step `0.01` doesn't leak 2-decimal noise into `$input.val()`/visible filter headers), and a fixed 5-position grid (avoiding absurd pip density on tiny-step sliders) replace ion's grid/value-bubble behavior.

## Review

Self-reviewed with the `code-review` skill (xhigh effort, 6 finder angles). Fixed: NaN/empty-string values silently reaching noUiSlider from malformed URL query params or corrupted input state, `onFinish`'s payload type changing from numbers to strings, unlabeled sub-tick pip clutter (noUiSlider auto-fills between explicit pip positions unless filtered), a wasted `updateOptions()` call on every from/to-only `update()`, and two small simplifications. Two low-severity items left as documented tradeoffs (double-init silently no-ops; a measured, not dynamically-derived, CSS margin reserving tooltip/pip space).

## Test plan

- [x] `npm run test` — full suite green (233 tests), including a new `range-slider.test.js` (18 tests) covering init, value sync/formatting, the `onFinish`-on-`change`-not-`update` invariant, `.update()` semantics, and the pip-clutter regression
- [x] Visually verified in headless Chrome against a live master baseline: routes search/filter sidebar (key level, rating — screenshots below) and the heatmap search sidebar (key level, item level, player deaths, duration)
- [x] Behaviorally verified in a real logged-in session: admin NPC-health page restores a cookie-saved key level across reload (was silently resetting to `min` without the fallback-to-`.val()` fix); the native `change` event fires exactly once on real interaction and never on a programmatic `.update()` (the mechanism `simulate.js`'s settings-cookie persistence depends on)

### Before / after (search filters sidebar)

Before (master):
before

After (this branch):
after


Wotuu (commented):
Ok

PR #3625, opened 2026-07-20
#3623 Keep local dev read-only against real S3 thumbnails restored from a production DB backup

### LINKED ISSUE #3623: Local dev environments can mutate real S3 thumbnail objects from a restored production DB backup
:robot:

## Background

Wotuu restores a production database backup locally to see how the site looks with real data.
That backup's `dungeon_route_thumbnails`/`files` rows can legitimately still have `File.disk =
's3_user_uploads'`, pointing at the real production S3 bucket. Those objects must be treated as
**strictly read-only** from a local environment: valid to read/display, never to delete or
overwrite. If a thumbnail needs to be regenerated locally, the result should land on the `public`
disk instead - the row is free to move to `public`, but the original S3 object it used to point at
must be left alone.

This was discovered as a side-quest while investigating `RepairBrokenThumbnailsTest` hangs during
work on #3349 (route poster cards, unrelated feature). Two separate problems surfaced:

1. **`RepairBrokenThumbnails` command/test hang**: `requeueThumbnailsMissingFromDisk()` scans every
   file-backed `dungeon_route_thumbnails` row and does `Storage::disk($file->disk)->exists(...)`.
   With a restored backup, ~226k of ~226k rows have `disk = 's3_user_uploads'`, so this makes a
   real S3 HTTP call per row - which hangs in a sandboxed dev environment with no egress. Once
   faked, a second bug surfaced: the final re-queue step ran one un-batched
   `whereIn('id', $allAffectedIds)` across up to 186k ids, hitting MySQL's "1390 Prepared statement
   contains too many placeholders".
2. **No read-only protection at all for local mutation of S3 thumbnails**: `File::deleteFromDisk()`
   unconditionally calls `Storage::disk($this->disk)->delete($this->path)`. Several ordinary local
   dev actions delete a `DungeonRouteThumbnail` (and cascade into `File::delete()`):
   `DungeonRoute::boot()`'s `deleting` hook (deleting a route deletes its thumbnails), and
   `ThumbnailService::attachThumbnailToDungeonRoute()` (regenerating a thumbnail deletes the
   superseded old one). Any of these, run locally against a restored backup, issues a **real S3
   DELETE** against production data. Separately, `ThumbnailService::doCreateThumbnail()` already
   had a narrower guard (`isRemoteDiskUnsafeForLocalGeneration()`) that refused to generate
   *any* new thumbnail at all when `FILESYSTEM_DISK` was S3 in a local environment - too blunt,
   since it blocks all local thumbnail regeneration rather than just protecting the read-only data.

## Fix (implemented, not yet committed - sitting in the `3349-route-poster-card` worktree)

- `RepairBrokenThumbnails.php`: batch the affected-route id list itself via `array_chunk(..., 1000)`
  before each `whereIn(...)->chunkById(...)` pass, instead of one giant `whereIn` across all ids.
- `RepairBrokenThumbnailsTest.php`: fake every configured disk in `setUp()` (not just
  `config('filesystems.default')`), so the existence-check scan never makes a real S3 call in tests.
- `File::deleteFromDisk()`: skip the real `Storage::disk()->delete()` call when
  `app()->environment('local')` and the disk's driver is `s3` - the DB row deletion still proceeds
  normally, only the physical remote object is left untouched.
- `ThumbnailService::doCreateThumbnail()` / `copyThumbnails()`: when the resolved disk is unsafe for
  local generation (real S3 disk + local env), redirect the write to the `public` disk instead of
  refusing outright, so local dev can freely regenerate thumbnails.
- New `tests/Feature/App/Models/FileTest.php` and a new regression test in
  `ThumbnailServiceTest.php` covering the read-only guard end-to-end.

## Remaining work

1. Move the above changes (already implemented) out of the `3349-route-poster-card` worktree onto
   their own branch/worktree for this issue, since they're unrelated to that PR's feature.
2. Commit, run `composer run analyse` + `composer run fix`, open a PR (`Closes #<this issue>`).
3. Open decision, not yet resolved: even with the placeholder-limit fix,
   `RepairBrokenThumbnailsTest`'s non-dry-run methods are still *slow* locally (300k+ queries
   against the full shared, ever-growing local dev DB - not a hang, just slow). Wotuu said "the
   pipeline seems to work fine" on CI, implying this is local-environment-only (CI's test DB is
   much smaller). Leave as-is (CI unaffected) unless Wotuu wants a real design change (e.g. a
   mockable repository seam) to also make it fast locally.

## Definition of done

- `php artisan test --compact` green for `FileTest`, `ThumbnailServiceTest`, and
  `RepairBrokenThumbnailsTest`.
- PR opened, `Closes #<this issue>` in the body.

Wotuu:
:robot:

Closes #3623

## Summary
- A local environment restoring a production database backup can have `dungeon_route_thumbnails`/`files` rows that legitimately still point at the real `s3_user_uploads` S3 disk. Those objects must stay strictly read-only from local: valid to read/display, never to delete or overwrite. If a thumbnail needs to be regenerated locally, the result now lands on the `public` disk instead of refusing outright or (worse) mutating the real S3 object.
- `File::deleteFromDisk()` now skips the real `Storage::disk()->delete()` call when running locally against a remote (S3-driver) disk — the DB row deletion still proceeds, only the physical object is left alone. This is the central fix: it protects every call site that deletes a thumbnail, including ones with no protection at all before (e.g. `DungeonRoute` deletion cascading into its thumbnails).
- `ThumbnailService::doCreateThumbnail()`/`copyThumbnails()` no longer refuse entirely when the default disk resolves to real S3 locally (previous behavior: silently do nothing) — they redirect the write to `public` so local dev can regenerate thumbnails freely.
- Also includes an unrelated fix found while chasing the same behavior: `RepairBrokenThumbnails`' final re-queue step batches the affected-route id list via `array_chunk(..., 1000)` instead of one giant `whereIn()`, which could exceed MySQL's prepared-statement placeholder limit; its test now fakes every configured disk (not just the default) so the command's existence-check scan never makes a real S3 call.
- Self-review (code-review skill, 8 finder angles) caught a real regression the first pass introduced: `RenderThumbnail` (a dev-only command that intentionally forces an isolated disk so test renders never land on the shared `public` disk) would have had that isolation silently defeated by the new redirect-to-public logic if run with `--disk=s3_user_uploads`. Fixed by rejecting remote `--disk` values outright. Also consolidated the byte-identical "is this disk unsafe to mutate locally" check that had been duplicated in `File` and `ThumbnailService` into one place, and hoisted a per-loop-iteration disk check out of `copyThumbnails()`.

## Known follow-up (not in this PR)
Self-review also flagged that `File::saveFileToDB()` (generic uploads, e.g. avatars via `HasIconFile`) has no equivalent guard — a local environment with `FILESYSTEM_DISK=s3_user_uploads` could still upload a brand-new file straight into the real bucket. This is a different category of risk (new junk data, not corruption of existing prod data) and out of scope for this PR, which is scoped to the thumbnail read-only requirement. Left as a note on #3623 rather than fixed here to avoid scope creep.

## Test plan
- [x] `php artisan test --compact` green for `FileTest`, `ThumbnailServiceTest`, `RenderThumbnailTest`, `RepairBrokenThumbnailsTest` (the last one takes ~56min locally against the full shared dev DB — confirmed passing, this is a known local-only slowness, not a regression; CI's test DB is much smaller)
- [x] `composer run fix` clean (no unrelated reformatting)

### DIFF
--- a/app/Console/Commands/DungeonRoute/RenderThumbnail.php
+++ b/app/Console/Commands/DungeonRoute/RenderThumbnail.php
@@ -45,6 +45,16 @@ class RenderThumbnail extends Command
         // Force the target disk so the render lands on the ISOLATED local disk (storage/app/private)
         // instead of the shared, bind-mounted public disk.
         $disk = (string)$this->option('disk');
+
+        // A remote (S3) disk defeats the whole point of this command - ThumbnailService silently
+        // redirects a real-S3 write to the public disk instead, which is exactly the shared disk
+        // this command exists to avoid touching.
+        if (config(sprintf('filesystems.disks.%s.driver', $disk)) === 's3') {
+            $this->error(sprintf('--disk=%s is a remote disk; this command only supports isolated local disks.', $disk));
+
+            return self::FAILURE;
+        }
+
         config(['filesystems.default' => $disk]);
 
         // The database is shared across the main stack and all worktrees, so persisting a thumbnail
--- a/app/Console/Commands/DungeonRoute/RenderThumbnail.php
+++ b/app/Console/Commands/DungeonRoute/RenderThumbnail.php
@@ -3,6 +3,7 @@
 namespace App\Console\Commands\DungeonRoute;
 
 use App\Models\DungeonRoute\DungeonRoute;
+use App\Models\File;
 use App\Models\Floor\Floor;
 use App\Service\DungeonRoute\ThumbnailServiceInterface;
 use Illuminate\Console\Command;
@@ -49,7 +50,7 @@ class RenderThumbnail extends Command
         // A remote (S3) disk defeats the whole point of this command - ThumbnailService silently
         // redirects a real-S3 write to the public disk instead, which is exactly the shared disk
         // this command exists to avoid touching.
-        if (config(sprintf('filesystems.disks.%s.driver', $disk)) === 's3') {
+        if (File::isRemoteDisk($disk)) {
             $this->error(sprintf('--disk=%s is a remote disk; this command only supports isolated local disks.', $disk));
 
             return self::FAILURE;
--- a/app/Console/Commands/DungeonRoute/RepairBrokenThumbnails.php
+++ b/app/Console/Commands/DungeonRoute/RepairBrokenThumbnails.php
@@ -90,8 +90,8 @@ class RepairBrokenThumbnails extends Command
      */
     private function requeueThumbnailsMissingFromDisk(ThumbnailServiceInterface $thumbnailService, bool $dryRun): void
     {
-        $affectedDungeonRouteIds  = [];
-        $brokenCustomThumbnailIds = [];
+        $affectedDungeonRouteIds  = collect();
+        $brokenCustomThumbnailIds = collect();
 
         // The model's default $with would also hydrate the unused floor relation for every row.
         $query = DungeonRouteThumbnail::query()
@@ -104,16 +104,16 @@ class RepairBrokenThumbnails extends Command
 
         $progressBar = $this->output->createProgressBar($total);
 
-        $query->clone()->chunkById(100, function ($thumbnails) use (&$affectedDungeonRouteIds, &$brokenCustomThumbnailIds, $progressBar): void {
+        $query->clone()->chunkById(100, function ($thumbnails) use ($affectedDungeonRouteIds, $brokenCustomThumbnailIds, $progressBar): void {
             foreach ($thumbnails as $thumbnail) {
                 /** @var DungeonRouteThumbnail $thumbnail */
                 $file = $thumbnail->file;
 
                 if ($file !== null && !Storage::disk($file->disk)->exists($file->path)) {
                     if ($thumbnail->custom) {
-                        $brokenCustomThumbnailIds[] = $thumbnail->id;
+                        $brokenCustomThumbnailIds->push($thumbnail->id);
                     } else {
-                        $affectedDungeonRouteIds[$thumbnail->dungeon_route_id] = true;
+                        $affectedDungeonRouteIds->push($thumbnail->dungeon_route_id);
                     }
                 }
 
@@ -124,17 +124,19 @@ class RepairBrokenThumbnails extends Command
         $progressBar->finish();
         $this->newLine();
 
-        if ($brokenCustomThumbnailIds !== []) {
+        $affectedDungeonRouteIds = $affectedDungeonRouteIds->unique();
+
+        if ($brokenCustomThumbnailIds->isNotEmpty()) {
             $this->warn(sprintf(
                 'Found %d custom thumbnail(s) whose disk object is missing - these cannot be repaired by re-queueing (thumbnail ID(s): %s).',
-                count($brokenCustomThumbnailIds),
-                implode(', ', $brokenCustomThumbnailIds),
+                $brokenCustomThumbnailIds->count(),
+                $brokenCustomThumbnailIds->implode(', '),
             ));
         }
 
-        $this->info(sprintf('Found %d route(s) with a thumbnail File whose disk object is missing.', count($affectedDungeonRouteIds)));
+        $this->info(sprintf('Found %d route(s) with a thumbnail File whose disk object is missing.', $affectedDungeonRouteIds->count()));
 
-        if ($affectedDungeonRouteIds === []) {
+        if ($affectedDungeonRouteIds->isEmpty()) {
             return;
         }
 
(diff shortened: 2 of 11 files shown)


app/Console/Commands/DungeonRoute/RepairBrokenThumbnails.php
Wotuu: Please convert both to Collection instances, that should clean the remaining code up the way I want it to be.

Wotuu (commented):
One small change.

app/Console/Commands/DungeonRoute/RepairBrokenThumbnails.php
Wotuu: :robot: Done — `$affectedDungeonRouteIds` and `$brokenCustomThumbnailIds` are now `Collection` instances built with `->push()`. `$affectedDungeonRouteIds` gets an explicit `->unique()` after the scan (replacing the old array-key dedup), and the id-batching loop now uses `$affectedDungeonRouteIds->chunk(1000)` directly instead of `array_chunk(array_keys(...), 1000)`.

app/Console/Commands/DungeonRoute/RenderThumbnail.php
```diff
@@ -45,6 +45,16 @@ public function handle(ThumbnailServiceInterface $thumbnailService): int
         // Force the target disk so the render lands on the ISOLATED local disk (storage/app/private)
         // instead of the shared, bind-mounted public disk.
         $disk = (string)$this->option('disk');
+
+        // A remote (S3) disk defeats the whole point of this command - ThumbnailService silently
+        // redirects a real-S3 write to the public disk instead, which is exactly the shared disk
+        // this command exists to avoid touching.
+        if (config(sprintf('filesystems.disks.%s.driver', $disk)) === 's3') {
```
Wotuu: :robot: **Third copy of the s3-driver check** — this PR's stated cleanup was consolidating the byte-identical "is this disk s3" check out of `File` and `ThumbnailService` into `File::isRemoteDiskProtectedFromLocalMutation()`. But this inline `config(sprintf('filesystems.disks.%s.driver', $disk)) === 's3'` re-implements the driver half of that same check a third time. It can't reuse `isRemoteDiskProtectedFromLocalMutation()` directly because that also gates on `environment('local')` and this command needs to reject S3 regardless of env. Consider extracting an env-agnostic `File::isRemoteDisk(string $disk): bool` that both `isRemoteDiskProtectedFromLocalMutation()` and this guard call, so there's one place that defines "remote/S3 disk".

tests/Feature/App/Service/DungeonRoute/ThumbnailServiceTest.php
```diff
@@ -54,24 +54,25 @@ private function restoreEnvAndDefaultDisk(array $original): void
...
             // Act
-            $result = $service->createThumbnail($dungeonRoute, 0);
-
-            // Assert
-            $this->assertNull($result);
+            // Falls through past the disk-safety guard onto the public disk instead of returning
+            // null; fails further down since there's no real dungeon/floor/puppeteer setup here,
+            // which is fine - this test only asserts the redirect (and its log call) happened.
+            $service->createThumbnail($dungeonRoute, 0);
+        } catch (Throwable) {
```
Wotuu, line 76: :robot: **Test name promises more than it verifies.** This test is named `redirectsToPublicDiskInsteadOfSkipping`, but it swallows every `Throwable` and asserts nothing about the write destination — the only real assertion is the mock's `doCreateThumbnailRedirectedRemoteDiskFromLocal` expectation, i.e. that the *log* fired. A regression where the redirect decision logs but the write still lands on the original S3 disk (e.g. someone reorders `$disk = 'public'` above/below the guard, or drops the reassignment) would keep this test green. The actual disk-write behavior is covered by `attachThumbnailToDungeonRoute_...leavesTheRemoteFileUntouched`, so this is a minor coverage gap, not a hole — but the name over-claims. Worth either renaming to reflect that it only asserts the redirect log, or narrowing the `catch` so an unexpected early throwable can't mask a redirect that never happened.

Wotuu:

app/Console/Commands/DungeonRoute/RenderThumbnail.php
Wotuu: :robot: Done — extracted `File::isRemoteDisk(string $disk): bool` (env-agnostic driver check), had `isRemoteDiskProtectedFromLocalMutation()` delegate to it, and switched `RenderThumbnail` to call it directly instead of reimplementing the `config(sprintf(...)) === 's3'` check inline.

tests/Feature/App/Service/DungeonRoute/ThumbnailServiceTest.php
Wotuu: :robot: Done — renamed the test to `createThumbnail_givenLocalEnvironmentAndRemoteDisk_logsRedirectToPublicDisk` and added a comment pointing at `attachThumbnailToDungeonRoute_givenLocalEnvironmentAndOldThumbnailOnRemoteDisk_leavesTheRemoteFileUntouched`, which is the test that actually covers the disk-write destination.

Wotuu (commented):
Ok

Issue #3627, opened 2026-07-20
MDT import/export does not handle raid markers (follow-up to #1431, #1453)


Wotuu:
## Background

#1431 restored the raid marker radial menu (the JS plugin backing right-click assignment had
been disabled since 2022) and #1453 made raid marker assignments survive a dungeon's mapping
version upgrade (mirroring the `KillZoneEnemy` npc_id/mdt_id approach). Raid markers are now
fully functional in the editor and persist correctly across mapping version bumps.

Wotuu asked on #1431: "double check if the MDT integration works properly for this
functionality. Can I import an MDT string with raid marker assignments and have them copied
over properly, and the other way around when exporting?"

Answer from investigating while working #1431/#1453: **there is currently zero MDT handling of
raid markers anywhere in the codebase.** This isn't a regression or a broken integration - it was
simply never built. Scoping this in would have blown past a single session for #1431/#1453, so
it's being filed here instead.

## Remaining work

Two independent integration points, neither of which currently reads or writes raid marker data:

1. **Export** - `app/Service/MDT/MDTExportStringService.php` serializes a `DungeonRoute` into a
   compressed MDT-importable string (object structure, Lua serialization - see the `mdt-export`
   skill). It has no code path that reads `DungeonRoute::enemyRaidMarkers()` or emits raid marker
   data into the exported objects.
2. **Import** - `app/Service/MDT/Import/MDTImportStringService.php` and `PullImporter.php` parse a
   compressed MDT string into a `DungeonRoute` (see the `mdt-import` skill). They have no code
   path that reads raid marker assignments out of the imported MDT Lua data and creates
   `DungeonRouteEnemyRaidMarker` rows.

## Key files

- `app/Service/MDT/MDTExportStringService.php`
- `app/Service/MDT/Import/MDTImportStringService.php`
- `app/Service/MDT/Import/PullImporter.php`
- `app/Models/DungeonRoute/DungeonRouteEnemyRaidMarker.php` (npc_id/mdt_id/enemy_id, post-#1453)
- `app/Http/Controllers/Ajax/AjaxEnemyController.php::setRaidMarker()` for the live-assignment
  shape to mirror

## Risks / notes

- MDT's own raid marker representation in its Lua format needs to be confirmed first (check a
  real MDT export/import fixture for whether/how it encodes marker assignments per clone).
- Should reuse the npc_id/mdt_id resolution pattern from #1453 rather than raw enemy IDs, so
[...]

Issue #3588, opened 2026-07-17
Pull lists not shown: jQuery 4.0 removed $.isNumeric, breaks jquery-bar-rating


Wotuu:
:robot: ### Bug

On production (v15.6.0) the pull lists do not render on dungeon map pages. The browser console shows:

```
Uncaught TypeError: t.isNumeric is not a function
    at t.fn.barrating (app-v15.6.0.js)
    ... activate (custom-v15.6.0.js:1658)
```

### Root cause

Dependabot bumped **jQuery 3.6.1 → 4.0.0** (#3560, commit `39be777c0`), which was merged *before* the v15.6.0 tag. jQuery 4.0 **removed the deprecated `$.isNumeric()`** method. The bundled `jquery-bar-rating@1.2.2` plugin still calls `$.isNumeric()` internally (upstream bug: antennaio/jquery-bar-rating#129), so it throws the moment `.barrating()` runs during map activation. Because the exception aborts the rest of map init, downstream rendering — including the pull lists — never runs.

`git tag --contains 39be777c0` = **v15.6.0, v15.6.1, v15.6.2**, so staging (v15.6.2) has the identical bug.

### Fix

1. Re-add a guarded `$.isNumeric` shim in `resources/assets/js/bootstrap.js` before any jQuery plugin loads (keeps jQuery 4.0.0).
2. Guard the live-session `.barrating()` call in `resources/assets/js/custom/inline/common/maps/map.js` behind a `MapContextLiveSession` check so it only runs for live sessions.

This is a compiled-asset change, so it ships as a new release (not an S3 hotfix).

### Follow-up

A separate issue should audit the other bundled legacy jQuery plugins for further jQuery-4 removals (`$.trim`, `$.isFunction`, `$.type`, ...).

Issue #3246, opened 2026-06-03
KillZone's getEnemiesAttribute needs to use eloquent instead


Wotuu:
Each DungeonRoute has 1..N KillZones attached to it. Each KillZone has 0..N enemies. If I recall correctly, due to the way mapping versions work I couldn't figure out how to use Eloquent in a proper way to simply be able to use $killZone->enemies and have it be correct. The `getEnemies()` function also suffers from this. 

`\App\Models\DungeonRoute\DungeonRoute::mapContextKillZones` also uses some custom queries which it shouldn't have to, I think, if it were able to use proper Eloquent. Performance is crucial in this part of the code, as is being able to resolve to the correct enemies. A KillZone stores an `npc_id` and an `mdt_id`, which resolve to an `enemy_id` based on the `mapping_version_id`.

If this code is cleaned up and performance improved, I consider this story done. For example, measure the performance of loading the same route (accessing `dungeonroute.view` route for example) to ensure we haven't regressed in performance.

Issue #1448, opened 2022-11-13
Pulls should use npc_id + mdt_id to save pulls with, instead of the raw enemy ID


Wotuu:
This makes it very easy to switch mapping versions of routes without having to re-create all the pulls.

PR #3630, opened 2026-07-20
#1431 Restore raid marker assignment + #1453 mapping-version safety

### LINKED ISSUE #1431: Assigning raid markers is not functional at the moment
I had to comment out a library I use in bootstrap.js because npm wouldn't compile my project without it. This needs to be looked into and resolved.

### LINKED ISSUE #1453: Assigning raid markers need to use mapping versions

Wotuu:
:robot:

Closes #1431
Closes #1453

## Summary

**Part 1 (#1431)** — the right-click radial menu for assigning raid markers depended on `zikes-circlemenu`, an old jQuery plugin whose `require()` was commented out in 2022 when it broke the webpack build (commit `e78958f43`), with the call site later also disabled (`21dbf561b`). Re-enabled both. Verified in headless Chrome against jQuery 4.0.0 (upgraded this month, #3588/#3590) — the vendor plugin's `.bind()`/`.on()`/`.data()` usage runs cleanly, the menu opens, a marker selection persists via `AjaxEnemyController::setRaidMarker` and renders.

**Part 2 (#1453)** — `DungeonRouteEnemyRaidMarker` stored a raw `enemy_id`, so any assignment silently broke across a mapping-version upgrade. Mirrors the existing `KillZoneEnemy` fix (#1448/#1454, #3246): `npc_id`+`mdt_id` become the source of truth, `enemy_id` the cached/derived pointer, re-resolved in `DungeonRouteService::upgradeMappingVersion()`.

## A bug this surfaced (and fixed)

Ran a multi-angle code review before opening this. It found that the mapping-version resolution join (mine, and the pre-existing `KillZoneEnemyRepository` pattern it mirrors) matched `mdt_id` with plain SQL `=`, which never matches `NULL` against `NULL`. ~2% of enemies (those placed outside an MDT import — vendors, quest NPCs, etc.) legitimately have a null `mdt_id`, so a raid marker assigned to one would get silently deleted as "orphaned" on the very next mapping version upgrade even though the NPC still existed. Confirmed against the shared dev DB: 36 of 2203 existing raid marker rows were affected. Fixed with MySQL's null-safe `<=>` operator, plus the migration backfill was skipping these rows entirely for the same reason.

The identical latent bug exists in the already-shipped `KillZoneEnemyRepository` (not touched by this MR) — checked its blast radius there: 11 of ~52M `kill_zone_enemies` rows, negligible, so I didn't file a separate follow-up for it, just flagging it here.

## Other review findings, considered and not applied

- **Shared trait for the "reset → re-resolve by mapping version → delete orphaned" pattern**, now duplicated between `KillZoneEnemyRepository` and the new `DungeonRouteEnemyRaidMarkerRepository`. Left as-is: the issue explicitly asked to mirror the existing pattern, and refactoring already-shipped, tested `KillZoneEnemyRepository` is a separate, riskier change deserving its own PR.
- **`DungeonRoute::cloneRelationsInto()` raid-marker branch** (checklist item from the issue): confirmed no special branch is needed — matches existing `KillZoneEnemy` precedent, which also doesn't re-resolve on clone. The regular clone path keeps the same `mapping_version_id` (no drift); the template-clone path has a pre-existing, equal-severity theoretical gap for both `KillZoneEnemy` and raid markers that isn't new here.
- **Unused composite index** on `(npc_id, mdt_id)`: kept it since the issue's own plan explicitly requested it (mirroring the legacy `kill_zone_enemies` migration).

## MDT round-trip

Wotuu asked (comment on #1431) whether MDT import/export handles raid markers. Answer: zero MDT handling exists for raid markers anywhere in the codebase — never built, not broken. Filed as a separate follow-up: #3627.

## Test plan

- [x] `AjaxEnemyControllerTest` — assignment persists `npc_id`/`mdt_id`, clearing deletes the row
- [x] `DungeonRouteEnemyRaidMarkerMappingVersionTest` — marker preserved across upgrade (normal + null-`mdt_id` enemy cases), marker dropped when the NPC no longer exists in the new mapping version
- [x] Manual verification in headless Chrome: right-click → menu opens → marker selected → persists → renders, full round trip through a real sandbox route, no console errors


app/Http/Controllers/Ajax/AjaxEnemyController.php
```diff
@@ -119,15 +119,23 @@ public function setRaidMarker(Request $request, DungeonRoute $dungeonRoute, Enem
...
         try {
             $raidMarkerName = $request->get('raid_marker_name', '');
+            $npcId          = $enemy->getMdtNpcId();
 
-            // Delete existing enemy raid marker
-            DungeonRouteEnemyRaidMarker::where('enemy_id', $enemy->id)->where('dungeon_route_id', $dungeonRoute->id)->delete();
+            // Delete existing enemy raid marker, matched by the durable npc_id/mdt_id identity
+            // rather than the mapping-version-scoped cached enemy_id, so a stale client reference
+            // (e.g. after a mapping version upgrade) can't leave a duplicate row behind
+            DungeonRouteEnemyRaidMarker::where('dungeon_route_id', $dungeonRoute->id)
```
Wotuu, line 143: :robot: **Correctness / data loss — over-delete by non-unique `(npc_id, mdt_id)`**

This delete now keys on `(dungeon_route_id, npc_id, mdt_id)` instead of the unique `enemy_id`. But `(resolved npc_id, mdt_id)` is **not unique within a mapping version** — I confirmed 19 such duplicate groups in the current DB (e.g. `mv=7 npc=95832 mdt_id=3` maps to 2 distinct enemies; these arise from facade/floor duplication, and it is *not* limited to the null-`mdt_id` case).

Consequence: if a route has raid markers on two distinct enemies that share the same `(npc_id, mdt_id)`, assigning/clearing a marker on one silently deletes the sibling's marker too — the `create` right after only re-inserts the clicked `enemy_id`, so the other enemy permanently loses its marker on the next map load. The old `where('enemy_id', ...)` delete could never do this. It is now impossible to keep independent markers on two same-identity enemies in one route.

app/Repositories/Database/DungeonRoute/DungeonRouteEnemyRaidMarkerRepository.php
```diff
@@ -5,11 +5,37 @@
...
+    {
+        // Null-safe equality (<=>) for mdt_id: enemies placed outside of an MDT import
+        // legitimately have a null mdt_id, and plain `=` never matches NULL against NULL.
+        DB::statement('
+            UPDATE dungeon_route_enemy_raid_markers drerm
+            JOIN enemies e
+                ON drerm.npc_id = COALESCE(e.mdt_npc_id, e.npc_id)
+                AND drerm.mdt_id <=> e.mdt_id
+                AND e.mapping_version_id = ?
+            SET drerm.enemy_id = e.id
```
Wotuu, line 40: :robot: **Correctness — `SET enemy_id = e.id` resolves to an arbitrary enemy for non-unique identities**

Because `enemies(npc_id, mdt_id)` is not unique within a mapping version (verified: 19 duplicate groups locally, all with non-null `mdt_id`), this `UPDATE ... JOIN` can match two `enemies` rows for a single raid marker. MySQL then assigns an arbitrary one of them to `enemy_id`.

After `upgradeMappingVersion`, a marker the user placed on a specific enemy instance can silently re-bind to the *other* same-identity enemy (potentially a different floor/pack). And when two markers in the route share the identity, both collapse onto whichever `e.id` the join happens to pick, effectively losing one. The removed exact `enemy_id` pointer had no such ambiguity. (Same latent ambiguity exists in the mirrored `KillZoneEnemyRepository`, but it is genuinely reachable here.)

app/SeederHelpers/RelationImport/Parsers/Relation/DungeonRouteEnemyRaidMarkersRelationParser.php
```diff
@@ -24,12 +28,28 @@ public function canParseRelation(string $name, array $value): bool
...
     {
+        $enemyIds = array_column($value, 'enemy_id');
+
+        /** @var Collection<int, Enemy> $enemies */
+        $enemies = Enemy::from(DatabaseSeeder::getTempTableName(Enemy::class))->whereIn('id', $enemyIds)->get()->keyBy('id');
+
         foreach ($value as $enemyRaidMarkerData) {
+            $enemy = $enemies->get($enemyRaidMarkerData['enemy_id']);
+            if ($enemy === null) {
+                throw new Exception(sprintf('Unable to find enemy with id %s', $enemyRaidMarkerData['enemy_id']));
```
Wotuu: :robot: **Robustness — one stale marker now aborts the entire `DungeonDataSeeder`**

The old parser inserted the raid marker with the fixture's raw `enemy_id` (dangling ids tolerated, no FK). This new `throw` turns a single unresolved `enemy_id` into an uncaught exception that aborts the whole `php artisan db:seed --class=DungeonDataSeeder` run.

This is dormant today (all `enemyRaidMarkers` arrays in `database/seeders/dungeondata/**` are currently `[]`), but it goes live the moment `mapping:save` exports a real raid marker whose enemy is later removed from that dungeon's mapping — global seeding would then hard-fail instead of skipping the one bad row. The sibling KillZone parser throws similarly, but raid markers can reference enemies that no kill zone does, so this is an incremental abort surface. Consider skipping (and logging) the row rather than aborting the seed.

tests/Feature/App/Service/DungeonRoute/DungeonRouteEnemyRaidMarkerMappingVersionTest.php
```diff
@@ -0,0 +1,164 @@
...
+
+        $route      = DungeonRoute::factory()->create(['dungeon_id' => $dungeon->id, 'mapping_version_id' => $existingMV->id]);
+        $raidMarker = $this->createRaidMarker($route, $enemy);
+
+        try {
+            /** @var Enemy $clonedEnemy */
+            $clonedEnemy = Enemy::where('mapping_version_id', $newMV->id)
+                ->where('npc_id', $enemy->npc_id)
+                ->where('mdt_id', $enemy->mdt_id)
+                ->firstOrFail();
```
Wotuu, line 58: :robot: **Test flakiness — assertion can diverge from the resolver for duplicate-identity enemies**

The expected clone is fetched via `->where('npc_id', ...)->where('mdt_id', ...)->firstOrFail()`, while production resolves `enemy_id` via an independent `UPDATE ... JOIN`. When `inRandomOrder()->first()` above happens to land on an enemy whose `(npc_id, mdt_id)` is one of the ~19 duplicate groups, `firstOrFail()` and the JOIN can pick *different* rows, so `assertEquals($clonedEnemy->id, $fresh->enemy_id)` fails intermittently. Pin a unique-identity enemy for the source (or assert membership in the candidate set) rather than equality against an independently-fetched `first()`.

tests/Feature/App/Service/DungeonRoute/DungeonRouteEnemyRaidMarkerMappingVersionTest.php
```diff
@@ -0,0 +1,164 @@
...
+                continue;
+            }
+
+            $enemy = Enemy::where('mapping_version_id', $mappingVersion->id)->whereNull('mdt_id')->inRandomOrder()->first();
+            if ($enemy !== null) {
+                return [$dungeon, $mappingVersion, $enemy];
+            }
+        }
+
+        throw new RuntimeException('Unable to find a retail dungeon with a null-mdt_id enemy on its current mapping version');
```
Wotuu: :robot: **Test robustness — hard `RuntimeException` when no null-`mdt_id` enemy exists in the seed**

This helper throws (erroring the test) if the seeded DB contains no enemy with a null `mdt_id`. That is exactly the state of the current main dev stack: I measured **0 of 81,672** enemies with a null `mdt_id` (both the dev and `phpunit` connections). So this test errors when run locally against the main stack, and only passes in CI because CI's seed happens to contain such enemies — a brittle dependency on seed-data composition.

Prefer `markTestSkipped(...)` over `throw new RuntimeException(...)` so the suite degrades gracefully where the precondition can't be met. (Also worth reconciling with the PR description's "~2% of enemies have a null `mdt_id`" claim, which doesn't hold on this data set.)

Wotuu:

Wotuu:
:robot: Found and fixed three issues while verifying this in the browser:

1. **Right-click did two things at once**: it opened both the enemy details modal and the raid marker circle menu. Shift+right-click now opens the circle menu; plain right-click keeps opening details (same as view mode). Tooltip updated to mention the shortcut.
2. **Circle menu icons (and the assigned marker on the enemy) always rendered as a star**, regardless of which marker was picked. Root cause: a Sass `background` shorthand in `assets.scss` was clobbering the per-icon `background-position` rules compiled earlier via `@use`. Split into `background-image`/`background-repeat`.
3. **Latent bug from #1138**: assigning a marker to an enemy that never had one before didn't render until the next page load, because the visual-rebuild callback on `enemy:set_raid_marker` was `.bind()`-ed but never invoked.

Verified in headless Chrome: plain right-click → details only; shift+right-click → circle menu only, with all 8 distinct marker icons; selecting one persists and renders correctly, including after a fresh page load.

Wotuu:
:robot: Two more issues from further testing, now fixed and pushed:

1. **Shift+right-click didn't work when clicking the marker icon rendered on top of an enemy** (only worked on the bare icon underneath). The marker overlay is a sibling of the enemy icon, not a descendant, so the click handler bound directly to the enemy icon never received it. Fixed by routing through Enemy's existing layer-level contextmenu listener (which reliably covers every sub-element) instead of a separate DOM binding.
2. **Deselecting a marker threw an uncaught TypeError** from Bootstrap's Tooltip internals — disposing a tooltip while its own hover-triggered transition was still in flight. Fixed by deferring tooltip disposal to after the circle menu's removal delay, once any transition has settled.

Verified in headless Chrome: shift+right-click now opens the menu from the overlay icon too, and repeated assign/deselect cycles (including hover-then-click, matching the original repro) produce zero console errors.

Wotuu:
:robot: Addressed all 5 findings from the cold review:

1. **Sibling marker collateral deletion** (`AjaxEnemyController`) — the npc_id/mdt_id identity isn't unique within a mapping version (19 duplicate groups exist). The delete now only falls back to identity-based matching when it currently resolves to exactly one enemy; otherwise it matches the exact `enemy_id` only.
2. **Arbitrary enemy resolution on upgrade** (`DungeonRouteEnemyRaidMarkerRepository`) — added a uniqueness guard so an ambiguous identity is left unresolved (and dropped by the existing orphan cleanup) instead of MySQL picking an arbitrary match.
3. **Seeder abort on one bad row** — now logs and skips instead of throwing; added test coverage (there were no existing tests for this class of seeder parser).
4. **Test flakiness from duplicate identities** — excluded ambiguous-identity enemies from the random pick.
5. **Hard failure when no null-mdt_id enemy exists** — changed to `markTestSkipped`; confirmed this is the actual cause of the `feature-other` CI failure on this PR.

Also found and fixed a pre-existing crash in `EnemyPatrol::cloneForNewMappingVersion` (missing null-safe operator, unrelated to this PR but was intermittently crashing this exact test suite since it exercises the same clone path on a randomly-picked dungeon) — stable across 5+ repeated runs now.

Ran a second review pass on my own fixes and it caught a real regression in fix #1 (a stale-client-after-upgrade case that the original delete handled correctly but my first pass broke) — fixed before pushing.

Separately noticed the `feature-controller` CI failure (`DungeonRouteControllerSaveNewTemporaryTest::saveNewTemporary_givenValidStartMapIcon_setsStartMapIconId`) is unrelated to raid markers or this PR's changes at all (no file overlap) — it depends on seeded "dungeon start" map icon data that's inconsistent between my local dev DB (none exist) and CI (some exist, but the test picks the wrong one). Didn't touch it since I can't reproduce CI's exact failure locally to verify a fix safely; flagging for separate investigation.

Wotuu:
:robot: Imported the string you provided, reproduced the error, and used it to build out full MDT raid marker support:

**Crash fix**: MDT strings exported without an affix week selected omit the `week` key entirely (yours did). Two call sites read `$decoded['week']` straight into an `int` parameter, throwing a `TypeError`. Both now accept `?int` and fall through to their existing "week not found" handling (current affix group; no rift offsets for that week).

**New feature — MDT raid marker import/export**: decoded your string's structure — MDT stores raid markers in `value.enemyAssignments`, shaped `{mdtNpcIndex: {mdtCloneIndex: raidTargetIndex}}`, addressing the same npc-index/clone-index space `pulls` already use. Added `RaidMarkerImporter` (mirrors `PullImporter`'s resolution pattern, including the three dungeon-specific clone-index hacks for MDT's duplicate-NPC-listing bug) and the matching export side in `MDTExportStringService`. Your string's 5 raid markers now import correctly (skull/star/triangle/star/square, verified against your exact data).

**Bonus bug found while stress-testing**: the Lua↔JSON bridge used for encode/decode can't tell a Lua table used as a map from one used as a sequential array. Since MDT indices are always ≥1, a set of clone/npc indices that happens to form a contiguous run starting at 1 silently round-trips as a 0-indexed JSON array instead of `{index: value}`, corrupting which enemy a marker resolves to. Detected via `array_is_list()` at both levels and corrected. Verified with 40+ randomized round-trip runs (multiple failures before the fix, 0 after) plus new test coverage.

app/Http/Controllers/Ajax/AjaxEnemyController.php
Wotuu: :robot: Fixed in c1d8db0b1 — `setRaidMarker` now computes `$identityIsUnique` (count of enemies in the route's mapping version matching `(npc_id, mdt_id)`) before including it in the delete's `orWhere`. When ambiguous, the delete falls back to matching this enemy's exact `enemy_id` only, so a sibling enemy sharing the identity can no longer have its marker deleted. Verified: relevant tests pass (`--filter=RaidMarker`, 7 passed / 1 expected skip).

app/Repositories/Database/DungeonRoute/DungeonRouteEnemyRaidMarkerRepository.php
Wotuu: :robot: Fixed in c1d8db0b1 — the `UPDATE ... JOIN` in `updateEnemyIdsByMappingVersion` now has a correlated-subquery uniqueness guard (`... = 1`) so it only sets `enemy_id` when the identity resolves to exactly one enemy in the target mapping version. Ambiguous identities are left `NULL` and then removed by the existing `deleteOrphanedByDungeonRouteId()` cleanup, instead of MySQL picking an arbitrary match.

app/SeederHelpers/RelationImport/Parsers/Relation/DungeonRouteEnemyRaidMarkersRelationParser.php
Wotuu: :robot: Fixed in c1d8db0b1 — an unresolvable `enemy_id` now logs a warning and `continue`s instead of throwing, so one stale marker row no longer aborts the whole `DungeonDataSeeder` run. Added `DungeonRouteEnemyRaidMarkersRelationParserTest` for coverage (none existed before).

tests/Feature/App/Service/DungeonRoute/DungeonRouteEnemyRaidMarkerMappingVersionTest.php
Wotuu: :robot: Fixed in c1d8db0b1 — the random enemy pick now excludes any identity with more than one match in the mapping version (mirrors the production resolution logic in `DungeonRouteEnemyRaidMarkerRepository`), so the `firstOrFail()` comparison can no longer diverge from what the `UPDATE ... JOIN` would resolve.

tests/Feature/App/Service/DungeonRoute/DungeonRouteEnemyRaidMarkerMappingVersionTest.php
Wotuu: :robot: Fixed in c1d8db0b1 — `getRetailDungeonWithNullMdtIdEnemy()` now calls `self::markTestSkipped(...)` instead of throwing a `RuntimeException`. Confirmed locally: the affected test now reports as skipped rather than erroring (current dev DB still has 0 null-`mdt_id` enemies, matching your measurement).

app/Service/MDT/Import/RaidMarkerImporter.php
```diff
@@ -0,0 +1,182 @@
...
+        if ($dungeon->key === Dungeon::DUNGEON_TOL_DAGOR && $npcIndex === 11) {
+            return $cloneIndex + 2;
+        }
+
+        if ($dungeon->key === Dungeon::DUNGEON_MISTS_OF_TIRNA_SCITHE && $npcIndex === 23) {
+            return $cloneIndex + 5;
+        }
+
+        return $cloneIndex;
+    }
```
Wotuu: I know of this hack (because I wrote it), is this now in two places? Because I'd rather it not be. Can it be put in a trait?

resources/assets/js/custom/enemyvisuals/enemyvisual.js
```diff
@@ -331,24 +331,38 @@ class EnemyVisual extends Signalable {
...
+                // here (unlike eagerly above): by now any in-flight show/hide transition has settled,
+                // either immediately (fadeOut = false) or after the 500ms delay below. Bootstrap keys
+                // instances by element in a plain Map (not a WeakMap), so skipping this would leak a
+                // Tooltip instance for every marker assigned/cleared this session.
+                $(this).find('[data-bs-toggle="tooltip"]').each(function () {
+                    let tooltip = bootstrap.Tooltip.getInstance(this);
+                    if (tooltip !== null) {
+                        tooltip.dispose();
+                    }
+                });
```
Wotuu, line 361: Can this tooltip cleanup thing be placed in `utils.js`? It sounds useful to call at other places, since I've seen floating tooltips before and rebuilding them all is indeed too expensive.

resources/assets/js/custom/models/enemy.js
```diff
@@ -803,7 +803,9 @@ class Enemy extends VersionableMapObject {
 
                 text = template($.extend({}, visualData, {
                     name: lang.get(this.npc.name),
-                    right_click_to_open_details: lang.get('js.right_click_to_open_details')
+                    right_click_to_open_details: lang.get('js.right_click_to_open_details'),
+                    show_raid_marker_shortcut: this.canOpenRaidMarkerMenu(),
+                    shift_right_click_to_assign_raid_marker: lang.get('js.shift_right_click_to_assign_raid_marker')
```
Wotuu: You don't need to pass `lang.get()` to handlebars. `getHandlebarsDefaultVariables()` should be included always - which automatically include all translations. Also - it's customary to suffix these translation keys with `_label`.

Wotuu (commented):
A few changes.

app/Service/MDT/Import/RaidMarkerImporter.php
Wotuu: :robot: Done — extracted into a shared `App\Service\MDT\Import\Traits\AppliesMdtCloneIndexHack` trait, used by both `RaidMarkerImporter` and `PullImporter` now.

resources/assets/js/custom/enemyvisuals/enemyvisual.js
Wotuu: :robot: Done — moved into a reusable `removeStrayTooltips()` helper in `util.js`.

resources/assets/js/custom/models/enemy.js
Wotuu: :robot: Done — dropped the manual `lang.get()` calls in favor of `getHandlebarsDefaultVariables()`, and renamed both keys in `lang/en_US/js.php` to the `_label` suffix convention (`right_click_to_open_details_label`, `shift_right_click_to_assign_raid_marker_label`), updating the handlebars template to match.

Wotuu (commented):
Ok

PR #3631, opened 2026-07-20
#3629 Generalize mapping:assignmdtids to run for all dungeons

### LINKED ISSUE #3629: Generalize mapping:assignmdtids to run for all dungeons
## Background

While reviewing #1431/#1453 (raid marker mapping-version safety), I found that resolving an
enemy's identity across a mapping-version upgrade (both the new `DungeonRouteEnemyRaidMarkerRepository`
and the pre-existing `KillZoneEnemyRepository` it mirrors) depends on `npc_id` + `mdt_id` matching.
Enemies with `mdt_id IS NULL` can still be matched (the repositories use MySQL's null-safe `<=>`
operator after a follow-up fix), but every enemy *should* have an `mdt_id` per the data model's own
assumption, and having one also lets these enemies participate correctly in MDT import/export
matching (`app/Service/MDT/Import/PullImporter.php`) going forward.

A read-only snapshot of `enemies` found **1555 rows with `mdt_id IS NULL` across 42
dungeon+mapping-version combinations** at the time. Two dungeons were extreme outliers (hundreds of
null rows each, repeated across multiple re-imported mapping versions) - Naxxramas and Karazhan
Crypts (Season of Discovery). Wotuu has already hand-fixed the worst offenders locally using
`php artisan mapping:assignmdtids` with its whitelist manually edited per run.

## The bug

`app/Console/Commands/Mapping/AssignMDTIDs.php` (signature `mapping:assignmdtids`) hardcodes a
`$dungeonWhitelist` array of exactly one or two active entries, with ~8 more commented out:

```php
$dungeonWhitelist = [
    Dungeon::RAID_NAXXRAMAS,
//            Dungeon::RAID_SERPENTSHRINE_CAVERN,
    //            Dungeon::DUNGEON_MOGU_SHAN_PALACE,
    // ... 6 more, all commented out
];
```

So despite its description ("Assigns MDT IDs to a mapping that doesn't have them yet" - no scoping
caveat), the command as shipped only ever processes whichever one or two dungeons happen to be
uncommented, and requires manually editing the source and redeploying/rerunning to cover another
dungeon. This is dev-scratchpad code that never got generalized.

## The ask

1. Replace the hardcoded whitelist with a proper `--dungeon=<key>` option (optional, defaults to
   processing every dungeon) so this becomes a real, reusable tool instead of requiring a source
   edit per run.
2. Add per-dungeon/mapping-version reporting (a before/after count of `mdt_id IS NULL` enemies, or
   at least how many got assigned per dungeon) so a run's effect is visible without a manual query
   afterward.
3. Verify the core assignment approach is still sound: it currently groups enemies by `npc_id`
   within a mapping version, finds the current max `mdt_id`, and assigns incrementing values to
   enemies missing one. This does not need to match MDT's own real clone numbering (confirmed via
   `app/Service/MDT/Import/PullImporter.php` - `mdt_id` is only ever read/matched against existing
   persisted values during import, never auto-derived), it only needs to be internally stable and
   unique per NPC within a mapping version. Worth double-checking this reasoning holds before
   running it broadly.
4. Run the improved command across all dungeons and report the before/after breakdown. It's
   expected that some dungeons/mapping versions won't fully resolve (e.g. if the gap isn't really
   "never assigned" but a deeper issue in that dungeon's own import history) - call those out
   rather than forcing them.

## Key files

- `app/Console/Commands/Mapping/AssignMDTIDs.php`
- `app/Logic/MDT/Data/MDTDungeon.php` (`getClonesAsEnemies()` - how real MDT clone/mdt_id data is
  derived from an actual MDT export, for comparison)
- `app/Service/MDT/Import/PullImporter.php` (how `mdt_id` is read/matched during import)

## Definition of done

- `mapping:assignmdtids` runs against any/all dungeons without a source edit.
- A before/after `mdt_id IS NULL` count per dungeon+mapping-version is reported.
- A feature test covers the assignment logic.

Wotuu:
:robot:

Closes #3629

## Summary

`app/Console/Commands/Mapping/AssignMDTIDs.php` hardcoded a `$dungeonWhitelist` array of one or
two active dungeons with ~8 more commented out, so despite its own description ("Assigns MDT IDs
to a mapping that doesn't have them yet") it only ever processed whichever entries happened to be
uncommented - requiring a source edit + rerun to cover a different dungeon.

- Replaced the whitelist with an optional `--dungeon=<key>` option (omit to process every
  dungeon), matching the existing `{--dungeon=}` convention used by `mdt:importnpcs`.
- Added a per-dungeon/mapping-version table reporting missing-before/missing-after `mdt_id` counts,
  so a run's effect is visible without a manual query afterward.
- Added feature test coverage (`AssignMDTIDsTest`): incrementing assignment continues from the
  existing max per `npc_id`, and `--dungeon` correctly scopes the run.

The assignment approach itself (group by `npc_id` within a mapping version, assign incrementing
values to enemies missing one) is unchanged - confirmed via `app/Service/MDT/Import/PullImporter.php`
that `mdt_id` is only ever read/matched against already-persisted values during an MDT string
import, never auto-derived, so this heuristic doesn't need to reproduce MDT's own clone numbering -
it only needs to be internally stable and unique per NPC within a mapping version (the actual need,
surfaced while fixing #1431/#1453's mapping-version-upgrade identity resolution).

## Before/after

Ran the improved command against every dungeon in the dev DB:

- **Before**: 259 enemies across 36 dungeon+mapping-version combinations had `mdt_id IS NULL`
  (two much larger outliers, Naxxramas and Karazhan Crypts (SoD) at ~250-350 each, had already been
  hand-fixed locally before this PR using the old whitelist mechanism).
- **After**: 0 remaining anywhere - every dungeon/mapping-version resolved fully in one run.

## Test plan

- [x] `composer run fix` clean (no unrelated reformats)
- [x] `AssignMDTIDsTest` passes (2 tests, 8 assertions)
- [x] Ran `php artisan mapping:assignmdtids` against all dungeons in the dev DB; verified
      `Enemy::whereNull('mdt_id')->count()` dropped to 0

### DIFF
--- a/app/Console/Commands/Mapping/AssignMDTIDs.php
+++ b/app/Console/Commands/Mapping/AssignMDTIDs.php
@@ -15,75 +15,91 @@ class AssignMDTIDs extends Command
      *
      * @var string
      */
-    protected $signature = 'mapping:assignmdtids';
+    protected $signature = 'mapping:assignmdtids {--dungeon= : Only process this dungeon (by key); omit to process every dungeon}';
 
     /**
      * The console command description.
      *
      * @var string
      */
-    protected $description = "Assigns MDT IDs to a mapping that doesn't have them yet.";
+    protected $description = "Assigns MDT IDs to enemies that don't have one yet, so they stay identifiable across a mapping version upgrade.";
 
     /**
      * Execute the console command.
      */
     public function handle(): int
     {
+        $dungeonKey = $this->option('dungeon');
+
+        $mappingVersionsQuery = MappingVersion::with(['enemies', 'dungeon']);
+        if ($dungeonKey !== null) {
+            $dungeon = Dungeon::where('key', $dungeonKey)->firstOrFail();
+            $mappingVersionsQuery->where('dungeon_id', $dungeon->id);
+        }
+
         /** @var Collection<int, MappingVersion> $mappingVersions */
-        $mappingVersions = MappingVersion::with([
-            'enemies',
-            'dungeon',
-        ])->get();
+        $mappingVersions = $mappingVersionsQuery->get();
 
-        $dungeonWhitelist = [
-            Dungeon::RAID_THE_EYE,
-            Dungeon::RAID_SERPENTSHRINE_CAVERN,
-            //            Dungeon::DUNGEON_MOGU_SHAN_PALACE,
-            //            Dungeon::DUNGEON_SCARLET_HALLS_MOP,
-            //            Dungeon::DUNGEON_SCARLET_MONASTERY_MOP,
-            //            Dungeon::DUNGEON_SCHOLOMANCE_MOP,
-            //            Dungeon::DUNGEON_SHADO_PAN_MONASTERY,
-            //            Dungeon::DUNGEON_SIEGE_OF_NIUZAO_TEMPLE,
-            //            Dungeon::DUNGEON_STORMSTOUT_BREWERY,
-            //            Dungeon::DUNGEON_TEMPLE_OF_THE_JADE_SERPENT,
-        ];
+        $rows       = [];
+        $totalCount = 0;
 
-        $count = 0;
         foreach ($mappingVersions as $mappingVersion) {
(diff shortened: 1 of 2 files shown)


Wotuu (commented):
Ok

PR #3635, opened 2026-07-21
#3632 Register combat log version 12.0.7

### LINKED ISSUE #3632: Register combat log version 12.0.7 (follow-up to #3624)
## Background

While triaging Staging's combat log parse failures via the new API from #3624/#3628, 493 of the
500 currently-unresolved `CombatLogParseFailure` rows (98.6%) share the same root cause:

```
Exception: Unable to find combat log version 22012000007!
```

thrown at `app/Logic/CombatLog/SpecialEvents/CombatLogVersion.php:70`. WoW client build `12.0.7`
computes `getVersionLong() == 22_012_000_007`, which isn't a key in
`App\Logic\CombatLog\CombatLogVersion::ALL` / `::RETAIL_ALL` — the table currently stops at
`RETAIL_12_0_5 = 22_012_000_005`. Every log line from a 12.0.7 client fails on the very first
`COMBAT_LOG_VERSION` special event.

## Risk investigated — and cleared

Registering a new combat log version is normally risky: parsing behaviour can be gated per
version, and unlisted versions could silently hit the wrong code path. That's not the case here:

- Every version-specific builder (`DamageBuilder`, `AuraAppliedBuilder`, `CombatantInfoBuilder`,
  `AdvancedDataBuilder`, `EncounterStartBuilder`, etc. — checked all of them) has a `default =>`
  arm in its `match($combatLogVersion)` that already falls through to the newest handler. An
  unregistered version doesn't hit a different code path than a registered-but-unhandled one — it
  just never gets *past* the single allowlist check in `CombatLogVersion::setParameters()` (and
  the duplicate check in `App\Service\CombatLog\ResultEvents\CombatLogVersion`).
- `Prefix::createFromEventName` / `Suffix::createFromEventName` dispatch on event *name*, not
  combat log version. `GenericDataBuilder::create()` ignores the version argument entirely.
- So `ALL`/`RETAIL_ALL` is purely a "known version" allowlist, not a schema-revision switch.

## Verified against real Staging data

Downloaded two full real runs from Staging via the new `/segments` API endpoint (run
`40653052`, 9 segments, ~203K lines; run `40712395`, 7 segments) and locally re-parsed every
segment end-to-end through `CombatLogServiceInterface::parseCombatLogToEvents()` with
`RETAIL_12_0_7` registered. **All 16 segments parsed with zero exceptions.** No new downstream
errors surfaced.

## Remaining work

In `app/Logic/CombatLog/CombatLogVersion.php`:

1. Add `public const int RETAIL_12_0_7 = 22_012_000_007;`
2. Add `self::RETAIL_12_0_7 => 15,` to both `RETAIL_ALL` and `ALL` (next revision number after 14).
3. Add a `retail-12-0-7` case to
   `getVersionLong_givenCombatLogVersionEvent_returnsCorrectVersionLong_DataProvider` in
   `tests/Unit/App/Logic/CombatLog/SpecialEvents/CombatLogVersion/CombatLogVersionTest.php`.

## Definition of done

- New/updated unit test green.
- MR opened to master, `Closes #3624`... actually this is a separate issue from #3624 (which was
  scoped to the read-only API endpoint only) — reference this issue number instead.

## Related

- A separate, unrelated cluster of 6/500 failures (`RioLogVersion` parameter-count errors) was
  found during the same triage — tracked separately, not blocking this fix.

Wotuu:
:robot: Closes #3632

## Summary
- Registers WoW client build `12.0.7` (`22_012_000_007`, revision `15`) in `App\Logic\CombatLog\CombatLogVersion::ALL`/`RETAIL_ALL`.
- This was the root cause of 493/500 (98.6%) of the currently-unresolved `CombatLogParseFailure` rows on Staging: `Exception: Unable to find combat log version 22012000007!`, thrown on the very first line of every affected log.
- Verified this is safe to register: every version-gated builder (`DamageBuilder`, `AuraAppliedBuilder`, `CombatantInfoBuilder`, `AdvancedDataBuilder`, etc.) already has a `default =>` arm that falls through to its newest handler for any unlisted version — the `ALL`/`RETAIL_ALL` table is purely a "known version" allowlist, not a schema-revision switch. Confirmed by re-parsing two full real Staging runs (16 segments, ~330K events total) end-to-end through `CombatLogServiceInterface::parseCombatLogToEvents()` with this change applied: zero exceptions, no new downstream failures.

## Test plan
- [x] Added `retail-12-0-7` case to `CombatLogVersionTest`'s data provider — green.
- [x] Full `CombatLog` test group (207 tests) green.
- [x] Manually re-parsed two real failing Staging runs (run 40653052, 9 segments; run 40712395, 7 segments) end-to-end with the fix applied — 0 failures.

### DIFF
--- a/.claude/skills/combatlog-parse-failure-triage/SKILL.md
+++ b/.claude/skills/combatlog-parse-failure-triage/SKILL.md
@@ -1,0 +1,198 @@
+---
+name: combatlog-parse-failure-triage
+description: Operational runbook to find, download, and reproduce real Staging/Production combat log parse failures before fixing them — the internal-team API endpoints, how to pull a failure's real Raider.IO log segments, the gzip/docker-cp reproduction recipe, and the safety checklist for shipping a fix. Use when triaging CombatLogParseFailure rows, chasing a "combat log parses are failing" report, or verifying a combatlog-parsing-internals fix against real production data. Pairs with combatlog-parsing-internals (how the parser itself works) and create-github-issue / worktree-docker (how to turn a diagnosis into a shipped MR).
+---
+
+# Combat Log Parse-Failure Triage
+
+Runbook for turning "some combat logs are failing to parse in Staging/Production" into a diagnosed,
+locally-reproduced, tested fix. Read `combatlog-parsing-internals` first if you haven't — this skill
+assumes you know the parser architecture and focuses on the *operational* loop: find → download →
+reproduce → fix → verify → ship.
+
+## 0. Credentials — get them fresh, never persist them
+
+The parse-failure API lives behind `api_role:admin` (HTTP Basic auth, Laratrust). **This repo is
+public** — never write a real username/password into a skill file, a commit, a script, or anywhere
+else that gets checked in. Each session, ask Wotuu for a current Staging (or Production, if that's
+what's being triaged) admin API service-account credential, use it only for that session, and remind
+him to rotate/delete it once you're done. Treat it like any other secret: environment/conversation
+only, never `.env`, never a file under version control (see `feedback_no_env_file` in memory).
+
+## 1. List unresolved failures
+
+```
+GET https://staging.keystone.guru/api/v1/combatlog/parse-failures
+```
+
+(swap the host for `keystone.guru` / whatever the production host is, for a prod triage). HTTP Basic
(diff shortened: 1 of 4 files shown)


Wotuu:

Wotuu (commented):
Ok

PR #3636, opened 2026-07-21
#3633 Treat RioLogVersion CLIENT_SESSION_ID as optional

### LINKED ISSUE #3633: RioLogVersion parameter-count parse failures — likely truncated log lines (follow-up to #3624)
## Background

While triaging Staging's combat log parse failures via the API from #3624/#3628, found a small
secondary cluster: 6 of 500 currently-unresolved `CombatLogParseFailure` rows are

```
InvalidArgumentException: Invalid parameter count for App\Logic\CombatLog\SpecialEvents\RioLogVersion - wanted 31-35, got 23
InvalidArgumentException: Invalid parameter count for App\Logic\CombatLog\SpecialEvents\RioLogVersion - wanted 31-35, got 29
```

(4 at `got 23`, 2 at `got 29`), all `season_id=17`, all `line_number=1` (the very first line of the
segment). Unrelated to the 12.0.7 registration bug fixed in #3632/MR #3635 (different exception
class, different code path) — one of the 12.0.7-affected runs (id 274/299) even shows this failure
*instead of* the version-registration one, since RioLogVersion's own param-count check throws before
the standalone `COMBAT_LOG_VERSION` allowlist check further down the file is ever reached.

## Refined finding (supersedes the original truncation-only guess)

`RioLogVersion` (`app/Logic/CombatLog/SpecialEvents/RioLogVersion.php`) requires 31-35 params for a
trash-variant `RIO_LOG_VERSION` line (`getParameterCount()=35`, `getOptionalParameterCount()=4` — the
4 optional params are `ENCOUNTER_ID`/`ENCOUNTER_STARTED_AT`, boss-only per the class's own docblock).

A **healthy** trash-variant line, pulled from a real successfully-parsed run (40653052, same
`season_id=17`, `APP_VERSION=4.11.2`):

```
RIO_LOG_VERSION,1,SEGMENT_TYPE,mplus_trash,APP_VERSION,4.11.2,PROCESSOR_VERSION,1,PLATFORM,win32,INSTANCE_ID,2811,DUNGEON_ID,558,SEGMENT_ID,1,CORRELATION_ID,...,CHALLENGE_MODE_STARTED_AT,1784511637186,TYPE,trash,CLIENT_SESSION_ID,7b173eb5-c12b-4c42-aae0-c13cfdd64f7b,COMBAT_LOG_VERSION,22,ADVANCED_LOG_ENABLED,1,BUILD_VERSION,12.0.7,PROJECT_ID,1
```

This is **exactly 31 params** — sitting right at the documented trash-variant floor — and it includes
*both* `CLIENT_SESSION_ID` and the trailing `COMBAT_LOG_VERSION,ADVANCED_LOG_ENABLED,BUILD_VERSION,
PROJECT_ID` quad. So per the class's own spec, both of those are part of the **mandatory** minimum for
a trash segment, not alternate/optional fields that vary by addon version. That rules out my original
"maybe it's just an older addon format" alternative and leaves genuine data loss (truncation or a
dropped-field addon bug) as the explanation — reinforced by the two clusters:

- **`got 23`** (4 failures, all `APP_VERSION=4.11.2`, same as the healthy sample above): line cuts off
  immediately after `CLIENT_SESSION_ID,<uuid>` — missing the entire trailing
  `COMBAT_LOG_VERSION,ADVANCED_LOG_ENABLED,BUILD_VERSION,PROJECT_ID` quad (8 params short of 31).
- **`got 29`** (2 failures, both `APP_VERSION=4.10.1`, an older addon build): line has the
  `COMBAT_LOG_VERSION` quad but is missing `CLIENT_SESSION_ID,<uuid>` entirely (2 params short of 31).

Both clusters are **internally consistent** (every occurrence in a cluster drops exactly the same
field(s), not a random cut point) — that argues against pure mid-write truncation and toward a
specific, reproducible condition in the Raider.IO addon's line-construction code where it sometimes
skips appending one particular trailing key/value block. Not conclusively provable from this side
(the addon isn't in this repo) without more samples or addon-side logs.

## Why this matters beyond just these 6 rows

`RioLogVersion` implements `HasCombatLogVersionInterface` — a successfully-parsed one becomes the
*active* `combatLogVersion` for every subsequent line in the segment
(`CombatLogService::parseCombatLog()`, `parsedEvent instanceof HasCombatLogVersionInterface`). Losing
the whole run to a hard failure on line 1 is a heavier cost than the missing field itself — especially
since a standalone `COMBAT_LOG_VERSION` special event later in the same segment can supply the version
info independently in many cases anyway (as seen in the healthy 12.0.7 sample from #3632's
investigation, which had a *second*, separate `COMBAT_LOG_VERSION` line at line 8).

## Possible fix directions (not yet designed/validated)

1. **Defensive tolerance**: treat `CLIENT_SESSION_ID` and the `COMBAT_LOG_VERSION` quad as
   independently-optional (like `ENCOUNTER_ID`/`ENCOUNTER_STARTED_AT` already are), lower the
   hard-required floor accordingly, and make `embeddedCombatLogVersion`/`buildVersion` nullable —
   `getVersionLong()` (`ComputesVersionLong` trait) would need a null-safe path so
   `CombatLogService::parseCombatLog()` simply doesn't update the active version off an incomplete
   `RioLogVersion` line (leaves whatever version was already active, or falls through to a later
   explicit `COMBAT_LOG_VERSION` line).
2. **Upstream**: if this turns out to be reproducible/confirmable as an addon bug (not truncation),
   flag it to whoever owns the Raider.IO addon rather than working around it here.

**Not implementing either yet** — this changes version-tracking semantics (silent partial-data
tolerance vs. loud failure), which is a different risk profile than a pure registry addition, and
deserves a second look before writing code. Small blast radius (6/500, 1.2%) means there's no urgency
forcing a rushed fix.

## Not yet done

- Get more samples (both clusters only have 2-4 examples each) to confirm the "always drops the exact
  same field(s)" pattern holds up, rather than being a coincidence of a small sample.
- Confirm whether a later explicit `COMBAT_LOG_VERSION` line reliably exists elsewhere in these
  specific failing segments (would derisk option 1 above — if it's always there, RioLogVersion's own
  embedded copy is genuinely redundant on failure).
- Decide on a fix direction and implement + verify against real failing segments end-to-end, same
  rigor as #3632 (re-parse real data with the fix applied before shipping).

## Key files

- `app/Logic/CombatLog/SpecialEvents/RioLogVersion.php`
- `app/Logic/CombatLog/SpecialEvents/Traits/ComputesVersionLong.php` (`getVersionLong()`)
- `app/Logic/CombatLog/CombatEvents/Traits/ValidatesParameterCount.php` (the wanted `min-max` check)
- `app/Service/CombatLog/CombatLogService.php::parseCombatLog()` (where `HasCombatLogVersionInterface`
  events update the active version)
- See the new `combatlog-parsing-internals` / `combatlog-parse-failure-triage` skills (added in #3632)
  for the general parser architecture and the download/reproduce recipe used to gather the above.

## Related

- Unrelated to #3632 (the dominant 12.0.7-registration failure cluster, MR #3635).

Wotuu:
:robot: Addresses part of #3633 — does NOT close it, see below.

## Summary
- `RioLogVersion`'s `RIO_LOG_VERSION` line parser required `CLIENT_SESSION_ID` unconditionally. Older Raider.IO addon builds (observed on `APP_VERSION 4.10.1`) never emit that field, which failed the parameter-count floor (`wanted 31-35, got 29`) and dropped the whole run — 2/500 of the currently-unresolved `CombatLogParseFailure` rows on Staging.
- Made `CLIENT_SESSION_ID` optional (same pattern already used for the boss-only `ENCOUNTER_ID`/`ENCOUNTER_STARTED_AT` fields). Confirmed nothing downstream reads `getClientSessionID()`, so tolerating its absence is low-risk.

## What this does NOT fix
#3633 also covers a second, riskier cluster ("got 23" — 4/500): a same-addon-version line missing the *trailing* `COMBAT_LOG_VERSION,ADVANCED_LOG_ENABLED,BUILD_VERSION,PROJECT_ID` block entirely. That block is the run's only source of the active combat log version, and verified there's no fallback `COMBAT_LOG_VERSION` line elsewhere in the affected segments — silently tolerating it would route the rest of the run through the wrong version-specific parser (e.g. `CombatantInfoV21` instead of `V22`) instead of failing loudly. Left as an open loud failure pending a safer fix; tracked in #3633.

## Test plan
- [x] Added a `mplus-trash-win32-missing-client-session-id` case to `RioLogVersionTest`'s data provider (real Staging raw line, `APP_VERSION 4.10.1`) — green.
- [x] Full `CombatLog` test group (207 tests) green.
- [x] Re-parsed 2 real Staging runs that previously failed with this exact error end-to-end (18 segments, ~561K events total) — zero failures with the fix applied.


tests/Unit/App/Logic/CombatLog/SpecialEvents/RioLogVersion/RioLogVersionTest.php
```diff
@@ -129,6 +129,28 @@ public static function parseEvent_givenRioLogVersionEvent_returnsCorrectValues_D
                 '12.0.5',
                 1,
             ],
+            'mplus-trash-win32-missing-client-session-id' => [
```
Wotuu, line 132: :robot: **Test coverage (minor):** The optional-parameter loosening (`getOptionalParameterCount()` → 6) applies to the boss variant too, so a boss segment from the same old addon build (`APP_VERSION 4.10.1`) missing `CLIENT_SESSION_ID` produces a **33-param** line (35 − ENCOUNTER fields are present, − CLIENT_SESSION_ID). That shape is now newly accepted but isn't exercised — only the trash/29-param path is. Since it's a real production shape from the exact build this PR targets, consider adding a `mplus-boss-...-missing-client-session-id` data-provider case (populated `ENCOUNTER_ID`/`ENCOUNTER_STARTED_AT`, `null` client session) to lock in that the boss branch parses correctly at the new floor. Not blocking — parsing is key-based and shared with the trash path.

Wotuu:

tests/Unit/App/Logic/CombatLog/SpecialEvents/RioLogVersion/RioLogVersionTest.php
Wotuu: :robot: Done — added a `mplus-boss-win32-missing-client-session-id` data-provider case mirroring the trash one but for the boss branch (populated ENCOUNTER_ID/ENCOUNTER_STARTED_AT, null client session), locking in that it parses correctly at the new floor.

Wotuu (commented):
Ok

PR #3641, opened 2026-07-21
#3640 Rename discover new() to newest() to avoid route-cache ParseError

### LINKED ISSUE #3640: Fix ParseError crash on /api/v1/routes/{gameVersion}/new (reserved-word controller method)
## Background

Production is throwing a 500 on `GET /api/v1/routes/{gameVersion}/new`:

```
[production.ERROR] HandlerLogging::uncaughtException
exceptionClass: ParseError
message: syntax error, unexpected token "new", expecting "("
url: http://keystone.guru/api/v1/routes/retail/new?count=1&offset=0
On laravel-serializable-closure://function new(
```

Root cause: `routes/api.php` registers this route with first-class-callable syntax:

```php
Route::get('new', new APIDungeonRouteDiscoverController()->new(...))->name('api.v1.discover.new');
```

`php artisan route:cache` (used in production) serializes that closure via
`laravel/serializable-closure`. When the cached route is later invoked, Laravel reconstitutes the
closure by `eval`-ing generated code — and since the controller method is literally named `new`,
the generated code is `function new(...) { ... }`, which is invalid PHP outside of class context
(`new` is a reserved word). The crash only shows up per-request in production once the route
cache is warm; it doesn't reproduce in tests (routes aren't cached under `runningUnitTests()`) or
at `route:cache` build time.

This is the same class of bug that previously bit a controller method named `list`.

## Fix

- Rename `APIDungeonRouteDiscoverController::new()` → `newest()`. The route path (`new`) and
  route name (`api.v1.discover.new`) stay unchanged — those are just strings, not callable
  identifiers, so the public API contract (URL) is unaffected.
- Update the route registration in `routes/api.php` to call `->newest(...)`.
- Rename the matching test methods in
  `tests/Feature/Controller/Api/V1/APIDungeonRouteDiscoverController/APIDungeonRouteDiscoverControllerTest.php`.
- Document the quirk in `.claude/CLAUDE.md` (routes registered via first-class-callable syntax
  must never be named after a PHP reserved word) so this doesn't recur a third time.

## Verification

- `php artisan test --compact tests/Feature/Controller/Api/V1/APIDungeonRouteDiscoverController/APIDungeonRouteDiscoverControllerTest.php`
  passes.
- Reproduced the actual mechanism directly via tinker: serialize/unserialize the closure with
  `Laravel\SerializableClosure\SerializableClosure` — fails with the old `new` method name,
  succeeds after the rename.
- `php artisan route:cache` succeeds and `route:list` shows the route unchanged
  (`api/v1/routes/{gameVersion}/new`).
- `composer run fix` / `composer run analyse` clean on the touched files.

## Definition of done

- [ ] Method renamed, route + tests updated, CLAUDE.md note added.
- [ ] Tests green.
- [ ] `composer run analyse` clean.
- [ ] Verified route caching no longer crashes on this endpoint.

Wotuu:
:robot: Closes #3640

## Summary
- `APIDungeonRouteDiscoverController::new()` → `newest()`. Routes registered with first-class
  callable syntax (`new FooController()->new(...)`) get serialized via
  `laravel-serializable-closure` when `route:cache` runs, and reconstituted by `eval`-ing
  generated code. A method literally named `new` produces invalid PHP
  (`function new(...)`), causing a per-request `ParseError` in production once the route cache
  is warm — this is exactly what production was hitting on `GET /api/v1/routes/{gameVersion}/new`.
- The route path (`new`) and route name (`api.v1.discover.new`) are unchanged — those are plain
  strings, not callable identifiers, so the public API URL is unaffected.
- Updated `routes/api.php` and the matching test method names.
- Documented the quirk in `.claude/CLAUDE.md` (this is the second time it's bitten a controller
  method — previously `list`) so it isn't reintroduced.
- Incidental fix: CI's `feature-controller` job was red on this branch from an unrelated flaky
  test, `DungeonRouteControllerSaveNewTemporaryTest::saveNewTemporary_givenValidStartMapIcon_setsStartMapIconId`.
  `DungeonRouteControllerCreateTestBase::getDungeonWithStartIcon()` assumed the newest
  (highest-id) dungeon-start `MapIcon` overall is on its own dungeon's current mapping version.
  That assumption broke once "bare" mapping versions (created only to hold floor union data,
  without a cloned map icon set) can become a dungeon's current version. Fixed the helper to walk
  candidates newest-first until one actually sits on its dungeon's current mapping version.

## Test plan
- [x] `php artisan test --compact tests/Feature/Controller/Api/V1/APIDungeonRouteDiscoverController/APIDungeonRouteDiscoverControllerTest.php` — 10 passed
- [x] Reproduced the actual failure mechanism via tinker: serializing/unserializing the route
      closure with `Laravel\SerializableClosure\SerializableClosure` fails with the old `new`
      method name and succeeds after the rename
- [x] `php artisan route:cache` succeeds; `route:list` shows `api/v1/routes/{gameVersion}/new`
      unchanged
- [x] `composer run fix` / `composer run analyse` clean on touched files
- [x] No UI change — no screenshots needed
- [ ] Watching CI to confirm `feature-controller` is green after the test-helper fix

### DIFF
--- a/routes/api.php
+++ b/routes/api.php
@@ -52,7 +52,7 @@ Route::prefix('v1')->group(static function () {
 
     Route::prefix('routes/{gameVersion}')->group(static function () {
         Route::get('popular', new APIDungeonRouteDiscoverController()->popular(...))->name('api.v1.discover.popular');
-        Route::get('new', new APIDungeonRouteDiscoverController()->new(...))->name('api.v1.discover.new');
+        Route::get('new', new APIDungeonRouteDiscoverController()->newest(...))->name('api.v1.discover.new');
         Route::prefix('{dungeon}')->group(static function () {
             Route::get('popular', new APIDungeonRouteDiscoverController()->dungeonPopular(...))->name('api.v1.discover.dungeon.popular');
             Route::get('new', new APIDungeonRouteDiscoverController()->dungeonNew(...))->name('api.v1.discover.dungeon.new');
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -133,6 +133,9 @@ review must start from a pre-reviewed, verified MR:
     }
 ```
 
+### Routes & controller method naming
+- Routes are registered with first-class callable syntax, e.g. `Route::get('new', new FooController()->new(...))`. Never name a controller method that's registered this way after a PHP reserved word (`new`, `list`, `print`, `echo`, `class`, `function`, ...). It parses and passes tests fine, but `php artisan route:cache` serializes the closure via laravel-serializable-closure, and reconstituting it later `eval`s code like `function new(...)` — invalid syntax, since a function can't be named after a reserved word outside class context. The crash only appears in production, per-request, once the route cache is warm and that request hits the route (`ParseError: syntax error, unexpected token "new"`), not at cache-build time or in tests (routes aren't cached under `runningUnitTests()`). Fix: rename the controller method (e.g. `new` → `newest`); the route path/name (`'new'`, `api.v1.discover.new`) can stay unchanged since that's just a string, not a callable identifier. This has bitten `new` (APIDungeonRouteDiscoverController) and `list` before.
+
 ### Queues
 - Use queued jobs for time-consuming operations with the `ShouldQueue` interface.
 
--- a/app/Http/Controllers/Api/V1/Public/Route/APIDungeonRouteDiscoverController.php
+++ b/app/Http/Controllers/Api/V1/Public/Route/APIDungeonRouteDiscoverController.php
@@ -59,7 +59,7 @@ class APIDungeonRouteDiscoverController extends Controller
      *     )
      * )
      */
-    public function new(
+    public function newest(
         APIOffsetPaginatedRequest $request,
         GameVersion               $gameVersion,
         DiscoverServiceInterface  $discoverService,
--- a/tests/Feature/Controller/Api/V1/APIDungeonRouteDiscoverController/APIDungeonRouteDiscoverControllerTest.php
+++ b/tests/Feature/Controller/Api/V1/APIDungeonRouteDiscoverController/APIDungeonRouteDiscoverControllerTest.php
@@ -142,7 +142,7 @@ final class APIDungeonRouteDiscoverControllerTest extends PublicTestCase
      * @throws Exception
      */
     #[Test]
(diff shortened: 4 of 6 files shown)


routes/api.php
```diff
@@ -52,7 +52,7 @@
 
     Route::prefix('routes/{gameVersion}')->group(static function () {
         Route::get('popular', new APIDungeonRouteDiscoverController()->popular(...))->name('api.v1.discover.popular');
-        Route::get('new', new APIDungeonRouteDiscoverController()->new(...))->name('api.v1.discover.new');
+        Route::get('new', new APIDungeonRouteDiscoverController()->newest(...))->name('api.v1.discover.new');
```
Wotuu, line 55: :robot: **Test coverage — no automated regression guard against this class of bug (third occurrence)**

This is the third time a route registered via first-class-callable syntax has been named after a PHP reserved word (`new` here, `list` previously, per the CLAUDE.md note added in this PR). The fix and the CLAUDE.md documentation both rely on a human remembering the rule; nothing fails the build if it's reintroduced. As the PR body itself notes, the normal test suite can't catch it because routes aren't cached under `runningUnitTests()`, so the existing `APIDungeonRouteDiscoverControllerTest` would still pass with the broken `new()` name.

Consider adding a cheap feature test that reproduces the exact production failure mechanism you already validated via tinker — iterate every registered route, and for each closure action round-trip it through `Laravel\SerializableClosure\SerializableClosure` (serialize **and** unserialize, since the `ParseError` surfaces at the `eval` on unserialize, not at serialize time). That turns "warm route cache + first request in prod" into a deterministic failing test the moment any reserved-word method name is registered this way, and would prevent a fourth occurrence.

Not blocking — the rename itself is correct and complete (I confirmed no other reserved-word first-class-callable route registrations remain on this branch).

Wotuu:

routes/api.php
Wotuu: :robot: Done — added `tests/Feature/Routes/RouteActionSerializationTest.php`, which iterates every registered closure-based route action and round-trips it through `serialize(new SerializableClosure($uses))` + `unserialize(...)->getClosure()`, the exact cycle `route:cache` performs in production. Verified it actually catches the regression: temporarily reverted `newest()` back to `new()` locally, confirmed the test failed with the exact production `ParseError` naming the specific broken route, then reverted back and confirmed it passes again.

Wotuu (commented):
Ok

Issue #3642, opened 2026-07-21
DatabaseSeeder's prepare/apply/cleanup loops silently skip everything after the first failure (follow-up to #3639)


Wotuu:
## Symptom

`php artisan db:seed --class=DungeonDataSeeder` fails deterministically in my local worktree with:

```
SQLSTATE[42S02]: Base table or view not found: 1146 Table 'keystone.guru.prod.floors_temp' doesn't exist
(insert into `floors_temp` ... values (1, 1, ..., dungeons.legion.arcway.floors.arcway, ...))
```

`Floor` is registered in `DungeonDataSeeder::getAffectedModelClasses()`, and its temp table should
have been created before any inserts run. It wasn't - and the seeder never reported an error about
it, it just barreled ahead into inserts against a table that doesn't exist.

## Root cause

Found while investigating #3639. `database/seeders/DatabaseSeeder.php::run()` has this pattern in
**three places** - prepare (~line 111-113), apply (~line 126-128), and cleanup (~line 141-143):

```php
$prepareFailed = false;
foreach ($affectedModelClasses as $affectedModel) {
    $prepareFailed = !$prepareFailed && !$this->prepareTempTableForModel($affectedModel);
}
```

This is broken short-circuit logic. Once `prepareTempTableForModel()` returns `false` for *any*
model in the list:

1. `$prepareFailed` becomes `true`.
2. On the *next* iteration, `!$prepareFailed` is `false`, so `false && ...` short-circuits -
   `prepareTempTableForModel()` is **never called** for any remaining model in the list.
3. Because of that short-circuit, the whole expression evaluates to `false`, so `$prepareFailed`
   gets **reset back to `false`** on that very same iteration.
4. By the end of the loop `$prepareFailed` is `false` again, so `if ($prepareFailed) { ...; break; }`
   never fires - the seeder proceeds as if every temp table was prepared successfully.

Net effect: one failed `prepareTempTableForModel()` call silently skips creating temp tables for
every model after it in `getAffectedModelClasses()`, and the seeder has no idea anything went
wrong until something tries to `INSERT` into a table that was never created - far away from the
actual failure, with a confusing error message. The identical bug pattern exists in the apply and
cleanup loops too (same masking, same silent-skip-of-remaining-items behavior).

I did not get far enough to identify *why* the first `prepareTempTableForModel()` call in the
chain actually failed in my environment (whatever model runs immediately before `Floor` in
[...]

PR #3643, opened 2026-07-21
#3639 Restore missing dungeon-start map icons on newly-added mapping versions

### LINKED ISSUE #3639: CI broken on master: DungeonRouteControllerSaveNewTemporaryTest start-map-icon mismatch
## Symptom

`php-tests (feature-controller)` is currently red on **every** PR based on recent master,
regardless of what the PR itself changes:

```
1) Tests\Feature\Controller\DungeonRouteController\DungeonRouteControllerSaveNewTemporaryTest::saveNewTemporary_givenValidStartMapIcon_setsStartMapIconId
Failed asserting that 642 is identical to 806.
```

The test posts a specific `dungeon_start_map_icon_id` (the highest-ID `MAP_ICON_TYPE_DUNGEON_START`
icon for an active dungeon's current mapping version, per `DungeonRouteControllerCreateTestBase::
getDungeonWithStartIcon()`) and asserts the created `DungeonRoute`'s `dungeon_start_map_icon_id`
matches what was posted. It doesn't — the saved value (642) differs from the posted one (806),
meaning `DungeonRouteController`'s save-new-temporary path is resolving to a *different* map icon
than the one actually submitted.

## Evidence this is a real, deterministic master-level regression — not this PR, not flaky

- Confirmed on **two unrelated PRs** with completely different diffs, both showing the exact same
  numbers (642 vs 806):
  - #3638 (a docs-only skill-file change, zero application code touched)
  - #1431 (`1431-restore-raid-marker-assignment`, an unrelated feature branch)
- Re-ran the #3638 job on the identical commit twice — failed identically both times. Not flaky.
- Since #3638's diff can't possibly cause a `DungeonRoute`/`MapIcon` behavior change, this must
  already be broken on master itself at whatever commit these branches are based on.

## Not yet done — needs investigation

- Bisect which recent master commit introduced this. Candidates from the last few hours of history
  (unconfirmed, not verified against this specific test):
  - `#3629` dungeon-data / `mapping:assignmdtids` backfill work (multiple commits: "Added floor
    unions and missing enemy MDT_IDs", "Backfill missing mdt_id values via mapping:assignmdtids",
    "Generalize mapping:assignmdtids to run for all dungeons")
  - `#3623`/`#3625` "Keep local dev read-only against real S3 thumbnails restored from a production
    DB backup"
  - Any other seeder/`map_icons`/`mapping_versions` JSON changes in that window
- Once the commit is found, determine whether the bug is in the seed data itself (a dungeon now has
  a start icon on a *different* mapping version than its current one, or a duplicate/ambiguous
  start icon) or in `DungeonRouteController`'s resolution logic (e.g. it silently falls back to a
  different icon instead of trusting the posted `dungeon_start_map_icon_id`, and that fallback path
  is now hitting for the dungeon this test happens to select).
- This is actively blocking CI on **every** open PR — worth prioritizing over normal queue.

## Key files

- `tests/Feature/Controller/DungeonRouteController/DungeonRouteControllerSaveNewTemporaryTest.php`
  (the failing test)
- `tests/Feature/Controller/DungeonRouteController/DungeonRouteControllerCreateTestBase.php`
  (`getDungeonWithStartIcon()` — how the test selects which map icon to post)
- `App\Http\Controllers\DungeonRouteController` — the save-new-temporary save path being tested

## Definition of done

- `php-tests (feature-controller)` green again on a fresh PR against master.
- Root cause documented (seed-data issue vs. controller logic issue) so it doesn't silently
  regress again.

Wotuu:
Closes #3639

## Summary

- `php-tests (feature-controller)` is red on every open PR: `DungeonRouteControllerSaveNewTemporaryTest::saveNewTemporary_givenValidStartMapIcon_setsStartMapIconId` fails with `Failed asserting that 642 is identical to 806` — but the actual failing assertion is inside the test helper `getDungeonWithStartIcon()`, which requires the highest-id dungeon-start map icon (across all active dungeons) to belong to its own dungeon's *current* mapping version.
- Root cause: commit `4fa54bf56` (#3629, floor-union/facade backfill) created several new "bare" mapping versions directly rather than through the normal clone-on-create flow, so none of them inherited their dungeon's `MAP_ICON_TYPE_DUNGEON_START` icon from the previous version. On a fresh seed, The Blinding Vale's old-version icon (id 642) is the only start icon that exists, but its now-current mapping version is 806 — exactly the numbers in the failure.
- Fixed by copying the existing start-icon position forward into each new mapping version's `map_icons.json`, for the 5 confirmed cases: Blinding Vale, Temple of Sethraliss, Murder Row, Den of Nalorakk, Ruby Life Pools.
- Also fixed an unrelated bug spotted along the way: Temple of Sethraliss's version-2 `map_icons.json` had a leftover entry still pointing at the old `mapping_version_id` (19 instead of 807).
- Altar of Fangs and Voidscar Arena were checked and are **not** part of this regression — neither has ever had a start icon on any mapping version.

## Verification

Could not run a full local `db:seed --class=DungeonDataSeeder` re-seed to verify end-to-end: it fails deterministically in my worktree on an unrelated dungeon (Legion Arcway), before reaching any of the changed files. That traces to a separate control-flow bug in `DatabaseSeeder::run()` (filed as #3642) — its prepare/apply/cleanup loops silently stop processing and mask the failure once any one model's temp-table step fails. Confirmed the shared dev DB was untouched by the failed attempts (temp-table-swap safety held).

Verified instead via direct structural comparison: each new JSON entry matches its sibling's exact shape, with `mapping_version_id` corrected to the new version's real id and `floor_id`/`lat`/`lng` copied unchanged (floors are dungeon-level, not mapping-version-scoped, so no remapping needed). For Ruby Life Pools specifically, confirmed folder `3` maps to `mapping_version_id: 804` via its `floor_unions.json`/`floor_union_areas.json`, despite no folder being named "8" to match its nominal version number.

CI's isolated fresh-seed run is the authoritative check here, per the issue's definition of done.

## Test plan

- [ ] `php-tests (feature-controller)` green on this PR
- [ ] No other jobs regress


Wotuu (commented):
Ok

PR #3644, opened 2026-07-21
#3633 Fall back to newest known version when RioLogVersion's version quad is missing

### LINKED ISSUE #3633: RioLogVersion parameter-count parse failures — likely truncated log lines (follow-up to #3624)

Wotuu:
:robot: Closes #3633

## Summary
- The remaining "got 23" cluster in #3633 (4/500 unresolved `CombatLogParseFailure` rows, all `season_id=17`, `APP_VERSION 4.11.2`): `RioLogVersion` lines sometimes drop the entire trailing `COMBAT_LOG_VERSION,ADVANCED_LOG_ENABLED,BUILD_VERSION,PROJECT_ID` quad — always together, never partially. This is the segment's only version source (confirmed no fallback `COMBAT_LOG_VERSION` line exists anywhere else in the affected runs), and `RioLogVersion` implements `HasCombatLogVersionInterface`, so a successful parse of it drives the *active* parse version for the rest of the segment.
- Previously left as an intentional loud failure because naively falling through to the parser's hardcoded starting default (`RETAIL_11_0_5`) would silently route the rest of the segment through a stale version-specific handler (e.g. `CombatantInfoV21` instead of the current `V22`) — a real correctness risk, not just lost metadata.
- New evidence changes the calculus: every other same-day segment sampled across the affected runs (both before and after the failures, both addon versions seen in the data) independently reports the exact same version — `CombatLogVersion::RETAIL_ALL`'s newest entry at the time. So instead of guessing, fall back to `array_key_last(CombatLogVersion::RETAIL_ALL)` when the quad is missing — the same "newest known version" concept `PollCombatLogRunsCommand` already uses elsewhere — decomposed back into the (combat log version, build version) pair `getVersionLong()` expects. `ADVANCED_LOG_ENABLED`/`PROJECT_ID` also independently default to their only-ever-observed values (`true`/`1`); neither has any downstream reader.
- Residual risk is narrow: a wrong guess only occurs right at a WoW patch boundary, and would likely surface as further, detectable param-count failures on other event types in the same run rather than silent corruption.

## Test plan
- [x] Added `mplus-trash-win32-missing-version-quad` / `mplus-boss-win32-missing-version-quad` cases to `RioLogVersionTest`'s data provider, using real truncated Staging raw lines from one of the 4 failing rows (run 40595534) — green.
- [x] Added a `getVersionLong_givenRioLogVersionEventMissingVersionQuad_fallsBackToNewestKnownVersion` test asserting the fallback resolves to `array_key_last(CombatLogVersion::RETAIL_ALL)`.
- [x] Full `CombatLog` test group (212 tests) green.
- [x] Re-parsed the real failing run (40595534, 7 segments) end-to-end with the fix applied — 145,054 events, 0 failures (previously failed on segment 1, line 1, with the exact `wanted 31-35, got 23` error). The other 3 got-23 rows (493/460/406) share the identical truncation signature per the API but weren't independently re-downloaded and reproduced this session — full source data pattern already confirmed across all 4 in issue triage.


Wotuu:
:robot: `php-tests (feature-controller)` is red here, but it's a known master-wide CI break unrelated to this diff — tracked in #3639, already fixed and green in MR #3643 (not yet merged). This MR should go green on its own once #3643 merges; not duplicating that fix here.

Wotuu (commented):
Ok

PR #3646, opened 2026-07-21
#3645 Restore missing dungeon-start map icons on 3 more bare mapping versions

### LINKED ISSUE #3645: 3 more dungeons missing dungeon-start icon on their current mapping version (follow-up to #3639)
Follow-up to #3639 / #3643. That fix covered the 5 dungeons Wotuu flagged by hand (Temple of
Sethraliss, Murder Row, Ruby Life Pools, Den of Nalorakk, Blinding Vale), all "bare" mapping
versions created by #3629's floor-union backfill script that never got a cloned dungeon-start
`MapIcon`.

After merging #3643, CI on several unrelated PRs (#3644, #3447, #3285, #3638, #3630) is still
failing `DungeonRouteControllerSaveNewTemporaryTest::saveNewTemporary_givenValidStartMapIcon_setsStartMapIconId`
with the same signature (`Failed asserting that <mv> is identical to <mv>` at
`DungeonRouteControllerCreateTestBase.php:92`).

Root cause: a full scan of all active dungeons' current mapping version for a dungeon-start
`MapIcon` turned up 3 more affected dungeons beyond the original 5:

- `cathedralofeternalnight` (dungeon id 3): current mv 3 (v1, game_version_id 1) has no start
  icon; mv 611 (v2) and mv 630 (v3), both game_version_id 7, do.
- `courtofstars` (dungeon id 4): current mv 614 (v6, game_version_id 1) has no start icon; mv
  615 (v7) also doesn't; mv 4/72/83/89/97 (v1-5, game_version_id 1) and mv 631 (v8,
  game_version_id 7) do.
- `the_temple_of_atal_hakkar` (dungeon id 96): current mv 223 (v2) has no start icon; mv 178
  (v1) does.

Floors are dungeon-scoped, not mapping-version-scoped, so the existing sibling version's icon's
`floor_id`/`lat`/`lng` is safe to copy onto the bare current version (verified each source floor
is non-facade and the dungeon's `default` floor).

Also found (separately, NOT part of this bug, leaving alone): `voidscar_arena`, `magtheridons_lair`,
`altar_of_fangs` have zero dungeon-start icons on ANY mapping version, ever — that looks like
these dungeons never had one to begin with rather than a regression, so no safe source coordinate
to copy. Flagging in case that's also unintentional, but not fixing blind.

Wotuu:
:robot: Closes #3645

## Summary
#3643 fixed the 5 dungeons Wotuu identified by hand for the #3639 CI failure
(`DungeonRouteControllerSaveNewTemporaryTest::saveNewTemporary_givenValidStartMapIcon_setsStartMapIconId`).
After merging it, several unrelated PRs kept failing the same test with the same signature
(`642 identical to 806`) — a full scan of every active dungeon's current mapping version for a
dungeon-start `MapIcon` turned up 3 more dungeons hit by the same #3629 backfill-script gap (a
"bare" mapping version cloned without its dungeon-start icon):

- `cathedralofeternalnight`: current mv 3 (v1) had none — copied floor/coordinates from mv 611 (v2).
- `courtofstars`: current mv 614 (v6) had none — copied from mv 631 (v8).
- `the_temple_of_atal_hakkar`: current mv 223 (v2) had none — copied from mv 178 (v1).

Floors are dungeon-scoped, not mapping-version-scoped, so each source icon's floor/lat/lng is
valid on the bare current version too — verified each source floor is non-facade and the
dungeon's `default` floor before copying.

`voidscar_arena`, `magtheridons_lair`, `altar_of_fangs` have zero dungeon-start icons on any
mapping version, ever — left untouched (no safe source to copy from); flagged in #3645 for Wotuu
to confirm whether that's intentional.

## Verification
- `mapping:save` export touched exactly the 3 targeted seeder JSON files, no drift elsewhere.
- Full round-trip: re-seeded via `db:seedone DungeonDataSeeder`, then re-scanned all active
  dungeons — only the 3 known "never had one" dungeons remain without a start icon on their
  current version.

## Test plan
- [ ] CI green (this is what should finally unblock #3644/#3447/#3285/#3638/#3630's shared red CI)

### DIFF
--- a/database/seeders/dungeondata/classic/the_temple_of_atal_hakkar/4/map_icons.json
+++ b/database/seeders/dungeondata/classic/the_temple_of_atal_hakkar/4/map_icons.json
@@ -76,5 +76,18 @@
         "permanent_tooltip": 0,
         "seasonal_index": null,
         "is_admin": true
+    },
+    {
+        "mapping_version_id": 223,
+        "floor_id": 237,
+        "dungeon_route_id": null,
+        "team_id": null,
+        "map_icon_type_id": 10,
+        "lat": -28,
+        "lng": 191.25,
+        "comment": null,
+        "permanent_tooltip": 0,
+        "seasonal_index": null,
+        "is_admin": true
     }
 ]
--- a/database/seeders/dungeondata/legion/cathedralofeternalnight/1/map_icons.json
+++ b/database/seeders/dungeondata/legion/cathedralofeternalnight/1/map_icons.json
@@ -24,5 +24,18 @@
         "permanent_tooltip": 0,
         "seasonal_index": null,
         "is_admin": true
+    },
+    {
+        "mapping_version_id": 3,
+        "floor_id": 8,
+        "dungeon_route_id": null,
+        "team_id": null,
+        "map_icon_type_id": 10,
+        "lat": -204.326,
+        "lng": 179.6193,
+        "comment": null,
+        "permanent_tooltip": 0,
+        "seasonal_index": null,
+        "is_admin": true
     }
 ]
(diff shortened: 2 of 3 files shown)


Wotuu (commented):
Ok

PR #3649, opened 2026-07-21
#3642 Fix DatabaseSeeder prepare/apply/cleanup loops silently skipping after first failure

### LINKED ISSUE #3642: DatabaseSeeder's prepare/apply/cleanup loops silently skip everything after the first failure (follow-up to #3639)

Wotuu:
:robot: Closes #3642

## Summary
- `DatabaseSeeder::run()`'s prepare/apply/cleanup loops used
  `$failed = !$failed && !$call()`, which resets the failure flag back to `false`
  whenever a later call in the loop succeeds, and short-circuits so remaining
  items are never even attempted once the flag trips. Net effect: one failed
  `prepareTempTableForModel()` (etc.) silently skips the remaining models, with
  no error reported — until something inserts into a temp table that was never
  created.
- Extracted the accumulation into a shared `allSucceeded()` helper that maps
  over every item unconditionally (`array_map` never short-circuits) and checks
  for any `false` result, so a mid-list failure can neither mask later calls nor
  be masked by them.

## Test plan
- [x] New unit test `tests/Unit/Database/Seeders/DatabaseSeederTest.php` covers:
  first-item failure still invokes every remaining item, all-succeed returns
  true, and (the specific bug) a failure confined to the *last* item is no
  longer swallowed.
- [x] Verified these tests fail against the original buggy accumulation logic
  and pass against the fix.
- [x] `#[Group('DungeonDataSeeder')]` tests (real prepare/apply/cleanup path
  against the actual seeder pipeline) still pass, confirming no regression to
  the happy path.
- [ ] Self-review via the `code-review` skill could not be invoked directly by
  the agent (`disable-model-invocation`); only a manual review was done. Run
  `/code-review` on this PR if an automated pass is wanted.

### DIFF
--- a/database/seeders/DatabaseSeeder.php
+++ b/database/seeders/DatabaseSeeder.php
@@ -204,8 +204,10 @@ class DatabaseSeeder extends Seeder
     }
 
     /**
-     * Invokes $callback for every item and reports whether any call failed. Every item is always
-     * attempted, regardless of an earlier failure (see #3642).
+     * Invokes $callback for every item and reports whether every call succeeded. Every item is
+     * always attempted, regardless of an earlier item returning false (see #3642). This guarantee
+     * only holds for callbacks that signal failure by returning false - a callback that throws
+     * still aborts the remaining items, since the exception propagates out of array_map().
      *
      * @param array<int, class-string>     $items
      * @param callable(class-string): bool $callback
--- a/database/seeders/DatabaseSeeder.php
+++ b/database/seeders/DatabaseSeeder.php
@@ -105,12 +105,12 @@ class DatabaseSeeder extends Seeder
             $affectedModelClasses = [];
 
             try {
-                $prepareFailed = false;
-
                 $affectedModelClasses = $seederClass::getAffectedModelClasses();
-                foreach ($affectedModelClasses as $affectedModel) {
-                    $prepareFailed = !$prepareFailed && !$this->prepareTempTableForModel($affectedModel);
-                }
+
+                $prepareFailed = !self::allSucceeded(
+                    $affectedModelClasses,
+                    fn(string $affectedModel): bool => $this->prepareTempTableForModel($affectedModel),
+                );
 
                 if ($prepareFailed) {
                     $this->command->error(sprintf('Preparing temp table for %s failed!', $seederClass));
@@ -122,10 +122,10 @@ class DatabaseSeeder extends Seeder
                     $this->call($seederClass);
                 });
 
-                $applyFailed = false;
-                foreach ($affectedModelClasses as $affectedModelClass) {
-                    $applyFailed = !$applyFailed && !$this->applyTempTableForModel($affectedModelClass);
-                }
+                $applyFailed = !self::allSucceeded(
+                    $affectedModelClasses,
+                    fn(string $affectedModelClass): bool => $this->applyTempTableForModel($affectedModelClass),
+                );
 
                 if ($applyFailed) {
                     $this->command->error(sprintf('Applying temp table for %s failed!', $seederClass));
@@ -137,10 +137,10 @@ class DatabaseSeeder extends Seeder
 
                 throw $e;
             } finally {
-                $cleanupFailed = false;
-                foreach ($affectedModelClasses as $affectedModelClass) {
-                    $cleanupFailed = !$cleanupFailed && !$this->cleanupTempTableForModel($affectedModelClass);
-                }
+                $cleanupFailed = !self::allSucceeded(
+                    $affectedModelClasses,
+                    fn(string $affectedModelClass): bool => $this->cleanupTempTableForModel($affectedModelClass),
+                );
 
                 if ($cleanupFailed) {
                     $this->command->error(sprintf('Cleaning up temp table for %s failed!', $seederClass));
@@ -201,6 +201,18 @@ class DatabaseSeeder extends Seeder
         return DB::connection($instance->getConnectionName())->statement(
             sprintf('DROP TABLE %s;', $tableNameNew),
         );
+    }
+
+    /**
+     * Invokes $callback for every item and reports whether any call failed. Every item is always
+     * attempted, regardless of an earlier failure (see #3642).
+     *
+     * @param array<int, class-string>     $items
+     * @param callable(class-string): bool $callback
+     */
+    private static function allSucceeded(array $items, callable $callback): bool
+    {
+        return !in_array(false, array_map($callback, $items), true);
     }
 
     public static function getTempTableName(string $className): string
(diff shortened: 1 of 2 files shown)


database/seeders/DatabaseSeeder.php
```diff
@@ -203,6 +203,18 @@ private function cleanupTempTableForModel(string $className): bool
...
+    /**
+     * Invokes $callback for every item and reports whether any call failed. Every item is always
+     * attempted, regardless of an earlier failure (see #3642).
+     *
+     * @param array<int, class-string>     $items
+     * @param callable(class-string): bool $callback
+     */
+    private static function allSucceeded(array $items, callable $callback): bool
+    {
+        return !in_array(false, array_map($callback, $items), true);
```
Wotuu, line 217: :robot: The `array_map` fix only defeats the short-circuit for callbacks that **signal failure by returning `false`**. A callback that **throws** still aborts `array_map` at that item, so the remaining items are never attempted — contradicting the docblock's "Every item is always attempted, regardless of an earlier failure (see #3642)".

This matters most in the `finally` cleanup loop: `cleanupTempTableForModel()` runs `DROP TABLE ..._temp`, which **throws** (MySQL error 1051, not a `false` return) when the temp table is missing. So if the first model's cleanup throws, later models' temp tables are left undropped — exactly the resilience #3642 is trying to guarantee — and the thrown exception in `finally` can also mask the original error. The same is true for `prepareTempTableForModel` (`CREATE TABLE ... LIKE`) failing by exception rather than a `false` return.

Not necessarily blocking (DDL failures throwing rather than returning `false` is the common case, and this throw-path abort predates the PR), but the new docblock over-promises. Consider either catching per-item inside the callback so a throw is converted to `false`, or softening the docblock to scope the guarantee to non-throwing callbacks.

database/seeders/DatabaseSeeder.php
```diff
@@ -203,6 +203,18 @@ private function cleanupTempTableForModel(string $className): bool
         );
     }
 
+    /**
+     * Invokes $callback for every item and reports whether any call failed. Every item is always
```
Wotuu: :robot: Minor: the summary "reports whether any call **failed**" reads inverted relative to the method name and return value. `allSucceeded()` returns `true` when **all** calls succeeded (none failed), and callers use it as `$prepareFailed = !self::allSucceeded(...)`. A future reader taking the sentence literally could assume `true` means "a call failed" and drop the `!`. Suggest rewording to e.g. "reports whether every call succeeded".

Wotuu:

database/seeders/DatabaseSeeder.php
Wotuu: :robot: Done — softened the docblock rather than changing behavior: it now states the "every item attempted" guarantee only holds for callbacks that signal failure by returning `false`, and explicitly calls out that a throwing callback (e.g. `cleanupTempTableForModel()`'s `DROP TABLE`) still aborts the remaining items via `array_map()`. Left the actual short-circuit-on-throw behavior unchanged since converting it to catch-and-continue would change how prepare/apply failures propagate (currently a thrown exception there is an intentional hard-abort, caught by the outer `catch (Exception $e)` and rethrown) — that felt like a larger behavioral change to make alongside a docblock-accuracy fix.

database/seeders/DatabaseSeeder.php
Wotuu: :robot: Done — reworded the summary to "reports whether every call succeeded" so it matches `allSucceeded()`'s name and return value instead of reading inverted.

Wotuu (commented):
Ok

PR #3652, opened 2026-07-21
#3642 Address code-review follow-ups on the prepare/apply/cleanup fix

### LINKED ISSUE #3642: DatabaseSeeder's prepare/apply/cleanup loops silently skip everything after the first failure (follow-up to #3639)

Wotuu:
:robot: Follow-up to #3649 (which fixed #3642 and is already merged).

## Summary
A code-review pass on #3649 surfaced a few things worth addressing before calling it settled:

- **Reliability fix**: `cleanupTempTableForModel()` now uses `DROP TABLE IF EXISTS` instead of a
  bare `DROP TABLE`. If an earlier model's prepare/apply step throws, its temp table was never
  created — but the `finally` block still calls cleanup for every model in the list. A bare `DROP
  TABLE` on a nonexistent table throws, and per PHP's `finally` semantics that new exception
  replaces the real in-flight exception the `catch` block was propagating, masking the actual root
  cause. `DROP TABLE IF EXISTS` makes cleanup tolerate this instead. (Verified pre-existing in the
  old buggy code too, not introduced by #3649 — but #3649 touched this exact code without fixing
  it, so folding it in here.)
- **Naming**: renamed `allSucceeded()` to `anyFailed()` and dropped the internal negation, so all
  3 call sites read directly (`$prepareFailed = self::anyFailed(...)`) instead of double-negating.
- **Docblock accuracy**: clarified that the "every item is always attempted" guarantee only holds
  when a callback signals failure by returning `false` — a callback that *throws* still aborts
  remaining items, since the exception propagates out of `array_map()`.
- **Test cleanup**: renamed tests to match `anyFailed()` and the repo's
  `[functionname]_given[Condition]_returns[ExpectedResult]` convention, replaced a parallel-array +
  `array_search` lookup with a direct associative map, and added a regression test proving cleanup
  no longer throws when a temp table was never created.

## Test plan
- [x] New test `cleanupTempTableForModel_givenTempTableWasNeverCreated_returnsTrueWithoutThrowing`
  — verified it fails against a plain `DROP TABLE` (no `IF EXISTS`) and passes with the fix.
- [x] Existing `anyFailed()` unit tests updated and pass.
- [x] `#[Group('DungeonDataSeeder')]` feature tests (real prepare/apply/cleanup pipeline) still pass.


Wotuu (commented):
Ok

Wotuu:

PR #3660, opened 2026-07-24
#128 Enemy forces checkpoints: how much you need before entering a corridor

### LINKED ISSUE #128: Show how much % in total is on a floor
This is helpful in places like Waycrest where you can then easily see how much % you need before advancing to the next level.

Wotuu:
:robot: Closes #128

> **This PR was revamped.** It previously showed the total enemy forces *on a floor*. That answers
> the wrong question, so the per-floor pill has been removed and replaced by what the issue is
> actually about. The first commit strips the old work back to the parts worth keeping.

## The problem

> *"We're at 80% and about to walk into this corridor. The corridor only holds 19%, so we need to
> pick up 1% before we go in or we end up short at the far end."*

A corridor is **not a floor**. It's an arbitrary group of enemies that may span several floors, and
what you want on screen is **how much you must already have before entering it**.

## What this adds

A new mapping-versioned model, **`EnemyForcesCheckpoint`** — a mapper-defined group of enemies (a
corridor, a wing) with an anchor point. Defined once per mapping version, so every route in that
dungeon gets it for free.

On the map it renders as a pill reading **`Need 80.8% before`**, following the existing **Enemy
number style** setting (`Need 235 EF before` in absolute mode).

### Cross-floor checkpoints

A checkpoint can span floors, but is anchored on one. Leaving it invisible on the other floors it
covers would hide it exactly where it matters, so on any other floor holding members it draws a
**satellite pill** at the centroid of that floor's members.

That centroid is safe to compute client-side *precisely because* split-floors coordinates are never
facade-converted. In the facade layout every enemy and the anchor itself sit on the one image, so a
satellite is never drawn and the single server-converted anchor suffices.

| Split-floors, anchor floor | Split-floors, the *other* floor the checkpoint spans |
| --- | --- |
| anchor | satellite |

Facade layout — one pill at the server-converted anchor:

facade

### Admin editing

Mirrors enemy patrols: place the marker, hit **Select enemies**, then click enemies (pack buddies
come along) to toggle membership. Lines are drawn from the checkpoint to each member.

Unlike patrols the selection is deliberately **not** stopped after one click — a corridor is many
enemies across several floors.

admin

#### Switching floors mid-selection — two bugs in `AdminDungeonMap`

This is the whole point of the feature, so I drove it in a real browser rather than trusting the
code read. It did **not** work, for two reasons, both `AdminDungeonMap` overriding `DungeonMap`
without forwarding to it:

1. `AdminDungeonMap::refreshLeafletMap()` took no arguments and called `super.refreshLeafletMap()`
   bare, so `clearMapState` fell back to its default of `true`. The `floorid:changed` handler passes
   `false` *specifically* so a selection survives a floor switch — the admin mapping page discarded
   that. It now forwards its arguments.
2. `AdminDungeonMap::_getMapControls()` replaces the base list wholesale and never calls `super`, so
   registering the new panel in `DungeonMap` did nothing on the only page that can edit checkpoints.

A **snackbar** is new here too (it was in the plan and I initially missed it; originally shipped as
a bespoke edit panel, replaced by a snackbar in the review round below to match the project's
existing "you're currently doing X" convention). While a selection is active it shows which checkpoint
is being edited and its running count/percentage, with a Done button. It exists precisely because
the checkpoint's own marker is floor-gated by `MapObject.shouldBeVisible()` and vanishes the moment you
switch floors — without it the admin is clicking enemies blind, with nothing on screen saying
selection mode is even still on.

Below: floor 61's checkpoint, still being edited after switching to floor 62 (The Cellar) - shown here
with the original edit-panel version, since replaced by the snackbar described below. The marker
itself is gone, the panel isn't, the enemies here are selectable, and the satellite pill shows.

cross-floor

`name` is **nullable**. A marker exists the moment it is placed and is saved right then, before the
mapper has had any chance to type anything — requiring a name made placing a checkpoint fail with a 422.
The name is filled in afterwards in the checkpoint's popup, which is where the **Select enemies** button
lives too. (Found by driving the actual draw tool in a browser, not by reading the code.)

## Things worth a careful look

- **`deleted`, not `deleting`.** The model releases its member enemies on `deleted`. Eloquent fires
  `deleting` through the dispatcher's `until()`, which halts on the first non-null result, and
  `SeederModel`'s listener always returns a bool — so a second `deleting` listener would never run
  at all. Without this, enemies keep a dangling `enemy_forces_checkpoint_id` and the next checkpoint handed
  that auto-increment id silently inherits them and reports enemy forces nobody assigned to it.
  (`EnemyPatrol` has exactly this dangling-FK behaviour today; for patrols it's cosmetic, here it
  would be a wrong percentage.)
- **The `$idMapping` second-pass FK re-link** in `MappingVersion::boot()`. Without it a cloned
  mapping version's enemies keep pointing at the *old* checkpoint, so the new version's checkpoint reports
  zero enemies while looking perfectly fine. This is the single most breakable part and has a
  dedicated test.
- **`MappingService::copyMappingVersionContentsToDungeon()` deliberately skips checkpoints.** That path
  doesn't copy enemies either, so a copied checkpoint would arrive empty and read 0%.
- **The `AdminDungeonMap::refreshLeafletMap()` fix changes behaviour beyond this feature.** Any map
  state now survives a floor switch on the admin mapping page, not just checkpoint selection — an admin
  in delete mode or MDT-enemy-selection who changes floor stays in that mode instead of being
  silently kicked out. That matches how the non-admin map has always behaved (the base class takes
  the argument for exactly this reason), and I believe the old behaviour was an accident of dropping
  the parameters rather than a decision — but it is the one change here that touches something
  outside enemy forces checkpoints, so it deserves a second opinion.
- **"Need 81% before" is only correct for a terminal / one-way checkpoint** (at 80, corridor holds 19,
  80+19 = 99 → short). That is the plain reading of the request and what a mapper defining these
  will mean; it is not "everything after this point".

## Tests

- **Vitest**: `getEnemyForcesForFloor` generalised to `getEnemyForcesForEnemies`, tests ported.
  Obsolete enemies stay *included* (unlike `KillZone.getEnemyForces()`) — an obsolete enemy is still
  standing there and still yields its forces. Full suite green (257).
- **PHPUnit**: clone-forward + FK re-link, delete-releases-enemies, and ajax
  create/delete/validation/authorization — plus one asserting that PUTting `enemy_forces_checkpoint_id`
  to the **enemy** endpoint actually persists the column, since that (not a dedicated endpoint) is
  what makes an enemy a member, and a 200 alone would look identical if the field were dropped.
  `--group=MappingVersion` and `--group=Controller` green.
  - The validation test asserts the `ValidationException` from the FormRequest rather than a
    rendered 422: how this app renders a ValidationException on `/ajax/` paths is a pre-existing,
    app-wide concern unrelated to this model.
  - Fixtures use `$dungeon->mappingVersions()->first()` rather than `getCurrentMappingVersion()` —
    `MappingVersion::boot()` clones from the highest `version` across *every* game version, while
    `getCurrentMappingVersion()` is scoped to one, and they are not always the same row.

**Also fixes an unrelated flake this PR's CI caught.** `DungeonRouteEnemyRaidMarkerMappingVersionTest`
still had one method (`...NpcNoLongerExistsInNewMappingVersion_deletesRaidMarker`) picking a random
retail dungeon and then a random enemy from it — the fix for #3679 guarded its two siblings but not
this one, so the same ~1-in-20 TypeError could still fire from a different method. It now uses the
same guarded helper (verified 5 consecutive green runs).

### Verified in a real browser (not just tests)

Beyond the screenshots above: placed a checkpoint with the draw tool (`POST … -> 201`), opened its
popup, ran the **Select enemies** flow and clicked an enemy — its two pack buddies came along
(`PUT …/enemy/{id} -> 200` ×3), the selection mode stayed active, and the pill updated live. Then
switched floors with the selection live and assigned again: members ended up spanning both floors
(`memberFloorIds: [62, 61]`). No page errors throughout. Verification data and the throwaway admin
user were cleaned up afterwards.

**Seeder round-trip** was run for real: `mapping:save` emits `enemy_forces_checkpoints.json` and
`enemies.json` carries `enemy_forces_checkpoint_id`. The exported dev seeder data was reverted
afterwards so it isn't in this diff.

One thing that looks wrong but isn't: the pill reads a higher percentage on the **admin mapping**
page than on explore for the same checkpoint. That page's map context has teeming on, so
`getEnemyForcesRequired()` returns `enemy_forces_required_teeming` (1000 vs 305 on Waycrest) — the
same denominator every other enemy-forces readout uses there.

## Notes

- Hidden by default in **route thumbnails** (`preview.blade.php`) and **embeds**; user-toggleable
  everywhere else via the existing show/hide dropdown.
- **Cache**: the payload lives in `dungeon_{id}_{mvid}_{style}`, which mapping edits do not
  invalidate today — admins use the drop-caches tool, same as for any other mapping change. Worth
  noting separately that `dropCaches()` clears the redis layer but not the `tmp_file` layer that
  `rememberLocal()` wraps it in; that is pre-existing and affects every mapping edit, so I left it
  alone rather than smuggling a behaviour change in here. Happy to file it.
- The facade `floor_id` trap that made the original per-floor approach fail is documented separately
  in #3682 (docs only, no code).
- **Echo**: the `enemyforcescheckpoint-changed` / `-deleted` broadcasts follow `MountableArea` exactly —
  emitted server-side, with no client listener. No admin mapping model (mountable areas, floor
  unions, patrols) has one; only route-editing models do. Deliberately consistent, not an oversight.
- Follow-up filed for the "gap vs your current route" readout: #3687.

## Review round (2026-07-27)

Addressed in a follow-up commit after Wotuu's first pass:

- Added the `js.enemyforcescheckpoint*` translation keys `mapobject.js` generates dynamically for
  every popup attribute (`name`/`faction`/`teeming`/`select_enemies` labels, plus the bare model
  name used as the popup title and draw-tool button caption) - all missing from the original diff.
- The draw tool now shows a cursor preview while placing a checkpoint (a static `fa-percent` glyph,
  matching the killzone/mapicon pattern) - `LeafletEnemyForcesCheckpointIcon` was previously an empty
  `divIcon` with no `html`, so nothing rendered under the cursor.
- Hovering a checkpoint's pill now highlights its member enemies, mirroring the existing `EnemyPatrol`
  mouseover/mouseout wiring in `enemyvisualmanager.js`.
- The admin edit panel described above is now a **snackbar** instead of a bespoke Leaflet control,
  consistent with every other transient "you're currently doing X" admin state (overpull
  selection, draw toolbar).
- Filed #3696 for a **pre-existing, not-introduced-by-this-PR** bug: the yellow dashed selection
  border doesn't neatly hug the enemy icon during any enemy-selection mode (patrol, MDT, or this
  checkpoint flow) - confirmed present on `master` today via the patrol flow too, likely BS4->BS5
  fallout.
- Whether hovering an enemy while selecting suppresses its tooltip: empirically A/B'd first
  (headless Chrome, simulated hover via `layer.fire('mouseover')`, `isTooltipOpen()` before/after)
  against both `EnemyPatrolEnemySelection` and `EnemyForcesCheckpointEnemySelection` on the
  pre-fix code - **neither suppressed the tooltip**, so this wasn't a regression #3660
  introduced. Wotuu confirmed he wants it suppressed everywhere regardless, so it's now fixed at
  the shared root rather than per-subclass: `MapState.disablesTooltips()` (false by default, true
  on `EditMapState` and on the enemy-selection states that want it). **Narrowed after review** - it
  briefly lived on the shared `EnemySelection` base, which also silenced tooltips during ordinary
  pull-building in the non-admin route editor (`EditKillZoneEnemySelection` /
  `SelectKillZoneEnemySelectionOverpull` extend the same base). It now sits on
  `EnemyPatrolEnemySelection`, `MDTEnemySelection` and `EnemyForcesCheckpointEnemySelection` only.
  - **First attempt crashed on the next enemy click.** Wiring `disablesTooltips()` through
    `Enemy.bindTooltip()`/`unbindTooltip()` on every `map:mapstatechanged` hit a real Leaflet
    defect: `Tooltip.js`'s `_addFocusListenersOnLayer` attaches a raw DOM `focus` listener that's
    re-added on every unbind but never removed, and unlike `_openTooltip` it has no null-guard on
    `this._tooltip` - so `Cannot set properties of null (setting '_source')` fired the moment a
    click (which focuses the marker, since Leaflet markers are keyboard-focusable) landed after
    an unbind. Fixed by never touching Leaflet's tooltip bind state for suppression: enemy
    tooltips get `bindTooltip({className: 'map_enemy_tooltip'})` unconditionally, and
    `DungeonMap.setMapState()` toggles one CSS class on the map container instead
    (`.map_enemy_tooltips_suppressed .map_enemy_tooltip { display: none !important; }`). Verified
    in a real browser: hides during any selection mode, reappears after, a checkpoint's own
    tooltip is unaffected, and repeated state cycles + a simulated native focus event no longer
    throw.
  - **That fix was correct but incomplete — the crash still reproduced.** It removed one of *two*
    independent unbind sites. `AdminEnemy._mapStateChangedEvent()` had its own copy of the same
    suppression idea, predating all of this: as soon as any `EnemySelection` started it called
    `unbindTooltip()` on every enemy within `getLatLngDistanceSquared() < 100` of the selected
    object. Those enemies were left with a null Leaflet tooltip while their icon kept the leaked
    `focus` listener, so clicking one — the enemies nearest the checkpoint, i.e. exactly the ones a
    mapper clicks — threw once per enemy. Captured from the real admin flow in headless Chrome
    (open the pill's popup → *Select enemies* → real mouse click on an enemy): **2 uncaught errors
    per click before, 0 after**, with the stack pinning `AdminEnemy._mapStateChangedEvent`.
    Three changes:
    1. `resources/assets/js/leaflet-tooltip-focus-fix.js` re-declares Leaflet's
       `_addFocusListenersOnLayer()` with the null-guard its sibling `_openTooltip()` already has,
       wired up in `bootstrap.js` next to the other self-owned shims. **This is the load-bearing
       one** — the app unbinds tooltips from still-live layers in several other places too
       (`MapObjectGroup#setLayerToMapObject`, `EnemyVisual`'s circle menu), and each leaves the same
       trap. Covered by `leaflet-tooltip-focus-fix.test.js`, which drives real Leaflet under jsdom;
       its first test asserts stock Leaflet *does* throw, so a future Leaflet bump that fixes this
       upstream turns that test red rather than leaving the shim silently pointless.
    2. `AdminEnemy._mapStateChangedEvent()` no longer touches tooltip bindings at all — the same
       feature implemented twice, and this copy was the trigger.
    3. `AdminEnemy.bindTooltip()` now passes `className: 'map_enemy_tooltip'`. Admin enemies build
       their own tooltip, so the CSS rule from the previous attempt matched **nothing** on the admin
       mapping page; without this, removing (2) would have silently brought the tooltips back during
       selection. Caught in verification, not by reading the code.
  - **Scope note worth a look:** because of (3), the admin mapping page now follows
    `MapState.disablesTooltips()` like every other map does — which means enemy tooltips are also
    suppressed in `EditMapState`, not just during enemy selection. (`DeleteMapState` and the draw
    tools are unaffected; they return `false`.) That matches the intent of `disablesTooltips()` as
    documented on the base class, but it *is* new on the admin page, where the enemy tooltip is the
    identification tool (`npc_id`, `enemy_id`, `mdt_id`, healths). This stays as-is; what *was*
    narrowed is the route-editor half - see above.

    Verified end to end in a real browser — no map state: tooltip visible; checkpoint selection
    active: `map_enemy_tooltip` present and `display: none`; after leaving the state: visible again;
    zero page errors throughout.

    tooltip suppression
- **Renamed `EnemyForcesRegion` → `EnemyForcesCheckpoint`** everywhere (model, migrations, events,
  controller, form request, repository, JS classes/files, CSS, translations, tests) - "region"
  implied a bounded area, but the model is just an anchor point + a flat list of member enemies,
  no shape at all. Not released anywhere yet, so the migrations were edited/renamed in place
  rather than expand/contract'd.

## Review round 2026-07-27 (commit `c41617e90`)

Five findings from the cold review, all addressed:

1. **`mapping:copy` left checkpoints on the source dungeon's floors** - `enemyForcesCheckpoints` added
   to `Copy.php`'s `$relations`. Documented in the `mapping-versioned-models` skill: that array is a
   fourth hardcoded model list, and no test covers the command.
2. **`copyMappingVersionContentsToDungeon()`'s comment overclaimed** - it justified skipping
   checkpoints with "this path doesn't copy enemies either", which holds for the bare-mapping-version
   caller but not for the MDT import path, where every package bump silently discards every
   checkpoint. **Decision: behaviour kept, comment rewritten to say so honestly, and tracked in
   #3702.** Nothing is lost today (the table ships in this PR, so there are no checkpoints anywhere);
   preserving membership needs an id map across a service boundary, a rule for checkpoints left empty
   by upstream enemy removals, and a test harness `MDTMappingImportService` does not have. Happy to
   implement it here instead if preferred.
3. **Test leaked 121 rows per run** into the shared test DB (query-builder delete skipped
   `MappingVersion`'s `deleting` cascade). Now a single Eloquent `->delete()`; measured 0 delta over
   two consecutive runs.
4. **Satellite pill and admin connection lines escaped the "Enemy forces checkpoints" toggle** - both
   were added straight to the leaflet map, and `refreshPill()` re-created the satellite on every floor
   switch while the group was hidden. Both now gate on the owning group's `isShown()` and redraw on
   its `visibility:changed`. Verified in headless Chrome, including against the pre-fix build to
   confirm the bug was real.
5. **Tooltip suppression narrowed** off the shared `EnemySelection` base (see above), verified in a
   real browser on both the route editor and the admin mapping page. Stray blank line in
   `dungeonmap.js` removed.

New vitest coverage: `enemyselection.tooltips.test.js` (6 tests) and `enemyforcescheckpoint.test.js`
(5 tests). Full vitest suite 272 passed; `tests/Feature/App/Model/Mapping/`,
`AjaxEnemyForcesCheckpointControllerTest` and `tests/Feature/App/Service/` all green; PHPStan clean.

| group shown | group hidden |
|---|---|
| shown | hidden |

## Review round 2026-07-27 evening (commit `6fc6a5898`)

Six items from Wotuu's pass, all addressed (details + verification data in the inline replies):

1. **Pill/edit-border misalignment** - the inner-pill CSS transform is gone; the divIcon now carries a measured `iconSize` + centered `iconAnchor` (measured *inside* the leaflet container - its em base font differs from body's), so Leaflet itself centers the marker, re-applies it on every element re-creation, and leaflet.draw's `_offsetMarker()` keeps working. Verified: pill center == anchor exactly, edit border box == pill + 4px per side.
2. **AdminEnemyPatrol copy-paste** - extracted `AdminEnemyConnections` (new file); both admin classes now only supply parent layer/center/latlngs.
3. + 4. **`instanceof` chains in `enemyvisual.js`** - replaced by `EnemySelection.drawsEnemyEditBorder()` (base `false`, `true` on the three admin selection states), pinned by new vitest cases alongside the `disablesTooltips()` ones. The similar-but-different chain in `icon.js` (different membership + semantics) is deliberately untouched, flagged in the reply.
5. **`MapObjectGroupManager` if-chain** - now a name → factory map with the editability rule stated once.
6. **Leaflet upstream issue + MR** - tracked in #3712 with the AI-transparency requirements recorded verbatim, to be run as its own task.

Also this round (earlier the same day): the `mapping-versioned-models` skill's checklist was rewritten into the complete end-to-end "add a new mapping-versioned model" flow with this PR as the worked example (`77a10adcd`, docs only).

Vitest 295 green; PHPStan clean (no PHP touched this round). Browser verification data (checkpoint + throwaway admin) cleaned up afterwards.

## Review round 2026-07-28 (commits `26f9a83f4`, `658588f57`)

**Reported:** assigning an enemy to a checkpoint briefly flashes that enemy's own edit popup;
`EnemyPatrolEnemySelection` does not do this.

Two independent causes, both in `Enemy`. The patrol flow dodges the second one only by accident - it
stops its map state on the first click, so nothing happens *during* a selection there.

1. **The unbind on selection start never ran.** `setPopupEnabled(false)` was guarded on
   `this.isPopupEnabled`, but that flag records what was last *asked for*, not what is bound:
   `save()`/`setSynced()` bind a popup by calling `_assignPopup()` straight through and never touch
   it. A freshly loaded enemy therefore sits there with a bound popup while the flag still reads
   `false`, the guard short-circuits, and the very first enemy click of a selection opens the edit
   popup - which the save's own `closePopup()` shuts again a moment later. That is the flash.
   Disabling is now unconditional; enabling keeps its guard.
2. **Every save during the selection re-bound it.** Each click saves the clicked enemy *and* each of
   its pack buddies, and each save runs `setSynced(true)` -> `_assignPopup()`, bypassing
   `setPopupEnabled()`. `Enemy._assignPopup()` now unbinds and returns while a map state is active,
   so a save cannot resurrect the popup.

**Scope is confined to the admin mapping page.** `Enemy.isEditable()` is `false` and the only other
subclass, `PridefulEnemy`, returns `false` from `isEditableByPopup()` - neither ever has a popup
bound, so the unconditional unbind is inert in the route editor. Only `AdminEnemy` is affected. The
added `layer.off('popupopen')` is safe for the same reason the rebind path already does it:
`MapObject._assignPopup()` is the only place in the codebase that registers a `popupopen` handler.

**Verified in headless Chrome** on the admin mapping page (real login, checkpoint placed, *Select
enemies*, real mouse clicks on a 3-enemy pack), counting Leaflet `popupopen` events on the same page
before and after the fix:

| | before | after |
|---|---|---|
| enemy popups opened during the selection (2 clicks) | **2** (one per click) | **0** |
| enemy popup bound after starting the selection | yes | no |
| enemy popup bound after a save mid-selection | yes | no |
| enemy popup bound again after leaving the selection | yes | yes |
| plain click reopens the popup afterwards | yes | yes |
| page errors | 0 | 0 |

The **patrol** flow was A/B'd on both builds and is byte-for-byte identical: one popup open on the
assigning click, which is the normal post-selection edit popup since the state has already ended.

No screenshot for this round: the defect is a transient flash, so a still frame cannot show its
absence - the `popupopen` counts above are the stronger evidence.

Vitest 300 green (4 new tests in `enemy.test.js` pinning both halves, including that the popup comes
back once the state ends - a too-broad guard would otherwise kill admin enemy popups permanently).
No PHP touched. Verification data (checkpoint + throwaway admin) cleaned up afterwards.


resources/assets/js/custom/mapcontrols/floorenemyforcescontrols.js
```diff
@@ -0,0 +1,198 @@
...
+     * Refreshes the UI to reflect the current per-floor enemy forces.
+     */
+    refreshUI() {
+        console.assert(this instanceof FloorEnemyForcesControls, 'this is not FloorEnemyForcesControls', this);
+
+        if (getState().isCurrentDungeonFacadeEnabled()) {
+            this._renderFacadeMarkers();
+        } else if (typeof this.statusbar !== 'undefined') {
+            let floorEnemyForces = this.map.enemyForcesManager.getEnemyForcesForFloor(getState().getCurrentFloor().id);
+            $(this.statusbar).find('.map_enemy_forces_floor_pill_value').html(this._formatValue(floorEnemyForces));
```
Wotuu: :robot: **Correctness** — in split-floors mode the pill renders unconditionally, so dungeons with no enemy forces get a permanent `Total: 0%` banner.

The control is pushed outside the speedrun `else if` in `dungeonmap.js`, so it is created for every map with `showControls.enemyForces` (which defaults to `true` in `common/maps/map.blade.php:82`) — including speedrun dungeons. I checked the mapping data: 13 of the 14 `speedrun_enabled` dungeons are raids with `enemy_forces_required = 0` and enemies summing to 0 EF (e.g. Naxxramas Classic, mapping version 524: 762 enemies, 0 total EF). `getFormattedPercentage()` returns `0` when `max === 0`, and percentage is the *default* number style, so those maps get a `Total: 0%` pill on every floor.

The facade path already guards against exactly this (`if (floorEnemyForces <= 0) { continue; }`, line 105) — the split path doesn't. Note that gating on `speedrun_enabled` would be the wrong fix: The Arcway is `speedrun_enabled` *and* has `enemy_forces_required = 392`, so the pill is meaningful there. Mirroring the zero-skip is the right fix.

Careful with the shape of it though — the statusbar node is created once in `onAdd`, so an early `return` here would leave the previous floor's value on screen after a floor switch. Toggle visibility instead:

```js
} else if (typeof this.statusbar !== 'undefined') {
    let floorEnemyForces = this.map.enemyForcesManager.getEnemyForcesForFloor(getState().getCurrentFloor().id);
    $(this.statusbar).find('.map_enemy_forces_floor_pill_value').html(this._formatValue(floorEnemyForces));
    $(this.statusbar).toggle(floorEnemyForces > 0);
}
```

resources/assets/js/custom/enemyforces/enemyforcesmanager.js
```diff
@@ -112,6 +112,39 @@ class EnemyForcesManager extends Signalable {
...
+        }
+
+        for (let key in enemyMapObjectGroup.objects) {
+            /** @type {Enemy} */
+            let enemy = enemyMapObjectGroup.objects[key];
+
+            // Mirror the filter used by KillZone.getEnemyForces() (skip obsolete) and additionally skip
+            // enemies that aren't actually present (wrong seasonal type/affix), so the total matches what
+            // is shown on the map.
+            if (enemy.floor_id === floorId && !enemy.isObsolete() && !enemy.shouldBeIgnored()) {
```
Wotuu: :robot: **Correctness** — excluding `isObsolete()` enemies contradicts this method's own documented semantics, and it also makes the pill go stale.

The docblock says the sum reflects "how much enemy forces is present on that floor, regardless of what is currently pulled". But `isObsolete()` means "was part of a route, but is no longer because we determined we should no longer pull this enemy after an overpull elsewhere" (`enemy.js:1118`/`setObsolete`) — the enemy is still standing on the floor and still yields enemy forces if you kill it. So it shouldn't be excluded from an "available on this floor" total. The inline comment justifies the filter by mirroring `KillZone.getEnemyForces()`, but that method filters obsolete precisely because it sums *pulled* enemies — the semantics this one explicitly disclaims.

The staleness is the symptom that proves it: `SelectKillZoneEnemySelectionOverpull._applyObsoleteEnemies()` (`mapstate/enemyselection/selectkillzoneenemyselectionoverpull.js:190`) re-flags every enemy's obsolete state after an overpull in the route editor, and `FloorEnemyForcesControls` only listens to `mapnumberstyle:changed`, `teeming:changed` and `map:mapobjectgroupsloaded` — so nothing recomputes the pill and it silently disagrees with the rest of the map. (`dungeonroutesearchsidebar.js:179` flips obsolete at runtime too.)

Dropping `!enemy.isObsolete()` fixes both without wiring up a new signal. Heads-up that `enemyforcesmanager.test.js` currently pins the opposite behaviour (`getEnemyForcesForFloor_givenObsoleteEnemy_excludesItFromSum`) and is green today, so that assertion needs inverting alongside the change.

scripts/build/custom-scripts-order.mjs
```diff
@@ -184,6 +184,7 @@ export const customScripts = [
     'resources/assets/js/custom/mapcontrols/dungeonspeedrunrequirednpcscontrols.js',
     'resources/assets/js/custom/mapcontrols/echocontrols.js',
     'resources/assets/js/custom/mapcontrols/enemyforcescontrols.js',
+    'resources/assets/js/custom/mapcontrols/floorenemyforcescontrols.js',
```
Wotuu: :robot: Nit (cosmetic, no functional impact): the `mapcontrols/` block is alphabetically ordered after its base class — `addisplaycontrols`, `drawcontrols`, `dungeonspeedrunrequirednpcscontrols`, `echocontrols`, `enemyforcescontrols`, `enemyvisualcontrols`, `factiondisplaycontrols`. `floorenemyforcescontrols.js` breaks that; it belongs after `factiondisplaycontrols.js`. Load order is correct either way since `mapcontrol.js` (line 181) precedes it.

Wotuu (commented):
:robot: Cold review of the diff. Nice, well-scoped frontend-only change: the split/facade split is handled properly, `cleanup()` unregisters everything it registers, the facade toggle reloads the page so there's no stale-mode risk, `L.divIcon({iconSize: null})` is safe (Leaflet skips width/height/margin when the size is null) and the CSS `!important` overrides cover it, `getCenteroid`/`getFormattedPercentage`/`tophorizontalcenter` all already exist, camelCase `:enemyForces` placeholders match existing `js.php` convention (`:baseHealth`), the handlebars template is picked up automatically by the `handlebars/` glob, the `module.exports` guard matches the existing pattern in `dungeonmap.js`/`util.js`, and the new file is registered in `custom-scripts-order.mjs`. CI is green.

Two correctness findings inline, plus one ordering nit. Nothing security-relevant (no backend/API surface, values come from `lang.get()` + numbers).

Wotuu:

resources/assets/js/custom/mapcontrols/floorenemyforcescontrols.js
Wotuu: :robot: Already fixed - the `$(this.statusbar).toggleClass('d-none', floorEnemyForces <= 0)` guard landed in `9abc42d745` ("Pin the pill to the visible map top, and show it on the explore page"), which was pushed after this comment was posted. The pill now hides itself (rather than early-returning, which would leave a stale value on screen after a floor switch) whenever the floor's enemy forces are 0. No further change needed here.

resources/assets/js/custom/enemyforces/enemyforcesmanager.js
Wotuu: :robot: Fixed in 81f06a5a5 - dropped the `!enemy.isObsolete()` filter from `getEnemyForcesForFloor()` so obsolete enemies (still standing on the floor, still worth their enemy forces) are counted again, matching the method's own docblock. Inverted `enemyforcesmanager.test.js`'s `getEnemyForcesForFloor_givenObsoleteEnemy_excludesItFromSum` to `_includesItInSum`, asserting the sum now includes the obsolete enemy's forces. Full JS suite green (24 files, 252 tests).

scripts/build/custom-scripts-order.mjs
Wotuu: :robot: Fixed in 81f06a5a5 - moved `floorenemyforcescontrols.js`'s entry to right after `factiondisplaycontrols.js` in `custom-scripts-order.mjs`, restoring alphabetical order in the `mapcontrols/` block.

resources/assets/js/custom/mapcontrols/floorenemyforcescontrols.js
Wotuu: :robot: Already moot: this file (`floorenemyforcescontrols.js`) and the entire per-floor pill it drew were removed by commit `de3226fe6` ("Strip the per-floor enemy forces pill back to its plumbing") — the feature was redesigned around `EnemyForcesRegion` instead of a floor total. The new `enemyforcesregioncontrols.js` control only shows its panel while a region is actively being edited (`isEditing` gate on `mapState instanceof EnemyForcesRegionEnemySelection`), so the "permanent Total: 0% banner" this comment describes no longer has anywhere to occur. No code change made for this finding.

resources/assets/js/custom/enemyforces/enemyforcesmanager.js
Wotuu: :robot: Already fixed on this branch: commit `81f06a5a5` ("Include obsolete enemies in the per-floor enemy forces total") dropped the `!enemy.isObsolete()` condition and rewrote the docblock to state the correct semantics explicitly. The method has since been generalised to `getEnemyForcesForEnemies()` (commit `de3226fe6`), and its test `getEnemyForcesForEnemies_givenObsoleteEnemy_includesItInSum` in `enemyforcesmanager.test.js` already pins the corrected (obsolete-inclusive) behaviour and is green. No further change needed.

scripts/build/custom-scripts-order.mjs
Wotuu: :robot: Fixed, on the current file. The originally-flagged `floorenemyforcescontrols.js` entry no longer exists (removed along with the file by `de3226fe6`), but its replacement, `enemyforcesregioncontrols.js`, had landed in the same out-of-order spot (after `enemyvisualcontrols.js` instead of before it). Moved it to sit right after `enemyforcescontrols.js` and before `enemyvisualcontrols.js`, restoring alphabetical order in the `mapcontrols/` block. Pushed in `91b771b47`.

app/Models/Mapping/MappingVersion.php
```diff
@@ -540,18 +574,20 @@ protected static function boot()
                 'enemyPatrols',
                 'mapIcons',
                 'mountableAreas',
+                'enemyForcesCheckpoints',
```
Wotuu, line 577: :robot: **`mapping:copy` to a different dungeon leaves checkpoints anchored on the source dungeon's floors.**

Checkpoints are wired into all three hardcoded lists in this `boot()` — good — but there is a fourth hardcoded model list this PR doesn't touch: the `$relations` array in `app/Console/Commands/Mapping/Copy.php` (~lines 74-84).

`mapping:copy <gameVersion> <source> <target>` goes `createNewMappingVersionFromPreviousMapping()` → `MappingVersion::create()` → this boot, so the checkpoints *are* cloned. `Copy.php` then re-points `floor_id` onto the target dungeon's floors for every relation in that array — `dungeonFloorSwitchMarkers`, `enemies`, `enemyPacks`, `enemyPatrols`, `mapIcons`, `mountableAreas`, `floorUnions`, `floorUnionAreas`, `npcEnemyForces` — but not `enemyForcesCheckpoints`.

After a cross-dungeon copy the cloned enemies land on the target floors (and correctly point at the cloned checkpoint, thanks to the new relink loop below), while the checkpoint itself keeps a `floor_id` belonging to the *source* dungeon. Concretely:

- `Floor::enemyForcesCheckpoints()` on the target dungeon returns nothing, so the pill never renders;
- `getDungeonId()` reports the source dungeon;
- `mapping:save` walks `$floor->enemyForcesCheckpointsForExport` (unscoped by mapping version), so the *source* dungeon's floor folder ends up exporting a checkpoint that belongs to the target dungeon's mapping version.

Adding `'enemyForcesCheckpoints'` to that `$relations` array should be all it needs.

app/Service/Mapping/MappingService.php
```diff
@@ -208,6 +208,10 @@ public function copyMappingVersionContentsToDungeon(
             $mountableArea->cloneForNewMappingVersion($targetMappingVersion);
         }
 
+        // Enemy Forces Checkpoints are deliberately NOT copied here. This code path (MDT import / copying a
```
Wotuu: :robot: **The reasoning here holds for one of the two call sites, not both — and the one it doesn't hold for is the one that runs routinely.**

`copyMappingVersionContentsToDungeon()` has exactly two callers:

1. `MappingVersionController@store` ("Add bare mapping version") — a deliberately enemy-less version. Skipping checkpoints there is clearly right, and that is the path this comment describes.
2. `createNewMappingVersionFromMDTMapping()` (line 80 of this file), which `MDTMappingImportService` runs on every MDT mapping-hash change. That path is **not** enemy-less: `MDTMappingImportService::importEnemies()` matches each imported MDT enemy to its predecessor via `getUniqueKey()` (npc_id + mdt_id) and explicitly carries mapper-owned fields across — `teeming`, `faction`, `required`, `skippable`, `kill_priority`, and on non-force imports `floor_id`/`lat`/`lng`.

So on that path a checkpoint would *not* arrive empty: the old→new enemy correspondence needed to preserve membership already exists and is already being used for five other mapper-owned fields. As written, every MDT package bump silently discards every enemy forces checkpoint in every retail dungeon, and mappers re-author them from scratch each time, with nothing logged.

Preserving it needs both halves: clone the checkpoints while keeping an old-id → new-id map (this method already does exactly that for `linked_dungeon_floor_switch_marker_id`), **and** add `enemy_forces_checkpoint_id`, remapped through that map, to the carry-over `$fields` list in `importEnemies()`.

Re-authoring on every MDT bump may well be acceptable for now — that's Wotuu's call. But it should be an explicit decision, and the rationale recorded here currently reads as settled while only covering call site 1.

tests/Feature/App/Model/Mapping/EnemyForcesCheckpointMappingVersionTest.php
```diff
@@ -0,0 +1,174 @@
...
+            $this->assertSame(
+                [$clonedCheckpoint->id],
+                $clonedEnemyCheckpointIds,
+                'Cloned enemies must point at the cloned checkpoint, not at the checkpoint of the previous mapping version.',
+            );
+        } finally {
+            if ($newMappingVersion !== null) {
+                Enemy::where('mapping_version_id', $newMappingVersion->id)->delete();
+                EnemyForcesCheckpoint::where('mapping_version_id', $newMappingVersion->id)->delete();
+                MappingVersion::where('id', $newMappingVersion->id)->delete();
```
Wotuu: :robot: **This leaks a whole retail dungeon's worth of cloned mapping rows into the shared test DB on every run.**

`MappingVersion::where('id', ...)->delete()` is a query-builder delete, so `MappingVersion::boot()`'s `static::deleting` cascade never fires. The two manual deletes above cover enemies and checkpoints — but the `created` hook also cloned `enemyPacks`, `enemyPatrols` (plus their polylines), `mapIcons`, `mountableAreas`, `floorUnions`, `floorUnionAreas` and `npcEnemyForces` into `$newMappingVersion`. Those rows survive, and there is no `RefreshDatabase` here, so they accumulate every time the test runs.

The sibling `DungeonRouteEnemyRaidMarkerMappingVersionTest` already does the right thing (`$newMV->delete();`). Replacing all three lines with:

```php
$newMappingVersion->delete();
```

lets the cascade clean everything up — including the checkpoints, thanks to the `enemyForcesCheckpoints()->delete()` this PR adds to that `deleting` chain, which incidentally gives that new line a bit of coverage here too.

resources/assets/js/custom/models/enemyforcescheckpoint.js
```diff
@@ -0,0 +1,336 @@
...
+            return;
+        }
+
+        let lat = 0, lng = 0;
+        for (let index in latLngs) {
+            lat += latLngs[index].lat;
+            lng += latLngs[index].lng;
+        }
+
+        this._satelliteLayerGroup = L.layerGroup().addTo(this.map.leafletMap);
```
Wotuu, line 413: :robot: **The satellite pill (and the admin connection lines) escape the map object group, so hiding "Enemy forces checkpoints" doesn't hide them.**

This adds the layer group straight to `this.map.leafletMap` rather than to the group's `layerGroup` / through `setMapObjectVisibility()`. `MapObjectGroup.setVisibility(false)` → `_updateVisibility(false)` → `setMapObjectVisibility(mapObject, false)` only ever removes `mapObject.layer`, so the satellite is untouched.

`enemyforcescheckpoint` is user-toggleable — `isUserToggleable()` defaults to true, and this PR adds `js.enemyforcescheckpoint_map_object_group_label`, which is precisely what `_setupMapObjectGroupVisibility()` renders into the Map Elements dropdown. And because `refreshPill()` is driven by `getState()`'s `floorid:changed` regardless of group visibility, a satellite is *re-created* on every floor switch while the group is toggled off. The choice is persisted in the `hidden_map_object_groups` cookie, so a user who hides the group once keeps getting orphan pills on every subsequent page load, with no way to get rid of them.

`AdminEnemyForcesCheckpoint.enemyConnectionsLayerGroup` has the same shape (`addTo(this.map.leafletMap)`) and the same problem on the admin mapping page.

Probably simplest to gate both on the owning group's `isShown()` and re-run `refreshPill()` on the group's `visibility:changed` signal.

resources/assets/js/custom/mapstate/enemyselection/enemyselection.js
```diff
@@ -7,6 +7,10 @@ class EnemySelection extends MapObjectMapState {
         return 'EnemySelection';
     }
 
+    disablesTooltips() {
```
Wotuu: :robot: **This reaches further than the admin page — it also silences enemy tooltips during ordinary pull-building in the route editor.**

`EditKillZoneEnemySelection` and `SelectKillZoneEnemySelectionOverpull` both extend `EnemySelection`, so they inherit this. Before this PR, `Enemy` only dropped its tooltip for `EditMapState` (`setTooltipEnabled(!(newMapState instanceof EditMapState))`), so hovering an enemy while picking it for a pull still showed npc name / health / enemy forces — the numbers a router uses to decide *whether* to click it.

The PR body flags the scope change on the admin mapping page's `EditMapState` and offers to narrow it, but doesn't mention this one, which lands on the main non-admin route editor. Worth confirming that "suppressed everywhere" was meant to include it. If not, the narrow version is to move `disablesTooltips() { return true; }` down onto the selection subclasses that actually want it (`EnemyPatrolEnemySelection`, `MDTEnemySelection`, `EnemyForcesCheckpointEnemySelection`) rather than the shared base.

(Unrelated nit while here: `resources/assets/js/custom/dungeonmap.js` picked up a stray blank line at the `AdminPanelControls` push in `_getMapControls()`.)

Wotuu:
:robot: Cold review (opus): 5 findings posted. Re-reviewed since the diff has grown substantially since the last round (full feature pivot to EnemyForcesCheckpoint).

app/Models/Mapping/MappingVersion.php
Wotuu: :robot: Fixed in `c41617e90` - added `'enemyForcesCheckpoints'` to `Copy.php`'s `$relations`, right after `'mountableAreas'` so it matches `boot()`'s ordering. `EnemyForcesCheckpoint` carries a plain `floor_id` and no second floor reference, so the generic `$attributes` path handles it with no special-casing (unlike `DungeonFloorSwitchMarker`/`FloorUnion`), and `MappingVersion` already has the relation so `->load($relations)` picks it up.

You were right that this is the worse failure mode of the two: not "checkpoint missing on the target" but "checkpoint present on the target while `mapping:save` exports it into the *source* dungeon's floor folder".

No test - there is no coverage for `mapping:copy` at all today, and standing one up means a real cross-dungeon copy against the shared dev DB. Instead I recorded the trap in the `mapping-versioned-models` skill, which previously only counted the three lists inside `MappingVersion::boot()` and the partial one in `copyMappingVersionContentsToDungeon()`. It now names `Copy.php`'s `$relations` as the fourth, and spells out why a model missing from it is worse off than one that was never cloned.

app/Service/Mapping/MappingService.php
Wotuu: :robot: You are right that the comment overclaimed, and it is rewritten in `c41617e90`. I did **not** implement the carry-over - reasoning below, and it is now tracked in #3702 rather than left implicit.

**Why the comment was wrong:** exactly as you describe. It justified skipping checkpoints with "this path doesn't copy enemies either", which is true for `MappingVersionController@store` and false for `createNewMappingVersionFromMDTMapping()`. The new comment names both call sites, states that the MDT path discards every checkpoint on every bump, and notes that the old to new enemy correspondence needed to preserve membership already exists in `importEnemies()` and is already used for five other mapper-owned fields.

**Why I left the behaviour alone (your call to overrule):**

1. **Nothing is lost today.** `enemy_forces_checkpoints` ships in this PR (`2026_07_27_000000_create_enemy_forces_checkpoints_table.php`), so there are zero checkpoints anywhere. This is a decision about future cost, not a live bug - which makes it a bad thing for me to settle unilaterally in a PR that has already been through three review rounds.
2. **It is not a one-liner.** `copyMappingVersionContentsToDungeon()` returns only a `MappingVersion`, so the id map has to cross a service boundary - and the method must keep *not* cloning checkpoints for call site 1, so it cannot simply always clone. The cheaper shape is to do the checkpoint cloning inside `importEnemies()` itself, which already holds both mapping versions; that is written up in the issue.
3. **There are real semantics to decide, not just plumbing.** An MDT bump is precisely when the enemy set changes. A checkpoint whose members all disappeared upstream would be carried over as an empty checkpoint - the exact failure the original comment was written to avoid - so "preserve membership" needs a companion rule for what to do with checkpoints that end up empty. And on `$forceImport = true`, `floor_id`/`lat`/`lng` are *not* carried over, so preserved membership can leave the anchor on a floor with no members (degrades to a satellite pill, not a crash).
4. **No harness exists.** `MDTMappingImportService` has no test coverage at all - the `MDTImportString*` tests cover the import/export *string* services. Shipping this untested into the path that rebuilds every retail dungeon's mapping is not a trade I would make without you asking for it.

All four points are in #3702 so none of it has to be re-derived. If you want it in this PR instead, say so and I will implement it as sketched.

tests/Feature/App/Model/Mapping/EnemyForcesCheckpointMappingVersionTest.php
Wotuu: :robot: Fixed in `c41617e90` - the three lines are now a single `$newMappingVersion->delete();`, matching `DungeonRouteEnemyRaidMarkerMappingVersionTest`. The second half of the `finally` (releasing and deleting the *original* checkpoint) is kept, since that one belongs to the pre-existing mapping version and no cascade touches it.

I measured the leak rather than just re-running the test. Row counts on the shared phpunit DB before and after one run of the **unfixed** file:

| table | delta |
|---|---|
| `enemy_packs` | +46 |
| `enemy_patrols` | +18 |
| `polylines` | +36 |
| `map_icons` | +1 |
| `mountable_areas` | +1 |
| `npc_enemy_forces` | +19 |

121 orphan rows per run, exactly the set you predicted. `mapping_versions`, `enemies` and `enemy_forces_checkpoints` were flat, since those three were being cleaned manually.

After the fix, two consecutive runs of the file: **every one of those counts identical, zero delta**. And as you noted, the cascade path now also exercises the `enemyForcesCheckpoints()->delete()` this PR adds to that `deleting` chain.

resources/assets/js/custom/models/enemyforcescheckpoint.js
Wotuu: :robot: Fixed in `c41617e90`, both the satellite pill and `AdminEnemyForcesCheckpoint.enemyConnectionsLayerGroup`.

`EnemyForcesCheckpoint` gained `getMapObjectGroup()` and `isMapObjectGroupShown()` (the method you were after is `MapObjectGroup.isShown()`, `leafletMap.hasLayer(this.layerGroup)`). `_refreshSatellitePill()` now removes any existing satellite and then bails when the group is hidden; `redrawConnectionsToEnemies()` got the same check next to its existing `shouldBeVisible()` one, since `MapObject.shouldBeVisible()` only covers floor/teeming/seasonal and knows nothing about the group toggle. The constructor registers on the group's `visibility:changed` and re-runs `refreshPill()` (which the admin subclass overrides to also redraw the lines), with a matching `unregister` in `cleanup()`.

I checked the load-order risk before relying on `isShown()` - if the group's `layerGroup` were not attached yet when the constructor's `map:mapobjectgroupsloaded` handler fires, gating on it would have silently broken first paint. It is attached: `groupIsShown` is `true` at first paint and the satellite draws.

Verified in headless Chrome on a real cross-floor checkpoint (anchored on Throne of the Tides floor 2, members on floor 1), driving the actual Map Elements dropdown entry rather than calling `setVisibility()`:

| step | `groupIsShown` | satellite on map |
|---|---|---|
| first paint | `true` | `true` |
| after hiding via Map Elements | `false` | `false` |
| floor switch away and back | `false` | `false` |
| after showing again | `true` | `true` |

And the same script against the pre-fix build, to confirm it was not a no-op: with `groupIsShown: false` the satellite was still `satelliteOnMap: true`, and survived the floor switch - the orphan pill you described.

Admin connection lines, on the admin mapping page: group shown gives `layerGroup` set and on the map; group hidden gives nothing drawn.

group shown
group hidden

Regression coverage in `resources/assets/js/custom/models/enemyforcescheckpoint.test.js` (5 tests), including the "hidden after being shown" case that pins the removal of an already-drawn satellite.

resources/assets/js/custom/mapstate/enemyselection/enemyselection.js
Wotuu: :robot: Good catch - "suppressed everywhere" was not intended to include pull-building, and narrowing is exactly right. Fixed in `c41617e90`: `disablesTooltips()` moved off the `EnemySelection` base onto `EnemyPatrolEnemySelection`, `MDTEnemySelection` and `EnemyForcesCheckpointEnemySelection` only. `EditKillZoneEnemySelection`, `SelectKillZoneEnemySelectionOverpull`, `DungeonRouteSearchEnemySelection` and `ViewKillZoneEnemySelection` now fall through to `MapState`'s `false` again. Left a note on the base saying why it must stay unoverridden.

This cannot bring back the crash the suppression originally worked around - that was the unbind/rebind approach, replaced by the `map_enemy_tooltips_suppressed` CSS class in `73dfeed6d`, and this only changes *which states* get the class.

Verified in headless Chrome, asserting on `.leaflet-container.classList.contains('map_enemy_tooltips_suppressed')`:

- **Non-admin route editor**, clicking `#killzones_new_pull` the way a user does: map state is `EditKillZoneEnemySelection`, class **absent**, and the enemies still carry their `enemy_tooltip` content. This is the path that regressed.
- **Admin mapping page**, checkpoint "Select enemies": class **present**.
- **Admin mapping page**, MDT enemy selection: class **present**.
- Clearing the map state: class **absent** again.
- No page errors on either page.

Also read straight off the prototypes in the live page, which is the whole invariant in one table:

| class | `disablesTooltips()` |
|---|---|
| `EnemySelection` | `false` |
| `EditKillZoneEnemySelection` | `false` |
| `SelectKillZoneEnemySelectionOverpull` | `false` |
| `DungeonRouteSearchEnemySelection` | `false` |
| `ViewKillZoneEnemySelection` | `false` |
| `EnemyPatrolEnemySelection` | `true` |
| `MDTEnemySelection` | `true` |
| `EnemyForcesCheckpointEnemySelection` | `true` |

Pinned in `resources/assets/js/custom/mapstate/enemyselection/enemyselection.tooltips.test.js` (6 tests): the base must not own a `disablesTooltips`, the three admin states must, and the two pull-building states must return `false`. It loads the real `MapState`/`MapObjectMapState` chain rather than stubbing the default, so it fails if the override creeps back up the hierarchy.

Stray blank line in `dungeonmap.js`'s `_getMapControls()` removed too.

resources/assets/css/sections/map.css
```diff
@@ -996,6 +1003,38 @@
...
+   default icon box and center the pill on its anchor point. */
+.map_enemy_forces_checkpoint_pill_icon {
+    width: auto !important;
+    height: auto !important;
+    margin: 0 !important;
+}
+
+.map_enemy_forces_checkpoint_pill_icon .map_enemy_forces_checkpoint_pill {
+    transform: translate(-50%, -50%);
+}
```
Wotuu: This causes the label to be misaligned with the border that you get when you hit Edit mode.

With this transform disabled:

Image

With it enabled:

Image

I understand why you did this but you need to find a different way so the yellow dotted line is aligned properly AND the label is aligned.

resources/assets/js/custom/admin/adminenemyforcescheckpoint.js
```diff
@@ -0,0 +1,145 @@
...
+     * Removes any existing UI connections to enemies.
+     */
+    removeExistingConnectionsToEnemies() {
+        console.assert(this instanceof AdminEnemyForcesCheckpoint, 'this is not an AdminEnemyForcesCheckpoint', this);
+
+        if (this.enemyConnectionsLayerGroup !== null) {
+            this.map.leafletMap.removeLayer(this.enemyConnectionsLayerGroup);
+            this.enemyConnectionsLayerGroup = null;
+        }
+    }
```
Wotuu, line 97: Correct me if I'm wrong - but this is a direct copy paste from AdminEnemyPatrol? If so - can we put this in a base class or in a separate class? I don't want duplicate code if we can avoid it.

resources/assets/js/custom/enemyvisuals/enemyvisual.js
```diff
@@ -406,7 +406,7 @@ class EnemyVisual extends Signalable {
 
             let isDeletable = mapState instanceof DeleteMapState && this.enemy.isDeletable();
             let isSelectable = (
-                (mapState instanceof MDTEnemySelection || mapState instanceof EnemyPatrolEnemySelection) && this.enemy.isSelectable()
+                (mapState instanceof MDTEnemySelection || mapState instanceof EnemyPatrolEnemySelection || mapState instanceof EnemyForcesCheckpointEnemySelection) && this.enemy.isSelectable()
```
Wotuu: Can we add a `drawsEnemyEditBorder(): bool` or something similar to the base EnemySelection class, and have it pass `true` for the above 3 classes? I'd like to get rid of these `instanceof` calls as much as possible as it makes it harder to maintain the site.

resources/assets/js/custom/enemyvisuals/enemyvisual.js
```diff
@@ -563,7 +563,7 @@ class EnemyVisual extends Signalable {
 
         let isDeletable = mapState instanceof DeleteMapState && this.enemy.isDeletable();
         let isSelectable = (
-            (mapState instanceof MDTEnemySelection || mapState instanceof EnemyPatrolEnemySelection) && this.enemy.isSelectable()
+            (mapState instanceof MDTEnemySelection || mapState instanceof EnemyPatrolEnemySelection || mapState instanceof EnemyForcesCheckpointEnemySelection) && this.enemy.isSelectable()
```
Wotuu: This is what I mean - now you gotta place it twice in random places in the code. That function would fix this.

resources/assets/js/custom/mapobjectgroups/mapobjectgroupmanager.js
Wotuu, line 71: Not your fault - but can we find a way to make this better? Idk, this long IF chain does not sit right with me. Either fix it or make a follow-up issue.

Wotuu:
:robot: Pushed `77a10adcd` (docs only, no code): rewrote the checklist in the `mapping-versioned-models` skill into the complete end-to-end flow for adding a new mapping-versioned model — schema/clone boot/`mapping:copy`, the `mapping:save` + seeder round-trip, the map-context payload, the ajax editor, and the previously-undocumented Leaflet front-end wiring (`constants.js`, map object group registration, `custom-scripts-order.mjs`, translation keys, and the automatic sidebar show/hide dropdown), using `EnemyForcesCheckpoint` from this PR as the worked example.

resources/assets/js/leaflet-tooltip-focus-fix.js
```diff
@@ -0,0 +1,61 @@
+// ---------------------------------------------------------------------------
```
Wotuu, line 1: This is something we ran into. We now have a local patch. Can we create an upstream issue + merge request so this can reach the Leaflet creators?

Ensure no recent issue was created with the same problem description so we don't spam their repository.

Wotuu: When you do this, clarify you're an AI agent posting on behalf of me, and link back to this very thread. I prefer full transparency - if they don't accept AI content then they can close it and that's the end of it.

Wotuu (commented):
A bunch of things need to be fixed.

resources/assets/css/sections/map.css
Wotuu: :robot: Fixed in `6fc6a5898`, with the transform gone entirely. The root cause was that the transform moved only the *painted* pill while the marker root element - the thing leaflet.draw puts the dashed border on - stayed at the anchor's top-left, so border and label could never both be right.

The pill's divIcon now carries an explicit **measured `iconSize` + centered `iconAnchor`** (`EnemyForcesCheckpoint._buildPillIcon()`): the markup is rendered off-screen, measured, and Leaflet itself centers the marker through its normal inline-margin mechanism. That also survives element re-creation (visibility toggles, floor switches re-build the icon from its options - a post-hoc inline-margin approach I tried first got wiped there), and it stays compatible with leaflet.draw's `_offsetMarker()`, which shifts those same margins by the border width when the edit border appears.

Two traps worth noting in the measurement: it has to happen **inside the leaflet container** (the pill's font-size is em-based and leaflet.css sets its own base font size, so a `document.body` measurement came out ~20% wide - caught in verification, not by reading), and the icon box also gets flex centering so a ±1px measurement drift stays symmetric.

Measured in headless Chrome on the admin mapping page: pill center == anchor **exactly** (also after a hide/show re-add), and in edit mode the border box is precisely pill + 4px per side with the pill center 0.5px from the anchor:

edit mode border hugging the label

resources/assets/js/custom/admin/adminenemyforcescheckpoint.js
Wotuu: :robot: Correct - it was a near-copy of `AdminEnemyPatrol` (same idea, small drift: patrol nests a second layer group and attaches to its map object group's `layerGroup`, checkpoint attached to the leaflet map directly). Fixed in `6fc6a5898`: the mechanical part - layer-group lifecycle, the polyline loop, the z-index, remove/redraw - now lives in one new class, `resources/assets/js/custom/admin/adminenemyconnections.js`, and both owners only supply what genuinely differs: the parent to attach to, the center latlng, and the member latlngs. Patrol's double-nested layer group is flattened away in the process; its lines still attach to the patrol group's own `layerGroup` (so the Map Elements toggle keeps covering them for free) while the checkpoint keeps its explicit `isMapObjectGroupShown()` gate.

Verified in headless Chrome on the admin mapping page: the checkpoint drew its 3 member lines and 4 of 7 patrols had connection lines through the shared class, zero page errors.

resources/assets/js/custom/enemyvisuals/enemyvisual.js
Wotuu: :robot: Done in `6fc6a5898` - `drawsEnemyEditBorder()` lives on the `EnemySelection` base returning `false`, with `true` overrides on exactly the three admin selection states (`EnemyPatrolEnemySelection`, `MDTEnemySelection`, `EnemyForcesCheckpointEnemySelection`). Both `enemyvisual.js` sites are now `mapState instanceof EnemySelection && mapState.drawsEnemyEditBorder() && this.enemy.isSelectable()`. New vitest cases in `enemyselection.tooltips.test.js` pin the whole invariant: base `false`, the three admin states `true`, and the pull-building states (`EditKillZoneEnemySelection`, `SelectKillZoneEnemySelectionOverpull`) `false` - so the border can't silently creep onto route-editor selections the way `disablesTooltips()` once did.

Verified live: during checkpoint enemy selection on the admin mapping page, 124 enemies carried the dashed border through the new method, zero page errors.

One related chain I deliberately did **not** touch: `models/icon.js:200` has `mapState instanceof MDTEnemySelection || mapState instanceof EnemyPatrolEnemySelection` - similar shape, but different membership (no checkpoint state) and different semantics (it switches *map icons* to their edit-mode look, not the enemy border). Folding it into `drawsEnemyEditBorder()` would change behavior for checkpoint selection, so that's your call - happy to do it as a follow-up if you want map icons consistent across all three selection modes.

resources/assets/js/custom/enemyvisuals/enemyvisual.js
Wotuu: :robot: Both call sites (`_buildVisual` and `refreshSize`) now go through `mapState.drawsEnemyEditBorder()` - the chain exists nowhere anymore. Fixed in `6fc6a5898`, see the reply above for the full picture.

resources/assets/js/custom/mapobjectgroups/mapobjectgroupmanager.js
Wotuu: :robot: Fixed it in `6fc6a5898` rather than deferring - the chain is now a name → factory map inside `_createMapObjectGroup()`:

```js
let factories = {
    [MAP_OBJECT_GROUP_ENEMY]: () => new EnemyMapObjectGroup(this, isMapAdmin),
    [MAP_OBJECT_GROUP_PATH]: () => new PathMapObjectGroup(this, !isMapAdmin),
    ...
};
let result = factories.hasOwnProperty(name) ? factories[name]() : null;
```

Built inside the method so class load order can't bite, and the editability rule that was scattered across fifteen branches is now stated once in a comment: mapping-owned groups are editable for admins, route-owned groups for everyone *but* admins, map icons always. Adding a future group is one line. The `console.assert` on unknown names is unchanged.

Verified: both the admin mapping page and the explore page construct all their groups through the new map with zero page errors.

resources/assets/js/leaflet-tooltip-focus-fix.js
Wotuu: :robot: Filed as #3712 to be run as its own task (per your instruction to handle it separately) rather than inline in this PR. The issue captures the full runbook: search Leaflet's tracker first so we don't file a duplicate, then one upstream issue + one PR with the null-guard mirroring `_openTooltip()`'s, referencing our shim and its stock-Leaflet-throws regression test as prior art. Your transparency requirements are recorded verbatim in the issue: the upstream posts must state they're written by an AI agent on behalf of you, link back to this thread, and if the Leaflet maintainers don't want AI-authored content they can simply close it. The definition of done also includes annotating `leaflet-tooltip-focus-fix.js` with the upstream links so the shim's retirement condition stays discoverable.

resources/assets/js/custom/models/enemyforcescheckpoint.js
```diff
@@ -0,0 +1,442 @@
...
+        container.innerHTML = html;
+        parent.appendChild(container);
+
+        let pill = container.firstElementChild;
+        let size = pill === null ? [0, 0] : [pill.offsetWidth, pill.offsetHeight];
+
+        parent.removeChild(container);
+
+        return size[0] > 0 ? size : null;
+    }
```
Wotuu, line 279: There's a lot of HTML and visual stuff going on in this class, can we split that off to a separate visual class instead, just like enemy.js and enemyvisual.js? That would separate the concerns a bit more.

OR if you think it's better, make a follow-up issue for a refactor to give every model the same treatment - split the logic from the visuals entirely. That'd be a big refactor but it can be worth it.

Wotuu (commented):
One thing

Wotuu:
:robot: **Enemy popup flashing while assigning enemies to a checkpoint** — fixed in `26f9a83f4`.

Two independent causes, both in `Enemy`, and `EnemyPatrolEnemySelection` dodges the second one only
by accident (it stops its map state on the first click, so nothing happens *during* a selection).

**1. The unbind on selection start never ran.** `setPopupEnabled(false)` was guarded on
`this.isPopupEnabled`, but that flag records what was last *asked for*, not what is bound:
`save()`/`setSynced()` bind a popup by calling `_assignPopup()` straight through and never touch it.
So a freshly loaded enemy sits there with a bound popup while the flag still reads `false`, the
guard short-circuits, and the very first enemy click of a selection opens the edit popup — which
the save's own `closePopup()` shuts again a moment later. That is the flash. Disabling is now
unconditional.

**2. Every save during the selection re-bound it.** Each click saves the clicked enemy *and* each of
its pack buddies; each save runs `setSynced(true)` → `_assignPopup()`, bypassing `setPopupEnabled()`
entirely. `Enemy._assignPopup()` now unbinds and returns while a map state is active, so a save
cannot resurrect the popup. This is the half the patrol flow never hit.

**Verified in headless Chrome on the admin mapping page** (real login, checkpoint placed, *Select
enemies*, real mouse clicks on a 3-enemy pack), counting Leaflet `popupopen` events:

| | before | after |
|---|---|---|
| enemy popups opened during the selection (2 clicks) | **2** (one per click) | **0** |
| enemy popup bound after starting the selection | yes | no |
| enemy popup bound after a save mid-selection | yes | no |
| enemy popup bound again after leaving the selection | yes | yes |
| plain click reopens the popup afterwards | yes | yes |
| page errors | 0 | 0 |

Also A/B'd the **patrol** flow on both builds: identical before and after (one popup open on the
assigning click, which is the normal post-selection edit popup — the state has already ended), so
nothing regressed there.

Vitest 300 green (4 new tests in `enemy.test.js` pinning both halves, including that the popup comes
back after the state ends — a too-broad guard would otherwise kill admin enemy popups permanently).
No PHP touched. Verification data (checkpoint + throwaway admin) cleaned up afterwards.

resources/assets/js/custom/models/enemyforcescheckpoint.js
Wotuu: :robot: Filed as #3716 to be run as its own task, per your "OR" option - a model/visual split is a real refactor worth its own review pass, and starting narrow (this class only, mirroring the Enemy/EnemyVisual pattern) rather than bundling it into this PR keeps this diff's own review scope from ballooning further. The issue leaves the door open for a maintainer to expand it to every model afterwards if that's still wanted.

Issue #170, opened 2019-05-19
Handle authorization better through Policies


Wotuu:
Something I've been meaning to revisit. Devise a way to handle  authorization properly on a bunch of pages using Policies.

PR #3663, opened 2026-07-25
#170 Add missing authorization gates to Ajax write endpoints

Wotuu:
:robot: Part of #170.

While auditing every controller for issue #170 (move authorization into policies, move
non-authorization out) I found a handful of write endpoints that never authorize at all. Those are
split out here as their own small MR so they can be reviewed and merged quickly, ahead of the larger
refactor.

## What this fixes

**`AjaxKillZoneController::store()` — missing `edit` gate when creating a kill zone**

`saveKillZone()` only re-authorizes when `$killZone->dungeonRoute !== null`, i.e. when *updating* an
existing kill zone. Creating a new one goes through `findOrNew(null)`, so `$killZone->dungeonRoute`
is null, the `edit` gate is skipped, and the only remaining check is `addKillZone` — which just
counts existing kill zones against the configured limit and ignores the user entirely.

The sibling controllers already get this right (`AjaxArrowController:46-47`,
`AjaxBrushlineController`, `AjaxPathController` all do `edit` then `addX`). KillZone now matches
them. The gate sits *outside* the surrounding try/catch on purpose.

**`AjaxKillZoneController` — `catch (Exception)` was catching `Mockery\Exception`**

The file imported `Mockery\Exception`, so `throw new Exception('Unable to save kill zone!')` and the
four `catch (Exception)` clauses all referred to the Mockery class rather than the global one. Two
consequences:

- `mockery/mockery` is in `require-dev`, so **in production the class does not exist** and that
  `throw` raised `Error: Class "Mockery\Exception" not found` — a 500 where a 404 was intended. The
  bug is invisible in tests, where Mockery *is* installed.
- Now that the import is corrected to the global `Exception`, the catches would also swallow
  `AuthorizationException` into a 404. Added an explicit rethrow ahead of them so a 403 stays a 403.

`AjaxPathController` and `app/Models/Dungeon.php` have the same bad import. I deliberately left them
alone: in `AjaxPathController` the `catch` sits inside a `DB::transaction` closure, so making it
actually fire would switch that path from "roll back, 500" to "commit partial write, 404" — a
behaviour change that deserves its own think. Filing a follow-up issue.

**`AjaxLiveSessionController::delete()` — no authorization whatsoever**

Any user holding the `user` role could end anyone's live session. Now authorizes `edit` against the
live session's *own* route rather than the one in the URL, so passing an unrelated route you do own
cannot be used to authorize the call.

**`AjaxMetricController::storeDungeonRoute()` — no gate**

The `@throws AuthorizationException` docblock was on the sibling `store()`, not this method. Now
authorizes `view`.

**`AjaxEchoController` — deleted**

`members()` was `dd($dungeonRoute);` — a debug leftover reachable at
`GET /ajax/echo/{dungeonRoute}/members`. Nothing can be consuming a `dd()` response, so the
controller and its route are removed rather than implemented.

## Tests

7 tests in `AjaxKillZoneControllerTest` (3 new), plus new
`AjaxLiveSessionControllerTest` and `AjaxMetricControllerTest`.

Each new deny-case was verified to actually fail without its fix, not merely to pass with it:

- `store_givenNewKillZoneOnAnotherUsersRoute_returnsForbidden` returns 201 with the gate removed.
- `storeAll_givenKillZoneOfAnotherUsersRoute_returnsForbidden` returns 404 with the rethrow removed
  (this one also covers the cross-route hijack guard inside `saveKillZone`).

Note on test setup: `DungeonRouteFactory` defaults `expires_at` to `now()->addHours(2)`, which makes
every factory route a **sandbox** route — and sandbox routes are editable by anyone by design. Every
authorization test here therefore sets `expires_at => null`, otherwise it would pass vacuously. The
live-session tests additionally use seeded users 3 and 4 rather than `User::factory()`, because that
route sits behind `role:user|admin` and a role-less factory user would be rejected by the middleware
before the policy ever ran.

## Verification

- `php artisan test --compact --filter="AjaxKillZoneControllerTest|AjaxLiveSessionControllerTest|AjaxMetricControllerTest"` — 11 passed
- Full CI `feature-controller` shard — 79 tests, 1 failure, pre-existing and environment-only:
  `AjaxEnemyControllerTest::setRaidMarker_givenRaidMarkerName_persistsNpcIdAndMdtId` fails locally
  because `2026_07_20_000000_add_npc_id_and_mdt_id_to_dungeon_route_enemy_raid_markers_table` is
  still pending on the shared dev database. Confirmed identical on unmodified `master`; CI migrates
  from scratch so it passes there.
- `composer run fix` and `composer run analyse` — clean.


app/Http/Controllers/Ajax/AjaxLiveSessionController.php
```diff
@@ -8,19 +8,29 @@
...
+     *
+     * @throws AuthorizationException
      */
     public function delete(Request $request, DungeonRoute $dungeonRoute, LiveSession $liveSession)
     {
+        // Prefer the live session's own route over the one in the URL, so that passing an unrelated
+        // route cannot be used to authorize against. Ending a session is an edit of that route.
+        $dungeonRoute = $liveSession->dungeonRoute ?? $dungeonRoute;
+
+        Gate::authorize('edit', $dungeonRoute);
```
Wotuu: :robot: **Finding (high): this `edit` gate locks the session's own creator out of ending their session.**

Starting a live session only requires `view`, not `edit` — [`LiveSessionController::create()`](https://github.com/RaiderIO/keystone.guru/blob/cf564090747836dfd8323f5d8d2644f8da7db35b/app/Http/Controllers/LiveSessionController.php#L37-L44) does `Gate::authorize('view', $dungeonroute)` and stores `'user_id' => Auth::id()`. So any logged-in user can legitimately start a live session on someone else's published route (the UI actively offers this: the "Start" button in [`header.blade.php`](https://github.com/RaiderIO/keystone.guru/blob/cf564090747836dfd8323f5d8d2644f8da7db35b/resources/views/common/maps/controls/header.blade.php#L153-L158) renders for any authenticated map viewer, and the "Stop" button on the live-session page renders unconditionally at [L120-L127](https://github.com/RaiderIO/keystone.guru/blob/cf564090747836dfd8323f5d8d2644f8da7db35b/resources/views/common/maps/controls/header.blade.php#L120-L127)).

With this gate, that creator now gets a 403 from their own Stop button: `edit` resolves to [`mayUserEdit()`](https://github.com/RaiderIO/keystone.guru/blob/cf564090747836dfd8323f5d8d2644f8da7db35b/app/Models/DungeonRoute/DungeonRoute.php#L860-L873), which is false for a non-owner/non-collaborator — `$liveSession->user_id` is never consulted. And since `expires_at` only gets set by this very endpoint, the orphaned session never expires; nobody except the route owner/an admin can ever end it.

Suggested shape:

```php
if ($liveSession->user_id !== Auth::id()) {
    Gate::authorize('edit', $dungeonRoute);
}
```

(The new tests don't catch this because in both of them the session creator and the route owner are the same user.)

app/Http/Controllers/Ajax/AjaxMetricController.php
```diff
@@ -24,11 +25,16 @@ public function store(APIMetricFormRequest $request, MetricServiceInterface $met
...
 
+    /**
+     * @throws AuthorizationException
+     */
     public function storeDungeonRoute(
         APIDungeonRouteMetricFormRequest $request,
         DungeonRoute                     $dungeonRoute,
         MetricServiceInterface           $metricService,
     ): Response {
+        Gate::authorize('view', $dungeonRoute);
```
Wotuu, line 46: :robot: **Note (informational, pre-existing line): this gate is bypassable via the ungated sibling `store()` above.**

The generic `POST /ajax/metric` endpoint in this same controller has no gate, and [`APIMetricFormRequest`](https://github.com/RaiderIO/keystone.guru/blob/cf564090747836dfd8323f5d8d2644f8da7db35b/app/Http/Requests/Metric/APIMetricFormRequest.php#L21-L31) validates `model_class` as a free-form `'string|null'` and `model_id` as a bare int — [`MetricService::storeMetric()`](https://github.com/RaiderIO/keystone.guru/blob/cf564090747836dfd8323f5d8d2644f8da7db35b/app/Service/Metric/MetricService.php#L20-L29) then writes them straight to the `metrics` table. So anyone can still record a metric against a route they cannot view by posting `model_class=App\Models\DungeonRoute\DungeonRoute` + a (sequential, enumerable) numeric `model_id` to `/ajax/metric`, sidestepping the `view` check added here.

Not a bug in the changed line — flagging it because the MR's stated scope is "write endpoints that never authorize at all", and `store()` is one, in this same file. Worth either folding into this MR (e.g. `Rule::in()` on `model_class` + resolving the model and gating `view` on it) or explicitly listing it for the follow-up refactor.

Wotuu:

app/Http/Controllers/Ajax/AjaxLiveSessionController.php
Wotuu: :robot: Fixed in 5450b9ad8. `delete()` now skips the `edit` gate when `(int) $liveSession->user_id === (int) Auth::id()` — the session's own creator can always end it, even when they're not the route owner/collaborator. Added `delete_givenSessionCreatorIsNotRouteOwner_endsTheLiveSession`, which failed before the fix (403) and passes after.

app/Http/Controllers/Ajax/AjaxMetricController.php
Wotuu: :robot: Fixed in 5450b9ad8. `store()` now resolves the `DungeonRoute` when `model_class === DungeonRoute::class` and applies the same `view` gate `storeDungeonRoute()` uses; non-DungeonRoute metrics (null or other classes) are unaffected. Added `store_givenUnviewableRouteReportedThroughGenericEndpoint_returnsForbidden` and `store_givenNonDungeonRouteModelClass_storesTheMetricWithoutAuthorization`.

Side note: writing that first test surfaced a latent bug in `APIMetricFormRequest` — `'int|null'`/`'string|null'` aren't valid Laravel rules (there's no "null" rule, only "nullable"), so any request to the generic endpoint threw a `BadMethodCallException`. Fixed alongside since it blocked the new tests and nothing previously exercised that path directly.

app/Http/Controllers/Ajax/AjaxKillZoneController.php
```diff
@@ -230,6 +230,11 @@ public function store(
     ): KillZone {
         $dungeonRoute = $killZone?->dungeonRoute ?? $dungeonRoute; // @phpstan-ignore nullsafe.neverNull
 
+        // Outside the try/catch on purpose: an authorization failure must surface as a 403, not be
+        // rewritten into the 404 below. saveKillZone() only re-authorizes when updating an existing
+        // kill zone, so without this a new kill zone could be created on any route.
+        Gate::authorize('edit', $dungeonRoute);
```
Wotuu: You said that there's no gate - but `saveKillZone()` function has it. I'd be fine with moving it here and strip the auth code away from the saving code. But now it does things twice.

app/Http/Controllers/Ajax/AjaxLiveSessionController.php
```diff
@@ -8,19 +8,33 @@
...
     {
+        // Prefer the live session's own route over the one in the URL, so that passing an unrelated
+        // route cannot be used to authorize against. Ending a session is an edit of that route.
+        $dungeonRoute = $liveSession->dungeonRoute ?? $dungeonRoute;
+
+        // The session's own creator may always end it, even if they aren't the route's
+        // owner/collaborator - otherwise a session started by e.g. a guest viewer can never expire.
+        if ((int)$liveSession->user_id !== (int)Auth::id()) {
+            Gate::authorize('edit', $dungeonRoute);
+        }
```
Wotuu: One thing to keep in mind - I can start a live session from a route that I did not create. I think this gating may be wrong as a result. We should gate based on who started the live session and see if they own the live session. If they do, great, otherwise, block the deletion. I don't think dungeonRoutes have anything to do with this.

Wotuu (commented):
2 comments

app/Http/Controllers/Ajax/AjaxLiveSessionController.php
Wotuu: :robot: Agreed, done in 75f1fd029 — added `LiveSessionPolicy::end()` that checks only `$liveSession->user_id === Auth::id()` (or admin), and the gate no longer resolves or references `$dungeonRoute` at all. One implementation note: I had to keep the `DungeonRoute $dungeonRoute` route parameter in the controller signature (just unused now) — dropping it entirely broke route-model-binding, since `CallableDispatcher`'s positional parameter resolution then handed the route's raw `dungeonRoute` segment to `$liveSession` instead once implicit binding had no reason to resolve it. Added a regression test pinning the actual bug: a route owner who did *not* start the session is now denied (previously allowed), plus one confirming an admin can still end any session.

app/Http/Controllers/Ajax/AjaxKillZoneController.php
Wotuu: :robot: Traced this through before making any change, since stripping it looked safe at first glance but isn't: `store()`'s own outer gate already reassigns `$dungeonRoute = $killZone?->dungeonRoute ?? $dungeonRoute` *before* calling `Gate::authorize`, so for `store()` specifically the inner check in `saveKillZone()` is genuinely redundant - same route, checked twice.

But `storeAll()` doesn't do that per-item reassignment: it authorizes once against the URL's `$dungeonRoute` at the top, then loops `saveKillZone()` over a batch of kill zone IDs from the request body without re-resolving each one's real route first. If one of those IDs belongs to a different route than the URL one, `storeAll()`'s outer check never catches it - only `saveKillZone()`'s internal re-check (line 86, which re-resolves via `$killZone->dungeonRoute`) does, and that's exactly the hijack case its own comment describes.

So I'd rather not strip it as-is - that would silently reopen the batch-endpoint hijack. Happy to do the bigger refactor (have `storeAll()` resolve+gate each kill zone's real route before calling `saveKillZone()`, then drop the internal check for good) if you'd like it as a follow-up, since it touches the batch-update loop's structure rather than being a one-line change - didn't want to make that call unprompted in the same pass as everything else here.

Wotuu: Ok yeah I'd still love this auth gate being done BEFORE the `store` functions get called. Let's make clear we validate first, and then cleanly store the result. Good calls.

Wotuu: :robot: Done in fb10f4e95 — pulled the ownership check out of `saveKillZone()` entirely into a new `authorizeKillZoneEdit()`, which resolves the kill zone's real route and authorizes it once, before `saveKillZone()` (or any storage side effect) runs. `store()` calls it directly; `storeAll()`'s batch loop calls it once per item so a hijacked kill zone id inside a batch is still caught per-item, not just against the URL's own route. `saveKillZone()`'s race-condition fallback (a concurrent request creating the kill zone between `findOrNew` and `create`) keeps its own re-authorize, since that one can only be resolved after the write itself. All 17 killzone tests still pass, including the existing hijack regression test (updated its stale comment to match).

Wotuu (commented):
Ok

PR #3665, opened 2026-07-25
#170 Remove dead and duplicated policy code, route views through the Gate

Wotuu:
:robot: Part of #170. Independent of #3663 — either can merge first.

Cleanup pass over `app/Policies/`, with **no intended behaviour change**. Everything here is either
a bug in code that never fires, dead code, or a view that duplicates a policy instead of asking it.

## Fixes

**`DungeonRoutePolicy::rate()` discarded the user the Gate gave it**

```php
public function rate(User $user, DungeonRoute $dungeonroute): bool
{
    return !$dungeonroute->isOwnedByUser();   // no argument
}
```

`isOwnedByUser()` defaults to `Auth::user()` when called with no argument, so the `$user` parameter
was ignored. It happens to be correct for a normal session request and wrong for anything else
(`Gate::forUser(...)`, a queued job, a second user in the same request). Now passes `$user`.

New test `rate_givenNonOwnerWhileOwnerIsAuthenticated_returnsAllowed` pins it — verified it returns
false (i.e. fails) against the old one-argument call.

## Deletions

**`TagPolicy::createTag`** — byte-identical to `TagCategoryPolicy::createTag` and **unreachable**.
The only call site is `Gate::authorize('create-tag', [$tagCategory, $model])`, and the Gate resolves
the policy from the *first* array element, so `TagCategoryPolicy` always wins.

Added `TagCategoryPolicyTest` before deleting, driving the ability through `$user->can('create-tag', …)`
rather than instantiating the policy — so it would catch the resolution silently moving elsewhere.

**`DungeonRoutePolicy::unpublish` and `::restore`** — zero call sites in `app/`, `routes/`,
`resources/` or `tests/`.

**`DungeonRoutePolicy::favorite` and both `Gate::authorize('favorite', …)` call sites**, removed
together. It returned unconditional `true`. Its only real effect was the non-nullable `User $user`
type hint denying guests — which the route's `['auth', 'role:user|admin']` middleware
(`routes/web.php:706`) already guarantees. Removing the method without the call sites would have
broken favouriting outright, so both go in the same commit.

**`LiveSessionPolicy::view` docblock** said "Determine whether the user can create a tag."

## Views now ask the policy

The view layer contained **zero** `@can` directives. Four map sidebar templates re-implemented
authorization by calling the model directly:

| File | Was | Now |
|---|---|---|
| `maps/controls/view.blade.php` | `@if($dungeonroute->mayUserEdit(Auth::user()))` | `@can('edit', $dungeonroute)` |
| `maps/controls/present.blade.php` | same | same |
| `maps/controls/elements/rating.blade.php` | `@if($dungeonroute->isOwnedByUser())` | `@cannot('rate', $dungeonroute)` |
| `maps/controls/header.blade.php` | `$dungeonroute?->mayUserEdit($user) ?? false` | `Gate::allows('edit', $dungeonroute)` |
| `maps/controls/header.blade.php` | `$dungeonroute->team->isUserMember(Auth::user())` | `@can('edit', $dungeonroute->team)` |

Each maps onto the policy method the corresponding endpoint already gates on, so the button and the
endpoint behind it can no longer disagree.

Guest behaviour is preserved in every case: `DungeonRoutePolicy::edit` takes `?User`, so a guest
still resolves to `mayUserEdit(null)` → `isSandbox()`; `TeamPolicy::edit` takes a non-nullable
`User`, so a guest is denied exactly as `isUserMember(null)` returned false.

## Verification

**Tests** — new `DungeonRouteViewSidebarTest` renders the real Blade and asserts the edit button and
rating dropdown appear for the owner and not for a non-owner (both directions, so an inverted or
always-true directive fails). It uses `User::factory()->create()` throughout, per this repo's
"never use a seeded user ID above 3" testing convention.

- `--group=Policy` — 26 passed
- `--filter=DungeonRouteViewSidebarTest` — 4 passed
- CI `feature-controller` shard — 79 tests, 1 failure, pre-existing and environment-only

**Visual** — A/B'd the map view page guest-side against the master stack in headless Chrome. The
authorization-driven controls are all inside `@auth`, so the decisive check is DOM measurement
rather than the screenshots; master and branch return identical values on both the changed route
page and an untouched homepage control:

```
MASTER  /route/…  {"navH":94,"teamLinks":0,"editLinks":0,"ratingDropdown":false}
BRANCH  /route/…  {"navH":94,"teamLinks":0,"editLinks":0,"ratingDropdown":false}
MASTER  /         {"navH":96,…}
BRANCH  /         {"navH":96,…}
```

(The two full-page screenshots differ only in nav font paint timing — `navH` is identical on both,
so it is not a layout change.) No page errors on either side; the one console error
(`SyntaxError: "undefined" is not valid JSON`) is present on master too.

## Deliberately not done

- **`DungeonRoutePolicy::publish`'s first guard is inert.** It denies unless
  `hasKilledAllRequiredEnemies()`, whose body is entirely commented out (`DungeonRoute.php:1122`) and
  which therefore always returns `true`. Whether to restore the check or drop it is a product call,
  not a cleanup — left untouched, filed separately.
- **Ability-name casing** (`schedule-publish` kebab vs `addKillZone` camel; Laravel's
  `formatAbilityToMethod` accepts both). All five camelCase abilities are the `add*` quota checks,
  which leave `DungeonRoutePolicy` entirely in the next MR of this series — renaming them here would
  be churn.

### DIFF
--- a/tests/Feature/Controller/DungeonRoute/DungeonRouteViewSidebarTest.php
+++ b/tests/Feature/Controller/DungeonRoute/DungeonRouteViewSidebarTest.php
@@ -1,0 +1,142 @@
+<?php
+
+namespace Tests\Feature\Controller\DungeonRoute;
+
+use App\Models\DungeonRoute\DungeonRoute;
+use App\Models\PublishedState;
+use App\Models\User;
+use PHPUnit\Framework\Attributes\Group;
+use PHPUnit\Framework\Attributes\Test;
+use Tests\TestCases\PublicTestCase;
+
+/**
+ * Covers the map sidebar's authorization-driven controls, which moved from direct model calls
+ * ($dungeonroute->mayUserEdit(Auth::user()), ->isOwnedByUser()) to @can/@cannot directives.
+ * Rendering the real Blade is the only way to catch a directive typo or an inverted condition.
+ */
+#[Group('Controller')]
+#[Group('DungeonRoute')]
+final class DungeonRouteViewSidebarTest extends PublicTestCase
+{
+    /**
+     * Users 3 and 4 are seeded with the "user" role and are not admins - an admin may edit every
+     * route, which would make the deny case pass vacuously.
+     */
+    private const int ROUTE_OWNER_USER_ID = 3;
+
+    private const int OTHER_USER_ID = 4;
+
+    #[Test]
+    public function view_givenOwner_rendersTheEditButton(): void
+    {
+        // Arrange
+        $route = $this->createRoute();
+
+        try {
+            $this->be(User::findOrFail(self::ROUTE_OWNER_USER_ID));
+
+            // Act
+            $response = $this->followingRedirects()->get($this->viewUrl($route));
+
+            // Assert
+            $response->assertOk();
+            $response->assertSee(route('dungeonroute.edit', [
+                'dungeon'      => $route->dungeon,
+                'dungeonroute' => $route,
+                'title'        => $route->getTitleSlug(),
+            ]), false);
+        } finally {
+            $route->delete();
+        }
+    }
+
+    #[Test]
+    public function view_givenNonOwner_hidesTheEditButton(): void
+    {
+        // Arrange
+        $route = $this->createRoute();
+
+        try {
+            $this->be(User::findOrFail(self::OTHER_USER_ID));
+
+            // Act
+            $response = $this->followingRedirects()->get($this->viewUrl($route));
+
+            // Assert
+            $response->assertOk();
+            $response->assertDontSee(route('dungeonroute.edit', [
+                'dungeon'      => $route->dungeon,
+                'dungeonroute' => $route,
+                'title'        => $route->getTitleSlug(),
+            ]), false);
+        } finally {
+            $route->delete();
+        }
+    }
+
+    #[Test]
+    public function view_givenOwner_showsTheCannotRateOwnRouteNotice(): void
+    {
+        // Arrange
+        $route = $this->createRoute();
+
+        try {
+            $this->be(User::findOrFail(self::ROUTE_OWNER_USER_ID));
+
+            // Act
+            $response = $this->followingRedirects()->get($this->viewUrl($route));
+
+            // Assert
+            $response->assertOk();
+            $response->assertSee(__('view_common.maps.controls.elements.rating.unable_to_rate_own_route'));
+        } finally {
+            $route->delete();
+        }
+    }
+
+    #[Test]
+    public function view_givenNonOwner_showsTheRatingOptions(): void
+    {
+        // Arrange
+        $route = $this->createRoute();
+
+        try {
+            $this->be(User::findOrFail(self::OTHER_USER_ID));
+
+            // Act
+            $response = $this->followingRedirects()->get($this->viewUrl($route));
+
+            // Assert
+            $response->assertOk();
+            $response->assertDontSee(__('view_common.maps.controls.elements.rating.unable_to_rate_own_route'));
+            $response->assertSee(__('view_common.maps.controls.elements.rating.your_rating'));
+        } finally {
+            $route->delete();
+        }
+    }
+
+    /**
+     * A published, non-sandbox route owned by ROUTE_OWNER_USER_ID. Sandbox routes (which the
+     * factory creates by default) are editable by anyone, so expires_at must be null here.
+     */
+    private function createRoute(): DungeonRoute
+    {
+        return DungeonRoute::factory()->create([
+            'author_id'          => self::ROUTE_OWNER_USER_ID,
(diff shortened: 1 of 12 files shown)


tests/Feature/Controller/DungeonRoute/DungeonRouteViewSidebarTest.php
```diff
@@ -0,0 +1,147 @@
...
+ */
+#[Group('Controller')]
+#[Group('DungeonRoute')]
+final class DungeonRouteViewSidebarTest extends PublicTestCase
+{
+    #[Test]
+    public function view_givenOwner_rendersTheEditButton(): void
+    {
+        // Arrange
+        $owner = User::factory()->create();
```
Wotuu, line 25: :robot: Minor, docs-only: the PR description says this test "uses seeded users 3 and 4 rather than `User::factory()`, since an admin may edit every route and a factory user has no role at all" — but the committed test uses `User::factory()->create()` throughout (correctly: the view page is public, so no role is needed, as the docblock on `createRoute()` says). The description appears to be from an earlier iteration and also contradicts the "never reference a seeded id above 3" guidance this PR adds to `.claude/skills/writing-tests/SKILL.md`. Worth updating the PR body so reviewers aren't misled about the test strategy. No code change needed.

Wotuu:
:robot: Cold review of the authorization cleanup, treated as high-risk (removed "dead" policy code that is still reachable would be a security regression). Every claim in the PR body was independently re-verified against the codebase:

- **`unpublish`/`restore` deletion** — confirmed zero call sites across `app/`, `routes/`, `resources/`, `tests/`; no `authorizeResource` usage and no `Gate::define` for these abilities anywhere. Genuinely dead.
- **`favorite` removal** — the two removed `Gate::authorize('favorite', ...)` calls were the only call sites; the routes are inside the `Route::middleware(['auth', 'role:user|admin'])` group ([routes/web.php#L705-L712](https://github.com/RaiderIO/keystone.guru/blob/9c812eb9bc2358266b8052873cddef4882d50b86/routes/web.php#L705-L712)), so guest denial is preserved and the unconditional-`true` policy removed nothing else.
- **`TagPolicy::createTag` unreachability** — only call site is `Gate::authorize('create-tag', [$tagCategory, $model])` in `AjaxTagController`; the Gate resolves the policy from the first array element (TagCategory → TagCategoryPolicy), whose `createTag` is byte-identical. The remaining `TagPolicy::edit`/`delete` still use all kept imports.
- **`@can('edit', $dungeonroute)` ≡ `mayUserEdit(Auth::user())`** — `DungeonRoutePolicy::edit(?User)` delegates directly to `mayUserEdit($user)`; no `Gate::before` hook and no policy `before()` exists, so no admin-bypass drift. Guest path identical (`mayUserEdit(null)` → `isSandbox()`).
- **`@can('edit', $team)` ≡ `isUserMember()`** — `TeamPolicy::edit` is exactly `$team->isUserMember($user)`; non-nullable `User` denies guests just as `isUserMember(null)` did; the block is guarded by `instanceof Team` so no null-team issue.
- **`@cannot('rate', ...)` guest safety** — the rating element's only include site is inside `@auth` in `view.blade.php`, so the non-nullable `User $user` on `rate()` never changes guest output; for authenticated users the inversion is exact.
- **`rate()` fix** — correct, and the new `rate_givenNonOwnerWhileOwnerIsAuthenticated_returnsAllowed` test genuinely pins the old bug.
- **Skill-doc claim** (CI users 1 = admin, 2 = internal_team, 3 = user) — matches `config/laratrust_seeder.php` role order and `LaratrustSeeder`'s per-role user creation.

No code-level findings. One minor docs-only inline comment posted: the PR body's description of the sidebar test's user strategy is stale relative to the committed code.

Wotuu:

tests/Feature/Controller/DungeonRoute/DungeonRouteViewSidebarTest.php
Wotuu: :robot: Confirmed — the test uses `User::factory()->create()` throughout (verified in `tests/Feature/Controller/DungeonRoute/DungeonRouteViewSidebarTest.php`), no seeded users. The PR description was stale/wrong, not the code. Corrected the body accordingly.

Wotuu: :robot: Checked — the PR description was already updated (2026-07-26T22:21:53Z, after this review) and now correctly says the test uses `User::factory()->create()` throughout, per the "never use a seeded user ID above 3" convention, rather than the stale "seeded users 3 and 4" claim. No further change needed. Leaving this thread unresolved for Wotuu to close.

Wotuu (commented):
Ok

PR #3667, opened 2026-07-25
#170 Move inline controller authorization into policies

Wotuu:
:robot: Part of #170. Independent of #3663 and #3665 — merge in any order.

This is the "pull authorization out of controllers and into policies" half of the issue. Each change
below moves a hand-rolled ownership or role check into the policy that already governs the
surrounding action, so the check has one home and the wrong-status-code problem goes away with it.

## New `UserPolicy`

There was no `UserPolicy`, so four checks about *users* had nowhere to live and were written inline —
each returning something other than 403:

| Ability | Replaces | Was |
|---|---|---|
| `update` | `AjaxUserController::store` self-ownership check | **400** for "this is not your account" |
| `delete` | `ProfileController::delete` admin-cannot-self-delete guard | `throw new Exception` → **500** |
| `makeRole` | `UserController::makeRole` super-admin escalation guard | **silently did nothing** — the form redirected back with no message and no role change |
| `revokeAdFreeGiveaway` | `giver_user_id !== $currentUser->id`, duplicated **verbatim** in `AjaxProfileController` and `AjaxTeamController` | **422** in both |

`makeRole` is the one worth a close look. It is a privilege-escalation guard: only names listed in
`keystoneguru.super_admins` may grant or revoke the admin role. The old code wrapped the whole method
in `if ($currentUser !== null)` and the admin branch in the super-admin check, so a non-super-admin
hitting it got a redirect and no feedback. It now denies with a message. The route was already behind
`role:admin`, so this was never an escalation path — but a failure that looks like success is its own
bug.

`AjaxUserController::store` also switches `User::where(...)->first()` + manual 400 to `firstOrFail()`,
so an unknown public key is now a **404** rather than a 400.

## New `MapIconPolicy`

`AjaxMapIconController` was the heaviest inline-authorization site in the repo — admin/non-admin
branching interleaved with `Gate::authorize`, and **two branches that threw a generic `Exception`**
(i.e. answered "you may not do this" with a 500):

- `createGlobal` — creating an icon with no dungeon route means editing the mapping itself; admin only
  (was `throw new Exception('Unable to save map icon!')`)
- `update` — a mapping icon (no route, no team) stays admin-only even when the request supplies a
  route the caller may otherwise edit (was the same generic `Exception`)
- `delete` — anything not attached to a route is admin-only (was `response(null, 403)`)
- `assignToTeam` — `$team->isUserCollaborator($user)`, deciding whether `team_id` may be set

**One asymmetry preserved rather than fixed:** `delete` is stricter than `update` — a *team* icon is
updatable by a collaborator but deletable only by an admin. That is exactly what the old code did
(`!$isAdmin && $mapIcon->dungeon_route_id === null` → 403, regardless of `team_id`). It looks like a
bug to me, but it is pre-existing and behaviour-changing to fix, so it is documented in the policy
and left alone. Say the word and I'll open an issue.

## `TeamPolicy` consolidation

Three methods ran `Gate::authorize(...)` and then repeated a *finer* check inline, hand-rolling
`response('Forbidden', 403)`:

- **`changeRole`** now takes the target user and the new role, and absorbs `Team::canChangeRole()`
  (which carried a `@TODO Should this go to a Policy?` marker). The controller becomes
  `Gate::authorize('change-role', [$team, $targetUser, $role])`. That rank ordering — nobody promotes
  past their own rank, nobody demotes someone above them — is now enforced by the gate, and pinned by
  a 6-case data provider.
- **`addRoute` / `removeRoute`** dropped their inline `canAddRemoveRoute($user)` branch entirely: it
  is *exactly* what `TeamPolicy::moderateRoute` already calls, so the branch was dead code that could
  never be reached with a false value.
- `AjaxDungeonRouteController::get`'s `// @TODO Policy?` + `abort(403, 'Unauthorized')` and
  `TeamController::store`'s inline `hasRole(ROLE_ADMIN)` field guard now go through
  `TeamPolicy::edit` and `TeamPolicy::changeRoutePublishing` respectively.

## `DungeonRoutePolicy::claim`

`DungeonRouteController::claim` had **no gate at all** — the rule lived inside
`DungeonRoute::claim()`, which returns false and does nothing when the route is not a sandbox route,
after which the controller redirected to the edit page regardless. Now `Gate::authorize('claim', …)`
denies up front; `DungeonRoute::claim()` itself is untouched by this diff and still self-guards on
`isSandbox()` as defense-in-depth, it is not a plain mutator.

## What deliberately stayed in the controllers

Quota and entitlement checks — `PatreonAdFreeGiveaway::getCountLeft()`, `hasPatreonBenefit()`,
`hasAdFreeGiveaway()` — remain inline as **422**s. They answer "you have run out", not "you are not
allowed", and the codebase already uses 422 for exactly this elsewhere.

## Tests

New `UserPolicyTest` (12), `MapIconPolicyTest` (11), `TeamPolicyTest` (8, incl. the rank-ordering data
provider), plus 2 `claim` cases added to `DungeonRoutePolicyTest`. `--group=Policy` goes from 26 to
**56 tests, all passing**. `UserPolicyTest` includes a `$user->can(...)` case to prove the new policy
is actually reachable through the Gate — policies are registered nowhere in this app (no
`AuthServiceProvider`, no `Gate::policy`), so a rename would silently detach one with every test
still green.

Note for whoever writes the next team test: `Team`'s `deleting` hook walks
`members->patreonAdFreeGiveaway`, which trips `preventLazyLoading`. Both new tests eager-load that
chain before deleting.

## Verification

- `--group=Policy` — 56 passed
- CI `feature-controller` shard — 79 tests, 1 failure, pre-existing and environment-only

No UI changes, so no visual verification: every touched path is an Ajax endpoint or a redirect.

### DIFF
--- a/app/Http/Controllers/Ajax/AjaxMapIconController.php
+++ b/app/Http/Controllers/Ajax/AjaxMapIconController.php
@@ -8,7 +8,6 @@ use App\Events\Models\ModelChangedEvent;
 use App\Http\Controllers\Traits\ChangesDungeonRoute;
 use App\Http\Requests\MapIcon\MapIconFormRequest;
 use App\Models\DungeonRoute\DungeonRoute;
-use App\Models\Laratrust\Role;
 use App\Models\MapIcon;
 use App\Models\Mapping\MappingModelInterface;
 use App\Models\Mapping\MappingVersion;
@@ -24,7 +23,6 @@ use Illuminate\Http\Response;
 use Illuminate\Support\Facades\Auth;
 use Illuminate\Support\Facades\Gate;
 use Override;
-use Teapot\StatusCode;
 use Teapot\StatusCode\Http;
 use Throwable;
 
@@ -64,15 +62,9 @@ class AjaxMapIconController extends AjaxMappingModelBaseController
         $validated                     = $request->validated();
         $validated['dungeon_route_id'] = $dungeonRoute?->id;
 
-        /** @var User|null $user */
-        $user = Auth::user();
-
-        $isUserAdmin = $user?->hasRole(Role::ROLE_ADMIN);
-        // Must be an admin to use this endpoint like this!
+        // No dungeon route means this icon is part of the mapping itself - admin only
         if ($dungeonRoute === null) {
-            if (!$isUserAdmin) {
-                throw new Exception('Unable to save map icon!');
-            }
+            Gate::authorize('createGlobal', MapIcon::class);
         } // We're editing a map comment for the user, carry on
         else {
             Gate::authorize('edit', $dungeonRoute);
@@ -88,23 +80,20 @@ class AjaxMapIconController extends AjaxMappingModelBaseController
             $validated,
             MapIcon::class,
             $mapIcon,
-            function (MapIcon $mapIcon) use ($coordinatesService, $validated, $user, $dungeonRoute, &$beforeModel) {
+            function (MapIcon $mapIcon) use ($coordinatesService, $validated, $dungeonRoute, &$beforeModel) {
                 // Set the team_id if the user has the rights to do this. May be null if not set or no rights for it.
                 $updateAttributes = [];
                 $teamId           = $validated['team_id'];
-                if ($teamId !== null) {
-                    $team = Team::find($teamId);
-                    if ($team !== null && $user !== null && $team->isUserCollaborator($user)) {
-                        $updateAttributes = [
-                            'team_id'          => $teamId,
-                            'dungeon_route_id' => null,
-                        ];
-                    }
+                if ($teamId !== null && Gate::allows('assignToTeam', [$mapIcon, Team::find($teamId)])) {
+                    $updateAttributes = [
+                        'team_id'          => $teamId,
+                        'dungeon_route_id' => null,
+                    ];
                 }
 
                 // Prevent people being able to update icons that only the admin should if they're supplying a valid dungeon route
-                if ($mapIcon->exists && $mapIcon->dungeon_route_id === null && $dungeonRoute !== null && $mapIcon->team_id === null) {
-                    throw new Exception('Unable to save map icon!');
+                if ($mapIcon->exists && $dungeonRoute !== null) {
+                    Gate::authorize('update', $mapIcon);
                 }
 
                 // The incoming lat/lngs are facade lat/lngs, save the icon on the proper floor
@@ -169,12 +158,10 @@ class AjaxMapIconController extends AjaxMappingModelBaseController
     {
         $dungeonRoute = $mapIcon->dungeonRoute;
 
-        $isAdmin = Auth::check() && Auth::user()->hasRole(Role::ROLE_ADMIN);
-        // Must be an admin to use this endpoint like this!
-        if (!$isAdmin && ($dungeonRoute === null || $mapIcon->dungeon_route_id === null)) {
-            return response(null, StatusCode::FORBIDDEN);
-        } // We're editing a map icon for the user, carry on
-        elseif ($dungeonRoute !== null) {
+        // Anything not attached to a dungeon route is part of the mapping - admin only
+        Gate::authorize('delete', $mapIcon);
+
+        if ($dungeonRoute !== null) {
             // Edit intentional; don't use delete rule because team members shouldn't be able to delete someone else's map comment
             Gate::authorize('edit', $dungeonRoute);
         }
--- a/app/Http/Controllers/Ajax/AjaxTeamController.php
+++ b/app/Http/Controllers/Ajax/AjaxTeamController.php
@@ -65,23 +65,15 @@ class AjaxTeamController extends Controller
      */
     public function changeRole(Request $request, Team $team)
     {
-        Gate::authorize('change-role', $team);
-
-        /** @var User $user */
-        $user = Auth::user();
         /** @var User $targetUser */
         $targetUser = User::where('name', $request->get('username'))->firstOrFail();
         $role       = $request->get('role');
 
-        // Only if the current user may do such a thing
-        if ($team->canChangeRole($user, $targetUser, $role)) {
-            $team->changeRole($targetUser, $role);
-            $result = response()->noContent();
-        } else {
-            $result = response('Forbidden', Http::FORBIDDEN);
-        }
+        Gate::authorize('change-role', [$team, $targetUser, $role]);
 
-        return $result;
+        $team->changeRole($targetUser, $role);
+
+        return response()->noContent();
     }
 
     /**
@@ -91,19 +83,12 @@ class AjaxTeamController extends Controller
      */
     public function addRoute(Request $request, Team $team, DungeonRoute $dungeonroute)
     {
+        // moderate-route resolves to TeamPolicy::moderateRoute, which is canAddRemoveRoute()
         Gate::authorize('moderate-route', $team);
 
-        /** @var User $user */
-        $user = Auth::user();
+        $team->addRoute($dungeonroute);
 
-        if ($team->canAddRemoveRoute($user)) {
-            $team->addRoute($dungeonroute);
-            $result = response()->noContent();
-        } else {
-            $result = response('Forbidden', Http::FORBIDDEN);
-        }
-
-        return $result;
+        return response()->noContent();
     }
 
     /**
@@ -113,19 +98,12 @@ class AjaxTeamController extends Controller
      */
     public function removeRoute(Request $request, Team $team, DungeonRoute $dungeonroute)
     {
+        // moderate-route resolves to TeamPolicy::moderateRoute, which is canAddRemoveRoute()
         Gate::authorize('moderate-route', $team);
 
-        /** @var User $user */
-        $user = Auth::user();
+        $team->removeRoute($dungeonroute);
 
-        if ($team->canAddRemoveRoute($user)) {
-            $team->removeRoute($dungeonroute);
-            $result = response()->noContent();
-        } else {
-            $result = response('Forbidden', Http::FORBIDDEN);
-        }
-
-        return $result;
+        return response()->noContent();
     }
 
     /**
@@ -220,16 +198,12 @@ class AjaxTeamController extends Controller
     {
         Gate::authorize('can-ad-free-giveaway', $team);
 
-        /** @var User $currentUser */
-        $currentUser = Auth::user();
-
         if ($user->patreonAdFreeGiveaway === null) {
             abort(422, 'Unable to remove ad-free giveaway - user does not have any at the moment.');
         }
 
-        if ($user->patreonAdFreeGiveaway->giver_user_id !== $currentUser->id) {
-            abort(422, 'Unable to remove ad-free giveaways that was not originally given by you.');
-        }
+        // Only the giver may take their own giveaway back
+        Gate::authorize('revokeAdFreeGiveaway', $user);
 
         $user->patreonAdFreeGiveaway->delete();
 
--- a/app/Http/Controllers/Ajax/AjaxTeamController.php
+++ b/app/Http/Controllers/Ajax/AjaxTeamController.php
@@ -3,6 +3,7 @@
 namespace App\Http\Controllers\Ajax;
 
 use App\Http\Controllers\Controller;
+use App\Http\Requests\Team\TeamChangeRoleFormRequest;
 use App\Http\Requests\Team\TeamDefaultRoleFormRequest;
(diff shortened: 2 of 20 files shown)


app/Http/Controllers/Ajax/AjaxMapIconController.php
```diff
@@ -88,23 +80,20 @@ public function store(
...
+                        'team_id'          => $teamId,
+                        'dungeon_route_id' => null,
+                    ];
                 }
 
                 // Prevent people being able to update icons that only the admin should if they're supplying a valid dungeon route
-                if ($mapIcon->exists && $mapIcon->dungeon_route_id === null && $dungeonRoute !== null && $mapIcon->team_id === null) {
-                    throw new Exception('Unable to save map icon!');
+                if ($mapIcon->exists && $dungeonRoute !== null) {
+                    Gate::authorize('update', $mapIcon);
```
Wotuu, line 96: :robot: **Behavioral loosening, not a status-code-only change.** The old guard here was

```php
if ($mapIcon->exists && $mapIcon->dungeon_route_id === null && $dungeonRoute !== null && $mapIcon->team_id === null) {
    throw new Exception('Unable to save map icon!');
}
```

which threw for **everyone — admins included** ($isUserAdmin was only consulted in the `$dungeonRoute === null` branch above). The PR body says this path "stays admin-only", but it was deny-all before; `MapIconPolicy::update` now lets admins through.

The consequence is not just the response code: `$validated['dungeon_route_id'] = $dungeonRoute?->id` was set at the top of `store()`, so an admin who PUTs a *mapping* icon (`dungeon_route_id === null`, `team_id === null`) via `/ajax/{dungeonRoute}/mapicon/{mapIcon}` now silently converts that shared mapping icon into a personal route icon — it disappears from the mapping for everyone. Previously that request 500'd for all callers, admin or not.

Admin-only, so low blast radius, but in a PR whose premise is "same decision, better status code" this one changes the decision. Suggest keeping this path deny-all (drop the admin escape from `update()` when both `dungeon_route_id` and `team_id` are null and a route is supplied), or explicitly calling the loosening out as intended in the PR body/policy docblock.

app/Http/Controllers/Ajax/AjaxTeamController.php
```diff
@@ -65,23 +65,15 @@ public function changeRoutePublishing(TeamRoutePublishingFormRequest $request, T
...
         $role       = $request->get('role');
 
-        // Only if the current user may do such a thing
-        if ($team->canChangeRole($user, $targetUser, $role)) {
-            $team->changeRole($targetUser, $role);
-            $result = response()->noContent();
-        } else {
-            $result = response('Forbidden', Http::FORBIDDEN);
-        }
+        Gate::authorize('change-role', [$team, $targetUser, $role]);
```
Wotuu, line 72: :robot: **Unvalidated `role` now produces a 500 before any authorization check.** `$role = $request->get('role')` is unvalidated (`Request`, no FormRequest), and the policy parameter is typed `string $role`. A missing `role` (null) or `?role[]=x` (array) makes the Gate's policy-method invocation throw a `TypeError` → 500.

This is reachable earlier than before: the old flow ran `Gate::authorize('change-role', $team)` (moderator check) *first*, so non-moderators got a 403 and only moderators could reach the equally-typed `canChangeRole(..., string $role)`. Now the TypeError fires at argument-binding time, before `isUserModerator()` in the policy body runs — so **any authenticated user** who supplies a valid `username` and omits/mangles `role` gets a 500 instead of the 403 they got before.

Per project convention this wants a FormRequest validating `username` and `role` (e.g. `role => required|string|in:...TeamUser::ALL_ROLES keys`); at minimum cast/validate `role` before the `Gate::authorize` call.

app/Policies/MapIconPolicy.php
```diff
@@ -0,0 +1,73 @@
...
+     */
+    public function delete(?User $user, MapIcon $mapIcon): Response
+    {
+        if ($mapIcon->dungeon_route_id !== null) {
+            return $this->allow();
+        }
+
+        return $user !== null && $user->hasRole(Role::ROLE_ADMIN) ?
+            $this->allow() :
+            $this->deny(__('policy.update_map_icon_admin_only'));
```
Wotuu: :robot: Nit: `delete()` reuses `policy.update_map_icon_admin_only` — "Only administrators may **change** map icons that are **not attached to a route or team**." Both halves are wrong for the documented stricter case this method exists to preserve: a non-admin denied deletion of a *team* icon is told the icon is not attached to a team, and the message says "change" rather than "delete". Suggest a dedicated `delete_map_icon_admin_only` key ("Only administrators may delete map icons that are not attached to a route.").

app/Policies/UserPolicy.php
```diff
@@ -0,0 +1,69 @@
...
+        return in_array($user->name, config('keystoneguru.super_admins', []), true) ?
+            $this->allow() :
+            $this->deny(__('policy.make_role_only_super_admins_may_grant_admin'));
+    }
+
+    /**
+     * Determine whether the user can revoke the ad-free giveaway the target user currently has.
+     * Only the giver may take their own giveaway back.
+     */
+    public function revokeAdFreeGiveaway(User $user, User $target): bool
```
Wotuu: :robot: Nit: this is the only new policy method whose denial loses its user-facing message. Both replaced call sites aborted 422 with a specific explanation (`controller.profile.error.remove_ad_free_giveaway_not_yours` / the inline team-controller string, both shown in the UI); returning `bool` here means both endpoints now respond with the generic "This action is unauthorized." Every other deny in this PR returns a `Response` with a translated message — suggest doing the same here (the existing `controller.profile.error.remove_ad_free_giveaway_not_yours` key can be reused), and double-checking the front-end handlers for these two endpoints tolerate 403-with-generic-message where they previously displayed the 422 message.

app/Http/Controllers/DungeonRoute/DungeonRouteController.php
```diff
@@ -405,13 +405,17 @@ public function copy(
...
+     * @throws AuthorizationException
+     */
     public function claim(
         Request      $request,
         Dungeon      $dungeon,
         DungeonRoute $dungeonroute,
         string       $title,
     ): RedirectResponse {
-        // Regardless of the result, try to claim the route
+        Gate::authorize('claim', $dungeonroute);
```
Wotuu, line 417: :robot: Nit, description/code drift: the PR body says "`claim()` is a plain mutator", but `DungeonRoute::claim()` is untouched by this diff and still self-guards on `isSandbox()` (returns false and skips the update for non-sandbox routes). Harmless — the gate now denies first and the internal guard is defense-in-depth — but either the body or the model should be brought in line so the next reader doesn't assume the model-side check was removed.

Wotuu (commented):
:robot: Cold review (fable), focused on auth-equivalence between the old inline checks and the new policies. Most mappings check out exactly: `TeamPolicy::edit` == the old `isUserMember` check in `AjaxDungeonRouteController::get`, `changeRoutePublishing` == the old `hasRole(ROLE_ADMIN)` guard (and it is user-only, so it is safe on the unsaved Team in `TeamController::store`), `moderateRoute` == `canAddRemoveRoute` so the dropped inline branches really were dead, the `change-role` ability has no other call sites so the policy signature change is safe, and the MapIcon delete/createGlobal paths are status-code-only changes (guests are handled by the nullable/non-nullable `User` parameters the same way the old `Auth::check()` guards did). 5 findings posted inline — 1 behavioral loosening (map icon update), 1 robustness regression (change-role TypeError), 3 nits.

Wotuu:

app/Http/Controllers/Ajax/AjaxMapIconController.php
Wotuu: :robot: Confirmed - the old inline guard threw for *everyone including admins* here, so this was a genuine permission expansion, not a status-code-only change. Fixed in `d931fe22a`: `MapIconPolicy::update()` now denies unconditionally when the icon is currently global (no route, no team), regardless of role - matching the old behaviour exactly. This is the only call site of this ability, so there's no other legitimate use of the admin-allow branch it was providing. Test `update_givenMappingIcon_returnsDeniedRegardlessOfRole` (renamed/inverted from the old admin-allowed assertion) pins it.

app/Http/Controllers/Ajax/AjaxTeamController.php
Wotuu: :robot: Confirmed - `role` reached `TeamPolicy::changeRole()`'s `string $role` parameter completely unvalidated. Added `TeamChangeRoleFormRequest` (validates `username` via `exists:users,name` and `role` via `Rule::in(TeamUser::ALL_ROLES)`), and `AjaxTeamController::changeRole()` now type-hints it instead of a plain `Request`. A missing/array/invalid role now fails validation instead of 500ing. New feature test `AjaxTeamControllerTest` covers the missing/array/invalid-role and unknown-username cases plus the happy path. Fixed in `d931fe22a`.

app/Policies/MapIconPolicy.php
Wotuu: :robot: Fixed in `d931fe22a` - `delete()` now uses its own `policy.delete_map_icon_admin_only` translation key instead of reusing `update_map_icon_admin_only`.

app/Policies/UserPolicy.php
Wotuu: :robot: Fixed in `d931fe22a` - `revokeAdFreeGiveaway()` now returns `Response` instead of `bool`, denying with the pre-existing `controller.profile.error.remove_ad_free_giveaway_not_yours` translation (same text the old inline checks in both callers used to show - that key was still present in all locales, just unused, since this PR's only removed call site was the one using it). Note the status code did change from the callers' old 422 to Gate::authorize's standard 403, since restoring the exact 422 would need callers to catch and remap the exception - happy to do that too if you'd rather keep 422, but the specific message was the actual regression flagged here. Updated the two tests asserting a raw bool return to use `->allowed()`/`->denied()`.

app/Http/Controllers/DungeonRoute/DungeonRouteController.php
Wotuu: :robot: Confirmed this is description drift, not a code issue - `DungeonRoute::claim()`'s internal `isSandbox()` guard is untouched and still self-guards, it's just now provably redundant since `DungeonRoutePolicy::claim()` (the only caller, via `Gate::authorize('claim', ...)` in the controller) checks the identical condition first. Left the model method as-is (defense in depth, and there's no harm in the duplication) rather than touching behavior for a documentation nit - the PR description could be corrected to say "the guard is now redundant" instead of "removed", but I'll leave that edit to you since it's just prose.

app/Http/Controllers/Ajax/AjaxMapIconController.php
Wotuu: :robot: Fixed — `MapIconPolicy::update()` now denies unconditionally when both `dungeon_route_id` and `team_id` are null, no admin escape, matching the docblock's stated intent.

app/Http/Controllers/Ajax/AjaxTeamController.php
Wotuu: :robot: Fixed — `changeRole` now takes a `TeamChangeRoleFormRequest` and reads `$request->validated('role')`, so an invalid/missing role 422s from validation before the Gate call instead of 500ing on a TypeError.

app/Policies/MapIconPolicy.php
Wotuu: :robot: Fixed — `delete()` now uses its own `policy.delete_map_icon_admin_only` key ("Only administrators may delete map icons that are not attached to a dungeon route."), distinct from `update()`'s message.

app/Policies/UserPolicy.php
Wotuu: :robot: Fixed — `UserPolicy::revokeAdFreeGiveaway()` now returns a `Response` carrying the `controller.profile.error.remove_ad_free_giveaway_not_yours` message instead of a bare bool.

app/Http/Controllers/DungeonRoute/DungeonRouteController.php
Wotuu: :robot: Fixed the PR body wording — it now says `DungeonRoute::claim()` is untouched by this diff and still self-guards on `isSandbox()` as defense-in-depth, rather than calling it a plain mutator.

Wotuu (commented):
Ok

PR #3668, opened 2026-07-25
#170 Return 404 for retired content and 422 for per-route caps

Wotuu:
:robot: Part of #170. **Stacked on #3663** — this branch is cut from `170-missing-authorization-gates`, so the diff below includes that MR's two commits until it merges; GitHub will narrow it automatically once #3663 lands. Review #3663 first, and merge it first.

(The PR targets `master` rather than the feature branch on purpose: the `php-tests`, `phpstan` and `php-cs-fixer` workflows only trigger on `pull_request` to `master`, so a feature-branch base would have skipped all PHP CI.)

This is the "policies that should not be policies" half of the issue. Two groups of checks lived in
`app/Policies/` but were not authorization at all, and in both cases the **status code was the
tell** — not the fact that they ignore `$user` (Laravel policies are allowed to do that).

## 404 for retired content

`DungeonPolicy`, `ExpansionPolicy`, `GameVersionPolicy` and `SeasonPolicy` were four one-method
policies whose entire body was an `active`-flag check:

```php
public function view(?User $user, Expansion $expansion): bool
{
    return $expansion->active;
}
```

They drove 23 `Gate::authorize('view', …)` calls across the two Discover controllers, so a retired
expansion answered **403**. But it is not withheld from *you* — it is gone for *everyone*. That is a
**404**. All four policies are deleted; the call sites become `abort_unless($x->active, 404)`.

**SEO note:** these are public, indexed pages. 403 → 404 is the intended change (Google treats 404 as
"gone", which is right for a retired expansion) but flagging it explicitly since it is externally
visible.

### Why not `resolveRouteBinding()` / a global scope

That was my first instinct and it is wrong here. I audited all 23 call sites first:

- `$gameVersion`, `$expansion`, `$dungeon` — always **route-bound**.
- `$season` — **never** route-bound. It is always a nullable query result
  (`Season::where('expansion_id', …)->where('index', $seasonIndex)->first()`), so a binding override
  would never fire for it. Worth noting on its own: the old code therefore called
  `Gate::authorize('view', null)` whenever the season index in the URL was bogus.
- `Dungeon` is route-bound across dozens of *non*-Discover routes — admin mapping, floor pages,
  `route/{dungeon}/{dungeonroute}`. Filtering on `active` at bind time would 404 the admin pages
  whose whole purpose is editing inactive dungeons.

So the check stays at the call site, where it only affects the Discover surface.

## 422 for per-route caps

`addKillZone` / `addBrushline` / `addPath` / `addArrow` / `addMapIcon` counted existing objects
against `config('keystoneguru.dungeon_route_limits.*')` and returned **403** on breach. Hitting a
configured cap is not "you are not allowed" — it is a request the server understood and won't
process, i.e. **422**. That is also what this codebase already returns for every other quota and
Patreon-benefit limit (`AjaxDungeonRouteController:572,602`, `AjaxProfileController:35-43`,
`AjaxTeamController:197-205`).

Moved into `app/Http/Controllers/Traits/EnforcesDungeonRouteLimits.php`, so the five near-identical
count/config/message triples become one helper plus a lookup table. **The five translation keys stay
in `lang/en_US/policy.php`** — only `en_US` is editable in this repo and the other locales are
managed externally, so renaming keys would orphan every existing translation. The file name is not
load-bearing.

### One interaction worth calling out

`Gate::authorize('addKillZone', …)` sits inside the private `saveKillZone`, which is wrapped in a
try/catch that rewrites failures into a generic 404. #3663 corrects that catch from `Mockery\Exception`
to the global `\Exception` — which means the new `abort(422)` (`HttpException extends \Exception`)
*would* be swallowed. So this MR extends the passthrough to
`catch (AuthorizationException|HttpException $deliberateResponse)`. This is precisely why the branch
is stacked rather than cut from master.

## Front-end check

The map editor renders `xhr.responseJSON.message` verbatim
(`resources/assets/js/custom/inline/layouts/app.js:143`), and its status switch has cases for
403/404/419 but **not** 422 — so if the message were lost, users would get the generic
"an error occurred" instead of "Unable to add more than 50 pulls to a single route."

`app/Exceptions/Handler.php:107` has an `HttpExceptionInterface` branch that *does* discard the
message in favour of `SymfonyResponse::$statusTexts[...]`, so this needed checking rather than
assuming. It only applies to `$request->isJson()` (request Content-Type) and `api/*`; the map editor
sends form-encoded bodies with `Accept: application/json`, which takes the `expectsJson()` path in
`parent::render()` and preserves the message. Verified against a real request:

```
JSON: {"message":"Unable to add more than 1 pulls to a single route."}
```

That assertion is now in the test suite (`assertJsonPath('message', …)`), so a future handler change
that swallows it fails a test rather than silently degrading the editor. Net user-visible change:
the notification prefix goes from `403:` to `422:`; the message itself is unchanged.

## Tests

- `DungeonRouteDiscoverInactiveContentTest` — 404 for an inactive expansion and an inactive dungeon,
  404 for an unknown season index, plus an active-expansion 200 so the 404s cannot pass for an
  unrelated reason. Uses the inactive rows **the seeder already provides** rather than flipping
  `active` on a live row, so a mid-test failure cannot leave the shared database dirty.
- `AjaxDungeonRouteLimitsTest` — 422 (with the right message) at the kill-zone cap, success below it,
  and 422 at the arrow cap. Caps are lowered via `config()` rather than by creating 50 rows.

## Verification

- `--filter="DungeonRouteDiscoverInactiveContentTest|AjaxDungeonRouteLimitsTest"` — 7 passed
- `--group=Policy` — 22 passed (was 26; the 4 deleted policies had no tests, the drop is
  `DungeonRoutePolicy`'s 5 removed quota abilities)
- CI `feature-controller` shard — 82 tests, 1 failure, pre-existing and environment-only


.claude/CLAUDE.md
```diff
@@ -49,6 +49,19 @@ the appearance of impersonating the user.
 dereferences the `@file` into its contents; lowercase `-f` would send the literal string
 `@<file>` as the body.
 
+### Stacked PRs: always target `master`
+
+`php-tests`, `phpstan` and `php-cs-fixer` are all `on: pull_request: branches: [master]`, so a PR
+opened against a *feature* branch silently skips every PHP check. It still shows green — `build`,
+`js-tests` and `php-cs-fixer` run — which looks like a pass until you count the checks (9 when a PR
+is wired up correctly, 3 when it is not).
```
Wotuu: :robot: The first sentence here misstates the `php-cs-fixer` trigger, and the paragraph contradicts itself two lines later: it first lists `php-cs-fixer` among the workflows that are `on: pull_request: branches: [master]` (i.e. skipped on a feature-branch base), then correctly counts it among the 3 checks that *do* run. The workflow itself has **no** base-branch filter — its header comment even says so explicitly ("No base-branch filter: stacked MRs ... must pass the same style gate"):

https://github.com/RaiderIO/keystone.guru/blob/8f40a823b53eb1a6d8dc94ec7c4845d73cb93586/.github/workflows/php-cs-fixer.yml#L3-L11

Only `php-tests` and `phpstan` are gated on `branches: [master]`. Since this section is durable guidance future agents will follow verbatim, the first sentence should drop `php-cs-fixer` from the gated list.

app/Policies/DungeonRoutePolicy.php
```diff
@@ -174,49 +174,4 @@ public function schedulePublish(User $user, DungeonRoute $dungeonRoute): Respons
 
         return $this->deny();
     }
-
-    public function addKillZone(?User $user, DungeonRoute $dungeonRoute): Response
```
Wotuu, line 162: :robot: Deleting these abilities (and the four `view` policies) leaves two in-repo skills describing an authorization surface that no longer exists, and `.claude/CLAUDE.md` says durable notes live in those skills — so a future security pass will be told to look for gates this PR removed:

- `security-review/SKILL.md` still lists `Season`, `Dungeon`, `Expansion`, `GameVersion` among the policies in `app/Policies/` (only 5 remain: DungeonRoute, LiveSession, Tag, TagCategory, Team), still documents ownership enforcement as `Gate::authorize('edit'|'view'|'addKillZone'|..., $dungeonRoute)`, and its "9 policies for 150+ models" reassurance is now off by four:

https://github.com/RaiderIO/keystone.guru/blob/8f40a823b53eb1a6d8dc94ec7c4845d73cb93586/.claude/skills/security-review/SKILL.md#L22-L27

- `ajax-map-editor-crud/SKILL.md` still says route-scoped endpoints are authorized via `Gate::authorize('edit', $dungeonRoute)` "(+ ability-specific gates like `addMapIcon`)":

https://github.com/RaiderIO/keystone.guru/blob/8f40a823b53eb1a6d8dc94ec7c4845d73cb93586/.claude/skills/ajax-map-editor-crud/SKILL.md#L65-L67

For a series that is specifically about tightening authorization semantics, the canonical security-model doc drifting out of sync is worth fixing in the same MR (the PR already updates `.claude/CLAUDE.md`, so docs are in scope).

Wotuu:

.claude/CLAUDE.md
Wotuu: :robot: Fixed in ae6fe0765 — corrected. `php-cs-fixer.yml` has no base-branch filter (its own header comment says so, and it runs on every `pull_request` event); only `php-tests` and `phpstan` are `on: pull_request: branches: [master]`. Line 54 now says "`php-tests` and `phpstan` are both..." instead of including `php-cs-fixer`, matching the "3 that always run" list two sentences later.

app/Policies/DungeonRoutePolicy.php
Wotuu: :robot: Fixed in ae6fe0765. The finding was actually broader than the visible diff on `DungeonRoutePolicy.php` alone — this PR also deletes `DungeonPolicy`, `ExpansionPolicy`, `GameVersionPolicy` and `SeasonPolicy` entirely (the 4 view policies) and adds the new `EnforcesDungeonRouteLimits` trait replacing the 5 deleted `add*` abilities. Updated `security-review/SKILL.md`: dropped the 4 removed policies from the list, corrected "9 policies" to "5 policies," replaced the `addKillZone`-style Gate example with a description of the new 422-via-trait cap mechanism, and added a pointer to the discover-controller 404 handling that replaced the view policies. Updated `ajax-map-editor-crud/SKILL.md`'s `addMapIcon` reference the same way.

.claude/CLAUDE.md
Wotuu: :robot: Fixed — the CLAUDE.md wording now correctly states only `php-tests` and `phpstan` are gated on `branches: [master]`, while `build`, `js-tests` and `php-cs-fixer` run regardless of base branch.

app/Policies/DungeonRoutePolicy.php
Wotuu: :robot: Fixed — the security-review skill's Authorization section now accurately lists only the 5 remaining policies (DungeonRoute, Team, LiveSession, Tag, TagCategory) and explains the Season/Dungeon/Expansion/GameVersion removal inline.

app/Http/Controllers/Traits/EnforcesDungeonRouteLimits.php
```diff
@@ -0,0 +1,56 @@
...
+{
+    public const LIMIT_KILL_ZONES = 'kill_zones';
+
+    public const LIMIT_BRUSHLINES = 'brushlines';
+
+    public const LIMIT_PATHS = 'paths';
+
+    public const LIMIT_ARROWS = 'arrows';
+
+    public const LIMIT_MAP_ICONS = 'map_icons';
```
Wotuu: Convert this to enums - I'm trying to limit `const` usage in this way.

app/Http/Controllers/DungeonRoute/DungeonRouteDiscoverController.php
```diff
@@ -87,8 +86,8 @@ public function discoverSeason(
 
         $season = Season::where('expansion_id', $gameVersion->expansion_id)->where('index', $seasonIndex)->first();
 
-        Gate::authorize('view', $gameVersion);
-        Gate::authorize('view', $season);
+        abort_unless($gameVersion->active, 404);
+        abort_if($season === null || !$season->expansion->active, 404);
```
Wotuu: Ok so I get that the point of this MR is to produce correct HTTP codes, and I agree. But I dislike this new notation. It pulls logic from the Policies, where I'd expect it, back into the controller. This auth logic is now also duplicated half a dozen times instead of it being in one place. Can this be handled differently? The Gate:: logic was fine but now it's being removed again.

Wotuu: I see that `$this->deny(<reason>, <code>)` in a `HandlesAuthorization` accepts a code as the second parameter. Should it then be a 404 there instead?

Wotuu (commented):
Needs work

app/Http/Controllers/Traits/EnforcesDungeonRouteLimits.php
Wotuu: :robot: Fixed in 8b5d09131. The `LIMIT_*` string constants are now a backed string enum, `App\Models\DungeonRoute\DungeonRouteLimitType` (cases `KillZones`, `Brushlines`, `Paths`, `Arrows`, `MapIcons`), placed alongside the other `DungeonRoute`-domain models. `EnforcesDungeonRouteLimits::LIMITS` keys off `DungeonRouteLimitType::X->value`, `abortIfDungeonRouteLimitReached()`'s `$type` parameter is now typed `DungeonRouteLimitType` instead of `string`, and every `Ajax*Controller` call site (`AjaxKillZoneController`, `AjaxBrushlineController`, `AjaxPathController`, `AjaxArrowController`, `AjaxMapIconController`) passes the enum case instead of the old constant.

app/Http/Controllers/DungeonRoute/DungeonRouteDiscoverController.php
Wotuu: :robot: Fixed in 8b5d09131. Restored `DungeonPolicy`, `ExpansionPolicy`, `GameVersionPolicy` and `SeasonPolicy::view()` (which this PR's first commit had deleted), and moved the `active` check back inside each of them. `DungeonRouteDiscoverController` and `DungeonRouteDiscoverExpansionSeasonController` go back to calling `Gate::authorize('view', $gameVersion)` / `Gate::authorize('view', $dungeon)` / `Gate::authorize('view', $expansion)` instead of the inline `abort_unless(...)`, so the auth logic lives in one place again instead of being duplicated across the discover methods.

For the nullable, non-route-bound `$season` lookup specifically, `Gate::authorize('view', [Season::class, $season])` resolves `SeasonPolicy` off the class name even when `$season` is `null`, which then gets passed straight through to `SeasonPolicy::view(?User $user, ?Season $season)`.

On the second comment: I checked, and `deny($message, $code)`'s second parameter is *not* an HTTP status - it just becomes the value `AuthorizationException::getCode()` returns; the actual HTTP status only comes from `Response::$status`, which `deny()` never touches. `deny($message, 404)` silently renders as a plain 403 (confirmed by a failing `DungeonRouteDiscoverInactiveContentTest` before I caught it). The mechanism you're looking for is `denyAsNotFound($message)` (or `denyWithStatus($status, $message)` for other codes) - that's what the four policies now use.

Wotuu (commented):
Ok

PR #3671, opened 2026-07-25
#3669 Fix Pickr.create being undefined, which aborted map editor initialisation

### LINKED ISSUE #3669: Pickr 1.10.0 CJS build breaks Pickr.create, aborting map editor sidebar initialisation
## Symptom

Every route editor page load throws an uncaught `TypeError: Pickr.create is not a function`.
Reproduced in headless Chrome against master (`e39ffd914`) on
`/route/{dungeon}/{key}/{title}/edit`; the error is also present in v15.7.0, so staging and
(once released) production are affected too.

## Cause

`resources/assets/js/bootstrap.js:111` does:

```js
window.Pickr = require('@simonwep/pickr');
```

The `@simonwep/pickr` 1.9.1 -> 1.10.0 bump (faee409a3, PR #3618) rebuilt the package with Vite and
`"type": "module"`. Its `exports` map is:

```json
".": {
    "import": "./dist/pickr.min.mjs",
    "require": "./dist/pickr.min.js",
    "types": "./types/pickr.d.ts"
}
```

Only the ESM build actually exports the class (`export { M as default }`). The CJS build resolved by
the `require` condition exports an empty object - verified in the app container:

```
type: object
ownNames: []
symbols: [ 'Symbol(Symbol.toStringTag)' ]
```

So `window.Pickr` is an empty object and `Pickr.create` is `undefined`.

## Blast radius

The throw happens inside `SettingsTabMap#activate()` (the `#edit_route_freedraw_options_color`
picker), which is called from `Map#activate()`. Everything after that call in `Map#activate()`
never runs:

- `this.settingsTabPull.activate()`
- the `floorid:changed` registration
- `_setupRatingSelection()`, `_setupFloorSelection()`, `_setupMapObjectGroupVisibility()`,
  `_setupEnemyVisualTypes()`, `_setupZoomControl()`, `_setupFavorite()`, `_setupLabelToggle()`

and because `InlineManager#activate()` never reaches
`this._activatedInlineCode[code.id] = code`, the whole dependency cascade for inline code depending
on `common/maps/map` is skipped as well.

Other `Pickr.create` call sites are equally dead: `settingstabpull.js`, `pullworkbench.js`,
`mapobject.js` (the map object colour attribute picker).

## Key files

- `resources/assets/js/bootstrap.js` (the `require`)
- `scripts/build/*` (esbuild resolve conditions, if the fix is made there)
- `resources/assets/js/custom/inline/common/maps/editsidebar/settingstabmap.js` (first throw site)

## Definition of done

- Route editor loads with no page errors in headless Chrome; the freedraw colour picker, pull colour
  picker and pull workbench picker all render and work.
- A test guards the export shape so a future Pickr bump cannot silently reintroduce this (the repo
  already has `resources/assets/js/jquery4-plugin-audit.test.js` as precedent for bundled-dependency
  contract tests).

Wotuu:
:robot: Closes #3669

## Problem

`@simonwep/pickr` 1.10.0 (bumped from 1.9.1 in #3618) declares `"type": "module"` while its
`require` export condition still points at `dist/pickr.min.js`, a file containing UMD code. Because
of the package type that file is parsed as ESM, so it exports nothing at all -
`require('@simonwep/pickr')` yields an empty module and `Pickr.create` is `undefined`. Only
`dist/pickr.min.mjs` (the `import` condition) exports the class.

The resulting `TypeError: Pickr.create is not a function` is thrown inside
`SettingsTabMap#activate()`, so everything after that call in `Map#activate()` never runs -
`settingsTabPull.activate()`, the `floorid:changed` registration and the whole `_setup*()` sidebar
wiring - and `InlineManager#activate()` never reaches `this._activatedInlineCode[code.id] = code`,
so every inline code depending on `common/maps/map` is skipped too.

This is in v15.7.0, so staging is affected.

## Fix

Import Pickr as ESM. `bootstrap.js` already mixes hoisted ESM imports (`laravel-echo`, `pusher-js`)
with in-place `require()` calls; the `window.Pickr` assignment stays exactly where it was, and Pickr
has no load-order dependency on jQuery.

## Verification

Headless Chrome, same editor page, master baseline (`:8008`) vs this branch (`:8106`):

| | master | this branch |
|---|---|---|
| page errors | `Pickr.create is not a function` | none |
| `typeof window.Pickr` | `object` | `function` |
| `typeof window.Pickr.create` | `undefined` | `function` |
| rendered `.pickr` widgets | 0 | 4 |

The visible symptom is the pull list, which sits on "Loading..." forever on master - the same class
of breakage as #3588.

**Before (master)**

before

**After (this branch)**

after

## Test

`resources/assets/js/pickr-export.test.js` bundles the exact usage from `bootstrap.js` with the
resolution-relevant esbuild settings from `scripts/build/app.mjs` and evaluates the output, plus a
static check that `bootstrap.js` doesn't reintroduce the `require()` form. Follows the
`jquery4-plugin-audit.test.js` precedent for bundled-dependency contract tests.

Confirmed the test actually catches the bug: switching the bundled snippet back to `require()` fails
with `expected 'object' to be 'function'`.

`npx vitest run`: 24 files, 248 tests, all green. No PHP touched.

## Note on the jsdom/esbuild shim

The test overrides `TextEncoder`/`TextDecoder`/`Uint8Array` with Node's implementations before
requiring esbuild - esbuild asserts `new TextEncoder().encode('') instanceof Uint8Array` on load,
which fails under jsdom because those come from different realms. Pickr itself needs the jsdom DOM,
so moving the file to the `node` environment isn't an option. Vitest isolates per test file, so the
override doesn't leak (full suite verified green).

### DIFF
--- a/resources/assets/js/bootstrap.js
+++ b/resources/assets/js/bootstrap.js
@@ -29,6 +29,12 @@ import Echo from 'laravel-echo'
 
 import Pusher from 'pusher-js';
 
+// Imported as ESM rather than require()d (#3669): since 1.10.0 the package declares
+// "type": "module", so its `require` export condition (dist/pickr.min.js) is parsed as ESM even
+// though it contains UMD code - the resulting module has no exports at all and Pickr.create is
+// undefined. Only the `import` condition (dist/pickr.min.mjs) exports the class.
+import Pickr from '@simonwep/pickr';
+
 window.Pusher = Pusher;
 
 function startLaravelEcho() {
@@ -108,7 +114,7 @@ window.d3 = require('d3'); // v3.5.14 since Pather uses an out-of-date version
 window.Pather = require('leaflet-pather');
 window.circleMenu = require('zikes-circlemenu');
 window.Noty = require('noty');
-window.Pickr = require('@simonwep/pickr');
+window.Pickr = Pickr; // See the ESM import above (#3669)
 window.AntPath = require('leaflet-ant-path');
 window.Grapick = require('grapick');
 window.simplebar = require('simplebar');
--- a/resources/assets/js/pickr-export.test.js
+++ b/resources/assets/js/pickr-export.test.js
@@ -1,0 +1,80 @@
+// ---------------------------------------------------------------------------
+// Regression coverage for #3669, in the spirit of the jquery4-plugin-audit
+// (#3590): guard a bundled dependency's export shape against a version bump
+// that silently breaks it.
+//
+// @simonwep/pickr 1.10.0 (bumped from 1.9.1 in #3618) declares "type": "module"
+// while its `require` export condition still points at dist/pickr.min.js, which
+// contains UMD code. Because of the package type that file is parsed as ESM, so
+// it exports nothing at all: `require('@simonwep/pickr')` yields an empty module
+// and `Pickr.create` is undefined. That threw
+// `TypeError: Pickr.create is not a function` inside SettingsTabMap#activate(),
+// which aborted the rest of Map#activate() and the InlineManager dependency
+// cascade behind it - i.e. most of the map editor sidebar never initialised.
+//
+// bootstrap.js therefore imports Pickr as ESM. This test bundles that exact
+// usage with the esbuild settings that drive module resolution in
+// scripts/build/app.mjs (bundle + format + platform) and evaluates the result,
+// so a future bump that breaks the export shape fails here rather than in
+// production.
+// ---------------------------------------------------------------------------
+
+const fs = require('node:fs');
+const path = require('node:path');
+const util = require('node:util');
+
+// esbuild asserts `new TextEncoder().encode('') instanceof Uint8Array` on load, which fails under
+// jsdom: its TextEncoder and its Uint8Array come from different realms. Line up all three on Node's
+// implementations before requiring esbuild. Pickr itself still needs the jsdom DOM, so switching
+// the whole file to the node environment isn't an option.
+global.TextEncoder = util.TextEncoder;
+global.TextDecoder = util.TextDecoder;
+global.Uint8Array = new util.TextEncoder().encode('').constructor;
+
+const {buildSync} = require('esbuild');
+
+const rootDir = path.resolve(__dirname, '..', '..', '..');
+
+/**
+ * Bundles a snippet with the resolution-relevant settings scripts/build/app.mjs uses for app.js,
+ * and returns the generated code.
+ */
+function bundle(contents) {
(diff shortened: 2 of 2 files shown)


Wotuu:
My only question is, how could this happen? I told you to merge dependabot MRs when they're safe, was the upgrade from 1.9.1 to 1.10.0 an incorrect symantic versioning thing? Should they have made it v2.0.0? Was it something else? Can we improve for the next release?

Wotuu:
:robot: Checked the package itself, not just our usage of it.

`@simonwep/pickr@1.9.1` had no `"type"`/`"exports"` fields at all — plain CJS resolution via `main`.
`1.10.0` added `"type": "module"` **and** an `exports` map at the package root:

```json
"type": "module",
"exports": {
  ".": {
    "import": "./dist/pickr.min.mjs",
    "require": "./dist/pickr.min.js",
    "types": "./types/pickr.d.ts"
  }
}
```

but `dist/pickr.min.js` (the `require` target) still contains the old UMD build, unchanged from
1.9.1. With `"type": "module"` set, Node/esbuild's resolver now parses that `.js` file as ESM
regardless of its actual UMD contents, so `require('@simonwep/pickr')` silently returns an empty
module — no error, no warning, just `Pickr.create === undefined`. This is the classic "dual package
hazard" packaging mistake: they exposed both `import`/`require` conditions but only actually ship an
ESM-compatible artifact. That's a breaking change for every CJS consumer, and by semver it
should've been a major bump (or they should've kept publishing a real CJS build under the `require`
condition) — 1.9.1 → 1.10.0 was mislabeled on their end, not a versioning mistake on our side.

Why it slipped through in #3618: nothing in the suite ever `require()`'d Pickr and called
`.create()` — `js-tests` never imported it, and the `build` check only bundles/resolves module
paths, it doesn't execute the resulting code, so the runtime-only breakage (empty exports at
`require()` time) had no way to surface. Green CI on a dependency bump PR isn't proof the bundle
still *works*, only that it *builds*.

For next time: this PR adds `resources/assets/js/pickr-export.test.js`, a standing regression test
that actually requires Pickr the way `bootstrap.js` does and asserts `Pickr.create` is callable —
so a future Pickr bump reintroducing this would fail CI directly. More generally, for any dependency
bump that touches a package's `"exports"`/`"type"` fields (visible in the diff of `package-lock.json`
or via `npm view <pkg> exports`), it's worth a quick headless-browser smoke of the page(s)
that use it before merging, not just green CI — same reasoning as the existing
`jquery4-plugin-audit.test.js` precedent for bundled-dependency contracts.

PR #3672, opened 2026-07-25
#3670 Rebind the enemy tooltip when nothing is bound, not only when the text changed

### LINKED ISSUE #3670: Enemy#bindTooltip silently skips rebinding when the layer has no tooltip bound
## Problem

`Enemy#bindTooltip()` (`resources/assets/js/custom/models/enemy.js`) only rebinds when the rendered
tooltip text changed:

```js
// Only rebind if the text has changed
if (this.tooltipText !== text) {
    this.tooltipText = text;
    this.unbindTooltip();
    this.layer.bindTooltip(text, {direction: 'top'});
}
```

The guard checks the *text*, not whether a tooltip is actually bound to the layer. So any code path
that unbinds the tooltip without also clearing `tooltipText` leaves the enemy permanently without a
tooltip: the next `bindTooltip()` recomputes the identical text, the guard short-circuits, and
nothing is ever rebound.

Verified in headless Chrome on `/{publicKey}`:

```js
enemy.layer.getTooltip()   // truthy
enemy.unbindTooltip();
enemy.layer.getTooltip()   // null
enemy.bindTooltip();
enemy.layer.getTooltip()   // still null  <-- silently skipped
```

`Enemy#setTooltipEnabled(false)` avoids this by explicitly resetting `this.tooltipText = ''`, which
shows the guard was known to be fragile - but that workaround has to be repeated at every call site.

## Current call sites that unbind without clearing `tooltipText`

- `EnemyVisual#_initCircleMenu()` `open:` handler - `self.enemy.unbindTooltip()`.
  Not currently observable: the handler then sets a `RaidMarkerSelectMapState`, which flips
  `canOpenRaidMarkerMenu()` to false and therefore *changes* the tooltip text
  (`show_raid_marker_shortcut`), so the guard happens to pass. It only works by accident - drop the
  shortcut hint from the tooltip and the enemy silently loses its tooltip on the next
  shift+right-click.
- `MapObjectGroup#setLayerToMapObject()` (`mapobjectgroup.js`) unbinds the old layer's tooltip on a
  layer swap; `tooltipText` survives.

## Fix

Make the guard account for a missing tooltip, e.g.:

```js
if (this.tooltipText !== text || !this.layer.getTooltip()) {
```

Leaflet's `unbindTooltip()` sets `_tooltip = null` and `getTooltip()` returns it, so this is a
reliable "is anything bound right now" check.

## Definition of done

- A vitest test asserts that `bindTooltip()` rebinds after an `unbindTooltip()` when the text is
  unchanged. Follow the global-script-model recipe documented at the top of
  `resources/assets/js/custom/models/killzone.test.js`.
- `npm test` green.

## Key files

- `resources/assets/js/custom/models/enemy.js`
- `resources/assets/js/custom/models/killzone.test.js` (test recipe)

Wotuu:
:robot: Closes #3670

## Problem

`Enemy#bindTooltip()` short-circuited on `this.tooltipText !== text`, which checks the rendered text
but not whether a tooltip is actually bound to the layer. Any code path that unbinds without also
resetting `tooltipText` therefore left that enemy without a tooltip for the rest of the session: the
recomputed text is identical, so the guard skipped the rebind and nothing ever bound it again.

Two live call sites unbind that way:

- `EnemyVisual#_initCircleMenu()`'s `open:` handler (#1431). Not currently observable, but only by
  accident: the handler then sets a `RaidMarkerSelectMapState`, which flips
  `canOpenRaidMarkerMenu()` to false and so *changes* the rendered text
  (`show_raid_marker_shortcut`), letting the guard pass. Drop that hint from the tooltip and every
  shift+right-clicked enemy silently loses its tooltip.
- `MapObjectGroup#setLayerToMapObject()` unbinds the old layer's tooltip on a layer swap while
  `tooltipText` survives.

`Enemy#setTooltipEnabled(false)` already worked around this by explicitly resetting `tooltipText`,
which is a good sign the guard was the wrong place to stop.

## Fix

Also rebind when nothing is bound. Leaflet's `getTooltip()` is `undefined` before the first bind and
`null` after `unbindTooltip()`, so it's a reliable "is anything bound right now" check.

## Verification

Headless Chrome, `explore/retail/pit-of-saron`, master baseline (`:8008`) vs this branch (`:8107`).
Both bind tooltips to all 230 enemies on load; the difference is the unbind/rebind round trip the
circle menu performs, after which the affected enemy is hovered:

| | master | this branch |
|---|---|---|
| enemies with a tooltip on load | 230 / 230 | 230 / 230 |
| `getTooltip()` after unbind + rebind | `null` | bound |
| tooltip visible on hovering that enemy | no | yes |

**Before (master) - hovering the enemy after the round trip shows nothing**

before

**After (this branch)**

after

## Tests

New `resources/assets/js/custom/models/enemy.test.js`, following the global-script model recipe
documented at the top of `killzone.test.js`. It covers:

- the regression: rebinds after an `unbindTooltip()` when the text is unchanged;
- that redundant rebinds are still skipped when a tooltip is already bound (the guard's original
  purpose);
- that a changed text still rebinds;
- that a null layer is a no-op.

Confirmed the regression test fails on the old guard (`expected null not to be null`).

`npx vitest run`: 24 files, 250 tests, all green. No PHP touched.

## Note

`enemy.js` gained the same guarded `module.exports` block `util.js` and `killzone.js` already carry -
it's a no-op in the browser (`module` is undefined there) and doesn't affect the concatenated bundle.

## Not the explore-page tooltip report

This was found while investigating a report of enemy tooltips not appearing on explore pages. It is
**not** that bug - that one reproduces on page load with no interaction, and I could not reproduce it
on any explore page. Filed and fixed here on its own merits.

### DIFF
--- a/resources/assets/js/custom/models/enemy.js
+++ b/resources/assets/js/custom/models/enemy.js
@@ -809,8 +809,13 @@ class Enemy extends VersionableMapObject {
                 text = lang.get('js.no_npc_found_label');
             }
 
-            // Only rebind if the text has changed
-            if (this.tooltipText !== text) {
+            // Only rebind if the text has changed - or if nothing is bound right now (#3670).
+            // Anything that unbinds the tooltip without also resetting tooltipText (the circle menu
+            // in EnemyVisual, a layer swap in MapObjectGroup#setLayerToMapObject) would otherwise
+            // leave this enemy without a tooltip permanently: the recomputed text is identical, so
+            // the guard short-circuits and the rebind never happens.
+            // Leaflet's getTooltip() is undefined before the first bind and null after unbindTooltip().
+            if (this.tooltipText !== text || !this.layer.getTooltip()) {
                 this.tooltipText = text;
 
                 // Remove any previous tooltip
@@ -1355,4 +1360,12 @@ class Enemy extends VersionableMapObject {
             this.visual = null;
         }
     }
+}
+
+// Guarded export for the test runner (Vitest). This is a no-op in the browser,
+// where `module` is undefined, so it does not affect the concatenated bundle.
+if (typeof module !== 'undefined' && module.exports) {
+    module.exports = {
+        Enemy,
+    };
 }
--- a/resources/assets/js/custom/models/enemy.test.js
+++ b/resources/assets/js/custom/models/enemy.test.js
@@ -1,0 +1,173 @@
+// ---------------------------------------------------------------------------
+// Regression coverage for #3670, following the global-script model recipe
+// documented at the top of killzone.test.js: define the globals enemy.js touches
+// at LOAD time before require()-ing it, then hand the class fake collaborators
+// instead of building a real DungeonMap.
+//
+// Enemy#bindTooltip() only rebinds when the rendered text changed. That guard
+// checked the text but not whether a tooltip was actually bound, so any code
+// path that unbinds without also resetting `tooltipText` (the raid marker circle
+// menu in EnemyVisual, a layer swap in MapObjectGroup#setLayerToMapObject) left
+// the enemy without a tooltip permanently - the recomputed text is identical, so
+// the guard short-circuited and nothing was ever rebound.
+// ---------------------------------------------------------------------------
+
+// 1a. Leaflet: enemy.js builds div icons and draw handlers at load time.
+global.L = {
+    divIcon: function () {
+    },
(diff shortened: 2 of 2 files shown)


Wotuu:

Wotuu (commented):
Ok

PR #3676, opened 2026-07-25
#1769 #1770 Creator directory with name search

### LINKED ISSUE #1769: Create discover route creator page

### LINKED ISSUE #1770: Add search function for route creators
If people have a favorite streamer (for example) they should be able to find them through this search function.

Wotuu:
:robot: Closes #1769
Closes #1770

Second of three MRs for the creator arc. **Stacked on #3675** — it targets `master`, so until #3675 merges this diff also contains #3675's commits. Review #3675 first.

Adds `/creators`: browse and search the people making routes. Behind the same `CreatorProfiles` flag as the podium, so it merges dark.

creator directory

creator search

## How listing works

Per your call: **automatic above a threshold, opt-out**. A user is listed when they have at least `keystoneguru.creators.min_published_routes` (default **3**) world-published routes and have not ticked *Hide me from the creator directory* on their profile. Ordered by published route count, then by id so pagination cannot repeat or skip anyone.

The eligibility query lives in `CreatorDirectoryService` rather than the controller, because the featured-creators row on the discover landing (MR 3) needs exactly the same rules — deriving them twice is how the two surfaces end up disagreeing about who counts as a creator.

## Three things worth a look in review

**The threshold and the displayed count come from one subquery.** `withCount()` produces `published_route_count`, and the threshold is applied with `having()` on that same alias rather than a second `whereHas()`. If they were separate subqueries they could drift — a card claiming "3 published routes" while the filter used different criteria.

**The search term is escaped before it reaches the LIKE.** Without `addcslashes($search, '%_\\')`, a search for `%` matches every creator on the site and `_` matches any single character. There is a test for exactly that.

**Unpublished routes never count.** Only `PublishedState::WORLD` counts towards the threshold, so a user with nothing but private drafts is not presented publicly as a route creator. Also tested.

## Verification

- **8 feature tests**, all passing: gating, above/below threshold, opt-out honoured, unpublished routes excluded, search filters, LIKE-wildcard escaping, and overlong search rejected.
- **PHPStan** clean, **php-cs-fixer** clean.
- **Browser-verified** in headless Chrome.

**Query cost measured, not assumed.** The debugbar shows 192 queries on `/creators`, which looks alarming until you compare: the homepage is 215 and `/routes` is 271 on the same stack. Instrumenting the service directly shows the directory itself costs **3 queries** — the pagination count, the creators query with its `withCount` subquery, and one batched load of the avatars. The rest is pre-existing site chrome, and this page is the lightest of the three.

## Note on your dev database

To make the screenshots meaningful I seeded 11 demo creators (`Dratnos`, `Yumytv`, `Naowh`, …) with world-published routes into the shared dev DB — before that, exactly one account qualified and the directory was a single tile. They are left in place so you can click around the feature when you wake up. To remove them:

```sh
docker compose exec -T app php artisan tinker --execute '
$names = ["Dratnos","Yumytv","Naowh","Quazii","Tettles","Growl","Jdotb","Meeres","Xerwo","Petko","Ellesmere"];
foreach (App\Models\User::whereIn("name", $names)->get() as $u) {
    App\Models\DungeonRoute\DungeonRoute::where("author_id", $u->id)->delete();
    $u->delete();
}'
```

## Deliberately not in scope

#1769 also asks for a **category filter** on the directory. Categories are #1771, which depends on route collections (#700); the directory ships with name search only and the filter lands with collections.

### DIFF
--- a/app/Service/Creator/CreatorDirectoryService.php
+++ b/app/Service/Creator/CreatorDirectoryService.php
@@ -2,20 +2,23 @@
 
 namespace App\Service\Creator;
 
-use App\Models\DungeonRoute\DungeonRoute;
-use App\Models\PublishedState;
 use App\Models\User;
+use App\Repositories\Interfaces\UserRepositoryInterface;
 use Illuminate\Contracts\Pagination\LengthAwarePaginator;
 use Illuminate\Database\Eloquent\Builder;
 
 class CreatorDirectoryService implements CreatorDirectoryServiceInterface
 {
+    public function __construct(private readonly UserRepositoryInterface $userRepository)
+    {
+    }
+
     /** @return LengthAwarePaginator<int, User> */
     public function paginateCreators(?string $search = null, ?int $perPage = null): LengthAwarePaginator
     {
         $perPage ??= (int)config('keystoneguru.creators.per_page');
 
-        return $this->buildListedCreatorsQuery()
+        return $this->userRepository->buildListedCreatorsQuery()
             ->when(
                 $search !== null && $search !== '',
                 static fn(Builder $builder): Builder => $builder->where(
@@ -26,51 +29,5 @@ class CreatorDirectoryService implements CreatorDirectoryServiceInterface
             )
             ->paginate($perPage)
             ->withQueryString();
-    }
-
-    /**
-     * Creators eligible for the directory, ordered by how many routes they have published.
-     *
-     * Listing is automatic above a threshold and opt-out, so the only people excluded are those
-     * below the bar and those who ticked hide_from_creator_directory.
-     *
-     * The threshold and the count rendered on each card come from the same aggregate, so they
-     * cannot drift apart.
-     *
-     * Deliberately a join against a pre-aggregated derived table rather than withCount() + having().
-     * A correlated withCount subquery cannot be filtered by an index, so MySQL had to evaluate one
-     * COUNT per non-opted-out user - a full scan of `users` - before it could discard anyone, and
-     * then materialise every matching `users.*` row (including the `bio` TEXT column) into a temp
-     * table to sort it. Grouping `dungeon_routes` first inverts that: the aggregate runs once
-     * against the published_state_id index, and `users` is then reached by primary key.
-     *
-     *   before  users type=ALL    key=NULL     rows=173  Using temporary; Using filesort
-     *   after   users type=eq_ref key=PRIMARY  rows=1
-     *
-     * The cost now scales with published routes and qualifying creators instead of with total
-     * registrations, which matters for a page intended to become public and is not cached.
-     *
-     * @return Builder<User>
-     */
-    private function buildListedCreatorsQuery(): Builder
-    {
-        $minPublishedRoutes = (int)config('keystoneguru.creators.min_published_routes');
-        $worldPublishedId   = PublishedState::ALL[PublishedState::WORLD];
-
-        $publishedRouteCounts = DungeonRoute::query()
-            ->selectRaw('author_id, COUNT(*) AS published_route_count')
-            ->where('published_state_id', $worldPublishedId)
-            ->groupBy('author_id')
-            ->havingRaw('COUNT(*) >= ?', [$minPublishedRoutes]);
-
-        return User::query()
-            ->select('users.*', 'published_routes.published_route_count')
-            ->joinSub($publishedRouteCounts, 'published_routes', 'published_routes.author_id', '=', 'users.id')
-            ->where('users.hide_from_creator_directory', false)
-            // The creator cards render the avatar
-            ->with(['iconfile'])
-            ->orderByDesc('published_routes.published_route_count')
-            // Stable tiebreak so pagination cannot repeat or skip a creator between pages
-            ->orderBy('users.id');
     }
 }
--- a/app/Service/Creator/CreatorDirectoryService.php
+++ b/app/Service/Creator/CreatorDirectoryService.php
@@ -1,0 +1,59 @@
+<?php
+
+namespace App\Service\Creator;
+
+use App\Models\PublishedState;
+use App\Models\User;
+use Illuminate\Contracts\Pagination\LengthAwarePaginator;
+use Illuminate\Database\Eloquent\Builder;
+
+class CreatorDirectoryService implements CreatorDirectoryServiceInterface
+{
+    /** @return LengthAwarePaginator<int, User> */
+    public function paginateCreators(?string $search = null, ?int $perPage = null): LengthAwarePaginator
+    {
+        $perPage ??= (int)config('keystoneguru.creators.per_page');
+
+        return $this->buildListedCreatorsQuery()
+            ->when(
+                $search !== null && $search !== '',
+                static fn(Builder $builder): Builder => $builder->where(
+                    'name',
+                    'like',
+                    sprintf('%%%s%%', addcslashes((string)$search, '%_\\')),
+                ),
+            )
+            ->paginate($perPage)
+            ->withQueryString();
+    }
+
+    /**
+     * Creators eligible for the directory, ordered by how many routes they have published.
+     *
+     * Listing is automatic above a threshold and opt-out, so the only people excluded are those
+     * below the bar and those who ticked hide_from_creator_directory.
+     *
+     * The route count is produced by a single withCount subquery and filtered with having() rather
+     * than a second whereHas subquery, so the threshold and the displayed count cannot drift apart.
+     *
+     * @return Builder<User>
+     */
+    private function buildListedCreatorsQuery(): Builder
+    {
+        $minPublishedRoutes = (int)config('keystoneguru.creators.min_published_routes');
+        $worldPublishedId   = PublishedState::ALL[PublishedState::WORLD];
+
+        return User::query()
+            ->where('hide_from_creator_directory', false)
+            ->withCount([
+                'dungeonRoutes as published_route_count' => static fn($query) => $query
+                    ->where('published_state_id', $worldPublishedId),
+            ])
+            // The route cards and creator cards both render the avatar
+            ->with(['iconfile'])
+            ->having('published_route_count', '>=', $minPublishedRoutes)
+            ->orderByDesc('published_route_count')
+            // Stable tiebreak so pagination cannot repeat or skip a creator between pages
+            ->orderBy('id');
+    }
+}
--- a/app/Service/Creator/CreatorDirectoryService.php
+++ b/app/Service/Creator/CreatorDirectoryService.php
@@ -2,6 +2,7 @@
 
 namespace App\Service\Creator;
 
+use App\Models\DungeonRoute\DungeonRoute;
 use App\Models\PublishedState;
 use App\Models\User;
 use Illuminate\Contracts\Pagination\LengthAwarePaginator;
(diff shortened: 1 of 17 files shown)


app/Service/Creator/CreatorDirectoryService.php
```diff
@@ -0,0 +1,59 @@
...
+
+        return User::query()
+            ->where('hide_from_creator_directory', false)
+            ->withCount([
+                'dungeonRoutes as published_route_count' => static fn($query) => $query
+                    ->where('published_state_id', $worldPublishedId),
+            ])
+            // The route cards and creator cards both render the avatar
+            ->with(['iconfile'])
+            ->having('published_route_count', '>=', $minPublishedRoutes)
```
Wotuu: :robot: **Eligibility filtering is O(all users) per request, not O(creators).**

`having('published_route_count', ...)` cannot use an index, so MySQL has to evaluate the `withCount` subquery for **every** non-opted-out user before it can discard anyone. `EXPLAIN` of the two queries this method produces (dev DB, 173 users):

```
-- page query
users            type=ALL  key=NULL  rows=173  Extra="Using where; Using temporary; Using filesort"
dungeon_routes   DEPENDENT SUBQUERY  key=dungeon_routes_author_id_index  ref=users.id

-- pagination count query (paginate() strips order/limit, then wraps in a derived table)
<derived2>       type=ALL
users            type=ALL  key=NULL  rows=173  Extra="Using where"
dungeon_routes   DEPENDENT SUBQUERY  key=dungeon_routes_author_id_index  ref=users.id
```

`key=NULL` on `users` in both: full table scan, one correlated `COUNT` per row, twice per page view, and the page query additionally materializes every matching `users.*` row (now including the new `bio` TEXT column) into a temp table to filesort it. `hide_from_creator_directory` is also unindexed, so it prunes nothing at the storage layer.

This is the cost the PR body's "the directory itself costs 3 queries" measurement doesn't capture — 3 queries, but two of them scan the entire user table. It's fine at 173 users and grows linearly with registrations, on an uncached page that is meant to eventually be public.

Cheapest halves-the-cost change: `simplePaginate()` instead of `paginate()` deletes the count query outright (the directory has no page-number UI requirement that needs a total). Beyond that, a denormalized `published_route_count` on `users` maintained when a route's `published_state_id` changes would make both the threshold and the ordering index-usable — and MR 3's featured-creators row will run the same query on the discover landing page, so whatever this costs gets paid twice.

Wotuu:

app/Service/Creator/CreatorDirectoryService.php
Wotuu: :robot: Agreed, and fixed in 205c4460a — thanks, this was a real gap in my own verification. I measured *query count* and reported "the directory itself costs 3 queries" as though that settled it; three queries, two of which full-scan `users`, is not cheap. Counting queries is not measuring cost.

I went further than either suggestion, because both left the scan in place. Rather than denormalising a counter (which needs maintenance on every `published_state_id` transition, and can drift), the query shape itself was the problem: aggregate `dungeon_routes` **first** and join that derived table, so `users` is reached by primary key instead of being scanned and filtered by a `having` that no index can serve.

EXPLAIN of the real Eloquent-generated queries, not hand-written SQL:

```
BEFORE - page query
users            type=ALL     key=NULL     rows=173  Using where; Using temporary; Using filesort
dungeon_routes   DEPENDENT SUBQUERY        ref=users.id

AFTER - page query
<derived2>       type=ALL     key=NULL     rows=114  Using temporary; Using filesort
users            type=eq_ref  key=PRIMARY  rows=1    Using where
dungeon_routes   type=ref     key=dungeon_routes_published_state_id_expires_at_index

AFTER - pagination count query
<derived2>       type=ALL     key=NULL     rows=114
users            type=eq_ref  key=PRIMARY  rows=1    Using where
dungeon_routes   type=ref     key=dungeon_routes_published_state_id_expires_at_index
```

The important part is that `users type=ALL key=NULL` is gone from **both** queries, and the temp table now materialises the derived aggregate (114 qualifying authors) rather than every matching `users.*` row including the `bio` TEXT column. Cost scales with published routes and qualifying creators, not with total registrations — which was your actual point.

Two deliberate deviations from your suggestions:

- **Kept `paginate()` rather than `simplePaginate()`.** The count query was half the problem *because* it scanned `users`; now that it is index-driven it is no longer the expensive half, and a browsable directory genuinely benefits from page numbers. If you would still rather drop the total it is a one-word change — say so and I will.
- **No denormalised `published_route_count` column.** It is the right tool if this ever needs to be index-*ordered* at much larger scale, but it buys nothing the join does not already give at this size and adds a write-path invariant to keep correct. Happy to file it as a follow-up if you want it queued rather than dismissed.

You were right that MR #3677 pays this twice — it shares `buildListedCreatorsQuery()`, so it inherits the fix. That branch has been rebased onto this one and its 13 tests pass.

Wotuu: :robot: Fixed — `buildListedCreatorsQuery()` now joins against a pre-aggregated derived table (grouped `dungeon_routes` by `author_id` with the published-route-count threshold applied there) instead of `withCount()`+`having()`, so `users` is reached by primary key (`eq_ref`) instead of a full table scan. Documented with the before/after EXPLAIN in the method's docblock.

Wotuu:
shared ancestor.

```sh
```

already in master.

updated 2026-08-04).

app/Service/Creator/CreatorDirectoryService.php
```diff
@@ -0,0 +1,76 @@
...
+     *
+     *   before  users type=ALL    key=NULL     rows=173  Using temporary; Using filesort
+     *   after   users type=eq_ref key=PRIMARY  rows=1
+     *
+     * The cost now scales with published routes and qualifying creators instead of with total
+     * registrations, which matters for a page intended to become public and is not cached.
+     *
+     * @return Builder<User>
+     */
+    private function buildListedCreatorsQuery(): Builder
```
Wotuu: This is more of an architectural question than anything. Right now, we have a service pattern where the controller will request a service, and the service will handle the business logic, etc. Right now, this service is pretty much just a proxy for some queries, and that's what we have repositories for. 

We could also have the controller directly request the repository, which I feel may bypass the pattern a bit, but it can work. We also have an optional third solution where we create a custom builder class, which is kind of what's already happening here, which I think is a good solution. It's a solution that I used at my previous job, but I fear that introducing that here introduces a third flavor. I kind of wanna spar a bit and figure out which pattern we should commit to and how. I'd like to have your opinion on that.

Wotuu: :robot: My opinion: keep it as a Service with the query built in a private method, exactly as this PR already has it. Don't move it to a bare repository call, and don't introduce a query-builder class as a new pattern.

First, a correction on the framing: `CreatorDirectoryService` isn't actually a proxy for repository queries — it doesn't touch a repository at all. `buildListedCreatorsQuery()` reads a config threshold, joins a pre-aggregated `dungeon_routes` subquery onto `users` (with a documented EXPLAIN improvement, `type=ALL rows=173 + filesort` → `type=eq_ref key=PRIMARY rows=1`), escapes the LIKE search term, eager-loads `iconfile`, and adds a stable `users.id` tiebreak so pagination can't skip/repeat rows across pages. That's real logic, just concentrated on the read side instead of the write side. The "proxy" description undersells it.

On repository vs. service: this query doesn't have a clean home in a repository. `UserRepository`/`DungeonRouteRepository` are each anchored to one model (`parent::__construct(User::class)`), but this is a `users` listing driven by an aggregate over `dungeon_routes` — it belongs to neither model's repository without either duplicating the query or having one repository reach into another's table. `DungeonRouteRepository` does host cross-model reads (`findSimilarRoutes`, `WeeklyRoute`, `KillZoneEnemyForces`), so repositories aren't strictly single-table — but they're still owned by one model, and there's no natural owner here. A Service is the right place for logic that spans models.

On the third-flavor worry: I checked every `*Builder*.php` in the codebase. They all exist already — `app/Service/CombatLog/Builders/` and `app/Logic/CombatLog/**/*Builder.php` — but every one of them is an *object-construction* builder (assembling a `DungeonRoute`/combat entity from parsed log events), not a query builder. There is currently zero precedent anywhere in the codebase for a builder class that wraps Eloquent query construction. So your instinct is correct: introducing one now genuinely would be a new, fourth flavor (Service / Repository / object-Builder / query-Builder), not reuse of an existing pattern. I'd avoid it.

The strongest precedent for what's already in this PR is `DiscoverService::popularBuilder()` (`app/Service/DungeonRoute/DiscoverService.php`) — a private method building an Eloquent query directly against `DungeonRoute::query()`, no repository involved, for the same kind of "browse/directory" read. `DungeonRouteSearchService` takes the opposite route and delegates to `DungeonRouteRepository::findSimilarRoutes()`/`findRoutes()` — but those are single-model, reusable queries, which is exactly when a repository method earns its keep.

So the dividing line I'd commit to: if a query is naturally scoped to one model and is (or is likely to become) reusable, it's a repository method. If it's cross-model, page-specific, or one-off, it's a private builder-style method inside the Service — promote it to its own class (mirroring `Service/CombatLog/Builders/`) only once a second caller actually needs it, not preemptively. `CreatorDirectoryService` is the second case today, and this PR already does the right thing.

Wotuu: I would say that the repository that owns the query is a model stat that it returns.

If I'm looking for a list of dungeon roots, I want to look in the dungeon root repository, and I don't care if it needs to join 10 tables to do it.

The results are dungeonroute, so it goes in a dungeonroute repository.

If you think that we can refactor it into the dungeon repository, we can do that. If you still think that's not a good idea and we should keep it as it is, you can keep it as it is. Either way, when you're done, we can merge this merge request.

Wotuu: :robot: Applying your rule literally: `buildListedCreatorsQuery()` returns `Builder<User>` — the `SELECT` is `users.*` (plus the aggregate column), joined against a `dungeon_routes` subquery, not the other way around. So under "the repository that owns the query is the model it returns," this belongs in `UserRepository`, not `DungeonRouteRepository`. Your `DungeonRoute` example reads as illustrative of the principle rather than a literal instruction for this query — happy to be corrected if you meant it literally.

Done: moved the method onto `UserRepository::buildListedCreatorsQuery()` (interface + implementation, following the existing `EnemyRepository`-style custom-method pattern), preserving the join shape, config threshold, LIKE escaping (still service-side, since that's request-driven filtering, not part of the listing itself), `iconfile` eager load, and the `users.id` tiebreak exactly. `CreatorDirectoryService` now injects `UserRepositoryInterface` and only owns pagination + the search filter.

Split the docblock along the interface/implementation line: the interface keeps the contract (threshold + opt-out semantics, ordering guarantee), the implementation keeps the EXPLAIN-verified join-strategy rationale, matching how custom repository methods are documented elsewhere in the codebase.

Added `tests/Feature/App/Repository/UserRepositoryTest.php` (7 tests) covering the threshold, the opt-out flag, the `iconfile` eager load, the `published_route_count` column surviving the move, and the ordering + tiebreak — nothing exercised those last two directly before, only indirectly through the controller test.

`composer run fix` / `composer run analyse` clean, all `CreatorDirectoryControllerTest` + new `UserRepositoryTest` tests green (15/15).

Pushed as `22d2abfc3`.

Issue #3650, opened 2026-07-21
Batch per-pull enemy-forces queries on discover leaderboard cards (follow-up to #3447)


Wotuu:
## Background

In #3447 (Route poster card / discover leaderboard rework), a cold review flagged an N+1 query
pattern in `DungeonRouteRepository::getEnemyForcesPerKillZone()`, used by both
`resources/views/common/dungeonroute/cardrow.blade.php` and `cardhero.blade.php` to render the
per-pull "fingerprint" bars on a route card.

Original finding: https://github.com/RaiderIO/keystone.guru/pull/3447#discussion_r3621311540

> **Efficiency — per-card N+1.** `getEnemyForcesPerKillZone()` issues one raw `DB::select` (a
> multi-join aggregate over `dungeon_routes`/`kill_zones`/`enemies`/`npcs`/`npc_enemy_forces`) per
> rendered card, and `popularBuilder()` does not eager-load `killZones`. Only the rendered HTML is
> cached, not the query result, so on a cold card cache one leaderboard page (up to 18 rows + up to
> 3 hero cards, same pattern in `cardhero.blade.php`) fires ~21 separate aggregate queries, scaling
> linearly with page size and repeated per cache variant (locale/admin).

This was deliberately deferred rather than fixed inline in #3447 — see the reply at
https://github.com/RaiderIO/keystone.guru/pull/3447#discussion_r3621458739 — because batching
requires restructuring how routes flow from the controller/service layer into these cards (today
each card independently resolves its own forces via `getEnemyForcesPerKillZone()`), and interacts
with the existing "only rendered HTML is cached, not the query result" caching behavior in the
same repository.

## Remaining work

- Batch the per-pull enemy-forces computation for an entire leaderboard page (up to 18 rows + up
  to 3 hero cards) into one grouped query instead of one query per card, OR derive the per-pull
  totals from an eager-loaded `killZones` relation on the routes collection passed into the cards.
- Likely touches: `app/Service/DungeonRoute/DevDiscoverService.php` /
  `DiscoverService` (wherever `popularBuilder()` lives), the controller(s) that pass routes into
  `cardrow.blade.php`/`cardhero.blade.php`, and
  `app/Repositories/Database/DungeonRoute/DungeonRouteRepository.php::getEnemyForcesPerKillZone()`.

## Risks / why it's non-trivial

- `popularBuilder()` currently does not eager-load `killZones`; adding it (or a batched forces
[...]

PR #3677, opened 2026-07-26
#1769 Featured creators rail on the reworked per-dungeon route page

Wotuu:
:robot: Part of #1769.

**Rewritten on 2026-08-16.** The original version of this MR put a featured-creators row on the discover *landing* (`dungeonroute.discover.discover`), built against that page's legacy `discover.panel` layout. It has been rewritten for the page people actually reach. #3675 and #3676 have since merged, so this no longer carries their commits.

## Why it moved

"Browse routes" does not land on the discover landing. It redirects by game version and then by the viewer's dungeon context onto the **per-dungeon overview** (`/routes/retail/<dungeon>`), and that page has since been rebuilt behind `DungeonRouteListRework` — a cinematic hero band, a ranked leaderboard, compact pagination (#3447, #3650, #3978). A grid of bordered creator cards on the old landing no longer describes where creators should appear, or what they should look like when they get there.

| | Before | After |
|---|---|---|
| Page | discover landing (`/routes/retail/season/N`) | per-dungeon overview (`/routes/retail/<dungeon>`) |
| Position | above the route lists | a one-line 94px rail above the hero band |
| Markup | `@include('creator.card')` in a `row-cols` grid | its own horizontal entries in a scrolling shelf |
| Heading | `h2.h4` + right-aligned "See all »" cell | a centred label line that is itself the directory link |
| Flags | `CreatorProfiles` | `CreatorProfiles` **and** `DungeonRouteListRework` |

featured creators rail

## The decisions worth reviewing

**It cannot sit between the hero band and the leaderboard**, which looks like the natural middle ground for a section that shouldn't outrank the routes. With no Raider.IO weekly routes the hero band is the top three of the same popular query, and the leaderboard's `startRank` continues from it — so a section wedged in there splits ranks 1‑2‑3 from 4‑onwards. It goes above the band instead.

**Opening the page is what makes it a rail rather than a section.** A six-across grid of tiles here would cold-open the page on creators and push the routes the visitor came for below the fold. One row of horizontal entries costs 94px and reads as a caption above the hero.

**It took two passes to stop reading as a list of tags**, and neither cause was size. The label sat inline ahead of the entries — `label: item item item` on one line is the grammar of a tag list. And each entry carried a single short string, because the first compaction pass moved the route count into a tooltip. So the label now has its own line, and the count is back under the name: it is the evidence that makes a featured creator worth a click, and two lines of real information is what makes an entry read as a person rather than a label. Entries also have a resting surface, not just a hover one.

**The entries are a shelf that scrolls, not a row that wraps** — at every width, not as a phone special case. With room, `safe center` centres them and no scrollbar appears; without, it scrolls with the next entry peeking in. Wrapping left an orphan at most widths (3+1 at 1280px, one-per-row and 288px tall on a phone), and the rail's usable width is not even monotonic across breakpoints — `discover col-xl-8` only applies from 1200px up, so 1199px is *wider* than 1280px — so chasing it with media queries does not converge. `flex-start` is declared before `safe center` because plain `center` plus overflow strands the leading entries in browsers without `safe`.

`featured_count` is 4, not 6, so the rail is one clean row at its 70rem measure. Purely a presentation call — it is one config line.

**The entries are deliberately not `creator.card`.** That tile is the *directory's* skin and its fixed vertical composition cannot compress to a rail. The rail borrows the leaderboard's vocabulary instead — page-background rows, neutral-gray hairlines, a gray hover wash, the same 70rem measure that `.dungeonroute_leaderboard` and `.discover_pagination` already share — with the avatar ring picking up `--theme-btn-accent` on hover/focus as the one accent moment. Same creator, different skin, the way a route already renders as a poster, a hero *and* a leaderboard row. `creator.card` itself is untouched, so the directory is unchanged.

**The focus ring is inset, not outset.** `overflow-x: auto` makes the shelf a scrollport on *both* axes, and an outline contributes nothing to the scrollable overflow region — so a positive `outline-offset` is simply clipped: the top 4px of every entry's ring, and the left edge of the first entry's at scroll position 0. Padding on the container does not close it, because the scrollbar gutter sits inside the padding box and eats `padding-bottom` exactly when the shelf overflows. `outline-offset: -2px` draws the ring inside the entry's own border box, where nothing can crop it at any width or scroll position:

focus ring

## Three things caught by screenshotting, not by tests

**Initials state their colour rather than inheriting it.** Each theme's `.theme.<name> a` rule (0,2,1) outweighs a single class on the entry (0,1,0), so `color: inherit` on a child of the entry link resolves to the theme's *link* colour — the initials rendered accent green under darkly and purple under vapor, a second accent competing with the hover ring and low-contrast lettering on the neutral disc. Same trap family as the `bg-body-*` note this MR adds to the `blade-expert` skill.

**A keyboard-only regression, invisible to the whole suite** — the clipped focus ring above. Verified fixed by tabbing to the entry in headless Chrome (real key presses, so `:focus-visible` genuinely applies rather than a scripted `.focus()`): `matchesFocusVisible: true`, computed ring `2px solid rgb(0, 188, 140)`, box inside the container on every side.

**A name clipped by `text-overflow` had no reveal path.** Reshaping into a rail moved the route count into the entry's `title` and dropped the name's own `title` in the same edit, and the CSS comment then documented the intent rather than the shipped markup. The tooltip now carries both, through a `view_creator.featured.entry_title` key so the join stays translatable.

## Backend

`getFeaturedCreators()` and `paginateCreators()` still share `buildListedCreatorsQuery()`, so a creator who opts out of the directory cannot still be featured; there is a test for exactly that.

The one behavioural change is that **`getFeaturedCreators()` is now cached** (`keystoneguru.creators.featured_ttl`, 1 hour). The shared query derives its published route counts with a `GROUP BY` over the whole `dungeon_routes` table, which materialises in full before `LIMIT` can take four rows off it — ~36 ms warm on a 507k-route database — and unlike the hero band the rail renders on *every* page of results, not just the first. The result varies by neither viewer nor dungeon nor page, so there is nothing to key on and no invalidation question. `CacheService::remember()` is a passthrough while caching is disabled, so development and the test suite are unaffected.

The composer binding stays. Its original justification ("three controller methods render the landing") no longer applies, but the pattern still buys the thing that matters: adding the rail to another discover surface is one `@include` that can never silently render empty because someone forgot to pass the data — the same reason `DiscoverSearchComposer` and `DiscoverAffixGroupComposer` exist.

## Verification

**6 feature tests**, all passing — eligibility, directory opt-out honoured on the rail too, limit respected, the rail rendered with both flags on, hidden with `CreatorProfiles` off, and hidden with `DungeonRouteListRework` off (so it cannot leak into the legacy layout it was never styled for). The limit test is mutation-checked: commenting out `->limit()` makes it fail.

**Browser-verified** in headless Chrome under **darkly, lux and vapor**, at **1600px, 1280px and 390px** — one row everywhere — every theme lives in the one class-scoped stylesheet, so swapping the `html` class is a faithful preview.

1280px (scrolls, with the fourth entry peeking) / lux / vapor:

lux
vapor

Mobile (390px) — the same shelf, 94px:

mobile

**PHPStan / php-cs-fixer** clean on the changed files. (Locally PHPStan reports 56 `class.notFound` errors, all in `app/Logic/MDT/IO/MDT2Codec.php` from `spomky-labs/cbor-php` missing in this worktree's vendor dir — pre-existing environment skew, unrelated to this diff.)

https://claude.ai/code/session_01F7ZNExCW4A9FkGuFr74XEr

### DIFF
--- a/resources/views/creator/featured.blade.php
+++ b/resources/views/creator/featured.blade.php
@@ -4,10 +4,18 @@ use App\Models\User;
 use Illuminate\Database\Eloquent\Collection;
 
 /**
- * The featured-creators row on the discover landing page.
+ * The featured-creators strip that closes the reworked per-dungeon route page.
+ *
+ * It sits below the leaderboard rather than above the routes on purpose: the person finding a route
+ * outranks the person who built one, so creators are the hand-off once the routes have been
+ * scanned, not the thing standing between the visitor and them.
+ *
+ * The entries are deliberately not creator.card - that tile is the directory's skin, and stacking a
+ * grid of cards under the leaderboard would reintroduce the visual weight the page has just spent
+ * two sections shedding. See creator-featured.css.
  *
  * featuredCreators is supplied by FeaturedCreatorsComposer rather than by the controller, because
- * three different controller methods render the discover landing.
+ * several controller methods render the pages this partial is included from.
  *
  * @var Collection<int, User> $featuredCreators
  */
@@ -15,24 +23,45 @@ use Illuminate\Database\Eloquent\Collection;
 $featuredCreators ??= collect();
 ?>
 @if($featuredCreators->isNotEmpty())
-    <div class="discover_panel px-xl-2">
-        <div class="row mt-4 align-items-center">
-            <div class="col">
-                <h2 class="h4 mb-0">
+    <div class="row mt-5 align-items-center discover_section_header">
+        <div class="col">
+            <h5 class="mb-0 text-center">
+                {{-- The directory link lives in the heading itself, matching the "Weekly routes" section above --}}
+                <a href="{{ route('creators.index') }}" title="{{ __('view_creator.featured.see_all') }}">
                     {{ __('view_creator.featured.title') }}
-                </h2>
-            </div>
-            <div class="col-auto">
-                <a href="{{ route('creators.index') }}">
-                    {{ __('view_creator.featured.see_all') }} &raquo;
+                    <i class="fas fa-angle-right" aria-hidden="true"></i>
                 </a>
-            </div>
+            </h5>
         </div>
+    </div>
 
-        <div class="row g-3 mt-1 row-cols-2 row-cols-md-4 row-cols-xl-6">
+    <div class="discover_creator_strip mt-2">
+        <div class="row g-2 row-cols-2 row-cols-sm-3 row-cols-lg-6">
             @foreach($featuredCreators as $creator)
+                <?php
+                // Set by the withCount() in CreatorDirectoryService; fall back rather than lazy-count per entry
+                $publishedRouteCount = $creator->published_route_count ?? 0;
+                ?>
                 <div class="col">
-                    @include('creator.card', ['creator' => $creator])
+                    <a href="{{ route('profile.view', ['user' => $creator]) }}" class="discover_creator_entry">
+                        @if($creator->iconfile !== null)
+                            <img src="{{ $creator->iconfile->getURL() }}"
+                                 alt="{{ $creator->name }}"
+                                 class="discover_creator_avatar"/>
+                        @else
+                            <div class="discover_creator_initials" aria-hidden="true">
+                                {{ $creator->initials }}
+                            </div>
+                        @endif
+
+                        <div class="discover_creator_name" title="{{ $creator->name }}">
+                            {{ $creator->name }}
+                        </div>
+
+                        <div class="text-body-secondary small">
+                            {{ trans_choice('view_creator.card.route_count', $publishedRouteCount, ['count' => $publishedRouteCount]) }}
+                        </div>
+                    </a>
                 </div>
             @endforeach
         </div>
--- a/resources/views/creator/featured.blade.php
+++ b/resources/views/creator/featured.blade.php
@@ -15,12 +15,12 @@ use Illuminate\Database\Eloquent\Collection;
  * two sections shedding. See creator-featured.css.
  *
  * featuredCreators is supplied by FeaturedCreatorsComposer rather than by the controller, because
- * several controller methods render the pages this partial is included from.
+ * several controller methods render the pages this partial is included from. It is deliberately not
+ * defaulted here: the composer is bound to this exact view, so an undefined variable means the
+ * binding is gone - which should be a loud error, not a section that silently stops rendering.
  *
  * @var Collection<int, User> $featuredCreators
  */
-
-$featuredCreators ??= collect();
 ?>
 @if($featuredCreators->isNotEmpty())
     <div class="row mt-5 align-items-center discover_section_header">
@@ -45,8 +45,9 @@ $featuredCreators ??= collect();
                 <div class="col">
                     <a href="{{ route('profile.view', ['user' => $creator]) }}" class="discover_creator_entry">
                         @if($creator->iconfile !== null)
+                            {{-- Decorative: the name is already the link's own text, right below it --}}
                             <img src="{{ $creator->iconfile->getURL() }}"
-                                 alt="{{ $creator->name }}"
+                                 alt=""
                                  class="discover_creator_avatar"/>
                         @else
                             <div class="discover_creator_initials" aria-hidden="true">
--- a/resources/views/creator/featured.blade.php
+++ b/resources/views/creator/featured.blade.php
@@ -1,0 +1,40 @@
+<?php
+
+use App\Models\User;
+use Illuminate\Database\Eloquent\Collection;
+
+/**
+ * The featured-creators row on the discover landing page.
+ *
+ * featuredCreators is supplied by FeaturedCreatorsComposer rather than by the controller, because
+ * three different controller methods render the discover landing.
+ *
+ * @var Collection<int, User> $featuredCreators
+ */
+
+$featuredCreators ??= collect();
+?>
+@if($featuredCreators->isNotEmpty())
+    <div class="discover_panel px-xl-2">
+        <div class="row mt-4 align-items-center">
+            <div class="col">
+                <h2 class="h4 mb-0">
+                    {{ __('view_creator.featured.title') }}
+                </h2>
+            </div>
+            <div class="col-auto">
+                <a href="{{ route('creators.index') }}">
+                    {{ __('view_creator.featured.see_all') }} &raquo;
+                </a>
+            </div>
+        </div>
+
+        <div class="row g-3 mt-1 row-cols-2 row-cols-md-4 row-cols-xl-8">
+            @foreach($featuredCreators as $creator)
+                <div class="col">
+                    @include('creator.card', ['creator' => $creator])
+                </div>
+            @endforeach
+        </div>
+    </div>
+@endif
--- a/resources/views/creator/featured.blade.php
+++ b/resources/views/creator/featured.blade.php
@@ -29,7 +29,9 @@ use Illuminate\Database\Eloquent\Collection;
  */
 ?>
 @if($featuredCreators->isNotEmpty())
-    <div class="discover_creator_rail mt-4">
+    {{-- A named landmark rather than a heading: dropping the heading row left seven sibling links
+         with nothing to announce or navigate to them by. What this is, is navigation. --}}
+    <nav class="discover_creator_rail mt-4" aria-label="{{ __('view_creator.featured.title') }}">
         {{-- The directory link is the rail's own label, so the rail needs no separate heading row --}}
         <a href="{{ route('creators.index') }}"
            class="discover_creator_rail_label"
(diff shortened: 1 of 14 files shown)


resources/views/creator/featured.blade.php
```diff
@@ -0,0 +1,40 @@
...
+                </h2>
+            </div>
+            <div class="col-auto">
+                <a href="{{ route('creators.index') }}">
+                    {{ __('view_creator.featured.see_all') }} &raquo;
+                </a>
+            </div>
+        </div>
+
+        <div class="row g-3 mt-1 row-cols-2 row-cols-md-4 row-cols-xl-8">
```
Wotuu: :robot: `row-cols-xl-8` is not a class Bootstrap generates, so the featured row never renders 8-across.

Bootstrap's `row-cols-*` utilities are generated from `$grid-row-columns`, which defaults to `6` (`node_modules/bootstrap/scss/_variables.scss:521`) and is not overridden anywhere under `resources/assets/sass/`. The compiled bundle confirms it:

```
$ grep -o "row-cols-xl-[0-9]*" public/css/app-<sha>.css | sort -u
row-cols-xl-1 ... row-cols-xl-6      # no -7, no -8
```

So `row-cols-xl-8` is inert and the row falls back to the last class that does exist, `row-cols-md-4`. Combined with `featured_count => 8` added in this same diff, the result at xl+ is **two rows of four**, and on mobile `row-cols-2` gives **four rows of two** stacked above the popular/new route lists.

The two knobs were clearly picked as a pair, so it is one fix rather than two: either

- `featured_count => 6` + `row-cols-xl-6` (valid, genuinely one row), or
- keep 8 and make the two-row layout deliberate with `row-cols-xl-4`.

The discover landing renders inside `col-xl-8 offset-xl-2` ([discover.blade.php#L1-L7](https://github.com/RaiderIO/keystone.guru/blob/58bdd857c112fbfb277406ff4916c381580188fc/resources/views/dungeonroute/discover/discover.blade.php#L1-L7)) — two thirds of the page width — so 8 cards across was going to be very cramped even if the class existed. That is why the suggestion is to lower the count rather than to hand-write a `row-cols-xl-8` rule.

For contrast, the directory partial sticks to valid values: [creator/directory.blade.php#L53-L55](https://github.com/RaiderIO/keystone.guru/blob/58bdd857c112fbfb277406ff4916c381580188fc/resources/views/creator/directory.blade.php#L53-L55) uses `row-cols-2 row-cols-md-3 row-cols-xl-4`.

Config side: [config/keystoneguru.php#L272-L276](https://github.com/RaiderIO/keystone.guru/blob/58bdd857c112fbfb277406ff4916c381580188fc/config/keystoneguru.php#L272-L276).

tests/Feature/Controller/FeaturedCreatorsTest.php
```diff
@@ -0,0 +1,171 @@
...
+        Feature::for($viewer)->activate(CreatorProfiles::class);
+
+        try {
+            // Act
+            $response = $this->actingAs($viewer)->followingRedirects()->get($this->discoverUrl());
+
+            // Assert
+            $response->assertOk();
+            $response->assertSee(__('view_creator.featured.title'));
+            $response->assertSee(__('view_creator.featured.see_all'));
```
Wotuu: :robot: This test's Arrange block is dead — it passes on the seeded DB's pre-existing creators, so it cannot distinguish "the featured row works" from "the seed data happens to contain creators".

The partial only guards on `$featuredCreators->isNotEmpty()`, and the row is `ORDER BY published_route_count DESC ... LIMIT 8`. The `$creator` arranged here sits at exactly `min_published_routes`, so by the test file's own reasoning a few lines up — *"a brand new creator at the threshold sits below the established ones"*, [FeaturedCreatorsTest.php#L26-L28](https://github.com/RaiderIO/keystone.guru/blob/58bdd857c112fbfb277406ff4916c381580188fc/tests/Feature/Controller/FeaturedCreatorsTest.php#L25-L29) — it will normally *not* be in the first 8. The two assertions are on the section heading and the "See all" link, both of which render for any non-empty result.

On the seeded test DB (204 users, 115 world-published routes here) there are already qualifying creators, so deleting the whole Arrange would leave the test green. That makes it a false-confidence test rather than a coverage gap: a regression that features the wrong people, or that drops `$creator` from the query entirely, still passes.

Cheap fix, in the spirit of the service-level tests above it: assert on the creator you arranged.

```php
$response->assertSee($creator->name);
```

(or keep the heading assertions and additionally assert `getFeaturedCreators(PHP_INT_MAX)` contains `$creator->id`, matching the pattern used by `getFeaturedCreators_givenACreatorAboveTheThreshold_includesThem`).

Note this does not apply to the negative test above it — `discover_givenFeatureInactive_doesNotRenderTheFeaturedRow` is genuinely load-bearing as written.

Wotuu: :robot: **Retracting this — it is wrong.** Please disregard the comment above.

I checked "does the seeded DB already contain qualifying creators?" against my local dev database, which has 12 of them (the 8th-ranked has 4 world-published routes, so a fresh creator at the threshold of 3 would indeed never make the top 8). That database is not what the test runs against.

CI builds the DB from scratch — `php artisan migrate` + `php artisan db:seed` + `LaratrustSeeder`, per [.github/actions/php-ci-setup/action.yml](https://github.com/RaiderIO/keystone.guru/blob/58bdd857c112fbfb277406ff4916c381580188fc/.github/actions/php-ci-setup/action.yml) — and none of the seeders create a qualifying creator. The only routes seeded are the 36 demo routes in `database/seeders/dungeondata/**/dungeonroutes.json`, and every one of them is `published_state_id` 1 or 3:

```
published_state_id counts: {1: 19, 3: 17}      # unpublished / world_with_link
author_id counts:          {1: 36}
```

`buildListedCreatorsQuery()` filters on `WORLD` (`published_state_id = 4`), so the seeded set contributes zero eligible creators. In CI the Arrange block is therefore fully load-bearing: delete it and `$featuredCreators` is empty, the `@if($featuredCreators->isNotEmpty())` guard suppresses the row, and both assertions fail. The test does exactly what its name says.

The one real residual is much smaller than what I claimed, and probably not worth changing: the assertions are on the section heading rather than on `$creator`, so the test's strength depends on the DB being freshly seeded. It is solid in CI and weakens on a long-lived dev database. Given CI reseeds every run, that seems fine as-is — and `assertSee($creator->name)` would be the wrong fix anyway, since it would be the assertion that breaks first on a dev DB.

Apologies for the noise.

Wotuu:
:robot: Cold review (opus): 1 finding posted. A second candidate (featured-row test asserting on the section heading rather than on the arranged creator) was posted and then retracted in-thread as a false positive — it assumed the CI test DB contains pre-existing qualifying creators, which it does not.

resources/views/creator/featured.blade.php
Wotuu: :robot: Follow-up: the verification screenshot in the PR body already shows this, which is worth pointing out because it means the browser check ran but the layout regression slipped through it.

`pr/1769-discover-featured.png` is **1585px wide**, comfortably inside the `xl` breakpoint (≥1200px), so it should show one row of eight. It shows two rows of four:

```
row 1:  AD Admin (58)   DR Dratnos (12)   YU Yumytv (9)   NA Naowh (7)
row 2:  Quazii (6)      Tettles (5)       Growl (5)       Jdotb (4)
```

That is `row-cols-md-4` taking effect because `row-cols-xl-8` does not exist — exactly the fallback described above, captured on the real page.

Not a criticism of the verification itself; the screenshot is genuinely the real page and it did its job. It is just that "8 tiles rendered" and "8 tiles rendered *in one row*" look similar enough at a glance that the check passed. Worth a note for the next featured-row tweak: confirm the row count, not just the tile count.

Wotuu: :robot: Fixed in 3e303b9f5: dropped `featured_count` to 6 (config/keystoneguru.php) and changed `row-cols-xl-8` to `row-cols-xl-6` (resources/views/creator/featured.blade.php) — verified against the compiled bundle that `row-cols-xl-6` exists and `-7`/`-8` do not. Went with the "6, one row" option per your preference; one clean row at xl+ instead of two cramped rows of four.

tests/Feature/Controller/FeaturedCreatorsTest.php
Wotuu: :robot: Fixed in 3e303b9f5: added an assertion that the arranged `$creator` is actually eligible via `getFeaturedCreators(PHP_INT_MAX)`, matching the pattern in `getFeaturedCreators_givenACreatorAboveTheThreshold_includesThem`. Went with that over a direct `assertSee($creator->name)` because the discover landing page also renders a "new routes" list that could independently surface the same creator name, which would make that assertion pass without ever proving the featured row rendered them.

Wotuu:
shared ancestor.

```sh
```

already in master.

updated 2026-08-04).

Wotuu:
:robot: Rebased onto current `master` (both predecessors, #3675 and #3676, are now merged).

Used the layered-rebase technique from the previous note: `git rebase --onto origin/master 77a86f7af HEAD`, dropping the duplicate predecessor commits.

One real conflict, in `app/Service/Creator/CreatorDirectoryService.php`: this branch's `#1769 Make creator eligibility index-usable...` commit optimized `buildListedCreatorsQuery()` in place in the service, but master's squashed #3676 had *also* relocated that same query (already optimized, identical logic) into `UserRepository::buildListedCreatorsQuery()`. Resolved by dropping the now-redundant service-local optimization and pointing `getFeaturedCreators()` at the repository method instead, matching current master's `paginateCreators()`.

Also repaired a commit-message artifact from the conflict resolution: `git rebase --continue`'s editor path stripped the leading `#1769 ...` line (read as a comment) and promoted the body's first paragraph to the subject. Fixed via cherry-pick-rebuild with `--cleanup=verbatim`.

Pre-existing, unrelated to this diff: `CreatorDirectoryControllerTest::index_givenACreatorAboveTheThreshold_listsThem` fails locally against the current worktree DB (75k seeded users push the pagination page past page 1) — this test file is untouched by this branch, entirely from the already-merged #3676.

CI running on the fresh push.

PR #3680, opened 2026-07-26
#3679 Fix the flaky raid-marker mapping-version test

### LINKED ISSUE #3679: Flaky test: DungeonRouteEnemyRaidMarkerMappingVersionTest fails ~1 run in 20 on any branch
:robot: `DungeonRouteEnemyRaidMarkerMappingVersionTest::upgradeMappingVersion_givenNpcStillExistsInNewMappingVersion_preservesRaidMarker` fails roughly **1 run in 20**, on any branch, with:

```
TypeError: createRaidMarker(): Argument #2 ($enemy) must be of type App\Models\Enemy, null given
  tests/Feature/App/Service/DungeonRoute/DungeonRouteEnemyRaidMarkerMappingVersionTest.php:142
  tests/Feature/App/Service/DungeonRoute/DungeonRouteEnemyRaidMarkerMappingVersionTest.php:51
```

Seen on an unrelated PR's `php-tests (feature-other)` job. Introduced with the raid-marker work in #3630.

## Root cause

The test picks a random retail dungeon **first**, and only **then** looks for an enemy uniquely
identified by its `(mdt_npc_id/npc_id, mdt_id)` pair on that dungeon's current mapping version:

```php
$dungeon = $this->getRetailDungeon();          // random
$enemy   = Enemy::where('mapping_version_id', $existingMV->id)
    ->whereRaw('( ... ) = 1')                  // uniquely identified only
    ->inRandomOrder()
    ->first();                                 // -> null for some dungeons
```

The uniqueness filter is correct and deliberate (it keeps the test aligned with what
`DungeonRouteEnemyRaidMarkerRepository` actually resolves). The problem is that it can legitimately
match nothing: measured against the seeded database, **3 of the 70 retail dungeons have zero
uniquely-identified enemies** —

- The Arcway
- Cathedral of Eternal Night
- Maw of Souls

In those dungeons every enemy shares its `(npc_id, mdt_id)` pair with at least one sibling. When
`getRetailDungeon()` draws one of them, `$enemy` is `null` and the test dies in `createRaidMarker()`.

3/70 ≈ **4%** per run, which matches the observed rate.

## Fix

Search for a dungeon that actually has a qualifying enemy instead of sampling one and hoping —
exactly the shape the same file already uses for `getRetailDungeonWithNullMdtIdEnemy()`. The helper
returns only once it has a non-null enemy, so it cannot hand back `null` by construction, and skips
if no dungeon qualifies at all.

## Why it matters now

It is not tied to any one branch, so it randomly reddens unrelated MRs and invites re-running CI
until it passes — which is exactly the habit that hides real failures.

Wotuu:
:robot: Closes #3679

Test-only, one file, branched straight off `master` so it can merge immediately and the three open creator MRs (#3675, #3676, #3677) pick it up from `master` rather than each carrying their own copy.

## Two failures, one root cause

`DungeonRouteEnemyRaidMarkerMappingVersionTest` resolves a raid marker across a mapping-version upgrade by matching on the `(mdt_npc_id/npc_id, mdt_id)` pair. That pair is **not unique** within a mapping version — some NPCs are placed twice under the same MDT clone index — so an ambiguous enemy cannot be re-resolved after the upgrade and its marker is dropped.

Both tests in the class picked an enemy without guaranteeing uniqueness, so both were data-dependent:

**1. `upgradeMappingVersion_givenNpcStillExistsInNewMappingVersion_preservesRaidMarker`** — picked a random retail dungeon *first*, then filtered for a uniquely-identified enemy. **3 of the 70** retail dungeons have none at all:

- The Arcway
- Cathedral of Eternal Night
- Maw of Souls

Drawing one of those returned `null` and the test died with `TypeError: createRaidMarker(): Argument #2 ($enemy) must be of type App\Models\Enemy, null given`. ~4% per run — this is what went red on #3675's `php-tests (feature-other)`.

**2. `upgradeMappingVersion_givenNpcWithNullMdtIdStillExistsInNewMappingVersion_preservesRaidMarker`** — applied no uniqueness filter at all. Only **152 of the 1549** null-`mdt_id` enemies in the seeded database are uniquely identified, so ~90% of draws picked an ambiguous one, the marker was dropped, and `assertNotNull($fresh)` failed.

The second one is worth flagging: **it fails 10 times out of 10 locally on plain `master`**, and has evidently only been passing in CI on friendlier seed data. It was going to start reddening MRs the moment that data shifted.

## Fix

Search shuffled dungeons for one that actually has a qualifying enemy, instead of sampling a dungeon and hoping — the shape the file already used for `getRetailDungeonWithNullMdtIdEnemy()`. Both helpers now only return once they hold a non-null enemy, so neither can hand back `null` by construction, and each skips if no dungeon qualifies.

The uniqueness filter itself is unchanged and deliberate: it keeps the tests aligned with what `DungeonRouteEnemyRaidMarkerRepository` actually resolves in production, rather than comparing against an arbitrary sibling. Constraining the null-`mdt_id` test the same way does not weaken it — that test is about the join being null-safe (`<=>` rather than `=`), which an unambiguous enemy exercises just as well.

## Verification

| | before | after |
|---|---|---|
| whole class, 15 consecutive local runs | ~9/15 failing | **0/15 failing** |
| null-mdt test alone on plain `master` | 10/10 failing | — |

PHPStan clean, php-cs-fixer clean. No production code touched.

## Note

I originally fixed only the first test, inside #3675. This MR hoists it out to `master` and adds the second fix, which only surfaced once the branch was rebuilt on current `master`. The copy in #3675 will be dropped so the fix lives in exactly one place.

### DIFF
--- a/tests/Feature/App/Service/DungeonRoute/DungeonRouteEnemyRaidMarkerMappingVersionTest.php
+++ b/tests/Feature/App/Service/DungeonRoute/DungeonRouteEnemyRaidMarkerMappingVersionTest.php
@@ -27,25 +27,9 @@ final class DungeonRouteEnemyRaidMarkerMappingVersionTest extends DungeonRouteSa
     public function upgradeMappingVersion_givenNpcStillExistsInNewMappingVersion_preservesRaidMarker(): void
     {
         // Arrange
-        $dungeon    = $this->getRetailDungeon();
-        $existingMV = $dungeon->getCurrentMappingVersion();
+        [$dungeon, $existingMV, $enemy] = $this->getRetailDungeonWithUniquelyIdentifiedEnemy();
         // Cloning a new mapping version also clones every enemy (same npc_id/mdt_id, new ids)
         $newMV = $this->createNewerMappingVersion($dungeon, $existingMV);
-
-        // npc_id/mdt_id is not guaranteed unique within a mapping version - a handful of NPCs are
-        // placed twice under the same MDT clone index. Excluding those keeps this test aligned
-        // with what production actually resolves (see DungeonRouteEnemyRaidMarkerRepository),
-        // instead of occasionally comparing against an arbitrary sibling via firstOrFail() below.
-        /** @var Enemy $enemy */
-        $enemy = Enemy::where('mapping_version_id', $existingMV->id)
-            ->whereRaw('(
-                SELECT COUNT(*) FROM enemies e2
-                WHERE e2.mapping_version_id = enemies.mapping_version_id
-                  AND COALESCE(e2.mdt_npc_id, e2.npc_id) = COALESCE(enemies.mdt_npc_id, enemies.npc_id)
-                  AND e2.mdt_id <=> enemies.mdt_id
-            ) = 1')
-            ->inRandomOrder()
-            ->first();
 
         $route      = DungeonRoute::factory()->create(['dungeon_id' => $dungeon->id, 'mapping_version_id' => $existingMV->id]);
         $raidMarker = $this->createRaidMarker($route, $enemy);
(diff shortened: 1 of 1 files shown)


Wotuu (commented):
Ok

PR #3681, opened 2026-07-26
#700 Route collections

### LINKED ISSUE #700: Route collections

Wotuu:
:robot: Closes #700

Fourth MR of the creator arc, stacked on #3675 → #3676 → #3677. Everything is behind the
**`CreatorProfiles`** Pennant flag, so merging changes nothing user-visible until the flag is
flipped. **Review/merge the three predecessors first** — until they merge, this diff contains
them.

> Note: the old `700-route-collections` branch (2023) is untouched — this is a fresh start, as you
> asked for in the issue. This branch is `700-add-route-collections`.

## What it does

A **collection** is a named, shareable list of a user's own routes — "my routes for this week" —
with the same visibility mechanics a route has: a `PublishedState` plus a `public_key`, so it can
be linked around exactly like a route.

- `/collections` — your collections, create / edit / delete
- `/collection/{public_key}` — the public page you link from Twitch/Discord
- Visible to: **Only me / My team / Everyone with the link / Everyone**, mirroring `DungeonRoute`
- "My collections" entry in the user dropdown (flag-gated)

## Decisions worth a look

- **Sharing a collection never publishes the routes in it.** The public page filters every member
  route through the viewer's own `mayUserView()`, so an unpublished route inside a public
  collection stays hidden. Covered by
  `view_givenAnUnpublishedRouteInAPublicCollection_hidesThatRoute`.
- **Own routes only.** A collection may only hold routes you authored (`Rule::exists(...)
  ->where('author_id', $userId)`, same guard the pinned routes use). Widening this later to other
  people's public routes is a one-line validation change — it just is not what the issue asked
  for.
- **Team sharing needs a team you are in.** `published_state=team` requires a `team_id`, and that
  team must be one you are a member of.
- **Naming**: `DungeonRouteCollection` (+ `DungeonRouteCollectionRoute` for the ordered coupling),
  matching its `DungeonRouteFavorite` / `DungeonRouteRating` siblings. `RouteCollection` was ruled
  out — both Laravel and Symfony ship a class by that name.
- **Limits**: 25 collections per user, 50 routes per collection, and a `create-collection` rate
  limiter alongside `create-team`.
- **No foreign keys**, so both delete paths clean up by hand: deleting a collection deletes its
  couplings, and `DungeonRoute::deleting()` now removes the route from any collection it was in.
- **No new JS.** Delete is a plain form post, and the route picker is a native multi-select — so
  nothing needs registering in the bundler.

## Not in this MR

- Categories on collections (#1771) — next MR.
- The category filter on `/creators` (finishes #1769) — the MR after that.
- Pinning collections to the profile podium — last, so #3675 stays stable while under review.

## Verification

- 21 new tests (14 feature, 7 unit), `composer run fix` and `composer run analyse` clean.
- Both pages verified in headless Chrome under a dark **and** a light theme; the light-theme navbar
  labels disappear on untouched pages too (client-side theme-swap artifact, confirmed against
  `/teams`).
- Self-reviewed against the diff; no findings left open.


Wotuu:
:robot: Headless-Chrome verification (worktree stack, logged in as admin with the flag on).

**`/collections` — your collections**

index dark

**`/collections/new` — create form**

new dark

**`/collections/{key}` — edit form**

edit dark

**`/collection/{key}` — the public page you share**

view dark

**Light theme (lux)** — same pages, checked for the `bg-body-*` white-on-white trap. The missing
navbar labels are an artifact of swapping the `html` theme class client-side; `/teams` on the
same branch renders identically.

index light

edit light

resources/views/common/collection/details.blade.php
```diff
@@ -0,0 +1,111 @@
...
+@include('common.general.messages')
+
+@isset($dungeonRouteCollection)
+    {{ html()->modelForm($dungeonRouteCollection, 'PATCH', route('collections.update', ['dungeonRouteCollection' => $dungeonRouteCollection]))->open() }}
+@else
+    {{ html()->form('POST', route('collections.savenew'))->open() }}
+@endisset
+
+<div class="mb-3{{ $errors->has('name') ? ' has-error' : '' }}">
+    {{ html()->label(__('view_collection.details.name') . '<span class="form-required">*</span>', 'name') }}
```
Wotuu: :robot: **Translation keys for a `common/` partial should live under `view_common.*`, not `view_collection.*`**

This partial is `resources/views/common/collection/details.blade.php`, but every key it uses is `view_collection.details.*`. [CLAUDE.md](https://github.com/RaiderIO/keystone.guru/blob/6a9f8948e866817986e23eb8e68b73ede11b6a4f/.claude/CLAUDE.md#L156-L158) says:

> For blade.php files, the translation keys matches exactly the file structure and name. For example, `resources/views/common/footer.blade.php` would have translation keys like `view_common.footer.copyright`.

The direct analogue in the repo follows that — `resources/views/common/team/details.blade.php` uses `view_common.team.details.*`:

https://github.com/RaiderIO/keystone.guru/blob/6a9f8948e866817986e23eb8e68b73ede11b6a4f/resources/views/common/team/details.blade.php#L20-L22

So this block should be `view_common.collection.details.*` in `lang/en_US/view_common.php`. The `view_collection.index.*` / `.new.*` / `.edit.*` / `.view.*` keys used by the page blades under `resources/views/collection/` are correct as-is — only the `details.*` group (used solely by this partial) is in the wrong file.

app/Http/Controllers/DungeonRoute/DungeonRouteCollectionController.php
```diff
@@ -0,0 +1,256 @@
...
+        Gate::authorize('edit', $dungeonRouteCollection);
+
+        /** @var User $user */
+        $user = Auth::user();
+
+        $dungeonRouteCollection->load(['dungeonRoutes']);
+
+        return view('collection.edit', [
+            'dungeonRouteCollection'  => $dungeonRouteCollection,
+            'ownDungeonRoutes'        => $this->getOwnDungeonRoutes($user),
```
Wotuu: :robot: **An admin who opens and saves someone else's collection silently empties it**

`mayUserEdit()` grants edit/delete to `ROLE_ADMIN` as well as the owner, so `Gate::authorize('edit', ...)` lets an admin through to this form:

https://github.com/RaiderIO/keystone.guru/blob/6a9f8948e866817986e23eb8e68b73ede11b6a4f/app/Models/DungeonRoute/DungeonRouteCollection.php#L142-L146

But the form and its validation are hard-scoped to `Auth::id()`, not to the collection's owner:

- `getOwnDungeonRoutes($user)` builds the `<select>` from **the admin's** routes (`where('author_id', $user->id)`), so none of the collection's actual routes appear as options and `in_array($ownDungeonRoute->id, $selectedDungeonRouteIds, true)` never matches - the picker renders with nothing selected.
- `DungeonRouteCollectionFormRequest::rules()` uses `$userId = Auth::id()` for `Rule::exists('dungeon_routes', 'id')->where('author_id', $userId)`, so the admin cannot re-select them either.

Submitting that form posts no `dungeon_routes`, `dungeonRoutes()` returns an empty collection, and `syncDungeonRoutes()` deletes every coupling and inserts none - the collection saves back empty, with no error shown.

https://github.com/RaiderIO/keystone.guru/blob/6a9f8948e866817986e23eb8e68b73ede11b6a4f/app/Http/Controllers/DungeonRoute/DungeonRouteCollectionController.php#L242-L245

Same mismatch on `team_id`: `Rule::exists('team_users', 'team_id')->where('user_id', $userId)` is checked against the admin, so a team-published collection cannot be saved by an admin at all unless they happen to be in that team.

No test covers the admin path (`update_givenAnotherUsersCollection_returnsForbidden` uses a plain user), so CI stays green. Either scope the form and the validation to `$dungeonRouteCollection->user_id` on the update path, or drop admins from `edit` in `mayUserEdit()` so the policy and the form agree.

resources/views/collection/view.blade.php
```diff
@@ -0,0 +1,58 @@
...
+@section('header-title')
+    {{ $dungeonRouteCollection->name }}
+@endsection
+
+@section('content')
+    <div class="card mb-4">
+        <div class="card-body">
+            <div class="text-body-secondary small mb-2">
+                <a href="{{ route('profile.view', ['user' => $dungeonRouteCollection->user]) }}">
+                    {{ __('view_collection.view.by_author', ['author' => $dungeonRouteCollection->user->name]) }}
```
Wotuu, line 30: :robot: **Deleting an account leaves its collections behind, and this page then 500s on a still-shareable public link**

`User::deleting()` cleans up `dungeonRoutes`, `reports`, `patreonUserLink` and team memberships, but nothing removes `dungeon_route_collections`:

https://github.com/RaiderIO/keystone.guru/blob/6a9f8948e866817986e23eb8e68b73ede11b6a4f/app/Models/User.php#L444-L449

`ProfileController::delete()` is a hard `$user->delete()`, and there are no foreign keys, so a world-published collection survives its author. Its `public_key` URL stays live (`mayUserView()` returns `true` for `world` / `world_with_link` without ever looking at the user), but both lines here dereference the now-null `user`: `route('profile.view', ['user' => null])` throws `UrlGenerationException`, and `$dungeonRouteCollection->user->name` is a null property read. The shared link 500s instead of 404ing.

The MR body enumerates two delete paths (collection deleted, route deleted); account deletion is the third. Adding `$user->dungeonRouteCollections()->delete()` to the `User::deleting()` hook covers it - the model's own `deleting()` hook won't fire on that mass delete, so the couplings need `DungeonRouteCollectionRoute::whereIn('dungeon_route_collection_id', ...)->delete()` too, or a per-model loop.

Wotuu:

resources/views/collection/view.blade.php
Wotuu: :robot: Sharpening one point above, because the MR body implies this path is covered and it isn't.

`$user->dungeonRoutes()->delete()` on [User.php#L446](https://github.com/RaiderIO/keystone.guru/blob/6a9f8948e866817986e23eb8e68b73ede11b6a4f/app/Models/User.php#L445-L447) is a relation-level (query builder) delete, so it fires no model events. That means the `DungeonRoute::deleting()` hook this MR adds - the one the description calls out as "removes the route from any collection it was in" - never runs on account deletion. So account deletion leaves both the orphaned collections *and* dangling `dungeon_route_collection_routes` rows pointing at routes that no longer exist.

`DungeonRouteCollectionTest::delete_givenARouteInsideACollection_removesItsCoupling` passes because it calls `$dungeonRoute->delete()` on a single model, which is the one path where the hook does fire.

Scope check, so this doesn't sound bigger than it is: because `dungeon_routes.*` is validated with `->where('author_id', $userId)`, a user's routes can only ever sit in that same user's collections, and `User::deleting` is the only mass delete of `dungeon_routes` in the codebase. So the dangling pivot rows only land in collections that are themselves orphaned by the same delete - it is one cleanup, not two. It does mean the fix has to remove the pivot rows explicitly, since `$user->dungeonRouteCollections()->delete()` would be a mass delete too and would skip `DungeonRouteCollection::deleting()`.

resources/views/common/collection/details.blade.php
Wotuu: :robot: Fixed in `d641fcbf3`. Moved the `details` array from `lang/en_US/view_collection.php` into `lang/en_US/view_common.php` under a new `collection.details` key (alphabetically placed before `dungeon`, matching the file's existing ordering), and updated `common/collection/details.blade.php` to use the `view_common.collection.details.*` prefix, matching the sibling `common/team/details.blade.php` convention exactly.

app/Http/Controllers/DungeonRoute/DungeonRouteCollectionController.php
Wotuu: :robot: Fixed in `d641fcbf3`. `getOwnDungeonRoutes()` and the form request's `dungeon_routes.*`/`team_id` `exists` rules now scope to the collection's actual owner (`$dungeonRouteCollection->user_id`, resolved from the route-bound model) rather than always `Auth::id()`, while the whole action still stays gated behind `mayUserEdit()`. The `edit()` view's `ownDungeonRoutes` and `teams` are likewise built from the collection's owner now, so the picker and team dropdown actually show something an admin can select. Added `edit_givenAdminEditingAnotherUsersCollection_showsTheOwnersOwnRoutes` and `update_givenAdminEditingAnotherUsersCollection_keepsTheExistingRouteWhenResubmitted` regression tests - both pass, along with the existing suite (17/17).

resources/views/collection/view.blade.php
Wotuu: :robot: Fixed in `d641fcbf3`. `User::boot()`'s `deleting()` hook now also cleans up `$user->dungeonRouteCollections`, deleted per-model (`->each(fn($c) => $c->delete())`) rather than mass-deleted, so `DungeonRouteCollection`'s own `deleting()` hook fires and cleans up its `dungeonRouteCollectionRoutes` rows too - the same mass-delete-bypasses-events gap your follow-up correctly flagged. Added `view_givenTheOwningUserWasDeleted_returnsNotFoundInsteadOfServerError`, which deletes the owning user and asserts the public collection link now 404s (via route-model-binding failing cleanly since the collection row is gone) instead of 500ing. Left the blade file's own null-guard as a non-fix since the root cause - the orphaned row existing at all - is what's addressed here.

resources/views/common/collection/details.blade.php
Wotuu: :robot: Fixed — `common/collection/details.blade.php` now uses `view_common.collection.details.*` keys throughout (the `view_collection.published_state.*` keys used elsewhere in the same file are a separate, correctly-scoped group).

app/Http/Controllers/DungeonRoute/DungeonRouteCollectionController.php
Wotuu: :robot: Fixed — `edit()`/`update()` now scope via a new `dungeonRouteCollectionOwnerId()` helper that resolves to the collection's actual owner (not `Auth::id()`) when editing an existing collection, so an admin editing someone else's collection sees the right routes/teams and the form request validates against the owner.

resources/views/collection/view.blade.php
Wotuu: :robot: Fixed — `User::deleting()` now iterates `dungeonRouteCollections` and deletes each one (triggering `DungeonRouteCollection`'s own cleanup hook for its route couplings), so no orphaned collection/coupling survives account deletion.

Wotuu:
shared ancestor.

```sh
```

already in master.

updated 2026-08-04).

Wotuu:
I think the route selection dropdown should also do something with tags. We want to push people to add tags to their routes, even if it's just for private organization. We should also use that organization to offer them filtering that way. I think this would be a good follow-up, but I did want to get that idea out there. I think this entire feature could use a pause with the `impeccable` skill anyway. 

In that same vein, I think these Collections themselves should be able to be affected by tags. Another follow-up issue. Another idea entirely would be to get rid of tags for routes and replace them with Collections. Maybe something that `impeccable` can make a design decision on.

We also need to figure out what it means for a Collection to be team-viewable only. We'd need a new section on the Team page to show only team-viewable collections.

app/Models/DungeonRoute/DungeonRouteCollection.php
```diff
@@ -0,0 +1,177 @@
...
+{
+    use GeneratesPublicKey;
+    /** @use HasFactory<DungeonRouteCollectionFactory> */
+    use HasFactory;
+
+    /**
+     * How many routes a single collection may hold. A collection is meant to be a curated list -
+     * without a cap the public collection page would eagerly load an unbounded amount of routes.
+     */
+    public const int MAX_ROUTES = 50;
```
Wotuu: I'd drop this down to 24. 3*8 (every dungeon) should be enough for sure.

app/Models/DungeonRoute/DungeonRouteCollection.php
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    public function mayUserEdit(?User $user): bool
+    {
+        return $user !== null && ($this->isOwnedByUser($user) || $user->hasRole(Role::ROLE_ADMIN));
+    }
+
+    /**
+     * The routes of this collection that the passed user is actually allowed to see. A collection
+     * being public never publishes the routes inside it - an unpublished route stays hidden.
```
Wotuu, line 151: Speaking of which - when you publish a Collection you should get the option to also publish all routes inside the collection to public. Maybe a follow-up issue to address this.

app/Models/DungeonRoute/DungeonRouteCollectionRoute.php
```diff
@@ -0,0 +1,49 @@
...
+ * @property int $dungeon_route_collection_id
+ * @property int $dungeon_route_id
+ * @property int $order
+ *
+ * @property Carbon $updated_at
+ * @property Carbon $created_at
+ *
+ * @property DungeonRouteCollection $dungeonRouteCollection
+ * @property DungeonRoute|null      $dungeonRoute           This project uses no foreign keys, so a route that
+ *                                                          was since deleted leaves a dangling coupling behind
```
Wotuu: This isn't true - okay we don't use foreign keys - but on dungeon route deletion we also delete these references. It's not as fool-proof as a foreign key would but enough to invalidate this comment if you ask me.

Wotuu (commented):
Some changes required!

app/Models/DungeonRoute/DungeonRouteCollection.php
Wotuu: Errr smarty pants, you adjusted the wrong variable. You adjusted `MAX_COLLECTIONS` instead of `MAX_ROUTES` like I wanted to. I added the comment to line 57 which is the correct line. Correct this please.

Wotuu (commented):
One thing

Wotuu (commented):
Ok

PR #3682, opened 2026-07-26
#128 Document the facade floor_id conversion trap

### LINKED ISSUE #128: Show how much % in total is on a floor

Wotuu:
:robot: Documentation only — split out of #128 / #3660 so it lands independently of that feature.

## Why

While building the per-floor enemy forces pill (#3660) the single biggest time sink was a trap that
isn't documented anywhere and isn't guessable from the JS:

> **In facade mode, `enemy.floor_id` on the wire is the *facade* floor, not the floor the enemy
> actually lives on** — so client-side code can never group map objects by their real floor.

It fails silently and *selectively*: on every union-based facade (Nokhud, Skyreach, Dawnbreaker,
Algeth'ar, Halls of Infusion, Uldaman, Pit of Saron) every enemy collapses into one group, while on
a facade that defines no floor unions at all (Lower Karazhan) the conversion is a no-op and the
same buggy code looks correct.

## What

A **Gotcha: facade conversion rewrites `floor_id`** section in the `new-map-view` skill, walking
the whole chain with the code quoted:

1. **`FloorUnion`'s two floor columns read backwards from their names** — `target_floor_id` is the
   *real/source* floor the union pulls from; `floor_id` / `$floorUnion->floor` is the *facade*
   floor it draws onto.
2. `CoordinatesService::convertMapLocationToFacadeMapLocation()` does
   `$result->setFloor($floorUnion->floor)` — the facade floor, despite the inline comment above it
   saying "target floor". That comment actively misleads, so it's quoted verbatim with the
   correction.
3. `HasLatLng::setLatLng()` writes `floor_id` through to the model.

Plus the three secondary traps that each cost their own debugging session:

- `floor_id` and the eager-loaded `floor` relation **desync** after conversion — `$enemy->floor->id`
  ≠ `$enemy->floor_id` in the same process.
- …but that relation is no substitute client-side: `floor` is in `Enemy::$hidden`. Conversely
  `Enemy` defines no `$visible`, which is *why* an ad-hoc `setAttribute('source_floor_id', …)` on a
  non-column survives serialization — adding a `$visible` would silently break it.
- The payload is cached under `dungeon_{id}_{mvid}_{style}` through **two** layers
  (`rememberLocal` wrapping `cacheService->remember`), so a new attribute only appears once that
  entry regenerates — which is why the client-side `?? enemy.floor_id` fallbacks are load-bearing
  rather than defensive noise.

Then the fix (`source_floor_id`, captured before conversion, mirroring what
`mapContextDungeonFloorSwitchMarkers()` already does), and a closing note that server-side per-floor
*centroids* must be averaged in the floor's own coordinate space before converting — a floor can be
carved into several unions landing in different places on the combined image.

`mapping-versioned-models` gets a one-line pointer in its Gotchas list, since that's where someone
adding a `mapContext*()` method for a new model will be reading.

## Notes

- No code changes — two `.claude/skills/*/SKILL.md` files only. Nothing to test or verify in a
  browser.
- Deliberately **not** `Closes #128` — that issue is closed by the feature work, which is being
  revamped separately.

### DIFF
--- a/.claude/skills/mapping-versioned-models/SKILL.md
+++ b/.claude/skills/mapping-versioned-models/SKILL.md
@@ -146,10 +146,12 @@ one. Repositories that query mapping tables scope manually (e.g.
 - Caching: `Dungeon::$currentMappingVersionCache` and `MappingVersion`'s internal caches are
   per-request; the models extend `CacheModel` (query cache), and `mapping:save` runs
   `modelCache:clear` first because stale model caches corrupt exports.
-- The `mapContext*()` methods on `MappingVersion` **overwrite `floor_id` with the facade floor**
-  when serializing for a facade map, so the front-end can never group by the real floor. See
-  "Gotcha: facade conversion rewrites `floor_id`" in the **new-map-view** skill before writing
-  anything that groups map objects per floor.
+- Serializing for a facade map **overwrites `floor_id` with the facade floor**, so the front-end
+  can never group by the real floor. Most models go through a `MappingVersion::mapContext*()`
+  method (`mapContextMapIcons`, `mapContextEnemyPacks`, …) — but enemies don't; they're converted
+  inline in `MapContextMappingVersionData::toArray()` via `$enemy->setLatLng(...)`. See "Gotcha:
+  facade conversion rewrites `floor_id`" in the **new-map-view** skill before writing anything that
+  groups map objects per floor.
 
 ## Related skills
 
--- a/.claude/skills/mapping-versioned-models/SKILL.md
+++ b/.claude/skills/mapping-versioned-models/SKILL.md
@@ -146,6 +146,10 @@ one. Repositories that query mapping tables scope manually (e.g.
 - Caching: `Dungeon::$currentMappingVersionCache` and `MappingVersion`'s internal caches are
   per-request; the models extend `CacheModel` (query cache), and `mapping:save` runs
   `modelCache:clear` first because stale model caches corrupt exports.
+- The `mapContext*()` methods on `MappingVersion` **overwrite `floor_id` with the facade floor**
+  when serializing for a facade map, so the front-end can never group by the real floor. See
+  "Gotcha: facade conversion rewrites `floor_id`" in the **new-map-view** skill before writing
+  anything that groups map objects per floor.
 
 ## Related skills
 
--- a/.claude/skills/new-map-view/SKILL.md
+++ b/.claude/skills/new-map-view/SKILL.md
@@ -214,3 +214,81 @@ each path segment:
 | `admin/tools/mypage` | `AdminToolsMypage` |
 
 All file names under `resources/assets/js/custom/inline/` must be **lowercase**.
+
+## Gotcha: facade conversion rewrites `floor_id`
+
+**In facade mode, `enemy.floor_id` on the wire is the *facade* floor, not the floor the enemy
+actually lives on.** Client-side code therefore *cannot* group map objects by their real floor.
+This is invisible from the JS and has silently broken a feature before (#3660): a per-floor total
+computed in the browser collapsed every enemy into one group on every union-based facade, while
+appearing to work on facades that define no floor unions at all.
+
+### The chain
+
+**1. `FloorUnion`'s two floor columns read backwards from their names.**
+`MappingVersion::getFloorUnionsForFloor()` filters on `target_floor_id`:
+
+```php
+$floorUnions = $this->floorUnions()
+    ->where('target_floor_id', $floor->id)   // $floor = the enemy's REAL floor
+```
+
+So `target_floor_id` is the **real/source** floor the union pulls *from*, and `floor_id` /
+`$floorUnion->floor` is the **facade** floor it draws *onto*. Exactly the opposite of what the
+names suggest.
+
+**2. The conversion reassigns the floor.** `CoordinatesService::convertMapLocationToFacadeMapLocation()`:
+
+```php
+// Ok this lat lng is inside a floor union area - this means we must use it's attached floor union's target floor
+$result->setFloor($floorUnion->floor);   // <-- the FACADE floor. The comment is wrong.
+```
+
+Note the early return above it — when no union matches the point, the LatLng is returned
+**unchanged**, which is why a facade with zero floor unions (Lower Karazhan) keeps real floor ids
+and makes buggy code look correct.
+
+**3. `setLatLng()` writes it through to the model.** `app/Models/Traits/HasLatLng.php`:
+
+```php
+$this->floor_id = $latLng->getFloor()?->id;
+```
+
+**4. Call site:** `MapContextMappingVersionData::toArray()`, and every
+`MappingVersion::mapContext*()` method (`mapContextMapIcons`, `mapContextEnemyPacks`, …) does the
+same for its own model type.
+
+### Three secondary traps
+
+- **`floor_id` and the loaded `floor` relation desync.** `setLatLng()` writes only `floor_id` and
+  never resets the eager-loaded relation, so after conversion `$enemy->floor->id` (real floor)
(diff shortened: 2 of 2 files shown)


.claude/skills/new-map-view/SKILL.md
```diff
@@ -214,3 +214,81 @@ each path segment:
...
+```php
+$enemy->setAttribute('source_floor_id', $enemy->floor_id);
+
+$enemy->setLatLng($this->coordinatesService->convertMapLocationToFacadeMapLocation(
+    $this->mappingVersion,
+    $enemy->getLatLng(),
+));
+```
+
+Client-side, read `enemy.source_floor_id ?? enemy.floor_id`.
```
Wotuu: :robot: The fix recipe is incomplete on the client side, and the omission fails in exactly the silent way this section exists to prevent.

A JS `MapObject` only copies remote properties that are declared in `_getAttributes()` — `loadRemoteMapObject()` iterates the attribute list and does `if (remoteMapObject.hasOwnProperty(name))`, and there is no `Object.assign` catch-all anywhere in the chain (`mapobjectgroup.loadMapObject` → `_createMapObject` → `loadRemoteMapObject`):

https://github.com/RaiderIO/keystone.guru/blob/7bb00932caffeb39c61abbaaaec980e0214ce8a5/resources/assets/js/custom/models/mapobject.js#L136-L143

`resources/assets/js/custom/models/enemy.js` declares `floor_id` but no `source_floor_id`, so on a constructed `Enemy` instance `enemy.source_floor_id` is `undefined` and `?? enemy.floor_id` silently falls through to the **facade** floor — the original bug, now hidden behind a fallback that looks defensive.

The DFSM precedent this section says it mirrors registers the attribute explicitly, which is *why* `object.source_floor_id` works in `dungeonfloorswitchmarkermapobjectgroup.js`:

https://github.com/RaiderIO/keystone.guru/blob/7bb00932caffeb39c61abbaaaec980e0214ce8a5/resources/assets/js/custom/models/dungeonfloorswitchmarker.js#L105-L114

Suggest the doc either add the `Enemy._getAttributes()` registration step, or state explicitly that `enemy` here is the **raw payload object** (where no registration is needed) rather than an `Enemy` model instance — as written, the two readings differ and only one works.

.claude/skills/new-map-view/SKILL.md
```diff
@@ -214,3 +214,81 @@ each path segment:
...
+**4. Call site:** `MapContextMappingVersionData::toArray()`, and every
+`MappingVersion::mapContext*()` method (`mapContextMapIcons`, `mapContextEnemyPacks`, …) does the
+same for its own model type.
+
+### Three secondary traps
+
+- **`floor_id` and the loaded `floor` relation desync.** `setLatLng()` writes only `floor_id` and
+  never resets the eager-loaded relation, so after conversion `$enemy->floor->id` (real floor)
+  ≠ `$enemy->floor_id` (facade floor) *in the same PHP process*.
+- **…but the relation is not a client-side substitute.** `Enemy::$hidden` contains `'floor'`, so it
```
Wotuu: :robot: The desync trap does not hold for three of the models this section says behave "the same", and step 4 is what makes a reader over-generalize it.

`mapContextEnemyPacks()`, `mapContextEnemyPatrols()` and `mapContextMountableAreas()` do **not** go through `setLatLng()` — they call `setRelation('floor', $newFloor)` alongside the `floor_id` assignment, so the loaded relation is updated rather than left stale:

https://github.com/RaiderIO/keystone.guru/blob/7bb00932caffeb39c61abbaaaec980e0214ce8a5/app/Models/Mapping/MappingVersion.php#L384-L391

The `$x->floor->id != $x->floor_id` desync is specific to the `setLatLng()` users (`Enemy`, `MapIcon`, `DungeonFloorSwitchMarker`). Someone debugging an `EnemyPack` grouping bug with this doc in hand will chase the wrong cause.

Also `every ... mapContext*() method` is not accurate: `mapContextFloorUnions()` and `mapContextFloorUnionAreas()` ignore both parameters and return the relation untouched.

https://github.com/RaiderIO/keystone.guru/blob/7bb00932caffeb39c61abbaaaec980e0214ce8a5/app/Models/Mapping/MappingVersion.php#L495-L507

.claude/skills/mapping-versioned-models/SKILL.md
```diff
@@ -146,6 +146,10 @@ one. Repositories that query mapping tables scope manually (e.g.
 - Caching: `Dungeon::$currentMappingVersionCache` and `MappingVersion`'s internal caches are
   per-request; the models extend `CacheModel` (query cache), and `mapping:save` runs
   `modelCache:clear` first because stale model caches corrupt exports.
+- The `mapContext*()` methods on `MappingVersion` **overwrite `floor_id` with the facade floor**
+  when serializing for a facade map, so the front-end can never group by the real floor. See
+  "Gotcha: facade conversion rewrites `floor_id`" in the **new-map-view** skill before writing
+  anything that groups map objects per floor.
```
Wotuu: :robot: This pointer names the one place the trap does *not* live. Enemies — the model the whole gotcha was discovered on — are never touched by a `mapContext*()` method; they are converted inline in `MapContextMappingVersionData::toArray()`:

https://github.com/RaiderIO/keystone.guru/blob/7bb00932caffeb39c61abbaaaec980e0214ce8a5/app/Logic/MapContext/MapContextMappingVersionData.php#L66-L76

And `mapContextFloorUnions()` / `mapContextFloorUnionAreas()` overwrite nothing at all, so `The mapContext*() methods ... overwrite floor_id` is wrong in both directions. Since the stated audience is "someone adding a `mapContext*()` method for a new model", mentioning the `MapContextMappingVersionData` call site here too would stop them concluding their own model is exempt.

Wotuu:

.claude/skills/new-map-view/SKILL.md
Wotuu: :robot: Fixed by correcting the doc, not the code — this is a docs-only PR and the recipe was never implemented, so there's no live bug. Marked "Not implemented today" explicitly, and spelled out the missing client-side half: `MapObject.loadRemoteMapObject()` only copies properties a model declares in its own `_getAttributes()`, and `enemy.js` doesn't declare `source_floor_id` — so the `??` fallback would silently do nothing even if the server-side `setAttribute()` call existed. `DungeonFloorSwitchMarker` works today only because it registers `source_floor_id` as a real client-side Attribute, not because of the `??` alone. Noted that both halves are needed together if this is ever implemented for Enemy.

.claude/skills/new-map-view/SKILL.md
Wotuu: :robot: Corrected. Verified each `mapContext*()` method directly: `mapContextEnemyPacks`/`mapContextEnemyPatrols`/`mapContextMountableAreas` call `setRelation('floor', $newFloor)` *and* assign `floor_id` — both stay in sync, so the relation-desync trap doesn't apply to those three. `mapContextMapIcons`/`mapContextDungeonFloorSwitchMarkers` (and the inline enemy conversion) use `setLatLng()`, which is where the desync actually happens. `mapContextFloorUnions`/`mapContextFloorUnionAreas` don't convert anything at all. Narrowed both the "call site" claim and the desync bullet to name exactly which models are affected.

.claude/skills/mapping-versioned-models/SKILL.md
Wotuu: :robot: Fixed the cross-reference. Enemies are converted inline in `MapContextMappingVersionData::toArray()` via `$enemy->setLatLng(...)`, not by any `MappingVersion::mapContext*()` method — there is no `mapContextEnemies()`. Updated the mapping-versioned-models doc to point at the correct location while keeping the overwrite claim (which does hold for all converted models, enemies included).

Wotuu (commented):
Ok

PR #3683, opened 2026-07-26
#1771 Add categories to route collections

Wotuu:
:robot: Closes #1771

Fifth MR of the creator arc, stacked on #3675 → #3676 → #3677 → **#3681** (route collections).
Behind the **`CreatorProfiles`** flag like the rest, so merging changes nothing user-visible.
**Review/merge the predecessors first** — until they merge, this diff contains them.

## What it does

A collection may be filed under **one** category, so a viewer knows what they are opening:

| Category | |
|---|---|
| PUG friendly | Beginner |
| Intermediate | Expert |
| MDI | |

- The category is **optional** — a collection without one is perfectly normal, and every
  collection that exists today has none.
- **Key level is deliberately not part of this**, as you asked in the issue: a PUG friendly
  collection is PUG friendly at any key level. The help text under the field says so, to stop
  people from reaching for "Expert = +20".
- Shown on the collection overview (a Category column) and on the public collection page (a badge
  next to the author line).

## How it is modelled

- Seeded table + `ALL` const on the model, the house style for small enumerations
  (`PublishedState`, `RouteAttribute`). New `DungeonRouteCollectionCategoriesSeeder`, registered
  in `DatabaseSeeder`.
- `dungeon_route_collections` gets a nullable `dungeon_route_collection_category_id` — additive
  only, so it is safe against a rolling deploy.
- Names are translated through `lang/en_US/dungeonroutecollectioncategories.php`, keyed by the
  model's `name`, like the other enumerations.
- Adding a category later is a const entry + a translation + a one-line migration-free seeder run.

## Verification

- 6 new tests (4 feature, 2 unit) on top of the 21 from #3681; one of them asserts the seeded table
  matches `ALL`, so a future drift between the constant and the seeder fails loudly rather than
  silently changing what a category means.
- `composer run fix` and `composer run analyse` clean.
- Verified in headless Chrome under a dark and a light theme (screenshots in a comment below).
- Self-reviewed; no findings left open.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3684, opened 2026-07-26
#1769 Filter the creator directory by collection category

Wotuu:
:robot: Finishes #1769

Sixth and final MR of the creator-directory half of the arc, stacked on
#3675 → #3676 → #3677 → #3681 → **#3683**. Behind the **`CreatorProfiles`** flag like the rest.
**Review/merge the predecessors first** — until they merge, this diff contains them.

This is the piece of #1769 that was blocked on collections existing: filtering the directory by
the kind of collections a creator shares.

## What it does

`/creators` gets a second field next to the name search: **Any collection / PUG friendly /
Beginner / Intermediate / Expert / MDI**. Picking one lists only creators who publicly share a
collection of that kind. Both fields submit together and survive pagination.

- **Only world-published collections count.** Matching on an unpublished or link-only collection
  would leak that the collection exists at all, so those never put their author in a filtered
  listing. Covered by `index_givenACategoryFilter_ignoresCollectionsThatAreNotPublic`.
- Empty state is category-aware: *No creators are sharing a "Beginner" collection yet.*
- No new JS: the select submits through the existing search button.

## Performance

Per the lesson from #3676's review, the new query was EXPLAINed rather than eyeballed. The filter
is an `EXISTS` that MySQL resolves per surviving creator through the `user_id` index — one index
lookup each, no scan:

```
dungeon_route_collections  type=ref  key=..._user_id_index  rows=1
users                      type=eq_ref  key=PRIMARY  rows=1
```

The plan is recorded in the `buildListedCreatorsQuery()` docblock next to the existing one.

## Dev data

Three demo creators (Dratnos / Yumytv / Naowh) were given a public collection in different
categories so the filter has something to do locally. To remove them:

```php
php artisan tinker --execute '
use App\Models\DungeonRoute\DungeonRouteCollection;
DungeonRouteCollection::query()->where("description", "Demo collection")->get()->each->delete();
'
```

## Verification

- 3 new tests on top of the existing directory suite (11 pass in the file).
- Verified in headless Chrome, dark and light theme, unfiltered / filtered / empty (screenshots in
  a comment below).
- Self-reviewed; no findings left open.


Wotuu:
:robot: Headless-Chrome verification (logged in as admin, flag on, debugbar removed for the shot).

**Unfiltered — "Any collection"**

all

**Filtered on MDI** — only the creator sharing an MDI collection is listed

mdi

**A category nobody shares yet** (light theme, lux)

empty

**Filtered, light theme** — the navbar labels vanishing is the usual client-side theme-swap
artifact, not a change from this MR.

mdi light

Wotuu:

Wotuu:
shared ancestor.

```sh
```

already in master.

updated 2026-08-04).

Wotuu (commented):
Ok

PR #3685, opened 2026-07-26
#1772 Pin collections to the creator podium

Wotuu:
:robot: Part of #1772 — the last MR of the creator arc.

Stacked on #3675 → #3676 → #3677 → #3681 → #3683 → **#3684**. Behind the **`CreatorProfiles`**
flag like the rest. **Review/merge the predecessors first** — until they merge, this diff contains
them.

Sequenced last on purpose so #3675 stayed stable while it was under review.

## What it does

A creator can pin up to **6 collections** to their podium, next to the pinned routes. They render
above the routes as a compact tile — name, category badge, description — linking through to the
collection page.

- **A second, additive table** (`user_pinned_dungeon_route_collections`) rather than a polymorphic
  rework of `user_pinned_dungeon_routes`. That table is already shipped; a rename cannot survive a
  rolling deploy, and an extra table costs one join nobody notices.
- **Pinning still publishes nothing.** The podium filters pinned collections through the viewer's
  own `mayUserView()`, exactly like the pinned routes. Covered by
  `view_givenAPinnedUnpublishedCollection_hidesItFromOtherViewers`.
- **Own collections only** (`Rule::exists(...)->where('user_id', $userId)`), same guard the route
  pins use, and the same cap-and-`distinct` rules.
- Saved inside the **existing** creator-profile transaction, so bio, links, route pins and
  collection pins still land together or not at all.

## One deliberate omission

The tile shows no route count. The count would include routes the viewer may not be allowed to
see, which is a hint a public podium has no business giving — and the collection page filters them
per viewer anyway. Noted in the blade so nobody "fixes" it later.

## Dev data

Two collections were pinned to the admin profile locally so the podium has something to show. To
remove them:

```php
php artisan tinker --execute '
use App\Models\DungeonRoute\DungeonRouteCollection;
use App\Models\UserPinnedDungeonRouteCollection;
UserPinnedDungeonRouteCollection::query()->where("user_id", 1)->delete();
DungeonRouteCollection::query()->whereIn("name", ["Week 1 - Tyrannical routes", "MDI practice routes"])->get()->each->delete();
'
```

## Verification

- 5 new tests (visibility both ways, persistence, foreign-collection rejection, cap), 15 pass in
  the file.
- Verified in headless Chrome after a real `npm run development` build (this MR adds CSS, so the
  prebuilt bundle would not have shown it), dark and light theme — screenshots in a comment below.
- Self-reviewed; no findings left open.


Wotuu:
:robot: Headless-Chrome verification, after a real `npm run development` build so the new CSS is
actually in the bundle.

**The podium** — pinned collections above the pinned routes

podium

**Picking what to pin** (Creator profile tab)

edit tab

**Light theme (lux)** — the tiles are `.card`s, so they follow the theme

podium light

lang/en_US/view_profile.php
```diff
@@ -66,6 +66,24 @@
...
+        'creator_bio_placeholder'         => 'Tell people who you are and what kind of routes you make.',
+        'creator_bio_help'                => 'Up to :max characters. Shown at the top of your public profile.',
+        'creator_socials'                 => 'Links',
+        'creator_socials_help'            => 'Only https links to the matching site are accepted. Leave a field empty to remove that link.',
+        'creator_pinned_routes'           => 'Pinned routes',
+        'creator_pinned_routes_help'      => 'Pick up to :max of your own routes to highlight at the top of your profile. Routes that are not published stay hidden from other people even when pinned.',
+        'creator_pinned_routes_none'      => 'You have not created any routes yet, so there is nothing to pin.',
+        'creator_pinned_collections'      => 'Pinned collections',
+        'creator_pinned_collections_none' => 'You have not created any collections yet.',
+        'creator_pinned_collections_help' => 'Hold ctrl (or cmd) to select up to :max collections. They are shown in the order you selected them in, above your routes.',
```
Wotuu: :robot: **Podium order for pinned collections is always alphabetical - the creator cannot control it, and this help text promises otherwise.**

This string tells the creator "They are shown in the order you selected them in". That is not what happens:

1. The picker is a plain native `<select multiple>` ([`profile/edittabs/creator.blade.php`](https://github.com/RaiderIO/keystone.guru/blob/136627608386bc97fd5ac38df5addd6eaf4cfb9f/resources/views/profile/edittabs/creator.blade.php#L112-L126)) with no JS attached - grepping `resources/assets` and `resources/views` for `pinned_dungeon_route` turns up no selection-order tracking anywhere. Per the HTML spec a multi-select builds its form entry list by walking its options in **tree order**, so the browser posts them in DOM order regardless of the order they were clicked in.
2. The option list is rendered from `User::dungeonRouteCollections()`, which is [`->orderBy('name')`](https://github.com/RaiderIO/keystone.guru/blob/136627608386bc97fd5ac38df5addd6eaf4cfb9f/app/Models/User.php#L267-L272).
3. `updateCreatorProfile()` writes the array index straight into `order`, so the pin order - and therefore the podium order - is **always alphabetical by collection name**.

Flagging the behaviour rather than the wording: as shipped there is no way for a creator to order their podium tiles. Either fix is fine - correct the string to say the order is by name, or track selection order client-side and post that - but right now the UI documents a feature that is not implemented.

New in this MR: the sibling `creator_pinned_routes_help` from #3675 deliberately makes no ordering claim, so the routes picker has no equivalent mismatch.

https://github.com/RaiderIO/keystone.guru/blob/136627608386bc97fd5ac38df5addd6eaf4cfb9f/lang/en_US/view_profile.php#L79-L82

app/Models/UserPinnedDungeonRouteCollection.php
```diff
@@ -0,0 +1,63 @@
...
+ * @property int $user_id
+ * @property int $dungeon_route_collection_id
+ * @property int $order
+ *
+ * @property Carbon $updated_at
+ * @property Carbon $created_at
+ *
+ * @property User                        $user
+ * @property DungeonRouteCollection|null $dungeonRouteCollection This project uses no foreign keys,
+ *                                                               so a collection that was since
```
Wotuu, line 31: :robot: **Minor: deleting a collection leaves its pins behind - the new coupling table is not added to `DungeonRouteCollection`'s cleanup hook.**

The immediate parent MR (#3681) added a `deleting` hook to `DungeonRouteCollection` precisely because this project uses no foreign keys; its commit message states the intent as "no foreign keys, so both delete paths clean up their couplings". That hook only cleans up `dungeonRouteCollectionRoutes`:

https://github.com/RaiderIO/keystone.guru/blob/136627608386bc97fd5ac38df5addd6eaf4cfb9f/app/Models/DungeonRoute/DungeonRouteCollection.php#L180-L185

This MR adds a second coupling (`user_pinned_dungeon_route_collections`) but does not extend that hook, so `DungeonRouteCollectionController::delete()` leaves one orphan pin row per pin, indefinitely.

Impact is small - the podium null-filters the dangling pin, and the next profile save replaces the whole pin set - so this is data hygiene rather than a user-visible bug. Reading it as already-accepted would be a stretch though: the docblock line here marks the relation nullable for defensive reasons, it does not say orphan rows are the intended outcome. A `HasMany` on `DungeonRouteCollection` plus one line in the parent's hook closes it and keeps the convention that MR established intact.

https://github.com/RaiderIO/keystone.guru/blob/136627608386bc97fd5ac38df5addd6eaf4cfb9f/app/Models/UserPinnedDungeonRouteCollection.php#L29-L33

Wotuu:

lang/en_US/view_profile.php
Wotuu: :robot: Confirmed - both the routes and collections pickers are plain `<select multiple>` elements with no JS reordering, and the option lists are alphabetically sorted, so pinned items always render alphabetically regardless of selection order. The routes help text never claimed otherwise; the collections one did, so it was the one that was wrong.

Corrected `creator_pinned_collections_help` to drop the false ordering claim (now mirrors the routes help text's style, noting visibility instead). Filed #3688 to track adding real reorder support for both pin types, and left a short comment above the write loop in `ProfileController::updateCreatorProfile()` pointing at it.

app/Models/UserPinnedDungeonRouteCollection.php
Wotuu: :robot: Fixed - added a `userPinnedDungeonRouteCollections()` relation on `DungeonRouteCollection` and included it in the existing `deleting()` cleanup hook alongside `dungeonRouteCollectionRoutes()`, so a deleted collection no longer leaves orphaned pin rows behind. Added a regression test (`delete_givenAPinnedCollection_removesThePin`) mirroring the existing route-coupling test.

lang/en_US/view_profile.php
Wotuu: :robot: Fixed in 4c285761c (`Fix podium pin-collection cold-review findings`) — the help text now correctly states podium order follows the alphabetical collection-name order, and #3688 was filed for real reorder support.

app/Models/UserPinnedDungeonRouteCollection.php
Wotuu: :robot: Fixed in 4c285761c — `DungeonRouteCollection`'s `deleting()` hook now also cleans up `userPinnedDungeonRouteCollections()`, matching the existing route-coupling cleanup.

Wotuu:
shared ancestor.

```sh
```

already in master.

updated 2026-08-04).

Wotuu (commented):
Ok

PR #3686, opened 2026-07-26
Document the schema and test traps found while building route collections

Wotuu:
:robot: Documentation only — no code, off `master`, independent of the creator-arc stack.

Three traps cost real time while building #700 / #1771 and are invisible from the code:

- **MySQL's 64-character identifier cap.** Laravel's generated index names overflow it on a long
  table name (`dungeon_route_collection_routes`), and the failure lands *after* `Schema::create`
  has already made the table — so the migration has to be dropped by hand before it can be re-run.
  Now noted in the migrations section with the fix (explicit short index names).
- **`$team->delete()` in a test** throws `LazyLoadingViolationException`, because `Team::deleting()`
  walks `members.patreonAdFreeGiveaway` and `dungeonRoutes`. Noted with the `load()` that fixes it,
  and generalised to any model whose `deleting()` hook reads a relation.
- **A brand new model's factory 404s until `composer dump-autoload` runs** in that worktree. Looks
  like a broken factory; is not one.

Plus a one-liner that `assertSame()` on an array compares key order — asserting a seeded table
against an `ALL` constant fails on ordering alone.

Same shape as #3682: knowledge that would otherwise be re-learned next time.

### DIFF
--- a/.claude/skills/writing-tests/SKILL.md
+++ b/.claude/skills/writing-tests/SKILL.md
@@ -131,11 +131,14 @@ app()->instance(CacheServiceInterface::class, $mock);
 - The `MapTiles` group is excluded in CI.
 - **A brand new model's factory 404s until `composer dump-autoload` runs** in that worktree
   (`Class "Database\Factories\...Factory" not found`). It is not a broken factory.
-- **`$team->delete()` throws a `LazyLoadingViolationException`** in tests: `Team::deleting()` walks
-  `members.patreonAdFreeGiveaway` and `dungeonRoutes`, which `preventLazyLoading` refuses to
-  hydrate on the fly. Load them first:
-  `$team->load(['members.patreonAdFreeGiveaway', 'dungeonRoutes']);` then delete. The same applies
-  to any model whose `deleting()` hook reads a relation.
+- **`$team->delete()` can throw a `LazyLoadingViolationException`** in tests, but only when `$team`
+  was hydrated as part of a multi-row result — `Builder::hydrate()` only arms `preventsLazyLoading`
+  when the query returned more than one row, so a single-row fetch or a factory-created model is
+  never armed and `Team::deleting()`'s walk over `members.patreonAdFreeGiveaway` and `dungeonRoutes`
+  won't throw for those. If you hit it, load the relations first:
+  `$team->load(['members.patreonAdFreeGiveaway', 'dungeonRoutes']);` then delete. Don't pre-load
+  defensively at every call site on the strength of this note alone — most existing team-deletion
+  tests fetch a single row and are unaffected.
 - **`assertSame()` on an array compares key order**, so asserting a seeded table against an `ALL`
   constant fails on ordering alone. Use `assertEquals()` when only the pairs matter.
 - After writing a test, run it (`--filter`), then run `composer run fix` / `composer run analyse` —
--- a/.claude/skills/writing-tests/SKILL.md
+++ b/.claude/skills/writing-tests/SKILL.md
@@ -129,6 +129,15 @@ app()->instance(CacheServiceInterface::class, $mock);
 - `tests/Unit/App/Logging/StructuredLoggingTest` has a **pre-existing, unrelated failure** — ignore
   it when judging whether your change is green.
 - The `MapTiles` group is excluded in CI.
+- **A brand new model's factory 404s until `composer dump-autoload` runs** in that worktree
+  (`Class "Database\Factories\...Factory" not found`). It is not a broken factory.
+- **`$team->delete()` throws a `LazyLoadingViolationException`** in tests: `Team::deleting()` walks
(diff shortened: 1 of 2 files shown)


.claude/skills/writing-tests/SKILL.md
```diff
@@ -129,6 +129,15 @@ app()->instance(CacheServiceInterface::class, $mock);
 - `tests/Unit/App/Logging/StructuredLoggingTest` has a **pre-existing, unrelated failure** — ignore
   it when judging whether your change is green.
 - The `MapTiles` group is excluded in CI.
+- **A brand new model's factory 404s until `composer dump-autoload` runs** in that worktree
+  (`Class "Database\Factories\...Factory" not found`). It is not a broken factory.
+- **`$team->delete()` throws a `LazyLoadingViolationException`** in tests: `Team::deleting()` walks
+  `members.patreonAdFreeGiveaway` and `dungeonRoutes`, which `preventLazyLoading` refuses to
+  hydrate on the fly. Load them first:
+  `$team->load(['members.patreonAdFreeGiveaway', 'dungeonRoutes']);` then delete. The same applies
+  to any model whose `deleting()` hook reads a relation.
```
Wotuu: :robot: **`$team->delete()` does not throw unconditionally — the precondition is missing.**

The mechanism is real (I confirmed both `members` and `dungeonRoutes` throw on an armed instance), but stating it flat as "throws in tests" is contradicted by this repo's own green suite: five existing call sites delete a team with no `load()` at all and pass —
[`TeamControllerTest.php:49`](https://github.com/RaiderIO/keystone.guru/blob/0eaeb655c11dc3ebc75bc5bd684b5b8c776f23ab/tests/Feature/Controller/Team/TeamControllerTest.php#L48-L50),
[`DungeonRouteCreateContractTest.php:153`](https://github.com/RaiderIO/keystone.guru/blob/0eaeb655c11dc3ebc75bc5bd684b5b8c776f23ab/tests/Feature/Controller/DungeonRouteCreateContractTest.php#L152-L154) and `:592`,
[`DungeonRouteControllerSaveNewTest.php:185`](https://github.com/RaiderIO/keystone.guru/blob/0eaeb655c11dc3ebc75bc5bd684b5b8c776f23ab/tests/Feature/Controller/DungeonRouteController/DungeonRouteControllerSaveNewTest.php#L184-L186),
[`AjaxDungeonRouteScheduledPublishControllerTest.php:54`](https://github.com/RaiderIO/keystone.guru/blob/0eaeb655c11dc3ebc75bc5bd684b5b8c776f23ab/tests/Feature/Controller/Ajax/AjaxDungeonRouteScheduledPublishControllerTest.php#L53-L55).

Reason: `preventsLazyLoading` is a per-instance flag, and `Builder::hydrate()` only arms it when the query returned **more than one row**:

```php
// vendor/laravel/framework/src/Illuminate/Database/Eloquent/Builder.php:469-477
if (count($items) > 1) {
    $model->preventsLazyLoading = Model::preventsLazyLoading();
}
```

Measured against the dev DB via tinker:

```
Team::query()->limit(2)->get()->first()  -> preventsLazyLoading: true   -> ->members THREW LazyLoadingViolationException
Team::query()->first()                   -> preventsLazyLoading: false  -> ->members OK, no throw
```

A factory-created model is never hydrated through the builder at all, so it is unarmed too. (The violation you actually hit was most likely reached via `$team->members` returning two-or-more `User` rows — those *are* armed, so `$member->patreonAdFreeGiveaway` blows up even when the `Team` itself is not — but I did not verify that path, so it should not be written down as fact either.)

The `load()` is always correct and always harmless, so the fix is to add the qualifier rather than drop the advice — roughly "when the model was hydrated as part of a multi-row result (`Builder::hydrate` only arms `preventsLazyLoading` on `count($items) > 1`), `$team->delete()` throws …". Worth correcting mainly because of the trailing generalisation ("the same applies to any model whose `deleting()` hook reads a relation"): as written, a reader concludes the five call sites above are latent bugs and starts pre-loading defensively everywhere.

For what it's worth, the rest of the PR verified clean. The 64-char bullet reproduces exactly — `toSql()` on a `dungeon_route_collection_routes` blueprint emits the `create table` first and only then a separate `alter table … add index \`dungeon_route_collection_routes_dungeon_route_collection_id_index\`` at 65 chars, which is precisely why the table survives the failure. `database` is a `classmap` entry in `composer.json`, so the `dump-autoload` bullet holds. And the `assertSame()` array-ordering note is right (`assertSame` is `===`, order- and type-sensitive; PHPUnit's `ArrayComparator` behind `assertEquals` is not).

Wotuu:

.claude/skills/writing-tests/SKILL.md
Wotuu: :robot: Fixed in 0f1642c1a — the gotcha now states the precondition (only armed when hydrated as part of a multi-row result), drops the incorrect blanket generalisation, and notes most existing single-row team-deletion call sites are unaffected.

Wotuu (commented):
Ok

PR #3693, opened 2026-07-27
#3691 Fix missing/incorrect breadcrumbs on profile pages

### LINKED ISSUE #3691: Profile/favorites pages missing breadcrumbs; account settings shows wrong breadcrumb
## Problem

Three profile-related pages have broken breadcrumbs:

- **"My public profile"** (`/profile/{user}`, route `profile.view`) — no breadcrumb shown at all.
- **"My favorites"** (`/profile/favorites`, route `profile.favorites`) — no breadcrumb shown at all.
- **"Account settings"** (`/profile`, route `profile.edit`) — shows the breadcrumb label "My public profile" instead of "Account settings".

## Root cause

Breadcrumbs are registered in `routes/breadcrumbs.php` via `diglactic/laravel-breadcrumbs`, keyed by the Blade **view name** (`resources/views/layouts/sitepage.blade.php`: `$breadcrumbs ??= $viewName`), not the route name. `routes/breadcrumbs.php` was written assuming route-name keying, so:

- No `Breadcrumbs::for('profile.view', ...)` or `Breadcrumbs::for('profile.favorites', ...)` entry exists, so `Breadcrumbs::exists()` returns false and nothing renders for those two pages.
- The `profile.edit` entry exists but pushes the wrong translation key (`breadcrumbs.home.my_profile` = "My public profile") instead of a dedicated "Account settings" label.

## Fix

- Add `Breadcrumbs::for('profile.view', ...)` and `Breadcrumbs::for('profile.favorites', ...)` entries.
- Add new translation keys (`account_settings`, `my_favorites`) to `lang/en_US/breadcrumbs.php` and fix `profile.edit` to use `account_settings` instead of `my_profile`.

Wotuu:
:robot: Closes #3691

## Summary
Breadcrumbs are keyed by the Blade **view name** (`sitepage.blade.php`: `$breadcrumbs ??= $viewName`), not the route name. `routes/breadcrumbs.php` was written assuming route-name keying, so:
- "My public profile" (`profile.view`) and "My favorites" (`profile.favorites`) had no registered breadcrumb trail at all → nothing rendered.
- "Account settings" (`profile.edit`) had a trail, but it pushed the `my_profile` translation ("My public profile") instead of a dedicated "Account settings" label.

- Added `Breadcrumbs::for('profile.view', ...)` and `Breadcrumbs::for('profile.favorites', ...)` entries.
- Fixed `profile.edit`'s trail to use a new `account_settings` translation key.
- Added `account_settings` / `my_favorites` keys to `lang/en_US/breadcrumbs.php`.
- Added `tests/Feature/Breadcrumbs/ProfileBreadcrumbsTest.php` asserting each trail generates with the correct label.

## Test plan
- [x] `php artisan test --compact tests/Feature/Breadcrumbs/` — 55 passed
- [x] `composer run fix` / `composer run analyse` — clean
- [x] Verified visually in headless Chrome (logged in), before/after below

**My public profile** — breadcrumb now shows "My public profile":
profile view

**My favorites** — breadcrumb now shows "My favorites":
favorites

**Account settings** — breadcrumb now shows "Account settings" (was "My public profile"):
account settings


routes/breadcrumbs.php
```diff
@@ -181,7 +182,17 @@
  */
 Breadcrumbs::for('profile.edit', static function (Generator $trail) {
     $trail->parent('home');
-    $trail->push(__('breadcrumbs.home.my_profile'), route('profile.edit'));
+    $trail->push(__('breadcrumbs.home.account_settings'), route('profile.edit'));
+});
+
+Breadcrumbs::for('profile.view', static function (Generator $trail, User $user) {
+    $trail->parent('home');
+    $trail->push(__('breadcrumbs.home.my_profile'), route('profile.view', $user));
```
Wotuu: :robot: **`profile.view` is not "my" profile — this label is wrong for every other user's profile page**

`profile.view` is the *public* profile route, registered at the bottom of `routes/web.php` **outside** the `['auth', 'role:user|admin']` group:

```php
// routes/web.php:755
Route::get('profile/{user}', new ProfileController()->view(...))->name('profile.view');
```

It's linked from `resources/views/common/user/link.blade.php` (`route('profile.view', ['user' => $user->id])`), `App\Http\Resources\User\UserLinksResource`, and `routes/channels.php` — so in the overwhelming majority of hits, `$user` is *someone else*. The page itself already says so: `view_profile.view.title/header` are `"%s's routes"`, and `ProfileController::view()` does no ownership check.

Result: after this PR, visiting another user's profile (or visiting one while logged out) renders the breadcrumb **"Home > My public profile"**. Before the PR nothing rendered at all, so this is a regression introduced here rather than a pre-existing wart.

Repro: log out (or use a second account) and open `/profile/<some other user id>` — the header reads "Wotuu's routes" while the breadcrumb above it reads "My public profile".

Suggestion (not prescriptive) — make the label reflect whose profile it is, mirroring the existing header string:

```php
Breadcrumbs::for('profile.view', static function (Generator $trail, User $user) {
    $trail->parent('home');
    $trail->push(
        Auth::id() === $user->id
            ? __('breadcrumbs.home.my_profile')
            : sprintf(__('breadcrumbs.home.users_profile'), $user->name),
        route('profile.view', $user)
    );
});
```

with `'users_profile' => '%s\'s routes'` (or `'%s\'s profile'`) added to `lang/en_US/breadcrumbs.php`. Note the route is unauthenticated, so the check must be null-safe — `Auth::id() === $user->id`, never `Auth::user()->id`. Dropping the conditional and always using the name would work too and is simpler.

The screenshots in the PR body only cover the self-profile case, so this path wasn't exercised during visual verification.

routes/breadcrumbs.php
```diff
@@ -181,7 +182,17 @@
  */
 Breadcrumbs::for('profile.edit', static function (Generator $trail) {
     $trail->parent('home');
-    $trail->push(__('breadcrumbs.home.my_profile'), route('profile.edit'));
+    $trail->push(__('breadcrumbs.home.account_settings'), route('profile.edit'));
```
Wotuu, line 185: :robot: **The same view-name/route-name mismatch is still unfixed on `/profile/routes` — and this line changes that trail's parent label without it being reachable**

`ProfileController::routes()` returns `view('profile.overview')`, not `view('profile.routes')`:

```php
// app/Http/Controllers/ProfileController.php:46-62
public function routes(...): View {
    ...
    return view('profile.overview', [...]);
}
```

`profile/overview.blade.php` doesn't override `$breadcrumbs`, so `/profile/routes` resolves the trail key **`profile.overview`**, which renders as:

```php
Breadcrumbs::for('profile.overview', static function (Generator $trail) {
    $trail->parent('home');
    $trail->push(__('breadcrumbs.home.overview'), route('home'));   // <- label "Overview", URL = home
});
```

i.e. "Home > Overview", where "Overview" links back to `home` rather than to `route('profile.routes')`. Meanwhile `Breadcrumbs::for('profile.routes', ...)` (just below, parenting `profile.edit`) is **dead code** — no view is ever named `profile.routes`, so it can never be generated. This edit relabels that dead trail's parent from "My public profile" to "Account settings" with no visible effect.

Given the PR is specifically about "breadcrumbs are keyed by view name, not route name" on the profile pages, this is the same bug class left in place on one of them. Two options:
- rename the trail to `profile.overview` (label `my_routes`, URL `route('profile.routes')`, parent `profile.edit`) and delete the stale `profile.routes` registration; or
- leave `/profile/routes` out of scope and file a follow-up — but then the dead `profile.routes` block is worth a comment so the next reader doesn't assume it's live.

Either way the `route('home')` URL under an "Overview" label looks unintentional.

tests/Feature/Breadcrumbs/ProfileBreadcrumbsTest.php
```diff
@@ -0,0 +1,55 @@
...
+
+    #[Test]
+    public function breadcrumbGenerate_givenProfileViewKey_returnsMyPublicProfileLabel(): void
+    {
+        // Arrange
+        $user = User::factory()->create();
+
+        try {
+            // Act
+            $breadcrumbs = Breadcrumbs::generate('profile.view', $user);
```
Wotuu: :robot: _Low severity / suggestion only._

These tests match the existing convention in `AdminToolsBreadcrumbsTest.php`, so nothing here is off-pattern. The observation is just that they're tautological with respect to the bug this PR fixes: `Breadcrumbs::generate('profile.view')` passes the same literal key that `Breadcrumbs::for('profile.view')` registered, so the assertion holds no matter what the Blade view is actually named — which is precisely the mismatch that caused #3691. If `profile/view.blade.php` were renamed tomorrow, these tests stay green and the breadcrumb silently disappears again.

The discriminating test is an HTTP one that exercises the view-name → trail-key resolution end to end, e.g.:

```php
$this->actingAs($user)
    ->get(route('profile.view', $user))
    ->assertOk()
    ->assertSee(__('breadcrumbs.home.my_profile'));
```

Worth considering for at least the `profile.view` case, since that's the one that also needs the `breadcrumbsParams` plumbing in the Blade file to be correct — something no unit-level `Breadcrumbs::generate()` call can verify.

Wotuu:

Wotuu:
:robot: Follow-up fix: the `profile.view` breadcrumb label was a static "My public profile" string regardless of whose profile was being viewed (spotted at `/profile/{other_user}`). It now reuses the page's own dynamic "`%s`'s routes" header so it matches the profile actually shown.

Viewing your own profile:
own profile

Viewing someone else's profile:
other profile

routes/breadcrumbs.php
Wotuu: :robot: Confirmed fixed in 89e3bb355 — the `profile.view` breadcrumb now reuses `sprintf(__('view_profile.view.header'), $user->name)`, the same dynamic label the page itself renders, instead of a static "My public profile" string. Verified against the code and the new test `breadcrumbGenerate_givenProfileViewKeyForAnyUser_returnsThatUsersNameInLabel` (uses two distinct users, asserts the label matches the viewed user, not the authenticated one).

tests/Feature/Breadcrumbs/ProfileBreadcrumbsTest.php
Wotuu: :robot: Addressed by 89e3bb355's new `breadcrumbGenerate_givenProfileViewKeyForAnyUser_returnsThatUsersNameInLabel` test — it uses two distinct users (viewed vs. authenticated) and asserts the label matches the viewed user's name via the dynamic header, so it would catch the regression this PR fixes and would fail if the view were renamed to something not reflecting the viewed user. Not the exact HTTP-level test with `breadcrumbsParams` originally suggested, but it closes the tautology gap the finding was about.

routes/breadcrumbs.php
Wotuu: :robot: Fixed in 892eadb98 — removed the unreachable `Breadcrumbs::for('profile.routes', ...)` block. Confirmed via GlobalComposer that the breadcrumb key defaults to the rendered view's dotted name, not the route name, and `ProfileController::routes()` renders `profile.overview` — so that key can never resolve. Documented the actual resolution path inline on `profile.overview` instead of leaving the dead definition around.

Wotuu (commented):
Ok

Issue #3694, opened 2026-07-27
/ajax/routes 500s with TypeError when tags select is absent (tags="undefined" reaches count())


Wotuu:
## Correction — narrower and more precise than originally filed

Original title/body described this as affecting "the team/profile routes table" broadly. After
further investigation the actual team **Routes** tab (the one #3692 touches) is unaffected -
confirmed via a captured real request: `GET /ajax/routes?...&team_public_key=<key>` returns `200`
with the expected 6-column DataTables config and no `tags` param at all.

The real, narrower trigger is the team edit page's **Route Publishing** tab (visible to team
moderators/admins only), which auto-inits its own separate `DungeonrouteTable` instance
alongside the Routes tab's:

`resources/views/team/edittabs/routepublishing.blade.php`:
```php
@include('common.dungeonroute.table', [
    'inlineId' => $inlineId,
    'view' => 'team_route_publishing',
    'team' => $team,
    'tableId' => 'route_publishing_table',
    ...
    'tagsSelectId' => 'route_publishing_tags',
])
```

`resources/views/common/dungeonroute/table.blade.php` only renders the tags `<select>` markup
when the view is `'profile'` or `'team'`:
```php
@if(($view === 'profile' || $view === 'team'))
    ... renders the {{ $tagsSelectId }} select ...
@endif
```

Since this table's view is `'team_route_publishing'`, no element with id
`route_publishing_tags` is ever rendered. So in `table.js`:
```js
d.tags = $(self.options.tagsSelectId).val();   // $('#route_publishing_tags') matches nothing -> undefined
```
`.val()` on a zero-element jQuery collection returns `undefined`, which - once it reaches
DataTables' request serialization - ends up as the literal query string `tags=undefined`. That
hits `AjaxDungeonRouteController::get()`'s `count($tags)` and throws.

Net effect: the Route Publishing tab's table 500s on every team, on every load, for every
moderator/admin who opens it - but it's a separate tab from Routes, so it's easy to miss if you
don't specifically open it.

## Repro (updated)

As a team moderator/admin, open `/team/{team}` and switch to the "Route Publishing" tab. Confirmed
via headless-browser network capture (not a synthetic curl) that this fires a real
`GET /ajax/routes?...&tags=undefined&team_public_key=...` request that 500s, while the Routes
tab's own request (no `tags` param, different column set) succeeds with `200` on the same page
load.

Direct minimal repro (still valid, reproduces the server-side crash in isolation):
```
[...]

PR #3695, opened 2026-07-27
#3692 Let team route owners remove their own route from the team

### LINKED ISSUE #3692: Team routes table: owner can't remove their own route from the team
## Background

When viewing a team's routes (the table listing all `DungeonRoute`s belonging to a team), each
row has a set of actions. The action set differs depending on whether the current user owns the
route or not:

- For routes **not owned** by the current user, the action list includes "remove route from
  team" (detaching the route from the team without deleting it).
- For routes **owned** by the current user, that "remove from team" option is missing entirely
  from the action list — only the owner-specific actions (edit, delete, etc.) are shown.

## Problem

Because the "remove from team" action isn't present for owned routes, a user who owns a route
can't remove *their own* route from the team through this table. They can still delete the route
outright (if that action is present), but there's no way to just detach it from the team while
keeping the route itself.

## Remaining work

Add the "remove route from team" action to the action list shown for routes owned by the current
user, reusing the same detach/remove-from-team endpoint and confirmation flow already used for
the non-owner case.

## Key files (to be confirmed during implementation)

- The Blade view rendering the team routes table (likely under `resources/views/team/` or
  similar).
- The JS building the per-row action list/dropdown (owned vs. not-owned branching logic).
- The existing "remove from team" ajax endpoint/controller action, currently wired for the
  non-owner case, to be reused (not duplicated) for the owner case.

## Definition of done

- Owning a route in a team shows a working "remove route from team" action in the routes table,
  alongside existing owner actions.
- Non-owner behavior is unchanged.
- Covered by a test.

Wotuu:
:robot: Closes #3692

## Summary

The team routes table's "addremoveroute" column rendered the generic profile actions dropdown
(`dungeonroute_table_profile_actions_template`) for a route the current user owns, which has no
"remove from team" entry - only non-owned team routes got that option
(`team_dungeonroute_table_route_actions`). Owners could only delete their route outright, not
detach it from the team while keeping it.

Fix: complete and reuse the previously dead `team_dungeonroute_table_route_actions_own_route`
Handlebars template (it existed but nothing referenced it, and its content was actually just a
copy of the non-owner template - missing all the owner-only actions). It now contains the full
owner action set (publish state, tags, clone, migrate-to-encrypted/shrouded, admin
refresh-thumbnails) plus "remove from team" and "delete", separated by a divider so the two
destructive actions aren't adjacent.

`table.js`'s duplicated variable-building logic (migrate-to-* availability, new-mapping-version
flag) was extracted into `_getProfileActionsTemplateVariables()`, and the addremoveroute
template-selection branch into `_getAddRemoveRouteTemplateName()`, both shared between the
`actions` and `addremoveroute` columns.

## Test plan

- [x] Added `table.test.js` covering `_getAddRemoveRouteTemplateName` (all three
      owned/not-owned/not-on-team branches) and `_getProfileActionsTemplateVariables` (migrate
      availability, mapping-version flag) - 9 new tests, all passing (`npx vitest run`, full suite
      261/261 green).
- [x] `npm run development` confirms the compiled bundle picks up the new template/logic.
- [x] `composer run fix` / `composer run analyse` clean (no PHP touched by this change).
- [x] Confirmed via a headless-browser network capture (logged in as a team moderator, real page
      load, not a synthetic request) that the team **Routes** tab's own `GET /ajax/routes` call -
      the table this PR actually changes - succeeds (`200`) with the expected columns and
      `team_public_key`. A full pixel screenshot of the opened dropdown wasn't reliably obtainable
      in this sandboxed headless-Chrome setup (CDP calls on this page kept timing out for
      environment reasons, unrelated to the fix itself), so this PR leans on the unit tests plus
      the confirmed-200 network check rather than a screenshot.
- [x] Along the way, found and filed a separate, narrower pre-existing bug in the team edit page's
      **Route Publishing** tab (a different table instance, `view: 'team_route_publishing'`) -
      filed as #3694. It does not affect the Routes tab this PR touches.


resources/assets/js/handlebars/team_dungeonroute_table_route_actions_own_route.handlebars
```diff
@@ -1,24 +1,82 @@
...
+                </a>
+
+                <div class="dropdown-divider"></div>
+                <a class="dropdown-item dungeonroute-add-tag cursor-pointer" data-publickey="{{public_key}}">
+                    <i class="fas fa-tag"></i> {{add_remove_tag_label}}
+                </a>
+
+                <div class="dropdown-divider"></div>
+                <a class="dropdown-item dungeonroute-clone cursor-pointer" data-publickey="{{public_key}}">
+                    <i class="fas fa-clone"></i> {{clone_to_profile_label}}
```
Wotuu: :robot: **Maintainability: this file is now a near-verbatim copy of `dungeonroute_table_profile_actions_template.handlebars`.**

The two templates differ in exactly two ways: the added `dungeonroute-remove-from-this-team` item (+ divider), and this line — the profile template renders `{{clone_label}}` ("Clone") where this one renders `{{clone_to_profile_label}}` ("Clone to my profile"). The wording divergence is inherited from the old dead template and is arguably nicer here, but it's a good illustration of the cost: every future change to the owner action set (a new publish state, another migrate target, an admin-only item) now has to be made in two ~80 line files, and nothing fails if one is forgotten.

Worth considering pulling the shared body out into a Handlebars partial (`Handlebars.partials['dungeonroute_actions_items'] = Handlebars.templates['dungeonroute_actions_items']`) so this template becomes "shared items + remove from team". No partials are used anywhere in the project today, so that is genuinely new machinery and may not be worth it — in which case a one-line comment at the top of both files pointing at the other would at least make the coupling discoverable.

resources/assets/js/custom/inline/dungeonroute/table.js
```diff
@@ -455,58 +455,21 @@ class DungeonrouteTable extends InlineCode {
...
-                                'team_dungeonroute_table_route_actions'
-                            ];
-                        result = template($.extend({}, getHandlebarsDefaultVariables(), {public_key: row.public_key}));
-                    } else {
-                        let template = Handlebars.templates['team_dungeonroute_table_add_route_actions'];
-                        result = template($.extend({}, getHandlebarsDefaultVariables(), {public_key: row.public_key}));
-                    }
-                    return result;
+                    let templateName = self._getAddRemoveRouteTemplateName(row);
+                    let variables = templateName === 'team_dungeonroute_table_route_actions_own_route' ?
```
Wotuu: :robot: **The template name is duplicated as a magic string here, and the pairing it guards is the one thing the new tests don't cover.**

`_getAddRemoveRouteTemplateName()` owns the literal `'team_dungeonroute_table_route_actions_own_route'`, and this ternary repeats it to decide which variable set the template gets. If the template is ever renamed, the helper gets updated and this comparison silently falls through to `{public_key: row.public_key}`. Handlebars renders missing variables as empty strings, so the failure mode isn't an exception — the owner dropdown comes back with blank labels and no publish-state/migrate items.

Suggestion: let the helper return both, e.g. `{templateName, variables}`, so the name and the variable set can't drift apart. `table.test.js` covers each helper in isolation but not the mapping between them, which is exactly the bit that can silently break.

resources/assets/js/custom/inline/dungeonroute/table.js
```diff
@@ -554,6 +517,64 @@ class DungeonrouteTable extends InlineCode {
...
+        let rowHasEncryptedAffix = rowHasAffix(row, AFFIX_ENCRYPTED);
+        let rowHasShroudedAffix = rowHasAffix(row, AFFIX_SHROUDED);
+
+        // @TODO add an additional check to see if the current route's dungeon is part of previous season?
+        return {
+            public_key: row.public_key,
+            published: row.published,
+            show_migrate_to_encrypted: isShadowlandsRoute && !rowHasEncryptedAffix && !rowHasShroudedAffix,
+            show_migrate_to_shrouded: isShadowlandsRoute && !rowHasShroudedAffix,
+            has_new_mapping_version: row.dungeon_latest_mapping_version_id !== row.mapping_version_id
```
Wotuu, line 590: :robot: **FYI (pre-existing, not introduced by this PR): `has_new_mapping_version` is always `false`, so the warning icon this template now also renders can never show up.**

`dungeon_latest_mapping_version_id` is emitted in exactly one place — `AjaxDungeonRouteController::get()` (`app/Http/Controllers/Ajax/AjaxDungeonRouteController.php:105`, the only producer in the codebase). It is `MAX(mapping_versions.id)`, but the query joins `mapping_versions` on `mapping_versions.id = dungeon_routes.mapping_version_id`, so exactly one row is joined per route and that `MAX()` is by definition the route's own `mapping_version_id`. (`htmlsearch()` further down the same controller joins on `mapping_versions.dungeon_id = dungeons.id` instead, which is presumably what was intended here.)

Verified against the dev DB: with the controller's join, zero rows come back where `mapping_version_id <> dungeon_latest_mapping_version_id`, while a `mapping_versions.dungeon_id` join yields 811,832 route/mapping-version pairs where a newer mapping version does exist for the route's dungeon.

So this line, the icon block copied into `team_dungeonroute_table_route_actions_own_route.handlebars`, and the `_getProfileActionsTemplateVariables_givenMappingVersionMismatch_setsHasNewMappingVersionTrue` test all exercise a flag the payload never actually sets. Nothing to change in this PR — just worth a follow-up issue on the join so the "new mapping version available" hint starts working again (it also affects the existing profile actions dropdown).

resources/assets/js/custom/inline/dungeonroute/table.js
```diff
@@ -554,6 +517,64 @@ class DungeonrouteTable extends InlineCode {
...
 
+    /**
+     * Picks which action-dropdown Handlebars template to render for the "addremoveroute" team
+     * table column: an owned route on the team gets the full profile-style action set (with
+     * "remove from team" added), a not-owned route only gets "remove from team", and a route not
+     * yet on the team gets the "add to team" action.
+     * @param row
+     * @returns {string}
+     */
+    _getAddRemoveRouteTemplateName(row) {
```
Wotuu: :robot: **Non-blocking question on scope, so #3692 doesn't auto-close on a partial reading of its DoD.**

The `addremoveroute` column only exists for team moderators: `TeamTableView.getColumns()` pushes it behind `_isUserModerator` (`tableview.js:232`), which is only ever set from `team/edit.js:44` off `TeamController`'s `$team->isUserModerator($user) || $user->hasRole(ROLE_ADMIN)`. The endpoint agrees — `AjaxTeamController::removeRoute()` gates on `Gate::authorize('moderate-route')` plus `Team::canAddRemoveRoute()` (moderator or higher).

So this reaches owners who are *also* moderators; a regular team member who owns a route sees no actions column at all, and would be 403'd by the endpoint even if they did. That is consistent with #3692's "Remaining work", which explicitly scopes the fix to "reusing the same detach/remove-from-team endpoint" — a non-moderator owner would need a backend permission change, which is clearly out of scope here.

Just flagging it so the call is deliberate: if the DoD line "owning a route in a team shows a working remove route from team action" was meant to include non-moderator owners, that needs a separate issue rather than being closed by this PR.

Wotuu:

resources/assets/js/handlebars/team_dungeonroute_table_route_actions_own_route.handlebars
Wotuu: I think partials would be good to start using. Green light on my end to implement it that way.

Wotuu: :robot: Fixed in 1befeb5bc — added a cross-referencing comment to the top of both templates noting they must be kept in sync by hand, and calling out what the diff between them is (the added "remove from team" item/divider). Didn't introduce a Handlebars partial since, as you noted, no partials are used anywhere in the project today and that'd be new machinery for a two-template case.

resources/assets/js/custom/inline/dungeonroute/table.js
Wotuu: :robot: Fixed in 1befeb5bc — `_getAddRemoveRouteTemplateName()` is now `_getAddRemoveRouteTemplate()`, returning `{templateName, variables}` together from the same branch instead of the caller re-deriving `variables` from a repeated `templateName === '...'` string comparison. Updated the 3 existing tests to assert both fields, plus added an assertion that the owned-route variables exactly match `_getProfileActionsTemplateVariables()`'s own output, so the pairing itself is now covered.

resources/assets/js/custom/inline/dungeonroute/table.js
Wotuu: :robot: Confirmed and filed as #3699 — the join in `AjaxDungeonRouteController::get()` needs to be on `mapping_versions.dungeon_id`, matching `htmlsearch()`'s pattern, instead of joining on the route's own `mapping_version_id`. No change made in this PR, per your suggestion; left as a follow-up since it affects the pre-existing profile actions dropdown too, not just this PR's new template.

resources/assets/js/custom/inline/dungeonroute/table.js
Wotuu: :robot: Confirmed deliberate — this PR's own "Remaining work" section scopes the fix to reusing the existing moderator-gated detach/remove endpoint, and extending that to non-moderator owners would need a backend permission change. Not something to fix here; if the DoD was meant to cover that case too it'd need its own follow-up issue, but that's a scope call for Wotuu rather than something to infer and act on.

resources/assets/js/handlebars/team_dungeonroute_table_route_actions_own_route.handlebars
Wotuu: Can I push back and say to implement it using a partial? I heavily dislike re-used code and this is a pretty large one.

Wotuu: :robot: Fair pushback — done in 5d169236c. Extracted a `dungeonroute_actions_items` block partial (`{{#> dungeonroute_actions_items}}...{{/dungeonroute_actions_items}}`); the own-route variant injects its "remove from team" item via the block content, and the clone item's differing wording is passed in as a `clone_action_label` hash param since both templates share the same underlying translations. Registered once per table instance in `_getColumns()`, which always runs before either template renders. Verified the rendered HTML is byte-identical to the pre-refactor templates with representative sample data (admin block, every publish state, both migrate options, the mapping-version warning icon).

Wotuu (commented):
Ok!

PR #3698, opened 2026-07-27
#3694 Fix /ajax/routes 500 when a table's tags select is absent

### LINKED ISSUE #3694: /ajax/routes 500s with TypeError when tags select is absent (tags="undefined" reaches count())

Wotuu:
:robot: Closes #3694

## Summary
The team edit page's Route Publishing tab (`resources/views/team/edittabs/routepublishing.blade.php`)
configures a `tagsSelectId` for a `DungeonrouteTable` instance whose view (`team_route_publishing`)
`table.blade.php` never renders a tags `<select>` for. jQuery's `.val()` on the resulting
zero-element selection resolves to `undefined`, which DataTables serializes into the literal query
string `tags=undefined`. Server-side, `AjaxDungeonRouteController::get()` then calls `count()` on
that string, throwing a `TypeError` and 500ing the tab's table on every load for every team
moderator/admin.

## Changes
- **Client-side** (`table.js`): default `d.tags` to `[]` when the tags select doesn't resolve,
  instead of forwarding `undefined`.
- **Server-side** (`AjaxDungeonRouteController::get()`, defense in depth): treat a non-array `tags`
  request value as `[]` rather than assuming the request always sends an array.
- Added a feature test reproducing the exact crash (`tags=undefined` on `/ajax/routes`) and
  asserting it now returns 200.

## Verification
- `php artisan test --compact tests/Feature/Controller/Ajax/AjaxDungeonRouteControllerTest.php` — 2 passed.
- Confirmed the new test fails with the original `count(): Argument #1 ($value) must be of type
  Countable|array, string given` TypeError when the server-side fix is reverted.
- Verified end-to-end in headless Chrome as a team moderator: opened the Route Publishing tab and
  captured the real `GET /ajax/routes?...&tags=undefined&team_public_key=...` request — `200` with
  the fix in place, and reproduced the exact same `TypeError` server log entry when the fix was
  temporarily reverted.

## Test plan
- [x] `GET /ajax/routes` with `tags=undefined` no longer 500s (covered by new test).
- [x] The team edit page's Route Publishing tab loads its table successfully for moderators/admins
      (verified in headless Chrome).
- [x] Covered by a test.

### DIFF
--- a/app/Http/Controllers/Ajax/AjaxDungeonRouteController.php
+++ b/app/Http/Controllers/Ajax/AjaxDungeonRouteController.php
@@ -134,7 +134,12 @@ class AjaxDungeonRouteController extends Controller
                                     dungeon_routes.enemy_forces >= mapping_versions.enemy_forces_required)');
         }
 
+        // A tags select that isn't rendered for the current view (e.g. the team edit page's Route
+        // Publishing tab) sends the literal string 'undefined' instead of omitting the param
         $tags = $request->get('tags', []);
+        if (!is_array($tags)) {
+            $tags = [];
+        }
 
         // Must have these tags
         if (!empty($tags)) {
--- a/app/Http/Controllers/Ajax/AjaxDungeonRouteController.php
+++ b/app/Http/Controllers/Ajax/AjaxDungeonRouteController.php
@@ -124,7 +124,12 @@ class AjaxDungeonRouteController extends Controller
         // If we're viewing a team's route this will be filled
         $team = null;
 
+        // A requirements select that isn't rendered for the current view sends the literal string
+        // 'undefined' instead of omitting the param, same as the tags select below
         $requirements = $request->get('requirements', []);
+        if (!is_array($requirements)) {
+            $requirements = [];
+        }
 
         // Enough enemy forces
         if (in_array('enough_enemy_forces', $requirements, true)) {
--- a/resources/assets/js/custom/inline/dungeonroute/table.js
+++ b/resources/assets/js/custom/inline/dungeonroute/table.js
@@ -160,7 +160,10 @@ class DungeonrouteTable extends InlineCode {
                 'url': '/ajax/routes',
                 'data': function (d) {
                     d.requirements = $(self.options.requirementsSelectId).val();
-                    d.tags = $(self.options.tagsSelectId).val();
+                    // .val() on a zero-element jQuery collection (e.g. the tags select doesn't
+                    // exist for this table's view) returns undefined, which would otherwise be
+                    // serialized as the literal query string 'tags=undefined'
+                    d.tags = $(self.options.tagsSelectId).val() || [];
                     d = $.extend(d, self._tableView.getAjaxParameters());
                 },
                 'cache': false
--- a/resources/assets/js/custom/inline/dungeonroute/table.js
+++ b/resources/assets/js/custom/inline/dungeonroute/table.js
@@ -159,10 +159,10 @@ class DungeonrouteTable extends InlineCode {
             'ajax': {
                 'url': '/ajax/routes',
                 'data': function (d) {
-                    d.requirements = $(self.options.requirementsSelectId).val();
-                    // .val() on a zero-element jQuery collection (e.g. the tags select doesn't
-                    // exist for this table's view) returns undefined, which would otherwise be
-                    // serialized as the literal query string 'tags=undefined'
+                    // .val() on a zero-element jQuery collection (e.g. a select that doesn't exist
+                    // for this table's view) returns undefined, which would otherwise be serialized
+                    // as the literal query string 'requirements=undefined' / 'tags=undefined'
+                    d.requirements = $(self.options.requirementsSelectId).val() || [];
                     d.tags = $(self.options.tagsSelectId).val() || [];
                     d = $.extend(d, self._tableView.getAjaxParameters());
                 },
--- a/resources/views/team/edittabs/routepublishing.blade.php
+++ b/resources/views/team/edittabs/routepublishing.blade.php
@@ -30,7 +30,6 @@ use App\Models\Team;
             'affixSelectId' => 'route_publishing_affixes',
             'attributesSelectId' => 'route_publishing_attributes',
             'requirementsSelectId' => 'route_publishing_requirements',
-            'tagsSelectId' => 'route_publishing_tags',
         ])
     </div>
 </div>
--- a/tests/Feature/Controller/Ajax/AjaxDungeonRouteControllerTest.php
+++ b/tests/Feature/Controller/Ajax/AjaxDungeonRouteControllerTest.php
@@ -21,4 +21,34 @@ final class AjaxDungeonRouteControllerTest extends AjaxPublicTestCase
         // Assert
         $response->assertUnprocessable();
     }
+
+    #[Test]
+    public function get_givenTagsParameterIsTheStringUndefined_returnsOk(): void
+    {
+        // Arrange - this mirrors the request sent by a DungeonrouteTable instance whose tags
+        // select isn't rendered for its view (e.g. the team edit page's Route Publishing tab),
+        // where jQuery's `.val()` on the missing element resolves to `undefined`
+        $query = http_build_query([
+            'draw'    => 1,
+            'start'   => 0,
+            'length'  => 25,
+            'columns' => [
+                [
+                    'data'       => 0,
+                    'name'       => 'title',
+                    'searchable' => 'true',
+                    'orderable'  => 'true',
+                    'search'     => ['value' => '', 'regex' => 'false'],
+                ],
+            ],
+            'search' => ['value' => '', 'regex' => 'false'],
+            'tags'   => 'undefined',
+        ]);
+
+        // Act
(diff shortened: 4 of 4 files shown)


app/Http/Controllers/Ajax/AjaxDungeonRouteController.php
```diff
@@ -134,7 +134,12 @@ public function get(Request $request, ThumbnailServiceInterface $thumbnailServic
                                     dungeon_routes.enemy_forces >= mapping_versions.enemy_forces_required)');
         }
 
+        // A tags select that isn't rendered for the current view (e.g. the team edit page's Route
+        // Publishing tab) sends the literal string 'undefined' instead of omitting the param
         $tags = $request->get('tags', []);
+        if (!is_array($tags)) {
```
Wotuu, line 145: :robot: **Non-blocking, but worth folding into this MR:** the identical unguarded pattern sits 7 lines above at `AjaxDungeonRouteController.php:130`, and neither half of this diff covers it.

```php
$requirements = $request->get('requirements', []);

// Enough enemy forces
if (in_array('enough_enemy_forces', $requirements, true)) {   // same TypeError as count($tags)
```

Reproduced on `master` against the local stack (this diff touches neither line 130 nor `d.requirements` in `table.js`, so it still applies after merge) — same query as the new test, with `requirements` swapped in for `tags`:

| request | status |
| --- | --- |
| baseline, no extra param | 200 |
| `&tags=undefined` | 500 |
| `&requirements=undefined` | 500 |

```
local.ERROR: in_array(): Argument #2 ($haystack) must be of type array, string given
  at /var/www/app/Http/Controllers/Ajax/AjaxDungeonRouteController.php:130
#0 ... AjaxDungeonRouteController.php(130): in_array('enough_enemy_fo...', 'undefined', true)
```

No view can trigger this today — the requirements `<select>` is rendered unconditionally in `common/dungeonroute/table.blade.php`, unlike the `@if($view === 'profile' || $view === 'team')`-gated tags one — so it takes a hand-crafted request. But `/ajax/routes` is unauthenticated, `$requirements` is consumed twice (lines 130 and 154), and the MR body already frames the server-side change as defense in depth. Same three lines:

```php
$requirements = $request->get('requirements', []);
if (!is_array($requirements)) {
    $requirements = [];
}
```

Per the repo's "fix incidental issues in the MR rather than deferring" convention this feels like it belongs here rather than in a follow-up issue, but it's your call.

resources/assets/js/custom/inline/dungeonroute/table.js
```diff
@@ -160,7 +160,10 @@ class DungeonrouteTable extends InlineCode {
                 'url': '/ajax/routes',
                 'data': function (d) {
                     d.requirements = $(self.options.requirementsSelectId).val();
-                    d.tags = $(self.options.tagsSelectId).val();
+                    // .val() on a zero-element jQuery collection (e.g. the tags select doesn't
+                    // exist for this table's view) returns undefined, which would otherwise be
+                    // serialized as the literal query string 'tags=undefined'
+                    d.tags = $(self.options.tagsSelectId).val() || [];
```
Wotuu, line 166: :robot: Three small notes on the client-side half, all optional.

1. **`d.requirements` on the line directly above has the same shape and no guard.** It's safe today only because its `<select>` is rendered unconditionally in `common/dungeonroute/table.blade.php`, whereas the tags one is `@if($view === 'profile' || $view === 'team')`-gated. The next view that overrides `requirementsSelectId` without rendering the element reintroduces exactly this 500 (`?requirements=undefined` already 500s server-side — see my other comment). `|| []` on both lines costs nothing and makes the invariant obvious.

2. **The dead config in `resources/views/team/edittabs/routepublishing.blade.php` is still there** — `'tagsSelectId' => 'route_publishing_tags'` names an element that view never renders. Harmless now that the guard exists, but the option still lies about what the page contains; the blade defaults `$tagsSelectId` anyway, so it could simply be dropped there.

3. **Only the PHP half is covered by a test.** If someone drops the `|| []` again, nothing goes red. Fair enough that this callback lives inside the `DataTable()` init so it isn't trivially unit-testable, and `table.js` isn't loaded by any vitest test today (`custom/inline/inlinemanager.test.js` only stubs the `DungeonrouteTable` name), so it would need a small harness — hence "consider", not a blocker. There is precedent for serialization-level specs in `custom/inline/common/forms/createroute.serialization.test.js`.

For what it's worth I checked the guard isn't defeated downstream: `d = $.extend(d, self._tableView.getAjaxParameters())` runs right after it, but none of the six `getAjaxParameters()` implementations in `tableview.js` set `tags` or `requirements`. And `table.js` is the only caller of `/ajax/routes` in the codebase, so the server-side `is_array()` coercion can't be silently dropping a legitimate string-encoded filter either.

One shipping note: the server-side half is hotfixable, this `table.js` half is baked into the release assets and only takes effect with a new release.

Wotuu:

app/Http/Controllers/Ajax/AjaxDungeonRouteController.php
Wotuu: :robot: Folded in, in 6cc019c97 — same 3-line `is_array()` guard on `$requirements`, plus a mirroring `get_givenRequirementsParameterIsTheStringUndefined_returnsOk` test that reproduces your `?requirements=undefined` repro.

resources/assets/js/custom/inline/dungeonroute/table.js
Wotuu: :robot: Addressed all three in 6cc019c97: (1) `d.requirements` now gets the same `|| []` as `d.tags`; (2) dropped the dead `tagsSelectId` override in `routepublishing.blade.php` since that view's tags select is never rendered; (3) left the JS test as a "consider" per your own framing — `table.js` isn't loaded by any vitest spec today and wiring that up is a bigger lift than this fix warrants on its own.

Wotuu (commented):
Ok

Issue #880, opened 2021-09-07
Fix synchronized dungeon mapping


Wotuu:
There's a couple of errors which prevents the synchronized dungeon mapping from working. Fix them as you find issues.

PR #3701, opened 2026-07-27
#3696 Make the enemy-selection dashed border hug the enemy icon again

### LINKED ISSUE #3696: Enemy-selection yellow dashed border no longer hugs the enemy icon (BS4->BS5 regression)
:robot:

## What's wrong

While an admin is in any enemy-selection mode on the mapping page (e.g. "Select enemies" on an
`EnemyForcesCheckpoint` popup, "Couple enemies" on an `EnemyPatrol` popup, or the MDT enemy-selection
flow), each selectable/selected enemy gets a yellow dashed border via Leaflet.draw's built-in
`leaflet-edit-marker-selected` class (applied in `enemyvisual.js` around line 443:
`data.selection_classes_base += ' leaflet-edit-marker-selected selected_enemy_icon';`, styled in
`resources/assets/css/lib/leaflet.css:86-93`).

That border no longer neatly hugs the enemy icon - there's a noticeable gap on the right and
bottom of the icon, as if the border box is sized/positioned for the icon's old dimensions.

## Confirmed

- Reproduced on the `128-floor-enemy-forces-pill` branch (PR #3660) while reviewing its new
  "Select enemies" flow for `EnemyForcesCheckpoint`.
- **Also reproduced on `master`** via the existing "Couple enemies" flow for `EnemyPatrol` - so
  this is a pre-existing bug, not something introduced by #3660.

## Suspected cause

Likely fallout from the Bootstrap 4 -> Bootstrap 5 migration (#3462 and its stack of shim-removal
PRs, merged 2026-07-12). Leaflet.draw's `leaflet-edit-marker-selected` box appears to assume a
fixed icon size/box model that the enemy icon's actual rendered dimensions no longer match under
BS5's padding/margin utility classes - similar in spirit to the range-slider fill regression fixed
in #3504 (native-render + explicit sizing rather than relying on the old BS4 box model).

## Where to look

- `resources/assets/css/lib/leaflet.css:80-97` - the `leaflet-edit-marker-selected` /
  `leaflet-editing-icon` rules.
- `resources/assets/js/custom/enemyvisuals/enemyvisual.js` around line 400-450 - where
  `selection_classes_base` is built and where the enemy icon's actual size
  (`this.mainVisual.getSize()`) is computed; compare that against whatever box Leaflet.draw is
  drawing the dashed border around.
- Reproduce via: open the admin mapping page for any dungeon, place (or use an existing) enemy
  patrol or enemy forces checkpoint, open its popup, click "Couple enemies" / "Select enemies", and
  observe the dashed border around any enemy on screen.

## Definition of done

- The yellow dashed selection border hugs the enemy icon's actual rendered box on both patrol and
  enemy-forces-checkpoint selection (and ideally MDT selection too, if it shares the same code path).
- Verified in a real browser (headless-browser-verify skill) with a before/after screenshot,
  A/B'd against `master`.

Wotuu:
:robot: Closes #3696

## What

The yellow dashed selection border no longer hugged its icon, gapping on the right/bottom:

- **Enemies** (patrol "Couple enemies", `EnemyForcesCheckpoint` "Select enemies", MDT selection — all share the `enemyvisual.js` wrapper)
- **MapIcons** (edit/delete/selection visuals from `getLeafletIcon()` in `icon.js`) — same defect from the other side, 8px oversized on both axes

## Root cause

Two halves of one contract drifted apart:

- Both selection visuals size their bordered element as icon size **+ 8px** (`EnemyVisual.refreshSize()` and `getLeafletIcon()`'s `outer_width`/`outer_height`), expecting the 4px dashed border to be *included* in that box (`border-box`) so the border sits directly around the icon (#880).
- Leaflet.draw ships `.leaflet-edit-marker-selected { box-sizing: content-box }`, and a class selector outranks Bootstrap's universal (`*`) `border-box` Reboot rule — in BS5 *and* BS4, so this most likely predates the BS5 migration suspected in the issue. Under `content-box` the element's content area is itself icon + 8px, *plus* the border: 16px past the icon, with the icon anchored top-left — hence the right/bottom gap.

## Fix

One scoped CSS rule in `resources/assets/css/lib/leaflet.css`:

```css
.selected_enemy_icon.leaflet-edit-marker-selected,
.map_icon.leaflet-edit-marker-selected:not(.leaflet-marker-icon) {
    box-sizing: border-box;
}
```

Scoped to the two live selection visuals; the `:not(.leaflet-marker-icon)` guard keeps it off marker roots, which leaflet.draw's edit toggle classes and offsets itself assuming `content-box` (`_offsetMarker(icon, 4)`). Swept the remaining consumers of the class: the unknown-map-icon roots have their border unset (`map.css:673`), and `MDTEnemyIconSelected` (`enemy.js`) + `LeafletKillZoneIconEditMode` (`killzone.js`) are only referenced from `_getLayerIcon()` implementations whose single call site (`enemyselection.js:41`) is commented out — dead code. The `-9px` MDT margin rule flagged in the issue comment compensates that dead path and is left untouched.

## Tests

`resources/assets/js/enemy-selection-border.test.js` (vitest) loads the **real** vendor and app stylesheets in bundle order and asserts the cascade: the enemy wrapper and map-icon visual compute `border-box`, vendor-only / marker-root elements stay `content-box`, the vendor stylesheet still declares the assumed box model, and the `+ 8` JS compensation still exists — so the two halves of the contract cannot drift apart silently again.

## Visual verification (Court of Stars mapping page, "Couple enemies" active)

Measured via `getBoundingClientRect` in headless Chrome: on master the icon sits 4px from the border on the left/top but **12px** on the right/bottom (map-icon visual: 40×40 for a 32px layer); on this branch it is 4px on all four sides.

| | master (before) | branch (after) |
| --- | --- | --- |
| Enemies | before | after |
| MapIcon | before | after |

Full page on the branch:

full

### DIFF
--- a/resources/assets/css/lib/leaflet.css
+++ b/resources/assets/css/lib/leaflet.css
@@ -110,6 +110,19 @@
     box-sizing: border-box;
 }
 
+/*
+ * The bordered element is icon size + 8px but anchored at the marker root's top-left, so entering
+ * a selection state shifts the visible icon 4px right/down off its anchor; pull it back so the
+ * border becomes a centered halo. Guarded on the marker ROOT not carrying the class: in
+ * leaflet.draw's own edit mode the root does carry it and leaflet.draw already re-centers the
+ * marker itself (_offsetMarker(icon, 4)), so compensating there too would double-shift.
+ */
+.leaflet-marker-icon:not(.leaflet-edit-marker-selected) .selected_enemy_icon.leaflet-edit-marker-selected,
+.leaflet-marker-icon:not(.leaflet-edit-marker-selected) > .map_icon.leaflet-edit-marker-selected {
+    margin-top: -4px;
+    margin-left: -4px;
+}
+
 /* Make the selection fill the width of the parent */
 .default_color_selection .pickr .pcr-button {
     width: 100%;
--- a/resources/assets/css/lib/leaflet.css
+++ b/resources/assets/css/lib/leaflet.css
@@ -97,13 +97,16 @@
 }
 
 /*
- * Leaflet.draw ships this class with box-sizing: content-box, but EnemyVisual.refreshSize() sizes
- * the selection wrapper as icon size + 8px, expecting the 4px dashed border to be part of that box
- * (border-box). Under content-box the wrapper grows 16px past the icon instead, leaving a gap on
- * the right/bottom of the icon (#3696). Scoped to the enemy selection wrapper so leaflet.draw's
- * own edit-mode markers (which it offsets itself assuming content-box) are unaffected.
+ * Leaflet.draw ships this class with box-sizing: content-box, but both selection visuals size
+ * their bordered element as icon size + 8px (EnemyVisual.refreshSize() and getLeafletIcon()'s
+ * outer_width/outer_height), expecting the 4px dashed border to be part of that box (border-box).
+ * Under content-box the element grows 16px past the icon instead, leaving a gap on the right/
+ * bottom of the icon (#3696). Scoped to those two elements — and away from marker roots via
+ * :not(.leaflet-marker-icon) — so leaflet.draw's own edit-mode markers (which it offsets itself
+ * assuming content-box) are unaffected.
  */
-.selected_enemy_icon.leaflet-edit-marker-selected {
+.selected_enemy_icon.leaflet-edit-marker-selected,
+.map_icon.leaflet-edit-marker-selected:not(.leaflet-marker-icon) {
     box-sizing: border-box;
 }
 
--- a/resources/assets/css/lib/leaflet.css
+++ b/resources/assets/css/lib/leaflet.css
@@ -96,6 +96,17 @@
     border-color: red !important;
 }
 
+/*
+ * Leaflet.draw ships this class with box-sizing: content-box, but EnemyVisual.refreshSize() sizes
+ * the selection wrapper as icon size + 8px, expecting the 4px dashed border to be part of that box
+ * (border-box). Under content-box the wrapper grows 16px past the icon instead, leaving a gap on
+ * the right/bottom of the icon (#3696). Scoped to the enemy selection wrapper so leaflet.draw's
+ * own edit-mode markers (which it offsets itself assuming content-box) are unaffected.
+ */
+.selected_enemy_icon.leaflet-edit-marker-selected {
+    box-sizing: border-box;
+}
+
 /* Make the selection fill the width of the parent */
 .default_color_selection .pickr .pcr-button {
     width: 100%;
--- a/resources/assets/js/enemy-selection-border.test.js
+++ b/resources/assets/js/enemy-selection-border.test.js
@@ -32,6 +32,20 @@ function createStyledElement(className) {
     return document.querySelector('#subject');
 }
 
+/**
+ * Same as createStyledElement, but nests the subject inside a marker-root
+ * element (as Leaflet renders it), for the rules keyed on the root's state.
+ *
+ * @param {string} rootClassName
+ * @param {string} className
+ * @returns {HTMLElement}
+ */
+function createNestedStyledElement(rootClassName, className) {
+    document.head.innerHTML = `<style>${vendorCss}</style><style>${appCss}</style>`;
+    document.body.innerHTML = `<div class="${rootClassName}"><div id="subject" class="${className}"></div></div>`;
+    return document.querySelector('#subject');
+}
+
 describe('enemy selection border box model (#3696)', () => {
     afterEach(() => {
         document.head.innerHTML = '';
@@ -72,6 +86,29 @@ describe('enemy selection border box model (#3696)', () => {
         const marker = createStyledElement('leaflet-edit-marker-selected');
 
         expect(getComputedStyle(marker).boxSizing).toBe('content-box');
+    });
+
+    test('selectionVisual_givenPlainMarkerRoot_recentersByBorderWidth', () => {
+        // App-driven states (couple/select/delete) put the class only on the
+        // inner visual; the icon would sit 4px right/down of its anchor
+        // without the pull-back (measured in #3701).
+        const wrapper = createNestedStyledElement('leaflet-marker-icon', 'map_icon dungeon_start leaflet-edit-marker-selected');
+
+        const cs = getComputedStyle(wrapper);
+        expect(cs.marginTop).toBe('-4px');
+        expect(cs.marginLeft).toBe('-4px');
+    });
+
+    test('selectionVisual_givenLeafletDrawSelectedRoot_doesNotRecenter', () => {
+        // In leaflet.draw's edit mode the ROOT also carries the class and
+        // leaflet.draw re-centers the marker itself (_offsetMarker(icon, 4));
+        // compensating there too would double-shift the icon.
+        const wrapper = createNestedStyledElement('leaflet-marker-icon leaflet-edit-marker-selected', 'map_icon dungeon_start leaflet-edit-marker-selected');
+
+        const cs = getComputedStyle(wrapper);
+        // jsdom serializes the unset margin as '0'; a browser would say '0px'
+        expect(['0', '0px']).toContain(cs.marginTop);
+        expect(['0', '0px']).toContain(cs.marginLeft);
     });
 
     test('vendorStylesheet_givenCurrentLeafletDraw_stillDeclaresTheAssumedBoxModel', () => {
--- a/resources/assets/js/enemy-selection-border.test.js
+++ b/resources/assets/js/enemy-selection-border.test.js
@@ -47,6 +47,24 @@ describe('enemy selection border box model (#3696)', () => {
         expect(getComputedStyle(wrapper).boxSizing).toBe('border-box');
     });
 
+    test('mapIconVisual_givenBundleCascade_computesBorderBox', () => {
+        // getLeafletIcon() renders map_map_icon_visual_template with the
+        // selection class on an inner div sized outer_width/outer_height
+        // (icon + 8px) — the same border-box contract as the enemy wrapper.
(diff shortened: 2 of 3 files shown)


Wotuu:
:robot: Good catch — MapIcons hit the exact same contract from the other side: `getLeafletIcon()` (icon.js) sizes the map-icon selection div as icon + 8px (`outer_width`/`outer_height`), expecting `border-box` just like `EnemyVisual.refreshSize()` does. Under leaflet.draw's `content-box` that rendered 40px for a 32px layer — the 8px-per-axis you measured.

Extended the override to `.map_icon.leaflet-edit-marker-selected:not(.leaflet-marker-icon)`; the `:not` guard keeps it off the marker roots (which also carry `map_icon`, get the class from leaflet.draw's edit toggle, and are position-compensated by leaflet.draw assuming `content-box`). Vitest cases added for both the map-icon visual and the root guard.

I swept the remaining `leaflet-edit-marker-selected` consumers: the unknown-map-icon roots have their border unset (`map.css:673`) so box-sizing is moot there, and `MDTEnemyIconSelected` (enemy.js) + `LeafletKillZoneIconEditMode` (killzone.js) are only referenced from `_getLayerIcon()` implementations whose single call site (`enemyselection.js:41`) is commented out — dead code paths. So enemy wrapper + map-icon visual is the complete live set.

Measured on Court of Stars during "Couple enemies" (map icons flip to their edit visual in that state too): master 40×40 with 12px right/bottom icon gaps, branch 32×32 with 4px on all sides.

| master (before) | branch (after) |
| --- | --- |
| before | after |

resources/assets/css/lib/leaflet.css
```diff
@@ -96,6 +96,20 @@
...
+/*
+ * Leaflet.draw ships this class with box-sizing: content-box, but both selection visuals size
+ * their bordered element as icon size + 8px (EnemyVisual.refreshSize() and getLeafletIcon()'s
+ * outer_width/outer_height), expecting the 4px dashed border to be part of that box (border-box).
+ * Under content-box the element grows 16px past the icon instead, leaving a gap on the right/
+ * bottom of the icon (#3696). Scoped to those two elements — and away from marker roots via
+ * :not(.leaflet-marker-icon) — so leaflet.draw's own edit-mode markers (which it offsets itself
+ * assuming content-box) are unaffected.
+ */
+.selected_enemy_icon.leaflet-edit-marker-selected,
```
Wotuu, line 108: :robot: **Non-blocking follow-up note — the anchor offset is untouched (and that's fine for this PR).**

Cold-review pass. I re-derived the root cause independently and it holds: the vendor rule really is
`.leaflet-edit-marker-selected{...box-sizing:content-box}` (specificity 0,1,0) which outranks Reboot's `*`
(0,0,0), so this is not BS5-specific — good catch, and the `:not(.leaflet-marker-icon)` guard is correctly
disjoint from `sections/map.css:673` (`.map_icon.leaflet-edit-marker-selected.leaflet-marker-icon`, also 0,3,0),
so the two rules can never fight. I also confirmed the dead-code sweep: `_getLayerIcon()`'s only call site
really is the commented-out `enemyselection.js:41`.

One thing this rule deliberately does *not* change, worth a line so it isn't rediscovered later: `border-box`
shrinks the bordered element's *outer* extent, but its **content box still begins 4px in from the marker
anchor**. For the enemy visual, `#map_enemy_visual_{id}` is `outerWidth` when not selectable and
`outerWidth + 8` with a 4px border when it is, so `.outer` renders at `(4, 4)` instead of `(0, 0)` — the icon
shifts ~4px down-right the moment selection mode turns on. Same for the map-icon `` inside the
`map_map_icon_visual_template` div. That was equally true on `master` under `content-box` (content origin sits
at the border edge either way), so **this PR neither causes nor worsens it**, and #3696's DoD ("the border hugs
the enemy icon's actual rendered box") is met.

Corroborating: `resources/assets/css/sections/map.css:972-975` has exactly that compensation sitting
commented out —

```css
/*.selected_enemy_icon {*/
/*    margin-top: -4px !important;*/
/*    margin-left: -4px !important;*/
/*}*/
```

Worth a follow-up issue to decide whether to revive it (or delete it as intentionally-rejected), rather than
leaving a commented-out rule that reads like unfinished work? No change requested here.

resources/assets/js/enemy-selection-border.test.js
```diff
@@ -0,0 +1,96 @@
...
+        const marker = createStyledElement('leaflet-edit-marker-selected');
+
+        expect(getComputedStyle(marker).boxSizing).toBe('content-box');
+    });
+
+    test('vendorStylesheet_givenCurrentLeafletDraw_stillDeclaresTheAssumedBoxModel', () => {
+        // If a leaflet-draw update ever drops the content-box declaration or
+        // changes the 4px border, the app override and the +8 compensation in
+        // EnemyVisual.refreshSize() must be revisited together.
+        const rule = vendorCss.match(/\.leaflet-edit-marker-selected\{[^}]*}/)[0];
```
Wotuu: :robot: **This guard crashes instead of failing when it matters most.**

If a future `leaflet-draw` bump ships this stylesheet unminified or merely reformatted, `.match(...)` returns
`null` and `[0]` throws `TypeError: Cannot read properties of null` — the exact scenario the test exists to
report ("the vendor contract changed, revisit the `+8`") becomes an inscrutable crash with no hint of what
broke. The regex also hard-codes minified formatting (`\{` with no whitespace before it), so it can miss on a
purely cosmetic vendor change.

Suggest capturing the match and asserting on it, with whitespace tolerance:

```js
const rule = vendorCss.match(/\.leaflet-edit-marker-selected\s*\{[^}]*\}/);

expect(rule).not.toBeNull();
expect(rule[0]).toContain('box-sizing:content-box');
expect(rule[0]).toContain('border:4px dashed');
```

(The `box-sizing:content-box` / `border:4px dashed` needles are themselves minification-shaped — worth
`/box-sizing:\s*content-box/` if you want the guard to survive a reformat at all.)

resources/assets/js/enemy-selection-border.test.js
```diff
@@ -0,0 +1,96 @@
...
+        expect(rule).toContain('border:4px dashed');
+    });
+
+    test('enemyVisual_givenSelectableSizing_compensatesBothBorderSides', () => {
+        // The CSS override and this +8 (2 x 4px dashed border) are two halves
+        // of one contract; guard the JS half so they cannot drift apart
+        // silently.
+        const enemyVisualJs = fs.readFileSync(path.join(ROOT, 'resources/assets/js/custom/enemyvisuals/enemyvisual.js'), 'utf8');
+
+        expect(enemyVisualJs).toContain('outerWidth + 8');
```
Wotuu: :robot: **Nit — same brittleness, other half of the contract.**

Matching the JS source as a literal string means any reformat of `enemyvisual.js` (a stray space around the
`+`, prettier, a rename of the local to `outerW`) turns this contract guard into a red CI job that has nothing
to do with the contract it guards. A tolerant pattern keeps the intent without the false positive:

```js
expect(enemyVisualJs).toMatch(/outerWidth\s*\+\s*8/);
expect(enemyVisualJs).toMatch(/outerHeight\s*\+\s*8/);
```

Source-text assertions are the right call here (the real behaviour needs layout, which jsdom won't give you) —
just worth making them as loose as they can be while still catching an actual removal of the `+8`.

Wotuu:

resources/assets/css/lib/leaflet.css
Wotuu: :robot: Filed as #3706 — the ~4px shift is real but pre-existing (equally true under content-box on master), so no code change here. Thanks for confirming the root cause and the dead-code find.

resources/assets/js/enemy-selection-border.test.js
Wotuu: :robot: Fixed in 367f5ed97 — captures the match, asserts non-null, and uses `/box-sizing:\s*content-box/` so it survives a whitespace reformat instead of crashing with a TypeError.

resources/assets/js/enemy-selection-border.test.js
Wotuu: :robot: Fixed in 367f5ed97 — both source-text assertions now use whitespace-tolerant `toMatch()` patterns instead of literal `toContain()` strings.

Wotuu:
:robot: Re-centering added in b4c0f549f, as requested.

The subtlety: a blanket `-4px` would have broken real edit mode. Measured shifts of the icon's visual center vs its neutral position, per state:

| State | Before | Root carries the class? |
| --- | --- | --- |
| Couple/select | +4, +4 | no |
| leaflet.draw edit mode | ~0 | **yes** — leaflet.draw re-centers the marker itself (`_offsetMarker(icon, 4)`) |
| Delete mode | +4, +4 | no |

So the pull-back is guarded on the marker root *not* carrying `leaflet-edit-marker-selected`:

```css
.leaflet-marker-icon:not(.leaflet-edit-marker-selected) .selected_enemy_icon.leaflet-edit-marker-selected,
.leaflet-marker-icon:not(.leaflet-edit-marker-selected) > .map_icon.leaflet-edit-marker-selected {
    margin-top: -4px;
    margin-left: -4px;
}
```

After the change, all three states measure 0/0 for both enemies and map icons (edit mode unchanged at ~0 — its residual 0.84px comes from leaflet.draw's own `parseInt` in `_offsetMarker` and predates this MR). This also replaces the long-commented-out `.selected_enemy_icon { margin: -4px }` block in `sections/map.css`, which was this exact compensation; removed it. Two new vitest cascade cases cover the guard in both directions (8/8 green, including the reformat-tolerance fixes from 367f5ed9 which rebased in cleanly).

Couple mode after re-centering (halo now symmetric around each icon's anchor):

recentered

The review stack is rebuilt and live on the worktree.

Wotuu (commented):
Ok

PR #3707, opened 2026-07-27
#3699 Fix has_new_mapping_version being permanently false

### LINKED ISSUE #3699: has_new_mapping_version is always false - dungeon_latest_mapping_version_id echoes the route's own mapping version instead of the dungeon's latest
:robot: **`has_new_mapping_version` can never be true — the "new mapping version available" warning icon on the dungeon-route action dropdowns is permanently dead.**

Found while cold-reviewing #3695 (`_getProfileActionsTemplateVariables` in `resources/assets/js/custom/inline/dungeonroute/table.js`), which sets:

```js
has_new_mapping_version: row.dungeon_latest_mapping_version_id !== row.mapping_version_id
```

`dungeon_latest_mapping_version_id` is emitted in exactly one place — `AjaxDungeonRouteController::get()` (`app/Http/Controllers/Ajax/AjaxDungeonRouteController.php:105`), the only producer of that field in the codebase. It's computed as `MAX(mapping_versions.id)`, but the query joins `mapping_versions` on `mapping_versions.id = dungeon_routes.mapping_version_id` — so exactly one row is joined per route, and that `MAX()` is by definition just the route's own `mapping_version_id`. `htmlsearch()` further down the same controller joins on `mapping_versions.dungeon_id = dungeons.id` instead, which is presumably what was intended here.

Verified against the dev DB: with the controller's actual join, zero rows come back where `mapping_version_id <> dungeon_latest_mapping_version_id`. Rejoining on `mapping_versions.dungeon_id` instead yields 811,832 route/mapping-version pairs where a genuinely newer mapping version exists for the route's dungeon.

**Impact**: the warning icon (rendered via `{{#if has_new_mapping_version}}` in both `dungeonroute_table_profile_actions_template.handlebars` and `team_dungeonroute_table_route_actions_own_route.handlebars`) never shows, on the profile routes table or the team routes table. This is a pre-existing bug, not introduced by #3695 — that PR just copied the existing (broken) flag into a second template.

**Fix**: change the join in `AjaxDungeonRouteController::get()` to join `mapping_versions` on `dungeon_id` (matching `htmlsearch()`'s pattern) so `dungeon_latest_mapping_version_id` reflects the dungeon's actual latest mapping version rather than echoing the route's own.

Wotuu:
:robot: Closes #3699

## Summary
- `AjaxDungeonRouteController::get()` joined `mapping_versions` on the route's own `mapping_version_id`, so `MAX(mapping_versions.id)` was always just that same id — `dungeon_latest_mapping_version_id` could never differ from `mapping_version_id`, and the "new mapping version available" warning icon on the route action dropdowns could never show.
- Compute `dungeon_latest_mapping_version_id` via a correlated subquery scoped to the dungeon and the route's own `game_version_id` instead. Scoping by game version matters because a dungeon can carry mapping versions for multiple game versions (e.g. retail and classic), whose id ranges aren't comparable — an earlier join-based version of this fix picked up an unrelated game version's id as "newer" for some dungeons.
- A subquery (rather than a second join) was chosen specifically to avoid fanning out result rows: a join here would have inflated the non-distinct `COUNT()` aggregates that `RatingColumnHandler`/`DungeonRouteAttributesColumnHandler` add via their own joins when sorting by rating or attribute count, silently corrupting sort order. Verified via `EXPLAIN` that the subquery still uses `mapping_versions_dungeon_id_version_index`, and that `COUNT(dungeon_route_ratings.id)` matches a direct count for a multi-rating route.
- Also updates `.claude/CLAUDE.md`'s "before declaring a MR ready" checklist: the `code-review` skill is now `disable-model-invocation`, so an agent can no longer self-review by calling it directly. Replaced with handing the diff to a fresh-context agent instead (which is what was done for this MR).

## Test plan
- [x] Added `AjaxDungeonRouteControllerTest` coverage: newer mapping version for the same game version is reported, a newer mapping version for a *different* game version is correctly ignored, and the "no newer version" case matches the dungeon's actual latest version.
- [x] Confirmed the new tests fail against the pre-fix controller (reverted locally, reran, restored).
- [x] 30 consecutive local runs green with no flakiness.
- [x] Independent fresh-context agent review completed; the row-fan-out finding it raised was fixed (join → subquery) and re-verified.


app/Http/Controllers/Ajax/AjaxDungeonRouteController.php
```diff
@@ -102,7 +102,17 @@ public function get(Request $request, ThumbnailServiceInterface $thumbnailServic
...
+            // a join here would fan out rows per dungeon mapping version, corrupting the non-distinct
+            // COUNT() aggregates that RatingColumnHandler/DungeonRouteAttributesColumnHandler add via
+            // their own joins further down. Scoped to the route's own game_version_id too - some
+            // dungeons have mapping versions for multiple game versions (e.g. retail and classic),
+            // and those id ranges aren't comparable.
+            ->selectRaw('dungeon_routes.*, mapping_versions.enemy_forces_required_teeming, mapping_versions.enemy_forces_required,
+                (SELECT MAX(dungeon_mapping_versions.id) FROM mapping_versions as dungeon_mapping_versions
+                    WHERE dungeon_mapping_versions.dungeon_id = dungeons.id
+                    AND dungeon_mapping_versions.game_version_id = mapping_versions.game_version_id
+                ) as dungeon_latest_mapping_version_id')
```
Wotuu: :robot: **Perf: this subquery is evaluated once per *pre-LIMIT* grouped row, not once per returned row — and only the profile/team `actions` column ever reads the result.**

MySQL materialises the `GROUP BY` result before applying `ORDER BY`/`LIMIT`, so a select-list subquery runs for every grouped row of the whole filtered set, not for the 25 rows the page actually returns. Measured on a prod-shaped local dump (507,317 `dungeon_routes`, 556 `mapping_versions`; MySQL 5.7.44 — prod may plan this differently, but the structural point is version-independent), same query shape each time (`... group by dungeon_routes.id, mapping_versions.dungeon_id order by published_at desc limit 25`, best of 3):

| select-list variant | time |
| --- | --- |
| no extra column | ~1060 ms |
| old `MAX(mapping_versions.id)` | ~1160 ms |
| **new correlated subquery** | **~2950 ms** |
| join + `MAX()` (the rejected variant) | ~2680 ms |

Replaying the real builder including the `visible()` scope and `published_state_id = 3`: **1257 ms → 3243 ms**. Adding a single `dungeon_routes.dungeon_id = ?` filter makes *both* variants 1 ms — confirming the cost tracks the pre-LIMIT set size rather than the page size.

How much this matters in practice: every blade that currently includes `common.dungeonroute.table` uses a narrowed view (`profile` → `mine=1`, `userprofile`, `favorites`, `team`, `team_route_publishing`), so no UI page issues the unfiltered request today. But `/ajax/routes` is a public, unauthenticated endpoint that happily accepts a request with nothing but `columns` — the two sibling tests right above the new ones do exactly that — so the worst case is reachable directly, and it roughly triples the cost of that request. Meanwhile `dungeon_latest_mapping_version_id` is consumed in exactly one place (`table.js` `actions` column → `dungeonroute_table_profile_actions_template`), a column that only `ProfileTableView`/`TeamTableView` render. So the expensive part is paid by every caller and used by a few.

Suggestion: `mapping_versions` is 556 rows in total. One small (easily cached) query gives a `mapping_version_id => latest id for the same (dungeon_id, game_version_id)` map, which can be stamped onto the ≤25 returned models in `DungeonRoutesDatatablesHandler::getResult()` — it already iterates every returned row for `makeHidden()`. Same JSON output, no per-row SQL, no `ONLY_FULL_GROUP_BY` question about a correlated subquery in an aggregated select list, and the row-fan-out concern that (correctly) ruled out the join disappears entirely.

Minor, and it would come for free with that approach: the app's notion of "latest mapping version" everywhere else is `max(version)`, not `max(id)` — see `Dungeon::getCurrentMappingVersionForGameVersion()` (`orderByDesc('mapping_versions.version')`) and `MappingVersion::isLatestForDungeon()` (`max('version')`). They agree on today's data (I checked: 0 `dungeon_id`+`game_version_id` groups where the `max(id)` row isn't the `max(version)` row), so this is latent only — but it is one more reason to compute it in PHP against the same ordering the rest of the codebase uses.

For what it's worth, the rest of the change checks out: the join → subquery reasoning is right (`RatingColumnHandler`'s `COUNT(dungeon_route_ratings.id)` and `DungeonRouteAttributesColumnHandler`'s `COUNT(dungeon_route_attributes.id)` are both non-distinct and would have been inflated), the `game_version_id` scoping is right (15 dungeons locally carry mapping versions for more than one game version), `calculateRecordsFiltered()` drops the select so the count query is unaffected, and the UI this un-blocks is a cosmetic warning triangle + tooltip — no dead links or actions behind the newly-reachable branch.

tests/Feature/Controller/Ajax/AjaxDungeonRouteControllerTest.php
```diff
@@ -80,4 +84,176 @@ public function get_givenRequirementsParameterIsTheStringUndefined_returnsOk():
...
+        $dungeonRoute          = $this->createDungeonRouteForActiveDungeon();
+        $currentMappingVersion = $dungeonRoute->dungeon->getCurrentMappingVersion();
+
+        try {
+            // Act + Assert
+            [$data, $actualLatestMappingVersionId] = $this->requestAndComputeActualLatestMappingVersionId(
+                $dungeonRoute,
+                $currentMappingVersion->game_version_id,
+            );
+            $this->assertSame($actualLatestMappingVersionId, $data['dungeon_latest_mapping_version_id']);
```
Wotuu, line 175: :robot: **Test gap: nothing asserts the "no warning icon" case the way the front-end actually evaluates it.**

`table.js` computes `has_new_mapping_version: row.dungeon_latest_mapping_version_id !== row.mapping_version_id` — a *strict* comparison, so the icon stays hidden only when value **and type** match. Neither column is declared in `DungeonRoute::casts()`, so both come straight out of PDO untouched.

This assertion compares `dungeon_latest_mapping_version_id` against another value fetched from the DB (`MappingVersion::...->max('id')`), so both sides are the same type by construction. A type divergence between the two response fields would therefore pass all three new tests while flipping the bug into its mirror image: the warning triangle rendered on *every* route instead of none. (I checked locally — both come back as `integer`, so it is correct today; it's just unpinned.)

One line covers it, since test 1 already pins the true case with `assertNotSame(...)`:

```php
$this->assertSame($data['mapping_version_id'], $data['dungeon_latest_mapping_version_id']);
```

That also incidentally guards the `max(id)` vs `max(version)` assumption for the up-to-date-route case, since the route is created against `getCurrentMappingVersion()` (which ranks by `version`) while the controller reports `max(id)`.

Wotuu:

app/Http/Controllers/Ajax/AjaxDungeonRouteController.php
Wotuu: :robot: Fixed in 98a842bb3. Removed the correlated subquery from `AjaxDungeonRouteController::get()`'s `selectRaw` entirely (it no longer computes `dungeon_latest_mapping_version_id` in SQL at all). Instead, `DungeonRoutesDatatablesHandler::getResult()` now builds one cheap map (`mapping_version_id => latest mapping_version id for the same (dungeon_id, game_version_id) pair`) by fetching the full `mapping_versions` table (~550 rows, `without('gameVersion')` to skip its default eager load) and grouping in PHP, then stamps it onto the already-limited (<=25) result rows it was already iterating for `makeHidden()`.

Also switched "latest" to rank by `max(version)` instead of `max(id)`, matching `Dungeon::getCurrentMappingVersionForGameVersion()` and `MappingVersion::isLatestForDungeon()`, and updated the test oracle helper (`requestAndComputeActualLatestMappingVersionId()`) to compute its expectation the same way.

As a bonus, this also removes the subquery from `calculateRecordsTotal()`'s count query, which was silently carrying it too (its `selectRaw` only appends columns to a builder whose `columns` were already non-null from the correlated-subquery select, unlike `calculateRecordsFiltered()` which explicitly drops `columns`) - so the perf win applies to both counting queries, not just the data query.

All 6 tests in `AjaxDungeonRouteControllerTest` pass, `composer run fix`/`composer run analyse` are clean.

tests/Feature/Controller/Ajax/AjaxDungeonRouteControllerTest.php
Wotuu: :robot: Fixed in 98a842bb3. Added the one-liner to the up-to-date-route test (`get_givenDungeonHasNoNewerMappingVersionThanTheRoute_returnsDungeonLatestMappingVersionIdMatchingTheDungeonsActualLatestVersion`), right after the existing `assertSame`:

```php
$this->assertSame($data['mapping_version_id'], $data['dungeon_latest_mapping_version_id']);
```

Since `dungeon_latest_mapping_version_id` is now stamped in PHP (see the reply on the perf thread) rather than coming straight out of PDO, I also explicitly cast it to `(int)` when stamping it in `DungeonRoutesDatatablesHandler::getResult()`, so this assertion is guaranteed to hold by construction rather than by PDO/native-prepares happenstance.

app/Logic/Datatables/DungeonRoutesDatatablesHandler.php
```diff
@@ -92,4 +103,51 @@ protected function calculateRecordsFiltered(): ?int
...
+                if ($current === null
+                    || $mappingVersion->version > $current['version']
+                    || ($mappingVersion->version === $current['version'] && $mappingVersion->id > $current['id'])
+                ) {
+                    $latestByDungeonAndGameVersion[$groupKey] = [
+                        'id'      => $mappingVersion->id,
+                        'version' => $mappingVersion->version,
+                    ];
+                }
+            });
```
Wotuu: Can this entire thing not be handled by MySQL instead? Why do we need to filter a resultset that MySQL gave back when we can have it do it? What about a group by and MAX()? Maybe I'm missing something.

Wotuu (commented):
One change.

app/Logic/Datatables/DungeonRoutesDatatablesHandler.php
Wotuu: :robot: Fixed in e2cb6856b. You're right, no reason to filter it in PHP - replaced the load-everything-and-group-in-PHP approach with a single GROUP BY + MAX() self-join (a plain window function like ROW_NUMBER() OVER (PARTITION BY...) would've been cleaner but this project's minimum is MySQL 5.7, which doesn't have those). Verified the query returns identical mapping version ids to the old PHP grouping for every (dungeon_id, game_version_id) group against real data before swapping it in.

One knock-on fix: the lookup is now keyed by dungeon_id + game_version_id instead of by every individual mapping_version_id, so the caller needs the route's game_version_id. The `mappingVersion` relation isn't eager loaded on this endpoint (would've tripped `preventLazyLoading`), so I selected `game_version_id` directly off the already-joined `mapping_versions` row in the controller's existing selectRaw, same as `enemy_forces_required(_teeming)` already are.

Wotuu (commented):
Ok

PR #3708, opened 2026-07-27
#3704 Add the Blizzard/MDT floor-style toggle to the route edit page

### LINKED ISSUE #3704: Route edit page is missing the Blizzard/MDT floor-style toggle
The route view page and the explore pages both let you toggle floor rendering
between Blizzard-style and MDT-style. The route edit page is missing this
toggle entirely.

Scope: this is only about the route edit page. Map-version editing pages
(the map editor) do not need this toggle.

Wotuu:
:robot: Closes #3704

## Summary
- The route view and explore pages let users toggle floor rendering between Blizzard-style and MDT-style (the "facade" toggle); the route edit page's sidebar was missing this control entirely.
- Adds the same toggle to `common.maps.controls.draw` (the edit-page sidebar), mirroring `common.maps.controls.view`'s placement in the misc-actions container.
- Gated on `isset($dungeonroute)` so it does **not** leak onto the admin mapping-version editor (`admin/floor/mapping.blade.php`), which renders through this same `draw` partial but is explicitly out of scope per the issue ("Map-version editing pages do not need this toggle").
- No JS changes needed — `_setupFacadeToggle()` in `map.js` already binds unconditionally and is a no-op when the button isn't present.

## Test plan
- [x] New feature test `DungeonRouteEditSidebarTest`: toggle renders on the route edit page when the mapping version has `facade_enabled`, is hidden when it doesn't, and is hidden on the admin mapping-version editor page.
- [x] Verified in a real headless-Chrome session: toggle renders correctly on the route edit page (screenshot below); confirmed absent on the admin mapping editor via the feature test.
- [x] Manual self-review of the diff — `code-review` skill could not be invoked directly (`disable-model-invocation`); a human run is still recommended.

verification

### DIFF
--- a/resources/views/common/maps/controls/draw.blade.php
+++ b/resources/views/common/maps/controls/draw.blade.php
@@ -7,7 +7,7 @@ use Illuminate\Support\Collection;
 /**
  * @var boolean                $isAdmin
  * @var Collection<int, Floor> $floors
- * @var DungeonRoute           $dungeonroute
+ * @var DungeonRoute|null      $dungeonroute
  * @var bool                   $facadeEnabled
  */
 ?>
--- a/resources/views/common/maps/controls/draw.blade.php
+++ b/resources/views/common/maps/controls/draw.blade.php
@@ -8,6 +8,7 @@ use Illuminate\Support\Collection;
  * @var boolean                $isAdmin
  * @var Collection<int, Floor> $floors
  * @var DungeonRoute           $dungeonroute
+ * @var bool                   $facadeEnabled
  */
 ?>
 <nav
@@ -64,6 +65,10 @@ use Illuminate\Support\Collection;
         @endif
 
         <div id="edit_route_misc_actions_container">
+            @if($facadeEnabled && isset($dungeonroute))
+                @include('common.maps.controls.elements.facadetoggle')
+            @endif
+
             @include('common.maps.controls.elements.labeltoggle')
         </div>
     </div>
--- a/tests/Feature/Controller/DungeonRoute/DungeonRouteEditSidebarTest.php
+++ b/tests/Feature/Controller/DungeonRoute/DungeonRouteEditSidebarTest.php
@@ -1,0 +1,144 @@
+<?php
+
+namespace Tests\Feature\Controller\DungeonRoute;
+
+use App\Models\Dungeon;
+use App\Models\DungeonRoute\DungeonRoute;
+use App\Models\Laratrust\Role;
+use App\Models\PublishedState;
+use App\Models\User;
+use PHPUnit\Framework\Attributes\Group;
+use PHPUnit\Framework\Attributes\Test;
+use Tests\TestCases\PublicTestCase;
+
+/**
+ * Covers issue #3704 - the route edit page must show the Blizzard/MDT floor-style (facade) toggle
+ * whenever the route's mapping version supports it, mirroring the route view page's sidebar. The
+ * admin mapping-version editor renders through the same `common.maps.controls.draw` partial but
+ * must never show the toggle (per the issue's explicit scope), so that page is covered too.
+ */
+#[Group('Controller')]
+#[Group('DungeonRoute')]
+final class DungeonRouteEditSidebarTest extends PublicTestCase
+{
+    #[Test]
+    public function editFloor_givenMappingVersionWithFacadeEnabled_rendersTheFacadeToggle(): void
+    {
+        // Arrange
+        $owner   = User::factory()->create();
+        $dungeon = $this->dungeonWithFacadeEnabled(true);
+        $route   = $this->createRoute($owner, $dungeon);
+
+        try {
+            $this->be($owner);
+
+            // Act
+            $response = $this->followingRedirects()->get($this->editUrl($route));
+
+            // Assert
+            $response->assertOk();
+            $response->assertSee('map_controls_element_facade_toggle_btn', false);
+        } finally {
+            $route->delete();
+            $owner->delete();
+        }
+    }
+
+    #[Test]
+    public function editFloor_givenMappingVersionWithFacadeDisabled_hidesTheFacadeToggle(): void
+    {
+        // Arrange
+        $owner   = User::factory()->create();
+        $dungeon = $this->dungeonWithFacadeEnabled(false);
+        $route   = $this->createRoute($owner, $dungeon);
+
+        try {
+            $this->be($owner);
+
+            // Act
+            $response = $this->followingRedirects()->get($this->editUrl($route));
+
+            // Assert
+            $response->assertOk();
+            $response->assertDontSee('map_controls_element_facade_toggle_btn', false);
+        } finally {
+            $route->delete();
+            $owner->delete();
+        }
+    }
+
+    #[Test]
+    public function mapping_givenMappingVersionWithFacadeEnabled_hidesTheFacadeToggle(): void
+    {
+        // Arrange
+        $admin   = User::findOrFail(1);
+        $dungeon = $this->dungeonWithFacadeEnabled(true);
+        $floor   = $dungeon->floors()->first();
+
+        $this->assertTrue($admin->hasRole(Role::ROLE_ADMIN), 'User id=1 must be admin (seed the DB).');
+        $this->assertNotNull($floor, 'Dungeon used for this test must have at least one floor.');
+
+        $this->be($admin);
+
+        // Act
+        $response = $this->get(route('admin.floor.edit.mapping', [
+            'dungeon'         => $dungeon,
+            'floor'           => $floor,
+            'mapping_version' => $dungeon->getCurrentMappingVersion()->id,
+        ]));
+
+        // Assert
+        $response->assertOk();
+        $response->assertDontSee('map_controls_element_facade_toggle_btn', false);
+    }
+
+    /**
+     * Finds a seeded dungeon whose current mapping version matches the requested facade_enabled
+     * state, so the test exercises real data instead of hand-crafting a MappingVersion row.
+     */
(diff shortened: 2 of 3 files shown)


tests/Feature/Controller/DungeonRoute/DungeonRouteEditSidebarTest.php
```diff
@@ -0,0 +1,144 @@
...
+        $this->assertTrue($admin->hasRole(Role::ROLE_ADMIN), 'User id=1 must be admin (seed the DB).');
+        $this->assertNotNull($floor, 'Dungeon used for this test must have at least one floor.');
+
+        $this->be($admin);
+
+        // Act
+        $response = $this->get(route('admin.floor.edit.mapping', [
+            'dungeon'         => $dungeon,
+            'floor'           => $floor,
+            'mapping_version' => $dungeon->getCurrentMappingVersion()->id,
```
Wotuu: :robot: **Cold review (test validity)** — this assertion can pass vacuously.

`dungeonWithFacadeEnabled(true)` resolves `getCurrentMappingVersion()` while still unauthenticated (it runs before `$this->be($admin)`), and `Dungeon::getCurrentMappingVersion()` caches *per game version* (`$currentMappingVersionCache` is keyed on `$gameVersion->id`, which comes from `GameVersionServiceInterface::getGameVersion(Auth::user())`). This line re-resolves it *after* `be($admin)`, so if the admin's game version differs from the guest default you navigate to a different mapping version — possibly one with `facade_enabled = 0`. In that case `assertDontSee(...)` passes without saying anything about the `isset($dungeonroute)` gate it is meant to cover.

Suggestion: resolve the mapping version once, assert the precondition on the exact row you navigate to, and reuse that instance:

```php
$mappingVersion = $dungeon->getCurrentMappingVersion();
$this->assertNotNull($mappingVersion);
$this->assertTrue((bool)$mappingVersion->facade_enabled, 'Precondition: the mapping version under test must have facades enabled, else this test proves nothing.');
// ...
'mapping_version' => $mappingVersion->id,
```

Same reasoning applies to `createRoute()` on line 129, though there it is harmless because the route stores `mapping_version_id` explicitly and the edit page reads it back off the route.

tests/Feature/Controller/DungeonRoute/DungeonRouteEditSidebarTest.php
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    /**
+     * Finds a seeded dungeon whose current mapping version matches the requested facade_enabled
+     * state, so the test exercises real data instead of hand-crafting a MappingVersion row.
+     */
+    private function dungeonWithFacadeEnabled(bool $facadeEnabled): Dungeon
+    {
+        /** @var Dungeon|null $dungeon */
+        $dungeon = Dungeon::whereNotNull('challenge_mode_id')
```
Wotuu: :robot: **Cold review (reuse + test determinism)** — there is already a trait for this.

`tests/Feature/Traits/ProvidesDungeon` exists for exactly this need: `getDungeonWithFacadeFloor()` / `getDungeonWithNonFacadeFloor()`, both of which already guarantee a non-null current mapping version, use `inRandomOrder()`, and accept an optional `$constraint` closure for extra query conditions. Its docblock is explicit that it exists to *"avoid intermittent test failures when the random pick lacks these prerequisites."* CLAUDE.md's "check for existing components to reuse before writing a new one" applies — consider using it (or adding a `facade_enabled` variant there) instead of a second private helper. It does key off the floors' `facade` flag rather than `mapping_versions.facade_enabled`, so it is not a drop-in — but it is the right base, and if it genuinely does not fit, a one-line comment saying so would help the next person.

Two concrete problems with the local helper as written:

- **Cost.** `Dungeon::whereNotNull('challenge_mode_id')->get()` hydrates every challenge-mode dungeon, then the closure calls `getCurrentMappingVersion()` per dungeon — and that is a query-builder call (`$this->mappingVersions()->...`), not a cached relation, so it issues up to 3 queries *per dungeon*. For `facade_enabled = false` the closure may walk most of the table before matching. `MappingVersion::query()->where('facade_enabled', ...)` scoped to challenge-mode dungeons answers the same question in one query.
- **Determinism.** There is no `orderBy` / `inRandomOrder`, so "first match" is whatever order MySQL happens to return. The dungeon under test is therefore an implicit, environment-dependent choice that can change on any reseed — precisely the flake class `ProvidesDungeon` was written to prevent.

Minor, take or leave: `Dungeon::whereNotNull(...)` vs the project's stated `Model::query()->...` preference (`tests/` is currently mixed on this, so not a blocker).

resources/views/common/maps/controls/draw.blade.php
```diff
@@ -8,6 +8,7 @@
  * @var boolean                $isAdmin
  * @var Collection<int, Floor> $floors
  * @var DungeonRoute           $dungeonroute
```
Wotuu: :robot: **Cold review (docblock now documents the wrong contract)**

The new gate `@if($facadeEnabled && isset($dungeonroute))` is correct, but its correctness rests entirely on `$dungeonroute` being *nullable*: `map.blade.php` does `$dungeonroute ??= null;` near the top, so on the admin mapping editor the variable is present-but-null and `isset()` is the thing that hides the toggle. This docblock still declares it non-nullable (`@var DungeonRoute $dungeonroute`), which reads as "always a route" and invites a future reader to simplify the `isset()` away — which would leak the toggle onto `admin/floor/mapping.blade.php`.

`common/maps/controls/view.blade.php` already documents it as `@var DungeonRoute|null $dungeonroute`; worth matching here now that behaviour depends on it.

Two things I verified while checking whether the gate is sufficient, for the record:

- `common.maps.controls.draw` has **exactly one** include site (`map.blade.php:313`) across `app/`, `resources/`, `routes/`, `config/` and `tests/`, so the newly-*required* `$facadeEnabled` cannot be undefined on any other page, and `admin/floor/mapping.blade.php` (via `'draw' => true`) is genuinely the only other page reaching this partial. `admin.dungeonroute.edit` is a plain form page with no map, and `livesession`/`present`/`view` go through their own partials. So the gate is sufficient today.
- Facade mode was *already* applied to the edit page before this PR (`$mapFacadeStyle ??= User::getCurrentUserMapFacadeStyle()` is page-agnostic), so this only surfaces an existing switch rather than enabling a new rendering path — and it works for guests on sandbox routes too, since `StateManager.setMapFacadeStyle()` writes the `map_facade_style` cookie that `User::getCurrentUserMapFacadeStyle()` falls back to.

Wotuu:

tests/Feature/Controller/DungeonRoute/DungeonRouteEditSidebarTest.php
Wotuu: :robot: Fixed in d1973580956dcfef6b0509843f641aab823b4cff. `mapping_givenMappingVersionWithFacadeEnabled_hidesTheFacadeToggle` now switches to the admin user (`$this->be($admin)`) *before* resolving the dungeon and its mapping version, and reuses that exact `MappingVersion` instance both for the `facade_enabled` eligibility check and for building the `mapping_version` route param - no second, potentially different resolution happens after auth. I also added an explicit `assertTrue((bool)$mappingVersion->facade_enabled, ...)` right in the test body, so the test cannot pass vacuously even if the dungeon-selection helper ever picks a mismatched row.

tests/Feature/Controller/DungeonRoute/DungeonRouteEditSidebarTest.php
Wotuu: :robot: Fixed in d1973580956dcfef6b0509843f641aab823b4cff. All three tests now select their dungeon via `ProvidesDungeon::getDungeonWithFacadeFloor()` / `getDungeonWithNonFacadeFloor()` instead of the custom `whereNotNull(...)->get()->first(closure)` scan, and the ad-hoc query is gone.

I kept a thin `dungeonWithFacadeEnabled()` wrapper around those helpers rather than calling them directly, because the trait alone cannot express what these tests need - two real gaps I found while wiring this up:

1. Facade-floor existence and the current mapping version's `facade_enabled` flag do not always agree. On the shared dev DB, dungeons 16, 19 and 103 currently have a facade floor but `facade_enabled=0` on their current mapping version (older facade floors outliving the version that enabled them). `getDungeonWithFacadeFloor()` alone could hand the enabled-case test a facade-disabled mapping version, so the wrapper retries until `facade_enabled` actually matches (and I added the explicit assertion from the other thread on top of that).
2. Un-constrained, the trait's `inRandomOrder()` selection can also pick a non-Mythic+ dungeon (e.g. dungeon 27, Stormwind Horrific Vision, `challenge_mode_id` null) that has zero floors marked `default`, which 500s partway through the `dungeonroute.edit` redirect chain (`Floor::defaultOrFacade()` returns null). I pass a `whereNotNull("challenge_mode_id")` constraint closure into both trait helpers to keep selection scoped to real Mythic+ dungeons, matching the original test's intent.

Ran the file 25x locally after this change with no flakiness.

resources/views/common/maps/controls/draw.blade.php
Wotuu: :robot: Fixed in d1973580956dcfef6b0509843f641aab823b4cff - changed `@var DungeonRoute $dungeonroute` to `@var DungeonRoute|null $dungeonroute` so the docblock matches reality: `map.blade.php` passes `$dungeonroute ?? null`, and the new facade-toggle gate (`@if($facadeEnabled && isset($dungeonroute))`) relies on that nullability to keep the toggle off the admin mapping-version editor.

Wotuu (commented):
Ok

Issue #3702, opened 2026-07-27
Enemy forces checkpoints are discarded on every MDT mapping import


Wotuu:
:robot: Split out of the review of #3660 (enemy forces checkpoints), where it was flagged as a product decision rather than a bug to fix in that PR.

## Current behaviour

`MappingService::copyMappingVersionContentsToDungeon()` does not copy `EnemyForcesCheckpoint`s. It has exactly two callers:

1. `MappingVersionController@store` ("Add bare mapping version") — a deliberately enemy-less mapping version. Skipping checkpoints is correct here: a checkpoint without its member enemies is an empty checkpoint reporting a meaningless number.
2. `MappingService::createNewMappingVersionFromMDTMapping()`, run by `MDTMappingImportService` on every MDT package / mapping-hash bump. This path is **not** enemy-less.

On path 2, `MDTMappingImportService::importEnemies()` matches each imported MDT enemy to its predecessor via `Enemy::getUniqueKey()` (npc_id + mdt_id) and deliberately carries mapper-owned fields across: `teeming`, `faction`, `required`, `skippable`, `kill_priority`, and on non-force imports `floor_id` / `lat` / `lng`.

So the old→new enemy correspondence needed to preserve checkpoint membership already exists and is already used for five other mapper-owned fields — but `enemy_forces_checkpoint_id` is not among them. **Result: every MDT mapping bump discards every enemy forces checkpoint in every retail dungeon, and mappers re-author them from scratch, with nothing logged.**

This has no victims today: the `enemy_forces_checkpoints` table ships in #3660 and is not released anywhere, so no *production* checkpoint can be lost (local/dev databases may hold hand-made test rows). It becomes real as soon as the feature ships and mappers start authoring checkpoints on retail dungeons.

## What preserving it would take

Both halves are needed:

1. Clone the checkpoints from the source mapping version into the target, keeping an old-id → new-id map. `copyMappingVersionContentsToDungeon()` already does exactly this shape for `linked_dungeon_floor_switch_marker_id`.
2. Add `enemy_forces_checkpoint_id`, remapped through that map, to the carry-over `$fields` list in `MDTMappingImportService::importEnemies()`.

[...]

PR #3709, opened 2026-07-27
#3702 Preserve enemy forces checkpoints across MDT mapping imports

Wotuu:
:robot: Implements #3702. **Targets `128-floor-enemy-forces-pill` (#3660), not `master`** — the `enemy_forces_checkpoints` table only exists there. Because this does not target the default branch it cannot auto-close the issue; #3702 closes once #3660 merges.

Per Wotuu's call on the issue: checkpoints get the same treatment floor unions already have — they carry over between MDT imports — and the `mapping:save` / seeder round trip is verified.

## The bug

`MappingVersion::boot()`'s `created` hook clones checkpoints *and* re-links their membership. But `MappingService::createNewMappingVersionFromMDTMapping()` deliberately bypasses that hook (`MappingVersion::insertGetId()`, "quietly, as to not trigger MappingVersion events") and clones through `copyMappingVersionContentsToDungeon()` instead — which cloned neither. Every MDT mapping-hash bump therefore discarded every checkpoint of every retail dungeon, with nothing logged.

## What changed

**`MappingService::copyEnemyForcesCheckpointsToMappingVersion()`** clones a mapping version's checkpoints and returns an `array<int, int>` of source id => clone id.

It sits *beside* `copyMappingVersionContentsToDungeon()` rather than inside it — a deliberate deviation from the floor-union shape, same behaviour, different location. Neither method copies enemies, so only the caller can re-link membership, and the id map has to reach `MDTMappingImportService`. Threading it through as a by-reference out-param was tried first and abandoned: the repo's `ErickSkrauch/align_multiline_parameters` fixer mangles `&$param` into `?array   &   $param`. Both existing callers now invoke it; the interface doc-block says every future caller should too.

**`MDTMappingImportService::importEnemies()`** translates each matched enemy's `enemy_forces_checkpoint_id` through that map. It could not simply join the `$fields` carry-over list next to `teeming`/`faction`/`required`/`skippable`/`kill_priority` — that list copies verbatim, which would attach the enemy to the *previous* mapping version's checkpoint. A predecessor whose checkpoint is not in the map logs a warning and lands unassigned rather than mis-assigned.

**Checkpoints left with no members** after the import — their enemies were removed upstream by MDT, or could not be matched — are deleted and logged, instead of surviving as an empty checkpoint reporting zero enemy forces. That is the failure mode the original "don't copy" comment was written to avoid. It is a query-builder delete on purpose: `SeederModel`'s `deleting` listener refuses the delete unless an admin is authenticated, which is never true for the scheduled import, and the model event it skips (`deleted`, which releases member enemies) is a no-op for a checkpoint with no members. The model cache is flushed by hand afterwards — laravel-model-caching's `CachedBuilder` only flushes on model events and `truncate()`, never on a builder delete.

## Intentional asymmetry between the two call sites

| | checkpoints cloned | memberless ones pruned |
|---|---|---|
| "Add bare mapping version" | yes | **no** |
| MDT mapping import | yes | **yes** |

A bare mapping version has no enemies at all, so *every* checkpoint it clones is memberless — pruning there would just reproduce today's "re-author from scratch". The mapper keeps their anchors and re-assigns members as they add enemies. On an MDT import the enemy set is repopulated, so a checkpoint that is still empty afterwards genuinely lost its members.

## Known and accepted: `--force` imports

`$forceImport = true` does not carry `floor_id` / `lat` / `lng` over. Membership can therefore survive onto an enemy that MDT moved to another floor, leaving the checkpoint anchored on a floor with no members. That degrades to a satellite pill (`EnemyForcesCheckpoint::_refreshSatellitePill()`), not a crash, and matches how the other carried-over fields already behave on a force import. Not addressed here.

## Seeder half (Wotuu's second ask)

`mapping:save` and `DungeonDataSeeder` were already wired for checkpoints in #3660 — but the seeder would have **died** the moment a single checkpoint existed: `EnemyForcesCheckpoint` uses the `SeederModel` trait, so `DatabaseSeeder::getTempTableName()` resolves it to `enemy_forces_checkpoints_temp`, and `flushModels()` inserts into that table — which was never created, because the model was missing from `DungeonDataSeeder::getAffectedModelClasses()`. Fixed, plus a new test that pins the invariant (`uses SeederModel` ⟺ `listed in getAffectedModelClasses()`) for every registered relation mapping — that is what would have caught this.

Export verified by round trip against the dev database: a checkpoint with a member enemy produced

```json
// legion/blackrookhold/1/enemy_forces_checkpoints.json
[{ "id": 58, "mapping_version_id": 294, "floor_id": 2, "name": "...", "lat": -54.89, "lng": 89.72 }]
// legion/blackrookhold/1/enemies.json
{ "id": 45996, ..., "enemy_forces_checkpoint_id": 58, ... }
```

The regenerated seeder JSON is **not** part of this MR (818 files of unrelated dev drift). The full `db:seed` back-import was deliberately not run either: 125 of 160 `dungeonroutes.json` had drifted to `[]` in this environment, so seeding would have wiped demo routes out of the shared dev database. The load path is covered by the invariant test above, the already-registered `EnemyForcesCheckpointRelationMapping`, and the fact that temp tables are `CREATE TABLE ... LIKE` (so they inherit the column automatically).

## Tests

- `CopyMappingVersionContentsEnemyForcesCheckpointTest` — the clone returns the id map, and `copyMappingVersionContentsToDungeon()` copies no enemies (which is *why* the map has to be returned).
- `MDTMappingImportEnemyForcesCheckpointTest` — `importEnemies()` with a stubbed `MDTDungeon`: the matched enemy lands on the *cloned* checkpoint, the emptied checkpoint is dropped. Driving the real `importMappingVersionFromMDT()` was rejected: it parses MDT's Lua and rewrites shared NPC/dungeon rows of the seeded test database, none of which this behaviour depends on.
- `DungeonDataSeederTest` — the seeder invariant described above.

Both feature tests deliberately pick a dungeon with no checkpoints of its own and borrow only enemies that are not already a member of one, so they neither clone nor clobber hand-made rows in the shared test database.

`composer run fix` and `composer run analyse` are clean; the four affected groups (`MappingVersion`, `EnemyForcesCheckpoint`, `MDT`, `DatabaseSeeder`) pass — 94 tests.

## MR-ready checklist

- **Visual verification**: N/A — no rendering change; the front-end checkpoint code is untouched.
- **Independent review**: not run by me (the review skill is not model-invocable in this session). Worth a `/code-review` pass before merge.
- **CI**: see checks below.

### DIFF
--- a/app/Http/Controllers/Dungeon/MappingVersionController.php
+++ b/app/Http/Controllers/Dungeon/MappingVersionController.php
@@ -49,12 +49,8 @@ class MappingVersionController extends Controller
                     $newMappingVersion,
                 );
 
-                // A bare mapping version has no enemies, so its checkpoints arrive without members - but the
-                // mapper who added the checkpoints should not have to draw them again (#3702)
-                $mappingService->copyEnemyForcesCheckpointsToMappingVersion(
-                    $currentMappingVersion,
-                    $newMappingVersion,
-                );
+                // A bare mapping version has no enemies, so enemy forces checkpoints - which are only
+                // meaningful once enemies are assigned to them - deliberately don't come along either (#3702).
             }
 
             Session::flash('status', __('controller.mappingversion.created_bare_successfully'));
--- a/app/Http/Controllers/Dungeon/MappingVersionController.php
+++ b/app/Http/Controllers/Dungeon/MappingVersionController.php
@@ -48,6 +48,13 @@ class MappingVersionController extends Controller
                     $currentMappingVersion,
                     $newMappingVersion,
                 );
+
+                // A bare mapping version has no enemies, so its checkpoints arrive without members - but the
+                // mapper who added the checkpoints should not have to draw them again (#3702)
+                $mappingService->copyEnemyForcesCheckpointsToMappingVersion(
+                    $currentMappingVersion,
+                    $newMappingVersion,
+                );
             }
 
             Session::flash('status', __('controller.mappingversion.created_bare_successfully'));
--- a/app/Http/Controllers/Dungeon/MappingVersionController.php
+++ b/app/Http/Controllers/Dungeon/MappingVersionController.php
@@ -48,9 +48,6 @@ class MappingVersionController extends Controller
                     $currentMappingVersion,
                     $newMappingVersion,
                 );
-
-                // A bare mapping version has no enemies, so enemy forces checkpoints - which are only
-                // meaningful once enemies are assigned to them - deliberately don't come along either (#3702).
             }
 
             Session::flash('status', __('controller.mappingversion.created_bare_successfully'));
--- a/app/Service/MDT/MDTMappingImportService.php
+++ b/app/Service/MDT/MDTMappingImportService.php
@@ -89,17 +89,18 @@ class MDTMappingImportService implements MDTMappingImportServiceInterface
             $newMappingVersion = $mappingService->createNewMappingVersionFromMDTMapping($dungeon, $gameVersion, $this->getMDTMappingHash($dungeon));
             $this->log->importMappingVersionFromMDTCreateMappingVersion($newMappingVersion->version, $newMappingVersion->id);
 
-            // Enemies are re-created from MDT below, so their checkpoint membership has to be re-linked by hand -
-            // which is why the checkpoints are cloned here rather than as part of the mapping version's contents
-            $enemyForcesCheckpointIdMapping = $mappingService->copyEnemyForcesCheckpointsToMappingVersion(
-                $currentMappingVersion,
-                $newMappingVersion,
-            );
-
             $mdtDungeon = new MDTDungeon($this->cacheService, $this->coordinatesService, $dungeon);
 
             try {
                 $this->log->importMappingVersionFromMDTStart($dungeon->id);
+
+                // Enemies are re-created from MDT below, so their checkpoint membership has to be re-linked by
+                // hand - which is why the checkpoints are cloned here rather than as part of the mapping
+                // version's contents
+                $enemyForcesCheckpointIdMapping = $mappingService->copyEnemyForcesCheckpointsToMappingVersion(
+                    $currentMappingVersion,
+                    $newMappingVersion,
+                );
 
                 $this->importDungeon($mdtDungeon, $dungeon, $newMappingVersion);
                 $this->importNpcs($newMappingVersion, $mdtDungeon, $dungeon, $gameVersion);
@@ -601,6 +602,10 @@ class MDTMappingImportService implements MDTMappingImportServiceInterface
         EnemyForcesCheckpoint::query()
             ->whereIn('id', $emptyEnemyForcesCheckpoints->pluck('id'))
             ->delete();
+
+        // laravel-model-caching only flushes on model events and truncate(), never on a builder delete, so the
+        // reads above would otherwise keep serving the checkpoints this just removed
+        new EnemyForcesCheckpoint()->flushCache();
     }
 
     /**
--- a/app/Service/MDT/MDTMappingImportService.php
+++ b/app/Service/MDT/MDTMappingImportService.php
@@ -9,6 +9,7 @@ use App\Logic\MDT\Entity\MDTPatrol;
 use App\Models\Dungeon;
 use App\Models\DungeonFloorSwitchMarker;
 use App\Models\Enemy;
+use App\Models\EnemyForcesCheckpoint;
 use App\Models\EnemyPack;
 use App\Models\EnemyPatrol;
 use App\Models\Faction;
@@ -88,6 +89,13 @@ class MDTMappingImportService implements MDTMappingImportServiceInterface
             $newMappingVersion = $mappingService->createNewMappingVersionFromMDTMapping($dungeon, $gameVersion, $this->getMDTMappingHash($dungeon));
             $this->log->importMappingVersionFromMDTCreateMappingVersion($newMappingVersion->version, $newMappingVersion->id);
 
+            // Enemies are re-created from MDT below, so their checkpoint membership has to be re-linked by hand -
+            // which is why the checkpoints are cloned here rather than as part of the mapping version's contents
+            $enemyForcesCheckpointIdMapping = $mappingService->copyEnemyForcesCheckpointsToMappingVersion(
+                $currentMappingVersion,
+                $newMappingVersion,
+            );
+
             $mdtDungeon = new MDTDungeon($this->cacheService, $this->coordinatesService, $dungeon);
 
             try {
@@ -95,7 +103,14 @@ class MDTMappingImportService implements MDTMappingImportServiceInterface
 
                 $this->importDungeon($mdtDungeon, $dungeon, $newMappingVersion);
                 $this->importNpcs($newMappingVersion, $mdtDungeon, $dungeon, $gameVersion);
-                $enemies = $this->importEnemies($currentMappingVersion, $newMappingVersion, $mdtDungeon, $dungeon, $forceImport);
+                $enemies = $this->importEnemies(
+                    $currentMappingVersion,
+                    $newMappingVersion,
+                    $mdtDungeon,
+                    $dungeon,
+                    $enemyForcesCheckpointIdMapping,
+                    $forceImport,
+                );
                 $this->importEnemyPacks($newMappingVersion, $mdtDungeon, $dungeon, $enemies);
                 $this->importEnemyPatrols($currentMappingVersion, $newMappingVersion, $mdtDungeon, $dungeon, $enemies);
                 $this->importMapPOIs($currentMappingVersion, $newMappingVersion, $mdtDungeon, $dungeon);
@@ -415,6 +430,9 @@ class MDTMappingImportService implements MDTMappingImportServiceInterface
     }
 
     /**
+     * @param  array<int, int>        $enemyForcesCheckpointIdMapping Source checkpoint id => cloned checkpoint id, as
+     *                                                                built by MappingService when it cloned the
+     *                                                                previous mapping version's checkpoints.
      * @return Collection<int, Enemy>
      */
     private function importEnemies(
@@ -422,6 +440,7 @@ class MDTMappingImportService implements MDTMappingImportServiceInterface
         MappingVersion $newMappingVersion,
         MDTDungeon     $mdtDungeon,
         Dungeon        $dungeon,
+        array          $enemyForcesCheckpointIdMapping,
         bool           $forceImport = false,
     ): Collection {
         // Get a list of new enemies and save them
@@ -461,8 +480,10 @@ class MDTMappingImportService implements MDTMappingImportServiceInterface
                 unset($mdtEnemy->enemy_id);
 
                 // Is group ID - we handle this later on
-                $mdtEnemy->enemy_pack_id      = null;
-                $mdtEnemy->mapping_version_id = $newMappingVersion->id;
+                $mdtEnemy->enemy_pack_id = null;
+                // Membership is restored below, but only for enemies we can match to their predecessor
+                $mdtEnemy->enemy_forces_checkpoint_id = null;
+                $mdtEnemy->mapping_version_id         = $newMappingVersion->id;
 
                 $existingEnemy = $currentEnemies->get($mdtEnemy->getUniqueKey());
                 if ($existingEnemy instanceof Enemy) {
@@ -509,6 +530,23 @@ class MDTMappingImportService implements MDTMappingImportServiceInterface
                         $updatedFields['seasonal_type'] = $existingEnemy->seasonal_type;
                     }
 
+                    // Enemy forces checkpoints were cloned into the new mapping version before we got here, so the
+                    // predecessor's checkpoint id must be translated - copying it verbatim would attach this enemy to
+                    // the previous mapping version's checkpoint (#3702).
+                    if ($existingEnemy->enemy_forces_checkpoint_id !== null) {
+                        $newEnemyForcesCheckpointId = $enemyForcesCheckpointIdMapping[$existingEnemy->enemy_forces_checkpoint_id] ?? null;
+
+                        if ($newEnemyForcesCheckpointId === null) {
+                            $this->log->importEnemiesCannotRecoverEnemyForcesCheckpoint(
+                                $mdtEnemy->getUniqueKey(),
+                                $existingEnemy->enemy_forces_checkpoint_id,
+                            );
+                        } else {
+                            $mdtEnemy->enemy_forces_checkpoint_id        = $newEnemyForcesCheckpointId;
+                            $updatedFields['enemy_forces_checkpoint_id'] = $newEnemyForcesCheckpointId;
+                        }
+                    }
+
                     $this->log->importEnemiesRecoverPropertiesFromExistingEnemy($mdtEnemy->getUniqueKey(), $updatedFields);
                 } else {
                     $this->log->importEnemiesCannotRecoverPropertiesFromExistingEnemy($mdtEnemy->getUniqueKey());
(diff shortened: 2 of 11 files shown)


app/Service/MDT/MDTMappingImportService.php
```diff
@@ -524,13 +563,51 @@ private function importEnemies(
...
 
         return $enemies;
     }
 
+    /**
+     * An MDT mapping bump is exactly when the enemy set changes. A checkpoint whose members were all removed
+     * upstream - or which lost them because we could not match them to their predecessor - would otherwise
+     * survive as an empty checkpoint reporting zero enemy forces, which is not what the mapper drew (#3702).
+     */
+    private function deleteEmptyEnemyForcesCheckpoints(MappingVersion $newMappingVersion): void
```
Wotuu: :robot: **The bare-mapping-version half of the fix does not survive the next MDT import — every checkpoint is silently wiped.**

The MR body's asymmetry table promises that on "Add bare mapping version" checkpoints are cloned and **not** pruned, so "the mapper keeps their anchors and re-assigns members as they add enemies". That promise expires at the next MDT bump:

1. `MappingService::copyMappingVersionToDungeon()` carries `mdt_mapping_hash` forward from the source, so the bare version stays hash-equal and no import fires immediately.
2. A bare mapping version has **zero** enemies (`copyMappingVersionContentsToDungeon()` copies none).
3. When MDT genuinely changes, `importMappingVersionFromMDT()` takes that bare version as `$currentMappingVersion`. `importEnemies()` builds `$currentEnemies` from it — empty. No `$existingEnemy` ever matches, so no imported enemy gets an `enemy_forces_checkpoint_id`.
4. Every cloned checkpoint is therefore memberless, and `deleteEmptyEnemyForcesCheckpoints()` deletes all of them.

So the anchors preserved in step 1 are exactly the ones destroyed in step 4 — and this is the state a bare version is *designed* to sit in (checkpoints drawn, members not yet re-assigned).

The prune's own stated intent already excludes this case: *"a checkpoint that is still empty afterwards genuinely lost its members"*. A checkpoint that never had members in the source version lost nothing. Suggest scoping the prune to clones whose **source** counterpart actually had members — both extra inputs are already parameters of `importEnemies()`:

```php
$hadMembers = $currentMappingVersion->enemyForcesCheckpoints()
    ->has('enemies')
    ->pluck('id')
    ->map(static fn(int $id): ?int => $enemyForcesCheckpointIdMapping[$id] ?? null)
    ->filter();
```

and only delete the currently-empty clones that appear in that set. One extra query, and it keeps the documented behaviour for the case the prune was actually written for (members removed upstream).

Worth noting alongside this: `importEnemiesDeleteEmptyEnemyForcesCheckpoint()` logs at `warning`, and Discord alerting is error-level — so a dungeon losing its entire checkpoint set is invisible to whoever ran `mdt:importmapping` unless they go read the log.

app/Service/Mapping/MappingServiceInterface.php
```diff
@@ -34,9 +34,25 @@ public function copyMappingVersionToDungeon(MappingVersion $sourceMappingVersion
 
     /**
      * Takes an existing mapping version's contents and applies it to another mapping version.
+     *
+     * Enemy forces checkpoints are NOT part of this: their members live in `enemies`, which this method does
+     * not copy either, so the caller has to be able to re-link membership itself. Clone them with
+     * copyEnemyForcesCheckpointsToMappingVersion() - every caller of this method should call that too (#3702).
```
Wotuu: :robot: **A doc-block is the only thing holding this pairing together, and the skill that exists precisely to stop this class of bug wasn't updated.**

Moving the checkpoint clone *out* of `copyMappingVersionContentsToDungeon()` is a reasonable call given only the caller can re-link membership — but it converts "one hardcoded model list to remember" into "one list **plus** a second method you must remember to call", enforced by nothing but this comment. Two concrete ways that regresses:

1. **A future third caller of `copyMappingVersionContentsToDungeon()`** silently drops checkpoints, which is exactly the bug #3702 was filed for. Nothing fails, nothing is logged; the checkpoints just aren't there.
2. **A future mapping-versioned model.** `.claude/skills/mapping-versioned-models/SKILL.md` is unchanged by this MR. Its wiring checklist step 4 (line 124) still reads *"`MappingService::copyMappingVersionContentsToDungeon()`: add a clone loop **if** the model must survive the MDT-copy path"*, and its Gotchas section (lines 140-141) still enumerates four hardcoded lists — none of which mention `copyEnemyForcesCheckpointsToMappingVersion()` or the fact that the MDT path now needs an id map threaded back to the caller. Anyone following that checklist after this merges lands their clone loop in the method that no longer covers checkpoints.

#3660 set the precedent here: when the `Copy.php` `$relations` trap was found in review, the fix landed *and* the trap was written into this same skill ("Don't forget `Copy.php`", lines 143-145), explicitly because the three lists in `MappingVersion::boot()` weren't the whole story. This MR adds a fifth place with the same failure shape and leaves the skill describing the pre-#3702 world.

Minimum ask: a short section in that skill naming the paired call and why the id map has to cross the service boundary. Optional but stronger: make the pairing unmissable in code — e.g. have `copyMappingVersionContentsToDungeon()` perform the clone and expose the id map (return a small DTO, or a `getLastEnemyForcesCheckpointIdMapping()`), so a caller that ignores it gets nothing worse than an unused value rather than missing rows.

tests/Feature/App/Service/MDT/MDTMappingImportEnemyForcesCheckpointTest.php
```diff
@@ -0,0 +1,199 @@
...
+                'importEnemies',
+            );
+            $importEnemies->invoke(
+                $this->app->make(MDTMappingImportServiceInterface::class),
+                $sourceMappingVersion,
+                $newMappingVersion,
+                $mdtDungeon,
+                $dungeon,
+                $enemyForcesCheckpointIdMapping,
+                true,
```
Wotuu: :robot: **The only test of the new re-link path runs `$forceImport = true` — the variant this MR explicitly puts out of scope.**

`mdt:importmapping {dungeon} {gameVersion} {--force}` defaults `$force` to `false`, and `importMappingVersionFromMDT()`'s parameter defaults to `false` too, so every real mapping bump takes the non-force path. This test hardcodes `true`.

To be clear about what that does and doesn't mean: the new `enemy_forces_checkpoint_id` translation sits *outside* the `if (!$forceImport)` guard, so the re-link logic itself is identical either way — this is a coverage gap, not a latent logic difference. But it is a gap in the wrong direction, because the MR body already carves force-imports out of its guarantees ("Known and accepted: `--force` imports … Not addressed here"). The net effect is that the one behaviour that *is* claimed to be fixed is only ever exercised through the mode that is disclaimed.

It also looks free to fix: `createMdtCloneOf()` copies `floor_id`/`lat`/`lng` verbatim off the predecessor, so on the non-force path the distance check computes 0, stays well under the 150 threshold, and the extra `floor_id`/`lat`/`lng` carry-over is a no-op. Passing `false` (or better, running both via a data provider) should pass unchanged and would additionally pin that the checkpoint translation coexists with the `$fields` carry-over rather than being shadowed by it.

Related, and worth being explicit about in the MR body rather than leaving implied: #3660's deferral point 4 was *"No harness exists — `MDTMappingImportService` has no test coverage at all … Shipping this untested into the path that rebuilds every retail dungeon's mapping is not a trade I would make."* Both new tests reach past `importMappingVersionFromMDT()` and drive `importEnemies()` by reflection, so that public entry point — the one the clone/import/prune ordering actually lives in, and where the checkpoint clone was moved to in `89fa98df` — is still uncovered. The reason given (a real import rewrites shared NPC/dungeon rows) is fair, but it means the ordering guarantee this MR depends on is asserted by construction, not by a test.

Wotuu:

Wotuu:
:robot: **Heads-up on CI coverage for this MR (and any MR stacked on #3660).**

`php-tests.yml` and `phpstan.yml` are both gated on `pull_request: branches: [master]`, so neither ran here — this MR targets `128-floor-enemy-forces-pill`. The three checks that did run (`build`, `js-tests`, `php-cs-fixer`) are green, but the PHP suite and PHPStan were never exercised by CI.

I ran both locally in the worktree instead:

- `composer run analyse` — **no errors**
- `php artisan test --exclude-group=MapTiles` (the full suite, matching what the CI shards cover) — **1187 passed, 1 skipped, 16846 assertions**, 0 failures

Worth knowing generally: any MR that stacks on another branch silently loses PHP test + static analysis coverage. Adding the base branch to those two workflow triggers, or switching them to an unfiltered `pull_request:` like `js-tests.yml`/`php-cs-fixer.yml` already use, would close the gap.

app/Service/MDT/MDTMappingImportService.php
Wotuu: :robot: Confirmed and fixed — this was a real bug, thank you for catching it.

`deleteEmptyEnemyForcesCheckpoints()` now takes the source mapping version and the id map, and only prunes a still-empty clone whose **source** counterpart genuinely had at least one member:

```php
$sourceIdsWithMembers = $currentMappingVersion->enemyForcesCheckpoints()
    ->has('enemies')
    ->pluck('id')
    ->all();
$prunableCloneIds = array_intersect_key($enemyForcesCheckpointIdMapping, array_flip($sourceIdsWithMembers));

$enemyForcesCheckpointsToPrune = $emptyEnemyForcesCheckpoints->whereIn('id', array_values($prunableCloneIds));
```

A checkpoint cloned by "Add bare mapping version" is memberless in the source by construction, so it's never in `$sourceIdsWithMembers` and now survives untouched — exactly the bare-mapping-version case this MR exists to preserve. The genuinely-emptied case (an enemy that MDT reports gone, or couldn't be matched) still has its source checkpoint show up with a member, so it's still pruned. Confirmed the existing test (`['Kept corridor']` assertion) still passes unchanged after the fix.

On severity: I kept the per-checkpoint line at `warning`, but added a new aggregate log call at `error` (Discord-alerting) — `importEnemiesPrunedEmptyEnemyForcesCheckpoints(dungeonId, count)` — fired once per import when at least one checkpoint is actually pruned. Post-fix, that only happens for a checkpoint that *did* have members and lost them, which is exactly the "you'd want to be pinged about this" bar from the structured-logging skill, so bumping it to error (rather than just adding a warning-level summary) seemed right.

New regression test added: `importEnemies_givenCheckpointThatNeverHadMembersInSource_survivesEvenThoughItIsEmptyAfterImport` in `MDTMappingImportEnemyForcesCheckpointTest.php`. It creates a checkpoint with zero enemies on the source mapping version (mirroring what "Add bare mapping version" produces), runs it through the full clone → `importEnemies()` pipeline with an MDT response that reports no relevant enemies, and asserts the checkpoint's clone still exists afterwards. Ran it (and the whole file) 4x back to back with no flakiness.

Files: `app/Service/MDT/MDTMappingImportService.php`, `app/Service/MDT/Logging/MDTMappingImportServiceLogging(Interface).php`, `tests/Feature/App/Service/MDT/MDTMappingImportEnemyForcesCheckpointTest.php`.

app/Service/Mapping/MappingServiceInterface.php
Wotuu: :robot: Addressed the required doc part; deliberately skipped the optional structural one.

**Required — done.** Updated `.claude/skills/mapping-versioned-models/SKILL.md`:
- "The two clone code paths" section now names `copyEnemyForcesCheckpointsToMappingVersion()` as a separate paired method (rather than describing checkpoints as merely "excluded"), and adds a new subsection explaining *why* it's separate (checkpoints' members live in `enemies`, which neither copy path touches) and why the id map has to cross the `MappingService` boundary back to the caller (only the caller can re-link membership).
- Checklist step 4 now mentions the paired-method pattern as an alternative to a plain clone loop, for a future model whose children live in a table `copyMappingVersionContentsToDungeon()` doesn't touch.
- Gotchas now lists this as the **fifth** hardcoded-list-shaped trap (previously only four were named), explicitly noting it's enforced by a doc-block comment only, and that the fix requires threading an id map back across the service boundary — which the other four don't need.

**Optional structural fix — skipped, on purpose.** I considered folding the clone into `copyMappingVersionContentsToDungeon()` itself and returning the id map from there (or via a getter), but `copyMappingVersionContentsToDungeon()` currently returns `MappingVersion` and that return value is used as-is by `createNewMappingVersionFromMDTMapping()` (returned straight through) and by both new tests. Changing its contract would touch the interface, both existing callers, and both regression tests added in this MR for what's currently a two-line doc-comment risk. Given the fix for finding 1 already makes the failure mode (forgetting the pairing) loudly visible via the new error-level log line whenever a whole checkpoint set goes missing, I judged the doc fix sufficient here and left the structural enforcement for a future MR if a second real model ever needs the same shape.

`.claude/skills/mapping-versioned-models/SKILL.md`.

tests/Feature/App/Service/MDT/MDTMappingImportEnemyForcesCheckpointTest.php
Wotuu: :robot: Fixed — converted the hardcoded `$forceImport = true` into a data provider covering both `true` and `false`:

```php
#[Test]
#[DataProvider('importEnemies_givenClonedEnemyForcesCheckpoints_relinksMatchedEnemiesAndDropsEmptiedCheckpoints_dataProvider')]
public function importEnemies_givenClonedEnemyForcesCheckpoints_relinksMatchedEnemiesAndDropsEmptiedCheckpoints(bool $forceImport): void
```

with the provider yielding `'force import' => [true]` and `'non-force import (the default; every real mapping bump takes this path)' => [false]`.

Ran it empirically rather than trusting the claim in the MR body: both cases pass, and I ran the file 4x back to back to rule out flakiness. Your read of why `false` is a no-op here checks out — `createMdtCloneOf()` builds the mock MDT enemy directly from the existing enemy's own `floor_id`/`lat`/`lng`, so on the non-force path the distance check compares the existing enemy's location against an identical location and always takes the "copy lat/lng over" branch; nothing in the checkpoint re-link logic (which sits outside the `if (!$forceImport)` block, as you noted) depends on which branch that takes.

I also added a second, separate test for the bare-mapping-version regression from the other finding on this MR (checkpoint with zero source members surviving an import) — that one only needed `forceImport = false` since it's testing the realistic path.

`tests/Feature/App/Service/MDT/MDTMappingImportEnemyForcesCheckpointTest.php`.

app/Http/Controllers/Dungeon/MappingVersionController.php
```diff
@@ -48,6 +48,13 @@ public function saveNew(
                     $currentMappingVersion,
                     $newMappingVersion,
                 );
+
+                // A bare mapping version has no enemies, so its checkpoints arrive without members - but the
+                // mapper who added the checkpoints should not have to draw them again (#3702)
+                $mappingService->copyEnemyForcesCheckpointsToMappingVersion(
+                    $currentMappingVersion,
+                    $newMappingVersion,
+                );
```
Wotuu: Get rid of this - bare mapping versions shouldn't have anything to do with enemies. In `copyMappingVersionContentsToDungeon` you can see that it there copies over everything that are hard-coded features of the map itself. Such as floor switches, mountable areas, floor unions. Things that literally don't change because the map is the map itself. Checkpoints focus on enemies and these enemies we do NOT want in the bare mapping. So the checkpoints should also not be there.

I see that further down the line when re-importing a dungeon from MDT this function is necessary - so it can live. But only called from there since it makes sense there - we want to reduce to the amount of admin required per new mapping import, so attempting to restore the checkpoints makes sense there.

Wotuu (commented):
One change

app/Http/Controllers/Dungeon/MappingVersionController.php
Wotuu: :robot: Fixed in 351eeacb5. Removed the `copyEnemyForcesCheckpointsToMappingVersion()` call from the "Add bare mapping version" branch entirely - agreed, a bare mapping version has no enemies to ever assign to a cloned checkpoint, so it would just sit permanently empty. The MDT re-import path (`MDTMappingImportService`) is the only remaining caller, since that path does recreate/re-link enemies afterwards. Also updated the interface doc-block and the mapping-versioned-models skill doc, both of which previously said every caller of `copyMappingVersionContentsToDungeon()` should pair the two calls - that's no longer true, it's caller-specific now.

app/Http/Controllers/Dungeon/MappingVersionController.php
```diff
@@ -49,12 +49,8 @@ public function saveNew(
...
                 );
 
-                // A bare mapping version has no enemies, so its checkpoints arrive without members - but the
-                // mapper who added the checkpoints should not have to draw them again (#3702)
-                $mappingService->copyEnemyForcesCheckpointsToMappingVersion(
-                    $currentMappingVersion,
-                    $newMappingVersion,
-                );
+                // A bare mapping version has no enemies, so enemy forces checkpoints - which are only
+                // meaningful once enemies are assigned to them - deliberately don't come along either (#3702).
```
Wotuu: Remove this comment.

Wotuu (commented):
Fix this comment and you can immediately merge this PR.

app/Http/Controllers/Dungeon/MappingVersionController.php
Wotuu: :robot: Removed in dace93e5d.

PR #3711, opened 2026-07-27
#3710 Fix the third flaky raid-marker mapping-version test

### LINKED ISSUE #3710: Fix remaining flaky raid-marker mapping-version test (3rd method missed by #3679)
:robot: #3679/#3680 fixed two of the three tests in `DungeonRouteEnemyRaidMarkerMappingVersionTest` for the same root cause (picking a random retail dungeon first, then querying for an enemy that may not exist on it — 3 of 70 retail dungeons have zero enemies).

The third test, `upgradeMappingVersion_givenNpcNoLongerExistsInNewMappingVersion_deletesRaidMarker`, still has the identical bug: `Enemy::where('mapping_version_id', $existingMV->id)->inRandomOrder()->first()` with no null-guarantee, on a dungeon picked via `getRetailDungeon()` before checking it has any enemies at all.

Surfaced by CI going red on PR #3663 (`php-tests (feature-other)`) with the exact same `TypeError: createRaidMarker(): Argument #2 ($enemy) must be of type App\Models\Enemy, null given` as #3679 described.

Fix: mirror the shuffle-and-search pattern `getRetailDungeonWithUniquelyIdentifiedEnemy()`/`getRetailDungeonWithNullMdtIdEnemy()` already use — a plain `getRetailDungeonWithEnemy()` helper (no uniqueness constraint needed for this test, since the delete-by-pair step already removes every sibling sharing that pair, matching or not).

Wotuu:
:robot: Closes #3710

Test-only, one file, branched straight off `master` so it merges independently of PR #3663 (whose CI run is what surfaced it).

## Same root cause as #3679/#3680, missed test

`DungeonRouteEnemyRaidMarkerMappingVersionTest` had two of its three tests fixed for this exact bug: `getRetailDungeon()` (or an equivalent random-first pick) selects a dungeon *before* checking it has any qualifying data, and 3 of the 70 retail dungeons (The Arcway, Cathedral of Eternal Night, Maw of Souls) have **zero enemies at all** on their current mapping version.

The third test, `upgradeMappingVersion_givenNpcNoLongerExistsInNewMappingVersion_deletesRaidMarker`, still did exactly that:

```php
$dungeon    = $this->getRetailDungeon();
$existingMV = $dungeon->getCurrentMappingVersion();
...
$enemy = Enemy::where('mapping_version_id', $existingMV->id)->inRandomOrder()->first();
```

Drawing one of the three empty dungeons returns a null `$enemy`, and `createRaidMarker()` throws `TypeError: Argument #2 ($enemy) must be of type App\Models\Enemy, null given`. This is exactly what went red on PR #3663's `php-tests (feature-other)` job — a branch with no relation at all to mapping versions or enemies, which is what gave it away as a pre-existing master-level flake rather than something caused by that PR's diff.

## Fix

Added `getRetailDungeonWithEnemy()`, mirroring the shuffle-and-search pattern `getRetailDungeonWithUniquelyIdentifiedEnemy()`/`getRetailDungeonWithNullMdtIdEnemy()` already use in this file — search shuffled dungeons for one that actually has an enemy, instead of sampling a dungeon and hoping. No uniqueness constraint is needed here (unlike the other two helpers): this test deletes every enemy sharing the picked enemy's `(npc_id, mdt_id)` pair in the new mapping version, so an ambiguous pick is deleted just as completely as a unique one.

## Verification

Ran the whole test class 8 consecutive times locally: the target test passed all 8/8 (previously ~1-in-20 failure rate). `composer run fix` (0 files touched) and `composer run analyse` both clean.

### DIFF
--- a/tests/Feature/App/Service/DungeonRoute/DungeonRouteEnemyRaidMarkerMappingVersionTest.php
+++ b/tests/Feature/App/Service/DungeonRoute/DungeonRouteEnemyRaidMarkerMappingVersionTest.php
@@ -93,12 +93,8 @@ final class DungeonRouteEnemyRaidMarkerMappingVersionTest extends DungeonRouteSa
     public function upgradeMappingVersion_givenNpcNoLongerExistsInNewMappingVersion_deletesRaidMarker(): void
     {
         // Arrange
-        $dungeon    = $this->getRetailDungeon();
-        $existingMV = $dungeon->getCurrentMappingVersion();
-        $newMV      = $this->createNewerMappingVersion($dungeon, $existingMV);
-
-        /** @var Enemy $enemy */
-        $enemy = Enemy::where('mapping_version_id', $existingMV->id)->inRandomOrder()->first();
+        [$dungeon, $existingMV, $enemy] = $this->getRetailDungeonWithEnemy();
+        $newMV                          = $this->createNewerMappingVersion($dungeon, $existingMV);
 
         $route      = DungeonRoute::factory()->create(['dungeon_id' => $dungeon->id, 'mapping_version_id' => $existingMV->id]);
         $raidMarker = $this->createRaidMarker($route, $enemy);
@@ -132,6 +128,42 @@ final class DungeonRouteEnemyRaidMarkerMappingVersionTest extends DungeonRouteSa
             'mdt_id'           => $enemy->mdt_id,
             'enemy_id'         => $enemy->id,
         ]);
+    }
+
+    /**
+     * A retail dungeon that actually has at least one enemy on its current mapping version.
+     *
+     * Picking a random dungeon and *then* querying for an enemy was flaky: 3 of the 70 retail
+     * dungeons (The Arcway, Cathedral of Eternal Night, Maw of Souls) have no enemies at all on
+     * their current mapping version, so roughly one run in twenty drew one of them, got a null
+     * enemy and died with a TypeError in createRaidMarker(). Search across dungeons instead,
+     * mirroring getRetailDungeonWithUniquelyIdentifiedEnemy() above.
+     *
+     * No uniqueness constraint is needed here (unlike the two helpers above): this test deletes
+     * every enemy sharing the picked enemy's (npc_id, mdt_id) pair in the new mapping version, so
+     * an ambiguous pick is deleted just as completely as a unique one.
+     *
+     * @return array{0: Dungeon, 1: MappingVersion, 2: Enemy}
+     */
+    private function getRetailDungeonWithEnemy(): array
+    {
+        $dungeons = Dungeon::whereNotNull('challenge_mode_id')->get()->shuffle();
+
+        foreach ($dungeons as $dungeon) {
+            /** @var Dungeon $dungeon */
+            $mappingVersion = $dungeon->getCurrentMappingVersion();
+            if ($mappingVersion === null) {
+                continue;
+            }
+
+            $enemy = Enemy::where('mapping_version_id', $mappingVersion->id)->inRandomOrder()->first();
+
+            if ($enemy !== null) {
+                return [$dungeon, $mappingVersion, $enemy];
+            }
+        }
+
+        self::markTestSkipped('Unable to find a retail dungeon with an enemy on its current mapping version');
     }
 
     /**
--- a/tests/Feature/App/Service/DungeonRoute/DungeonRouteEnemyRaidMarkerMappingVersionTest.php
+++ b/tests/Feature/App/Service/DungeonRoute/DungeonRouteEnemyRaidMarkerMappingVersionTest.php
@@ -35,9 +35,13 @@ final class DungeonRouteEnemyRaidMarkerMappingVersionTest extends DungeonRouteSa
         $raidMarker = $this->createRaidMarker($route, $enemy);
 
         try {
+            // Look up the clone by the same coalesced key updateEnemyIdsByMappingVersion()
+            // resolves markers by, not raw npc_id: getRetailDungeonWithUniquelyIdentifiedEnemy()
+            // only guarantees uniqueness on (COALESCE(mdt_npc_id, npc_id), mdt_id), so a sibling
+            // enemy can share the raw (npc_id, mdt_id) pair while having a different coalesced key.
             /** @var Enemy $clonedEnemy */
             $clonedEnemy = Enemy::where('mapping_version_id', $newMV->id)
-                ->where('npc_id', $enemy->npc_id)
+                ->whereRaw('COALESCE(mdt_npc_id, npc_id) = ?', [$enemy->getMdtNpcId()])
                 ->where('mdt_id', $enemy->mdt_id)
                 ->firstOrFail();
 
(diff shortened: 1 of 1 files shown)


tests/Feature/App/Service/DungeonRoute/DungeonRouteEnemyRaidMarkerMappingVersionTest.php
```diff
@@ -134,6 +130,42 @@ private function createRaidMarker(DungeonRoute $route, Enemy $enemy): DungeonRou
...
+     *
+     * Picking a random dungeon and *then* querying for an enemy was flaky: 3 of the 70 retail
+     * dungeons (The Arcway, Cathedral of Eternal Night, Maw of Souls) have no enemies at all on
+     * their current mapping version, so roughly one run in twenty drew one of them, got a null
+     * enemy and died with a TypeError in createRaidMarker(). Search across dungeons instead,
+     * mirroring getRetailDungeonWithUniquelyIdentifiedEnemy() above.
+     *
+     * No uniqueness constraint is needed here (unlike the two helpers above): this test deletes
+     * every enemy sharing the picked enemy's (npc_id, mdt_id) pair in the new mapping version, so
+     * an ambiguous pick is deleted just as completely as a unique one.
```
Wotuu: :robot: **The stated justification doesn't quite hold — the delete key and the resolution key are different columns.** No live flake today (0 instances on current seed data), so treat this as optional; but the reasoning as written is what the next reader will trust.

The test deletes by the raw `npc_id` column:

```php
Enemy::where('mapping_version_id', $newMV->id)
    ->where('npc_id', $enemy->npc_id)
    ->where('mdt_id', $enemy->mdt_id)
    ->delete();
```

`DungeonRouteEnemyRaidMarkerRepository::updateEnemyIdsByMappingVersion()` re-resolves on `COALESCE(e.mdt_npc_id, e.npc_id)`, and `createRaidMarker()` above stores `$enemy->getMdtNpcId()` (= `mdt_npc_id ?? npc_id`) on the marker. So a sibling enemy `Y` with `Y.npc_id != $enemy->npc_id` but `COALESCE(Y.mdt_npc_id, Y.npc_id) == $enemy->getMdtNpcId()` and the same `mdt_id` **survives the delete** — and if it is the only such survivor it satisfies the repository's `COUNT(*) = 1` guard, re-resolves the marker, and `assertNull($raidMarker->fresh())` fails. An ambiguous pick is therefore not automatically "deleted just as completely as a unique one": it depends on whether the ambiguity sits on the raw or on the coalesced key.

I measured how close that is: across all 71 current retail mapping versions (12,449 enemies, 25 of which have `mdt_npc_id != npc_id`) there are **0** enemies with such a survivor, so the test cannot fail on today's data. It is latent only — but latent data-dependence is precisely what this PR set out to remove, so it may be worth closing off now by deleting on the same key the resolver matches on:

```php
->whereRaw('COALESCE(mdt_npc_id, npc_id) = ?', [$enemy->getMdtNpcId()])
```

in place of `->where('npc_id', $enemy->npc_id)`. That also makes the simulated "NPC was removed from the map" exact rather than approximate.

Two small doc nits in this same block:

- "3 of the 70 retail dungeons (The Arcway, Cathedral of Eternal Night, Maw of Souls)" is now **4 of 71** — `Altar of Fangs` (Midnight) also has zero enemies on its current mapping version in the seeded database.
- Both "above" references point the wrong way: this helper is inserted *above* `getRetailDungeonWithUniquelyIdentifiedEnemy()` / `getRetailDungeonWithNullMdtIdEnemy()`, so it mirrors the helpers *below* it. (The sibling docblock gets the direction right with "… below".)

tests/Feature/App/Service/DungeonRoute/DungeonRouteEnemyRaidMarkerMappingVersionTest.php
```diff
@@ -134,6 +130,42 @@ private function createRaidMarker(DungeonRoute $route, Enemy $enemy): DungeonRou
...
+            }
+
+            $enemy = Enemy::where('mapping_version_id', $mappingVersion->id)->inRandomOrder()->first();
+
+            if ($enemy !== null) {
+                return [$dungeon, $mappingVersion, $enemy];
+            }
+        }
+
+        self::markTestSkipped('Unable to find a retail dungeon with an enemy on its current mapping version');
```
Wotuu, line 175: :robot: Adjacent finding, not a defect in this PR — but worth knowing before this pattern gets a third copy: **`markTestSkipped` in these helpers can silently switch a test off, and it already has.**

`getRetailDungeonWithNullMdtIdEnemy()` (added in #3680) can never find an enemy on current seed data. There are **0** enemies with a null `mdt_id` anywhere in the seeded database (81,665 enemy rows), and grepping the seeder JSON for `"mdt_id": null` across `database/seeders/dungeondata/` returns 0 hits, so CI's freshly-seeded database has none either. `upgradeMappingVersion_givenNpcWithNullMdtIdStillExistsInNewMappingVersion_preservesRaidMarker` consequently skips rather than runs (current `master`, local):

```
 ✓ upgrade mapping version given npc still exists in new mapping versi… 3.43s
 - upgrade mapping version given npc with null mdt id still exists in…  0.27s
 ✓ upgrade mapping version given npc no longer exists in new mapping v… 3.82s
 Tests:    1 skipped, 2 passed (5 assertions)
```

and this PR's own `php-tests (feature-other)` job reports `Tests: 509, Assertions: 1949, Skipped: 2` — so it is skipping in CI too, not only locally. (#3680's body cited "152 of the 1549 null-`mdt_id` enemies", which no longer matches the data at all; the null-safe `<=>` join coverage that test exists to protect is currently not exercised anywhere.)

For *this* new helper the fallback is effectively unreachable — it would take every retail dungeon having zero enemies — so there is nothing to fix on this line itself. The suggestion is for the family: have these helpers `throw new \RuntimeException(...)` the way `ProvidesDungeon::getDungeonWithNonFacadeFloor()` does, instead of skipping. A helper that can no longer find its data is a coverage regression, and a skip makes that invisible inside a green run. The null-`mdt_id` half probably deserves its own issue rather than scope creep here.

Wotuu:

tests/Feature/App/Service/DungeonRoute/DungeonRouteEnemyRaidMarkerMappingVersionTest.php
Wotuu: :robot: Good catch, confirmed. Fixed in dff7b5780: the delete now matches on `COALESCE(mdt_npc_id, npc_id)` (via `whereRaw`) instead of raw `npc_id`, mirroring exactly what `updateEnemyIdsByMappingVersion()` resolves markers by.

While verifying this I found the same raw-`npc_id` vs coalesced-key mismatch in the two clone lookups in the sibling tests above (`getRetailDungeonWithUniquelyIdentifiedEnemy()`'s uniqueness guarantee is on the coalesced key, not raw `npc_id`) - a query against current seed data found 24 enemy rows across retail mapping versions that share a raw `(npc_id, mdt_id)` pair with a sibling but have a *different* coalesced key, which is a live (not just latent) risk of `firstOrFail()` grabbing the wrong clone in those two tests. Fixed those the same way for consistency.

Also fixed the docblock nits: verified against current seed data it's now 4 of 71 retail dungeons (Altar of Fangs joined the list), and reworded the "above" references to correctly point at the helpers below (matching the sibling docblock's existing wording).

All confirmed via direct query against the current seed - 0 delete-condition mismatches exist today, so this closes off a latent gap rather than fixing an active flake. Ran the suite 5x back to back, all green (`composer run fix`/`analyse` also clean). Pushed as dff7b5780.

tests/Feature/App/Service/DungeonRoute/DungeonRouteEnemyRaidMarkerMappingVersionTest.php
Wotuu: :robot: Confirmed, not a defect in this PR - this is the pre-existing `getRetailDungeonWithNullMdtIdEnemy()` helper from #3680, and it's already tracked as issue #3713 (0 null-`mdt_id` enemies exist anywhere in the current seeded database, so that test's `markTestSkipped` path is what actually runs - CI's "1 skipped" reflects exactly this). No code change made here; leaving it to #3713.

Wotuu (commented):
Ok

PR #3718, opened 2026-07-28
#3705 Add feedback notification when (de-)favoriting a route from the header

### LINKED ISSUE #3705: Add feedback notification when (de-)favoriting a route from the header
(De-)favoriting a route via the header control on both the route view page and
the route edit page currently gives no feedback — the user can't tell whether
the action succeeded.

Add a notification (e.g. a toast) confirming the favorite/unfavorite action on
both pages.

Wotuu:
:robot: Closes #3705

## Summary
- The header favorite star (route view + edit pages, both rendered via `common.maps.controls.header` / `CommonMapsMap`) gave no feedback when clicked, so the user couldn't tell whether the favorite/unfavorite AJAX call succeeded.
- `_favorite()` in `map.js` now shows a success toast (reusing the existing `showSuccessNotification` used elsewhere, e.g. route/enemy reports) once the request resolves, with distinct added/removed copy.
- Errors are already handled by the global `defaultAjaxErrorFn` (`$.ajaxSetup`), so no separate error-path change was needed.
- New JS-consumed lang keys (`js.dungeonroute_favorite_added_success` / `_removed_success`) were added to `lang/en_US/js.php` — since these are baked into the release JS lang bundle, this change needs a full release, not a hotfix.

## Test plan
- [x] New Vitest unit test (`map.favorite.test.js`) covering both the add (POST) and remove (DELETE) paths and their respective toast text.
- [x] Full `npx vitest run` suite passes (302/302).
- [x] `composer run fix` / `composer run analyse` clean (no PHP logic changed, only a lang array).
- [x] Manually verified in headless Chrome on both the route **view** and **edit** pages: clicking the star toggles the favorite and shows the correct toast each time.

## Review notes
An independent fresh-context review found no blocking issues; approved. It flagged two pre-existing,
out-of-scope observations (not introduced by this change), left as follow-ups rather than fixed here:
- The `#favorite` hidden-input `change` handler (`map.js:455-456`) can never fire (hidden inputs don't
  match `:checked` / dispatch native `change`); the only live path is the `.favorite_star` click handler.
- The optimistic star-flip happens before the AJAX call resolves, with no request lock/debounce, so
  rapid double-toggling could race and/or leave the star visually out of sync on a failed request.

### DIFF
--- a/lang/en_US/js.php
+++ b/lang/en_US/js.php
@@ -278,6 +278,8 @@ return [
     'sidebar_enemy_skippable_info_label'                                   => 'A skippable enemy is an enemy that can be skipped without using invisibility potions or Rogue Shroud. These enemies may be excluded from your routes once you mark enemies as overpulled during a live session.',
     'user_report_enemy_success'                                            => 'Report sent successfully. Thank you for helping improve Keystone.guru!',
     'dungeonroute_report_enemy_success'                                    => 'Report sent successfully. Thank you for helping improve Keystone.guru!',
+    'dungeonroute_favorite_added_success'                                  => 'Route added to your favorites',
+    'dungeonroute_favorite_removed_success'                                => 'Route removed from your favorites',
     'tag_delete_success'                                                   => 'Tag removed successfully',
     'tag_create_success'                                                   => 'Tag created successfully',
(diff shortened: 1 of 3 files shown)


Wotuu (commented):
Ok

PR #3719, opened 2026-07-28
#3703 Suppress enemy tooltips while assigning a raid marker

### LINKED ISSUE #3703: Assigning a raid marker shows the enemy's tooltip
When assigning a raid marker to an enemy, the enemy's tooltip is shown during the
assignment. It shouldn't be — the tooltip should stay suppressed while a raid marker
is being assigned.

Note: this may already be fixed by #128.

This does not block the v15.7.1 production release.

Wotuu:
:robot: Closes #3703

## The bug

The raid marker circle menu is drawn on top of the enemy it belongs to, so any enemy tooltip shown
while it is open covers the items the user is aiming at. The clicked enemy's own tooltip was already
handled (`EnemyVisual` unbound its Leaflet tooltip when the menu opened), but a **neighbouring**
enemy's tooltip popped up as soon as the mouse passed over it on its way to a menu item.

So no, #128 did not already fix this — but it did introduce the mechanism this fix uses.

## The fix

`RaidMarkerSelectMapState` — the map state the circle menu already starts — now overrides
`disablesTooltips()`. `DungeonMap#setMapState()` toggles the `.map_enemy_tooltips_suppressed` class
off that, which suppresses *every* enemy tooltip on the map for as long as the menu is open.

That let the per-layer `unbindTooltip()`/`bindTooltip()` pair in `EnemyVisual` go — the exact
pattern #128 moved away from, since every Leaflet `unbindTooltip()` leaks a raw DOM `focus`
listener it never removes.

## Also fixed: the shortcut hint disappearing after an assignment

Reproducing the bug surfaced a second one in the same flow. `Enemy#bindTooltip()` caches the
rendered text and only rebinds when it changes, and the save that follows an assignment rebinds the
tooltip **while `RaidMarkerSelectMapState` is still active** — baking
`canOpenRaidMarkerMenu() === false` into the cached text. The result: after assigning a marker, that
enemy's tooltip permanently lost its *"Shift + right-click to assign a raid marker"* line.

The hint is now built from a new state-independent `supportsRaidMarkerMenu()` (edit mode, not an
admin enemy); `canOpenRaidMarkerMenu()` keeps the map-state check and still gates the context menu.

## Verification

Driven in real headless Chrome on a route in edit mode (Black Rook Hold), same enemy and same
neighbour on both sides.

| step | before | after |
| --- | --- | --- |
| hover the enemy | tooltip with hint | tooltip with hint |
| circle menu open, hover a neighbour | **neighbour tooltip covers the menu** | no tooltip |
| after assigning a marker, re-hover | **tooltip without the hint** | tooltip with hint |

Tests: `resources/assets/js/custom/mapstate/raidmarkerselectmapstate.tooltips.test.js` (new — pins
the override on this state and pins the bases as *not* overridden, so the suppression cannot be
pushed up the chain onto pull-building) and five cases added to `models/enemy.test.js` for the hint.
Full vitest suite green (308 tests); `composer run fix` / `analyse` clean.

## Review round (cold-context reviewer)

- **The circle menu could leave its map state active forever.** The radial lives inside the enemy's
  divIcon, so removing the enemy's layer takes the menu's DOM with it — and nothing cleaned up:
  the shown/hidden handler deliberately did nothing on hide, and `_cleanupCircleMenu()`'s fade-out
  path calls `.delay().queue()` on the selector result, a silent no-op once that DOM is gone. A
  floor switch hides every enemy on the old floor *and* deliberately keeps the map state, so this
  was easy to hit. Pre-existing (enemy popups and the raid marker shortcut stayed dead until
  reload), but this MR would have widened it to "no enemy tooltip anywhere". Fixed in both places
  and confirmed in headless Chrome — before: `RaidMarkerSelectMapState` still active after a floor
  switch, 0 tooltips visible; after: state `null`, tooltips back.
- Corrected a comment that blamed a save after assigning a marker — `assignRaidMarker()` never
  saves. The rebinds that actually baked the hint out are `EnemyVisual`'s own `bindTooltip()` call
  (removed here) and `MapObjectGroup#setLayerToMapObject()` on a floor switch during a map state.
- Split an `it.each` assertion that ignored its parameter into its own test.
- New `enemyvisuals/enemyvisual.test.js` covers the cleanup guard (verified red without the fix).

Reviewer notes explicitly checked and found clean: `circlemenu:open`/`circlemenu:close` have no
consumers; `buildVisual()`'s cleanup/re-init path is strictly better than before; `AdminEnemy`
overrides `bindTooltip()` wholesale so the admin page is unaffected; the `AdminEnemy` global is in
the same concatenated bundle so `instanceof` cannot ReferenceError.

Known and deliberate: the hint now also renders during pull-building, where shift+right-click does
nothing until the selection ends. It did before too (cached from the initial bind) — it is just
consistent now instead of accidental. Out of scope for #3703 as filed: non-enemy tooltips (map
icons, killzone index) can still overlap the menu, and touch users see a hint they cannot act on.

### DIFF
--- a/resources/assets/js/custom/enemyvisuals/enemyvisual.js
+++ b/resources/assets/js/custom/enemyvisuals/enemyvisual.js
@@ -285,22 +285,20 @@ class EnemyVisual extends Signalable {
                 refreshTooltips();
             },
             open: function () {
-                self.enemy.unbindTooltip();
                 self.signal('circlemenu:open');
 
+                // Tooltips would cover the menu that is drawn on top of the enemy - suppressed for as
+                // long as this state is active by RaidMarkerSelectMapState.disablesTooltips(), which
+                // covers every enemy on the map rather than just this one (#3703). Unbinding this
+                // enemy's Leaflet tooltip here instead is what #128 moved away from: every unbind
+                // leaks a raw DOM 'focus' listener Leaflet never removes.
                 self.map.setMapState(new RaidMarkerSelectMapState(self.map, self.enemy));
             },
             close: function () {
-                // Unassigned when opened
-                self.enemy.bindTooltip();
-
                 // Delete ourselves again
                 self._cleanupCircleMenu();
             },
             select: function (evt, index) {
-                // Unassigned when opened
-                self.enemy.bindTooltip();
-
                 // Assign the selected raid marker
                 self.enemy.assignRaidMarker($(index).data('name'));
 
--- a/resources/assets/js/custom/enemyvisuals/enemyvisual.js
+++ b/resources/assets/js/custom/enemyvisuals/enemyvisual.js
@@ -49,7 +49,12 @@ class EnemyVisual extends Signalable {
                 }
             } else {
                 // When an object is hidden, its layer is removed from the parent, effectively rendering its display nil.
-                // We don't need to do anything since if the visual is added again, we're going to re-create it anyways
+                // We don't need to do anything since if the visual is added again, we're going to re-create it anyways.
+                // The circle menu is the exception: it lives inside that same (now removed) DOM, so nobody can close
+                // it anymore, and RaidMarkerSelectMapState would stay active for the rest of the session - taking
+                // enemy popups, the raid marker shortcut and (since #3703) every enemy tooltip with it. A floor
+                // switch hides every enemy on the old floor, so this is not an edge case.
+                self._cleanupCircleMenu(false);
             }
         });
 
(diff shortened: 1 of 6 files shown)


Wotuu:
:robot: Headless-Chrome verification — the raid marker circle menu open on the same enemy, mouse hovering the same neighbouring enemy.

**Before** (master): the neighbour's tooltip pops up over the menu.

before

**After**: no tooltip, the menu stays fully readable.

after

Wotuu (commented):
Ok

Issue #3720, opened 2026-07-28
MappingVersion clone-on-create can pick the wrong previous version for dungeons on multiple game versions (follow-up to #3713)


Wotuu:
:robot: Surfaced while rewriting the fixture in `DungeonRouteEnemyRaidMarkerMappingVersionTest` for #3713.

`MappingVersion::boot()`'s `static::created(...)` clone-on-create hook (app/Models/Mapping/MappingVersion.php:549-618) picks the mapping version to clone from like this:

```php
$existingMappingVersions = $newMappingVersion->dungeon->mappingVersions()->get();
if ($existingMappingVersions->count() < 2) {
    return;
}
$previousMappingVersion = $existingMappingVersions[1];
```

`Dungeon::mappingVersions()` is `orderByDesc('mapping_versions.version')` with **no `game_version_id` filter**. `version` is assigned independently per `game_version_id` (see `MappingService::createNewBareMappingVersion()` / `createNewMappingVersionFromPreviousMapping()`: `version = currentMappingVersionForThatGameVersion->version + 1`), so for a dungeon that has mapping versions on more than one game version, the raw `version` numbers of different game versions interleave.

Example from the current seeded DB (retail = game_version_id 1, Legion Timewalking = game_version_id 7):

```
The Arcway:        gv7/v3, gv7/v2, gv1/v1
Pit of Saron:       gv1/v8 ... gv1/v1, gv3/v1
```

If a new mapping version is created for `game_version_id=1` on The Arcway (e.g. `version = 1 + 1`), the global `mappingVersions()` list ordered by `version` desc is `[gv1/v2 (new), gv7/v3, gv7/v2, gv1/v1]` — index `[1]` is `gv7/v3`, not the `gv1/v1` row the new version was actually derived from. The clone-on-create hook would then copy game_version_id 7's enemies/map icons/etc. into the new game_version_id 1 mapping version instead of the correct predecessor's.

**Not confirmed as an active production bug** — whether any real code path actually triggers this (`MappingService::createNewBareMappingVersion`, `createNewMappingVersionFromPreviousMapping`, `Copy` command, admin clone flows) depends on whether any of those are ever called for a dungeon that already has mapping versions on more than one `game_version_id`. Raising this as an invariant violation for someone with more context on the clone flows to judge severity, not asserting confirmed data corruption.

[...]

PR #3721, opened 2026-07-28
#3713 Fix silently-skipped null-mdt_id raid marker test

### LINKED ISSUE #3713: Null-mdt_id enemy coverage silently gone from DungeonRouteEnemyRaidMarkerMappingVersionTest
:robot: Surfaced by the cold review on PR #3711 (a fix for a different, adjacent flake in the same test file).

`getRetailDungeonWithNullMdtIdEnemy()` (added in #3680, closing #3679) can no longer find any qualifying data: there are **0** enemies with a null `mdt_id` anywhere in the current seeded database (81,665 enemy rows), and grepping the seeder JSON for `"mdt_id": null` across `database/seeders/dungeondata/` returns 0 hits — so CI's freshly-seeded database has none either.

As a result, `upgradeMappingVersion_givenNpcWithNullMdtIdStillExistsInNewMappingVersion_preservesRaidMarker` silently skips instead of running — both locally on `master` and in CI (the `php-tests (feature-other)` job reports `Skipped: 2`, one of which is this test). #3680's body cited "152 of the 1549 null-`mdt_id` enemies" at the time, which no longer matches current data at all — the null-safe `<=>` join behavior this test exists to protect (see its own comment: "enemies placed outside of an MDT import legitimately have a null `mdt_id`; the resolution join must be null-safe, not plain `=`") is currently not exercised anywhere.

Two things worth doing, roughly in order of value:
1. Make `markTestSkipped`-based helpers in this file **fail loudly instead of silently skip** when they can't find qualifying data — mirroring `ProvidesDungeon::getDungeonWithNonFacadeFloor()`'s `throw new \RuntimeException(...)` pattern. A helper that can no longer find its data represents a coverage regression; a skip makes that invisible inside an otherwise-green run.
2. Investigate why null-`mdt_id` enemies no longer exist in the seed data at all (a seeder/import change since #3680?), and either restore some null-`mdt_id` fixtures or rework the test to construct one directly rather than relying on it existing in shared seed data.

Not a regression from any specific PR — it's a slow drift in shared seed data undermining a test's coverage guarantee. Low urgency (no user-facing impact), but worth fixing before it's forgotten a second time.

Wotuu:
:robot: Closes #3713

## Summary
- Null-`mdt_id` enemies no longer exist anywhere in the seed data (Wotuu confirmed in #3713: it's now considered an invalid enemy state and every such enemy has since been backfilled a real `mdt_id`), so `getRetailDungeonWithNullMdtIdEnemy()` always fell through to `markTestSkipped()` - the null-safe (`` <=> ``) join in `DungeonRouteEnemyRaidMarkerRepository::updateEnemyIdsByMappingVersion()` that this test exists to protect was silently not exercised in CI.
- Rewrote the test to construct both the existing-mapping-version enemy and its new-mapping-version counterpart directly (mirroring `AssignMDTIDsTest::createEnemy()`) instead of searching shared seed data for one.
- Verified the rewritten test actually has teeth: temporarily reverted the `` <=> `` back to `` = `` in the repository and confirmed the test fails (`assertNotNull($fresh)` fails because the marker gets orphan-deleted), then reverted.
- Converted the file's other two `markTestSkipped()` calls to `throw new \RuntimeException(...)`, mirroring `ProvidesDungeon::getDungeonWithNonFacadeFloor()` - a helper that can no longer find its data now fails loudly instead of silently skipping (issue item 1).

## Why the test doesn't rely on `MappingVersion::boot()`'s clone-on-create
While building this I found that `createNewerMappingVersion()`'s underlying clone-on-create hook picks its "previous mapping version" to clone from by global `version` number with no `game_version_id` scoping. For dungeons with mapping versions on more than one game version (e.g. Timewalking dungeons like The Arcway), that can pick the wrong source, which made an earlier version of this fixture ~20-25% flaky. Filed as a separate issue since it's a pre-existing gap in shared test/clone infrastructure with its own blast radius, not something to fix inside a test-coverage MR: #3720. The rewritten test sidesteps it entirely by constructing both enemies explicitly rather than depending on what gets cloned.

## Test plan
- [x] `DungeonRouteEnemyRaidMarkerMappingVersionTest` run 25+ times in a row locally, all green (previously ~20-25% flaky with the clone-dependent version)
- [x] Negative control: reverting `` <=> `` to `` = `` makes the new test fail as expected
- [x] `composer run fix` clean (0 files changed)
- [x] `composer run analyse` clean (no errors)

### DIFF
--- a/tests/Feature/App/Service/DungeonRoute/DungeonRouteEnemyRaidMarkerMappingVersionTest.php
+++ b/tests/Feature/App/Service/DungeonRoute/DungeonRouteEnemyRaidMarkerMappingVersionTest.php
@@ -6,6 +6,7 @@ use App\Models\Dungeon;
 use App\Models\DungeonRoute\DungeonRoute;
 use App\Models\DungeonRoute\DungeonRouteEnemyRaidMarker;
 use App\Models\Enemy;
+use App\Models\Faction;
 use App\Models\Mapping\MappingVersion;
 use App\Models\RaidMarker;
 use App\Service\DungeonRoute\DungeonRouteServiceInterface;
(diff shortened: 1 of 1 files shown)


Wotuu (commented):
Ok

PR #3724, opened 2026-07-28
#3722 De-flake dungeon fixture selection once and for all

### LINKED ISSUE #3722: De-flake dungeon fixture selection in tests once and for all (follow-up to #3679, #3710)
:robot: Filed by Claude.

## Background

MRs keep hitting intermittent pipeline failures caused by tests that pick a random seeded dungeon
which then turns out not to satisfy what the test actually needs. Each time the fix has been to
patch `tests/Feature/Traits/ProvidesDungeon.php` or to hand-roll yet another picker in the test
file itself. Two such fixes landed in one week: #3679 / #3680 and #3710 / #3711.

Measured against the local seeded DB (155 dungeons; CI seeds the same JSON via `db:seed`):

| Failure mode | Rate | Real? |
| --- | --- | --- |
| The 20-draw retry loop exhausts | 0.02-0.05% per call | Minor |
| Helper returns a valid dungeon, test then finds no enemy in it | ~4% per call | This is what broke #3680 / #3711 |
| Helper returns a dungeon whose only qualifying floor is inactive | 0% today, purely data-luck | Latent |

So the retry loop is **not** the main bug. The dominant class is filtering that happens *after*
the helper returns, which is why hardening the four existing loops in isolation would not stop the
churn.

Three further findings:

1. `getDungeonWithNonFacadeFloor()` and `getDungeonWithFacadeFloor()` never check `floors.active`,
   while production consistently uses `floorsForMapFacade(...)->active()`
   (`ThumbnailService.php:289`, `RenderThumbnail::resolveFloors`,
   `DungeonRouteThumbnailRepository::hasFreshThumbnailForVariant`). Three call sites
   (`QueueThumbnailTest:26`, `ThumbnailServiceTest:520`, `ThumbnailServiceTest:673`) pass only
   because no seeded dungeon currently has an inactive-only floor set.
2. Three parallel implementations already exist, and two already converged on the correct idiom
   (scan a shuffled candidate set instead of sampling): `GeneratesDungeonRoutes.php:22` and
   `DungeonRouteEnemyRaidMarkerMappingVersionTest.php:157`. `ProvidesDungeon` is the last holdout.
3. `getCurrentMappingVersion()` depends on `Auth::user()` and is memoised per Dungeon instance
   (`app/Models/Dungeon.php:145,225-278`). When code under test re-fetches the dungeon
   (`DungeonRouteSaveService::persist():157` does `Dungeon::findOrFail()`) it can resolve a
   different mapping version than the helper validated.

## Remaining work

**1. Rewrite `ProvidesDungeon` around one engine** - a single `findDungeon()` with all-optional
named parameters (`facadeEnabled`, `minActiveFloors`, `maxActiveFloors`, `challengeMode`,
`speedrunEnabled`, `dungeonActive`, `requireDefaultFloor`, `minEnemies`, `minEnemyPacks`,
`gameVersion`, `shuffle`, `constraint`, `resolve`) returning
`array{0: Dungeon, 1: MappingVersion, 2: mixed}`.

- Exhaustive, not sampled: SQL-prefilter, then `->with('floors')->disableCache()->get()`
  (shuffled unless `shuffle: false`) and iterate. If a qualifying dungeon exists it *will* be found.
- `disableCache()` matters: `Dungeon` extends `CacheModel` and `MODEL_CACHE_ENABLED=true` in CI but
  `false` locally. `inRandomOrder()` builds a constant cache key, and
  `DungeonRouteSaveServiceSaveTest:913` constrains on `whereDoesntHave('dungeonRoutes', demo)`
  which changes as tests run.
- `resolve` is the key addition: a PHP-level predicate evaluated *inside* the scan, whose non-null
  return becomes element [2]. That is what lets "and it must have an enemy" stop being a
  post-return filter.
- Throws with a diagnostic naming where the pool emptied, rather than skipping.
- The four existing helpers stay as thin presets, so none of the ~37 existing call sites change.

**2. Converge every other picker onto it** - `GeneratesDungeonRoutes`,
`DungeonRouteEnemyRaidMarkerMappingVersionTest` (3 helpers), `DungeonFixtures` (kept deterministic
on purpose), `DungeonRouteEditSidebarTest` (its retry wrapper is dead code), plus the four fully
unguarded `Dungeon::inRandomOrder()->first()` pickers in `AjaxAdminCombatLogRouteControllerTest`,
`CombatLogRouteEnemyFailureTest`, `AdminToolsCombatLogControllerTest` and
`AjaxDungeonRouteControllerTest`.

**3. Lock it with `tests/Feature/Traits/ProvidesDungeonTest.php`** - asserts every preset returns a
dungeon that actually satisfies its guarantees, proves the scan skips bad candidates
deterministically (pin the pool to three known enemy-less dungeons plus one good one), and asserts
the no-match path throws a diagnostic.

## Risks

- `DungeonFixtures` is currently deterministic (`orderByDesc('id')->first()`), so its 10 call sites
  have never run against any other dungeon. It is deliberately kept deterministic here; randomising
  it belongs in a separate change so any fallout is attributable.
- Selection stays random-among-valid, so a call site that breaks on 1 dungeon out of 24 will not be
  caught by a handful of runs. An exhaustive pool pass over the heaviest consumers is needed during
  implementation.

## Definition of done

- `composer run fix` and `composer run analyse` clean (PHPStan level 6 covers `tests/`).
- Affected test files looped ~20x, plus an exhaustive pool pass over `DungeonRouteSaveServiceSaveTest`
  and `ThumbnailServiceTest`.
- Full suite green (`--exclude-group=MapTiles`), all five CI shards green.

Wotuu:
:robot: Closes #3722

## Why

MRs kept hitting intermittent pipeline failures from tests that pick a random seeded dungeon which then turns out not to satisfy what the test needs. The fix each time was to patch `ProvidesDungeon` or hand-roll yet another picker — two such fixes landed in one week (#3680, #3711).

Measuring first showed the 20-draw retry loop was **not** the main cause:

| Failure mode | Rate |
| --- | --- |
| 20-draw retry loop exhausts | 0.02–0.05% per call |
| Helper returns a valid dungeon, test then finds no enemy in it | ~4% per call — this is what broke #3680/#3711 |
| Helper returns a dungeon whose only qualifying floor is inactive | 0% today, purely data-luck |

The dominant class is filtering that happens **after** the helper returns, so hardening the four existing loops in isolation would not have stopped the churn.

## What changed

**One engine.** `ProvidesDungeon::findDungeon()` takes all-optional **named** arguments (`facadeEnabled`, `minActiveFloors`, `maxActiveFloors`, `challengeMode`, `speedrunEnabled`, `dungeonActive`, `requireDefaultFloor`, `minEnemies`, `minEnemyPacks`, `gameVersion`, `shuffle`, `constraint`, `resolve`) and returns `array{Dungeon, MappingVersion, mixed}`.

- **Exhaustive, not sampled** — SQL-prefilter, then scan the whole shuffled candidate set. It cannot fail to find a dungeon that exists. This is the idiom `GeneratesDungeonRoutes` and the raid-marker test had already converged on separately.
- **`resolve`** is the piece that stops the churn: a predicate evaluated *inside* the scan whose non-null return becomes element `[2]`. "…and it must have an enemy" is now part of choosing the dungeon, not a filter applied to one already chosen.
- **Returns the validated `MappingVersion`** — `getCurrentMappingVersion()` resolves against `Auth::user()` and memoises per instance, so re-resolving on a re-fetched `Dungeon` can yield a different row.
- **Throws with a diagnostic** naming where the pool emptied, e.g. `No seeded dungeon satisfies [challengeMode: true, minActiveFloors: 1, resolve]. 71 dungeon(s) matched the SQL criteria, all rejected during the scan (resolve: 71).`
- **`disableCache()`** — `Dungeon` is a `CacheModel` and model caching is on in CI, off locally. `inRandomOrder()` builds a constant cache key, and constraints like `whereDoesntHave('dungeonRoutes')` depend on rows other tests create.

The four existing helpers remain as thin presets, so **none of the ~37 existing call sites changed**.

**Two silent gaps closed.** `getDungeonWithNonFacadeFloor()` and `getDungeonWithFacadeFloor()` never checked `floors.active`, while production always counts via `floorsForMapFacade(...)->active()`. Floor counting now goes through that exact call.

**Everything converged.** `GeneratesDungeonRoutes`, the raid-marker helpers, `DungeonFixtures` (deleted; became a preset), `DungeonRouteEditSidebarTest`'s dead retry wrapper, and the four fully unguarded `Dungeon::inRandomOrder()->first()` pickers.

**New `ProvidesDungeonTest`** locks the guarantees — including a test that pins the pool to several dungeons that *cannot* satisfy a requirement plus exactly one that can, and asserts the suitable one comes back on every shuffle order.

## Verification

- New `ProvidesDungeonTest`: 32 tests. 360 tests green across every affected suite, after rebasing onto current master.
- **Exhaustive pool pass, not sampling.** A temporary hook pinned `candidate[i % poolSize]` first and the four heaviest consumers were run for `i = 0..101`. Because the modulo is per-criteria, that drives **every member of every pool** through the suite — the 24-dungeon `getRetailDungeon` pool, 49 `exactlyOne`, 53 `multiple`, 50 `facade`, 88 non-speedrun, 13 speedrun-with-difficulties, 11 no-demo-routes, 102 `nonFacade`. **0 failures / 102 positions.** The hook was removed and the suite re-run clean.
- Raid-marker tests run 8× consecutively, green each time.
- `composer run analyse` clean; `php-cs-fixer` touched only files in this diff.
## Second commit: `RepairBrokenThumbnails` scoping

While verifying, `RepairBrokenThumbnailsTest` turned out to take **~78 minutes** on a developer database (I initially misread it as a hang — it isn't, it's just that slow).

It runs the real maintenance command **unscoped**, and the command walks every `dungeon_route_thumbnails` row. Because the test's `setUp()` fakes every configured disk, every one of the 226,170 rows in the local dev database looks broken — so each test re-queued every route in the database and deleted real fileless rows. That is a test mutating data it never created. CI only escaped it because its database is freshly seeded and nearly empty.

The command now takes `--dungeon-route-id=*` and the test passes its own route. This is worth having independently — repairing a single reported route no longer means scanning the whole table — and there's a new test proving the scoping actually excludes routes it wasn't given.

**~78 minutes → 1.07 seconds.**

## Third commit: structured logging no longer leaks into PHPUnit output

Every service's structured log lines were being written to a console stream during CI runs — 82 in the `feature-other` shard alone — interleaved with the progress output and carrying ANSI colour codes.

The channel is a **static** on `StructuredLogging`, decided in the constructor from `runningInConsole() && !runningUnitTests() && app.type === 'local'`. `runningUnitTests()` needs the application to have bound its environment; a logger constructed before that happens answers "no" and latches `stderr` on for the whole process, every service included. Whether that happens depends on when the first logger is built — which is why it shows in CI but never locally.

It now resolves per log call, so the question is asked once the application can answer it. An explicit `setChannel()` still wins, and a locally-run artisan command still logs to `stderr` (verified via tinker). Two unit tests lock both halves.

*Caveat worth stating: this one could not be reproduced locally, so CI is the verification.*

## Two latent defects found and fixed on the way

**1. The raid-marker helpers assumed the wrong clone source.** They call `createNewerMappingVersion()` and then look up the cloned counterpart of an enemy on the *current* mapping version — but `MappingVersion`'s created hook clones from `$dungeon->mappingVersions()[1]` (`MappingVersion.php:560`), i.e. the dungeon's **highest-version** mapping version regardless of game version. Where those differ the clone comes from a different mapping entirely and the lookup finds nothing. #3721 sidestepped this for the null-mdt_id test by constructing both sides directly; the other two tests still rely on clone-on-create, so the helper now requires the two to coincide — as part of choosing the dungeon, so it cannot be discovered after the fact.

**2. `maxActiveFloors: 2` on `hasFreshThumbnailForVariant_givenOneFreshAndOneStaleFloorThumbnail_returnsFalse`.** That test seeds two thumbnail rows, but `hasFreshThumbnailForVariant()` first compares row count against the dungeon's active floor count — so on a dungeon with 3+ floors it returned false on the short-circuit and the staleness check it exists to test never ran.

> **Note on the null-mdt_id test.** I independently found that `upgradeMappingVersion_givenNpcWithNullMdtIdStillExistsInNewMappingVersion_preservesRaidMarker` was silently skipping on every run (#3629 backfilled every `mdt_id` out of the seed data), and had fixed it here by nulling a seeded enemy and restoring it. #3721 landed on master first with a better fix — it constructs a synthetic enemy and never touches seeded data. This branch is rebased onto that and **takes master's version wholesale**; only the two remaining search helpers are converged onto `findDungeon()`.

## Review

A fresh-context agent with no shared context reviewed the diff and raised six findings; all are addressed:

- **Blocker:** the `finally` restoring a mutated `mdt_id` sat behind three cascade-heavy deletes, so a throw in any of them would have left seeded data corrupted. That code is gone entirely — this branch now takes master's #3721 approach, which never mutates seeded data in the first place.
- **Real regression, now fixed:** `gameVersion:` was a preference rather than a filter. The deleted `DungeonFixtures` guaranteed in SQL that the dungeon had a mapping version *for that game version* carrying enemies; `getCurrentMappingVersion()` falls back to the acting user's version, then the default, then simply the highest. `ClassCompendiumControllerTest` depends on the old guarantee. Now enforced in SQL **and** verified against the resolved mapping version in the scan, with `gameVersion` reported in the diagnostic.
- Three near-vacuous assertions rewritten to be pool-constrained so the requirement under test is the only discriminator — `requireDefaultFloor` would have passed ~99% of the time with the filter removed (154/155 dungeons qualify), `dungeonActive` ~86%. Added the missing `speedrunEnabled`, `challengeMode: false` and `gameVersion` cases.
- Dropped a dead `->with('floors')` eager-load (nothing in the scan reads it, and the returned model is re-fetched).
- Softened a `disableCache()` comment that promised more than the code delivers.
- Documented that the presets deliberately drop the validated `MappingVersion`, and that `resolve()` receives the pre-reload instance.

## Not changed on purpose

- **`dungeonActive` defaults to "don't care".** Requiring `dungeons.active = 1` would drop 20 of the 102 dungeons the non-facade helper can return today; that is a behaviour change, so it is opt-in (`AjaxDungeonRouteControllerTest` opts in, where it matters).
- **`DungeonFixtures`' successor stays deterministic.** Its 10 call sites have only ever run against one dungeon; randomising them could surface latent per-dungeon assumptions as failures indistinguishable from a regression here. Worth doing — as its own change.


app/Console/Commands/DungeonRoute/RepairBrokenThumbnails.php
```diff
@@ -94,9 +117,11 @@ private function requeueThumbnailsMissingFromDisk(ThumbnailServiceInterface $thu
...
 
         // The model's default $with would also hydrate the unused floor relation for every row.
-        $query = DungeonRouteThumbnail::query()
-            ->whereHas('file')
-            ->without('floor');
+        $query = $this->scopeToRequestedDungeonRoutes(
+            DungeonRouteThumbnail::query()
+                ->whereHas('file')
+                ->without('floor'),
+        );
```
Wotuu: Use the repository class for this. Create a new function that optionally allows the passage of a dungeon route id. Cleaner than this.

Wotuu (commented):
One thing.

app/Console/Commands/DungeonRoute/RepairBrokenThumbnails.php
Wotuu: :robot: Fixed - added `filelessThumbnailsQuery()` and `fileBackedThumbnailsQuery()` to `DungeonRouteThumbnailRepositoryInterface`/`DungeonRouteThumbnailRepository`, each taking an optional array of dungeon route IDs to scope to. The command now injects the repository and calls these instead of building the query inline, and the private `scopeToRequestedDungeonRoutes()` helper is gone. Added repository-level tests for both new methods (scoped and unscoped) alongside the existing coverage.

tests/Feature/Traits/ProvidesDungeon.php
```diff
@@ -126,25 +290,60 @@ protected function getDungeonWithMultipleNonFacadeFloors(?Closure $constraint =
...
-            }
+    /**
+     * Returns a Mythic+ dungeon whose current mapping version actually carries enemies.
+     *
+     * Deliberately deterministic: this returns the same dungeon on every run, as the fixture it
+     * replaced did. Its callers have therefore only ever been exercised against one dungeon, and
+     * randomising them here would surface latent per-dungeon assumptions as failures that look
+     * indistinguishable from a regression. That is worth doing - but as its own change.
+     */
+    protected function getDungeonWithCurrentMappingVersionWithEnemies(?GameVersion $gameVersion = null): Dungeon
```
Wotuu, line 304: :robot: The `gameVersion` **default** silently dropped the Retail guarantee that the deleted fixture had.

`DungeonFixtures::getDungeonWithCurrentMappingVersionWithEnemies()` defaulted to Retail:

```php
public static function getDungeonWithCurrentMappingVersionWithEnemies(
    int $gameVersionId = GameVersion::ALL[GameVersion::DEFAULT_GAME_VERSION],
) {
    return Dungeon::…->whereHas('mappingVersions', static fn($q) => $q->where('game_version_id', $gameVersionId)->whereHas('enemies'))…
```

so *every* call site — including the ones that pass nothing — was guaranteed a dungeon with a Retail mapping version carrying enemies. Here the default is `null`, which `findDungeon()` reads as "no game version requirement at all". The nine argument-less call sites (`EnemyRepositoryTest` ×4, `NpcRepositoryTest` ×5) therefore lost that guarantee; only `ClassCompendiumControllerTest`, which passes the game version explicitly, still has it.

The PR body lists `gameVersion:` being a preference rather than a filter as a real regression that was found and fixed — this is the other half of that same guarantee, and "Not changed on purpose" only defends the determinism of this helper, not the loss of the game-version scoping. So it reads as an oversight rather than a decision.

No behaviour change today: I ran the old fixture query and the new scan side by side against a seeded database and both return dungeon 149 (`algeth_ar_academy_midnight`, `game_version_id = 1`). What changed is the guarantee, not the current result — a future higher-`id` Mythic+ dungeon whose *current* mapping version is Classic/Midnight would silently redirect these Retail-oriented repository tests onto a non-Retail mapping, which is exactly the "the draw succeeded, the test then failed on something the draw never guaranteed" shape this PR exists to eliminate.

Restoring it does not undo the improvement in the other direction (the new check is strictly stronger — it requires the *current* mapping version to carry the enemies, where the fixture only required *some* Retail mapping version to). It cannot be a parameter default since it is not a constant expression, but inside the body it is a one-liner:

```php
$gameVersion ??= GameVersion::getDefaultGameVersion();
```

Wotuu:

Wotuu (commented):
Ok

tests/Feature/Traits/ProvidesDungeon.php
Wotuu: :robot: Fixed - `getDungeonWithCurrentMappingVersionWithEnemies()` now does `$gameVersion ??= GameVersion::getDefaultGameVersion();` before calling `findDungeon()`, restoring the Retail guarantee. Verified: EnemyRepositoryTest, NpcRepositoryTest, ClassCompendiumControllerTest, and ProvidesDungeonTest all pass (48 tests), `composer run fix`/`analyse` clean.

PR #3729, opened 2026-07-28
#3723 Keep the raid marker circle menu from discarding or stranding a map state

### LINKED ISSUE #3723: Raid marker circle menu: reopen can discard another map state, and an async open can strand one
:robot: Follow-up to #3703 / #3719. A cold review of that MR's later commits raised three findings that
landed after the MR was already merged, so they never got addressed. None of them is a regression from
#3719 — two are narrow pre-existing holes it got close to, one is a test gap.

All line references are against `resources/assets/js/custom/enemyvisuals/enemyvisual.js` on master
(`decc9f047`).

### 1. `buildVisual()` reopens the circle menu — and its map state — over a state it did not start

`_cleanupCircleMenu()` now only ends `RaidMarkerSelectMapState` if that is still the active state, so it
cannot cancel a selection the user started while the menu was open. The reopen path does not honour the
same invariant:

```js
if (hadCircleMenu && reopenCircleMenu) {
    this._initCircleMenu(false);   // -> open callback -> setMapState(new RaidMarkerSelectMapState(...))
}
```

Repro: open the raid marker menu on an enemy, then start a map state whose
`shouldRebuildEnemyVisuals()` is `true` (`EditMapState`, `DeleteMapState`, `MDTEnemySelection`,
`EnemyPatrolEnemySelection`, `EnemyForcesCheckpointEnemySelection`). The `map:mapstatechanged` handler
rebuilds the visual, the cleanup correctly leaves that state alone, and the re-init then replaces it with
a `RaidMarkerSelectMapState` anyway — the state the user just started is silently discarded.

Fix: guard the reopen on `this.map.getMapState() === null || instanceof RaidMarkerSelectMapState`.

### 2. Hiding the enemy inside the same tick the menu opens strands the map state

`zikes-circlemenu` fires its `open` callback from a `setTimeout`, so
`setMapState(new RaidMarkerSelectMapState(...))` always lands one macrotask after `_initCircleMenu()`
returns. If the enemy is hidden inside that window (an Echo update, a faction/teeming toggle, a queued
floor switch), the hidden-branch cleanup runs while the state is still `null`: it nulls `_circleMenu`,
and because the icon is already out of the DOM the `$(this).remove()` matches nothing, so the plugin's
jQuery handlers on the detached `<ul>` survive. The pending timer then starts a `RaidMarkerSelectMapState`
with no menu and no `_circleMenu` handle — `_cleanupCircleMenu()` early-returns forever and
`canOpenRaidMarkerMenu()` (which needs `getMapState() === null`) blocks opening another one. Stranded for
the session: enemy popups, the raid marker shortcut and every enemy tooltip stay dead until reload.

Suggested fix: clean up off `self._circleMenu` (which *is* the radial's jQuery set and stays valid on a
detached node, so jQuery's `cleanData` strips the pending handler) instead of re-querying by id, and use
an explicit `document.contains(...)` check for the fade-out branch — `.length` is always 1 there, so it
cannot decide it. That also fixes the smaller leak below.

### 3. Bootstrap Tooltip instances leak when the menu's DOM is already gone

The disposal loop in `cleanupFn` re-queries with `$(this).find('[data-bs-toggle="tooltip"]')`, which
matches nothing once the radial has been removed from the DOM. Bootstrap keys instances by element in a
plain `Map`, not a `WeakMap` — exactly what the comment right above it guards against. Pre-existing, but
#3719 made that path routinely reachable (every enemy hide). Same fix as 2.

### 4. Test gap

The `hidden`-branch call to `_cleanupCircleMenu(false)` (the behavioural core of that commit) has no
coverage — mutating the line away leaves the whole JS suite green. The guards themselves *are* pinned by
`resources/assets/js/custom/enemyvisuals/enemyvisual.test.js`. Covering the wiring needs the registered
callback captured from a fake `enemy.register`, which means running `EnemyVisual`'s constructor with
`setVisualType()` stubbed — the prototype-call recipe used in that file deliberately avoids it.

Wotuu:
:robot: Closes #3723

Follow-up to #3703 / #3719: the three findings a cold review raised after that MR had already merged.

### 1. `buildVisual()` re-opened the menu over a map state it did not start

`_cleanupCircleMenu()` only ends `RaidMarkerSelectMapState` if that is still the active state, so a
selection the user started while the menu was open survives the cleanup. The re-open path did not
honour the same invariant: every map state whose `shouldRebuildEnemyVisuals()` is `true`
(`EditMapState`, `DeleteMapState`, `MDTEnemySelection`, `EnemyPatrolEnemySelection`,
`EnemyForcesCheckpointEnemySelection`) rebuilds the visual, and the re-init then replaced the state
the user had just started with a `RaidMarkerSelectMapState`.

Guarded with a new `_canReopenCircleMenu()` — `null` or a `RaidMarkerSelectMapState` only. Named to
keep it apart from `Enemy#canOpenRaidMarkerMenu()`, which gates opening a menu from scratch and
requires no map state at all.

### 2. Hiding the enemy inside the same tick the menu opens stranded the map state

`zikes-circlemenu` fires its `open` hook — which calls `setMapState(new RaidMarkerSelectMapState(...))`
— from a `setTimeout`, and registers that hook as a jQuery handler on the radial `<ul>` itself. If the
enemy was hidden inside that window (a floor switch, a faction/teeming toggle, an Echo update), the
cleanup re-queried `#map_enemy_raid_marker_radial_<id>` and matched nothing: Leaflet removes the icon
with a native `removeChild`, which never runs jQuery's `cleanData`, so the handler survived on the
detached node. The pending timer then started a map state for a menu that no longer existed, with
`_circleMenu` already nulled — `_cleanupCircleMenu()` early-returns forever after that, and
`canOpenRaidMarkerMenu()` blocks opening another one. Enemy popups, the raid marker shortcut and every
enemy tooltip stayed dead for the rest of the session.

Cleanup now runs off `self._circleMenu` — the plugin's own jQuery set, which stays valid on a detached
node, so `.remove()` strips the pending handler — and decides the fade-out with an explicit
`document.contains()` check, since that set's `length` is 1 either way.

### 3. Bootstrap Tooltip instances leaked once the menu's DOM was gone

The disposal loop re-queried the same way, so it matched nothing on a detached radial and leaked one
Tooltip instance per item (Bootstrap keys them by element in a plain `Map`, not a `WeakMap`) every
time an enemy was hidden with the menu open. Same fix.

### 4. Tests

- `enemyvisual.circlemenu.test.js` (new): real jQuery + the real `zikes-circlemenu` plugin against a
  real DOM, which is what findings 2 and 3 need — both are about the plugin's own jQuery state on a
  detached node, which a hand-rolled `$` stub cannot model. Verified red on master: the "hidden before
  the menu opens" test lands a `RaidMarkerSelectMapState` on a menu that is gone, and the tooltips are
  never disposed.
- `enemyvisual.test.js`: covers the re-open guard through `buildVisual()` itself (not just the helper),
  and closes the `hidden`-branch gap from the issue by running the constructor with `setVisualType()`
  stubbed. `_cleanupCircleMenu`'s existing tests now model the radial as an always-length-1 set on a
  real (attached or detached) DOM node, matching what the plugin actually hands back.
- Each guard was mutation-checked: removing it turns the matching test red.

Full JS suite: 326 passed.

No visual verification — this is interaction/state behaviour with no rendering change, and the menu
itself is unchanged. The scenarios are pinned by the tests above instead.

### Cold review

Reviewed by a fresh-context agent; it found no correctness defect in the fix, and independently
mutation-checked the tests (6 go red on a revert to master). Its findings, all addressed in the second
commit:

- The `document.contains()` guard's comment claimed a detached radial's `delay()/queue()` would never
  drain. That is **false** — verified empirically: jQuery's queue is keyed off the element's data, not
  the document tree. It was true on master, where `$radial` came from a selector and went *empty*, and
  the rewrite kept the old conclusion under a new mechanism. Reworded to the real reason (not sitting
  out 500ms with the plugin's handlers still on a radial nobody can see, where a pending `open` hook
  can start a map state in the meantime), along with the matching test name and comment. The guard
  itself stays.
- `cleanup()` faded the menu out, so a discarded visual would call `setMapState(null)` 500ms later —
  after it had already unregistered from `map:mapstatechanged`. Now cleans up synchronously; the
  redundant null check went with it.
- Documented why the synchronous cleanup in `buildVisual()` cannot re-enter it
  (`RaidMarkerSelectMapState#shouldRebuildEnemyVisuals()` is `false`).
- Test naming: the `shown`/`hidden` cases are constructor wiring, so they are named after the
  constructor now. Added the pull-building case (an `EnemySelection` state), which is how the re-open
  guard is most often reached in practice.

Re-opening the menu when a rebuild lands inside its 500ms *close* animation is a separate, pre-existing
hole — `_circleMenu` cannot express "closing" vs "open". Filed as #3730 rather than papered over here.

Full JS suite after the review round: 328 passed.

### DIFF
--- a/resources/assets/js/custom/enemyvisuals/enemyvisual.circlemenu.test.js
+++ b/resources/assets/js/custom/enemyvisuals/enemyvisual.circlemenu.test.js
@@ -1,0 +1,186 @@
+// Integration coverage for the raid marker circle menu's teardown (#3723).
+//
+// Unlike enemyvisual.test.js - which hands _cleanupCircleMenu() a hand-rolled jQuery stand-in to pin
+// its guards - these tests run the real jQuery and the real zikes-circlemenu plugin against a real
+// (jsdom) DOM, because both findings are about the plugin's own jQuery state on a *detached* node:
+//
+//  - The plugin registers its `open` hook as a jQuery handler on the radial <ul> itself and fires it
+//    from a setTimeout. Leaflet destroys the enemy's icon with a native removeChild, which does not
+//    run jQuery's cleanData, so that handler survives the enemy being hidden. Cleaning up by
+//    re-querying `#map_enemy_raid_marker_radial_<id>` matched nothing at that point, leaving the
+//    pending timer free to start a RaidMarkerSelectMapState for a menu that no longer exists - with
+//    _circleMenu already nulled, nothing could ever end that state again.
+//  - The same re-query made the Bootstrap Tooltip disposal loop a no-op, leaking one instance per
+//    tooltip item (Bootstrap keys them by element in a plain Map, not a WeakMap).
+//
+// Both are fixed by cleaning up off `self._circleMenu`, which *is* the plugin's own jQuery set and
+// stays valid on a detached node.
+
+// Real jQuery, replacing the minimal stub from test/setup.js. Must be global before the plugin is
+// required: it is an IIFE closing over the `jQuery` global.
+global.$ = global.jQuery = require('jquery');
+require('zikes-circlemenu');
+
+const fs = require('fs');
+const path = require('path');
+const HandlebarsRuntime = require('handlebars');
+
+global.Signalable = class Signalable {
+};
+
+// Stand-ins for the collaborators the map state constructors assert on. The fakes below are built as
+// instances of these so those assertions hold.
+global.DungeonMap = class DungeonMap {
+};
+global.MapObject = class MapObject {
+};
+
+// The real chain, so the map state the menu starts is the real one.
+const {MapState} = require('../mapstate/mapstate');
+global.MapState = MapState;
+const {MapObjectMapState} = require('../mapstate/mapobjectmapstate');
+global.MapObjectMapState = MapObjectMapState;
+const {RaidMarkerSelectMapState} = require('../mapstate/raidmarkerselectmapstate');
+global.RaidMarkerSelectMapState = RaidMarkerSelectMapState;
+
+const {EnemyVisual} = require('./enemyvisual');
+
+// The real radial markup, so the tooltip items the disposal loop looks for are the real ones.
+const radialTemplate = HandlebarsRuntime.compile(
+    fs.readFileSync(
+        path.join(__dirname, '../../handlebars/map_enemy_raid_marker_template.handlebars'),
+        'utf8'
+    )
+);
+
+global.Handlebars = {templates: {map_enemy_raid_marker_template: radialTemplate}};
+global.getHandlebarsDefaultVariables = () => ({});
+global.refreshTooltips = () => {
+};
+global.removeStrayTooltips = () => {
+};
+global.c = {map: {enemy: {calculateMargin: () => 4}}};
+
+/**
+ * Opens the circle menu on an enemy the way _openRaidMarkerMenu() does, without advancing any timers:
+ * the plugin fires its `open` hook - and with it setMapState() - from a setTimeout, so on return the
+ * menu is on screen but no map state has started yet.
+ */
+function makeOpenCircleMenu() {
+    const id = 42;
+
+    document.body.innerHTML = `<div id="map_enemy_visual_${id}"><div class="enemy_icon"></div></div>`;
+    const container = document.getElementById(`map_enemy_visual_${id}`);
+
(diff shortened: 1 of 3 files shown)


Wotuu (commented):
Ok

PR #3733, opened 2026-07-28
#3732 Enemy forces checkpoints follow the Pull number style setting

### LINKED ISSUE #3732: Enemy forces checkpoints follow the Enemy number style setting instead of the Pull number style setting
Enemy forces checkpoints display their value (percentage vs. raw enemy forces) based on the **Enemy number style** map setting, but they should follow the **Pull number style** setting instead.

Every other "how much enemy forces is this group worth" display - the pulls sidebar (`rowelementkillzone.js`), kill zone tooltips (`killzone.js`), enemy pack tooltips (`enemypack.js`) - reads `getKillZonesNumberStyle()` / listens to `killzonesnumberstyle:changed`. `getMapNumberStyle()` is for the per-enemy numbers drawn on the map (`enemyvisualmainenemyforces.js`). A checkpoint is a group total, so it belongs with the former.

### Affected call sites

- `resources/assets/js/custom/models/enemyforcescheckpoint.js`
  - pill markup (`_getPillHtml()`)
  - tooltip text (`bindTooltip()`)
  - `mapnumberstyle:changed` register/unregister
- `resources/assets/js/custom/mapstate/enemyselection/enemyforcescheckpointenemyselection.js`
  - admin edit snackbar (`_refreshSnackbar()`)
  - `mapnumberstyle:changed` register/unregister

### Expected

Toggling **Pull number style** updates the checkpoint pill, its tooltip and the admin edit snackbar; toggling **Enemy number style** leaves them alone.

### Note

The cookie defaults in `map.js` are `map_number_style: 'enemy_forces'` and `kill_zones_number_style: 'percentage'`, so this also changes the out-of-the-box checkpoint display from raw enemy forces to a percentage - which matches the pill's own `fa-percent` placement icon.

Introduced in #3660 (part of #128).

Wotuu:
:robot: Closes #3732

Enemy forces checkpoints displayed their value (percentage vs. raw enemy forces) based on the **Enemy number style** setting. They now follow **Pull number style** instead.

`getMapNumberStyle()` governs the numbers drawn on individual enemies (`enemyvisualmainenemyforces.js`). Every "how much enemy forces is this group worth" display - the pulls sidebar, kill zone tooltips, enemy pack tooltips - reads `getKillZonesNumberStyle()`. A checkpoint is a group total, so it belongs with the latter.

### Changes

| File | What |
| --- | --- |
| `models/enemyforcescheckpoint.js` | pill markup, tooltip text, and the `killzonesnumberstyle:changed` register/unregister pair |
| `mapstate/enemyselection/enemyforcescheckpointenemyselection.js` | admin edit snackbar + its register/unregister pair |
| `lang/en_US/view_common.php` | the Pull number style help text said "pulls sidebar" only |

The register/unregister pairs move together with the getters, so toggling **Pull number style** still repaints the pill, tooltip and snackbar live.

### Behaviour change worth a conscious ack

Both settings default to percentage in `statemanager.js`, but `map.js` seeds the cookies on every map page load as `map_number_style: 'enemy_forces'` / `kill_zones_number_style: 'percentage'`. So **any user who never touched the Pull toggle sees checkpoint pills flip from raw enemy forces to a percentage** - that's the default experience, and checkpoints already shipped (`8defc3f55` is in v15.8.0 / v15.8.1).

Percentage seems right for "how much you need before entering" (it matches the pill's own `fa-percent` placement icon), but it is a visible change for existing users rather than a silent one. Two notes:
- Cookie-less headless renders (thumbnails) are unaffected - both `statemanager.js` fallbacks are percentage, before and after.
- In a `pulls => false` embed the settings modal isn't rendered, so a viewer there has no toggle and is stuck with the default.

### Tests

- `enemyforcescheckpoint.test.js` - 4 new tests over `_getPillHtml()` and `bindTooltip()`
- `enemyforcescheckpointenemyselection.numberstyle.test.js` (new) - 2 tests over `_refreshSnackbar()`

Each test sets the two number style settings to **opposing** values, so it fails if the wrong one is consulted - verified by mutation (revert the fix: all 6 fail; re-apply: all pass). Full vitest suite: 322 passed. `composer run fix` / `analyse` clean.

Verified in real headless Chrome, both the cookie path and the live-repaint path - see the comments below.

### Also included: two skill doc updates

The last commit is docs only (`.claude/skills/headless-browser-verify`, `.claude/skills/worktree-docker`). Verifying this fix ran into three undocumented traps - a worktree has no `node_modules`, the main checkout usually can't serve a master baseline for a JS change, and `map.js`'s secure cookies mean the page can't write its own settings over plain HTTP (so state-manager setters silently no-op, and every map page logs one benign `JSON.parse` console error). Per `.claude/CLAUDE.md` that knowledge belongs in-repo, and it's adjacent enough to this MR to not be worth its own.

### DIFF
--- a/.claude/skills/headless-browser-verify/SKILL.md
+++ b/.claude/skills/headless-browser-verify/SKILL.md
@@ -20,6 +20,20 @@ mkdir -p .chrome-tmp && cp .claude/skills/headless-browser-verify/browse.js .chr
 
 No published ports, so every worktree stack can run its own chrome simultaneously.
 
+**In a worktree, `node_modules` must be a real directory, not a symlink.** `sh/worktree.sh create`
+does not provide one at all, and browse.js resolves `puppeteer` from `/var/www/node_modules` *inside*
+the container - where a host symlink into the main checkout does not resolve (only the worktree is
+mounted). A symlink is fine for `npx vitest` on the host but fails here with
+`Cannot find module 'puppeteer'`. Hardlink a copy instead - seconds, and it shares the main
+checkout's blocks:
+
+```sh
+cp -al /home/wouterkoppenol/Git/private/keystone.guru/node_modules ./node_modules
+rm -rf node_modules/.cache   # the only entries cp -al cannot link (root-owned); harmless
+```
+
+Delete it again before teardown.
+
 ## Running
 
 ```sh
@@ -59,6 +73,23 @@ both** to the PR, not just the branch shot.
   for master and `http://nginx/<page>` for the branch, same viewport.
 - Check the baseline is actually master: the page footer shows the built commit
   (`v15.3.2 (05d7dc)`); if the main checkout's assets are stale, rebuild there first.
+- **For a JS/CSS change, prefer an in-worktree baseline over the main stack.** The main checkout
+  often has no built asset for its own HEAD (`public/js/app-<sha>.js` missing), so its pages load
+  without JS and cannot serve a baseline at all — and building there to fix that touches the user's
+  checkout. Instead flip the changed files in place:
+
+  ```sh
+  git checkout HEAD~1 -- <changed files>   # or the merge-base commit
+  npm run development                       # host-side; ~10s incremental
+  # ...run the driver, save the "before" shots...
+  git checkout HEAD -- <changed files>
+  npm run development
+  ```
+
+  The asset filename is derived from **git HEAD**, which never moves during this, so nginx serves
+  the same URL throughout and only the bundle contents change. Confirm the restore with
+  `grep -c <new symbol> public/js/custom-<sha>.js` before moving on — a forgotten restore leaves the
+  worktree silently reverted.
 - **Exercise dynamic UI, not just page load.** JS-inserted content (map sidebar, dropdown menus)
   breaks differently from server-rendered HTML, and a regression can hit only the
   dynamically-inserted elements. Use `--click` and screenshot the opened/expanded state on both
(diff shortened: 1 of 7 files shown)


Wotuu:
:robot: **Visual verification** — real headless Chrome, `/explore/wotlk/pit-of-saron` (the Cavern checkpoint on mapping version 797), same page, same viewport, master build vs. branch build.

Both cookies are set explicitly on every run, always to **opposing** values, so the pill's text says which setting it consulted.

### Pull number style = Percentage, Enemy number style = Enemy forces

| Before (master) | After (this branch) |
| --- | --- |
| before | after |
| `504 EF required` — followed **Enemy** number style | `78.4% required` — follows **Pull** number style |

### Pull number style = Enemy forces, Enemy number style = Percentage

| Before (master) | After (this branch) |
| --- | --- |
| before | after |
| `78.4% required` — followed **Enemy** number style | `504 EF required` — follows **Pull** number style |

A clean inversion in both directions: master mirrors whatever **Enemy number style** is set to, the branch mirrors **Pull number style**. No page errors on either build.

Wotuu:
:robot: **Live-repaint verification** — the screenshots above prove the *getters* changed; this covers the other half of the diff, the signal registration, with no page reload.

On `/explore/wotlk/pit-of-saron`, both number-style getters are stubbed to `enemy_forces` (so whichever signal the pill listens to will flip it), then each signal is fired in turn:

| Step | Pill text |
| --- | --- |
| initial | `78.4% required` |
| after `mapnumberstyle:changed` (what the **Enemy number style** toggle emits) | `78.4% required` — unchanged ✅ |
| after `killzonesnumberstyle:changed` (what the **Pull number style** toggle emits) | `504 EF required` — repainted ✅ |

So the checkpoint is listening on the new signal and no stale listener remains on the old one.

Note for anyone repeating this: driving it through `setKillZonesNumberStyle()` instead does **not** work over plain HTTP. `Cookies.set` there carries `secure: true` outside a local environment, so the write is silently dropped and the getter keeps returning the old value — the pill then looks like it failed to repaint when in fact the setting never changed. Hence stubbing the getter and firing the signal directly, which is exactly what the setter does after writing.

The admin edit snackbar (`EnemyForcesCheckpointEnemySelection`) is covered by unit tests rather than the browser: `admin/floor/mapping.blade.php` passes `pulls => false`, so neither toggle is rendered on the only page that state exists on. That was equally true of the old `mapnumberstyle:changed` registration, so it is not a regression — but it does mean the snackbar's setting is not reachable from that page's UI either way.

Wotuu (commented):
Ok

PR #3736, opened 2026-07-29
#3730 Fix raid marker circle menu re-opening during its close animation

### LINKED ISSUE #3730: Raid marker circle menu re-opens if the visual is rebuilt during its close animation
:robot: Spotted during the cold review of #3729 (which fixes #3723). Pre-existing on master, and not a
regression from either MR — but it is the one hole left in the area #3729 covers, so it belongs with
that set.

`EnemyVisual#_cleanupCircleMenu()` fades the raid marker radial out over 500ms when called from the
plugin's `close` or `select` callbacks. `this._circleMenu` stays non-null for that whole window — it is
only nulled inside `cleanupFn`, once the queue runs.

`buildVisual()` decides whether to put the menu back from exactly that handle:

```js
let hadCircleMenu = this._cleanupCircleMenu(false);
// ...
if (hadCircleMenu && reopenCircleMenu && this._canReopenCircleMenu()) {
    this._initCircleMenu(false);
}
```

So a rebuild landing inside the fade-out sees `hadCircleMenu === true` for a menu the user just
dismissed. `_canReopenCircleMenu()` (added in #3729) does not help: the map state is still
`RaidMarkerSelectMapState` at that point, since ending it is also part of the queued cleanup. The menu
pops back open and re-starts the state.

Repro: open the raid marker menu on an enemy, click the enemy again to close it, and within 500ms
trigger a rebuild — `killzone:attached`, `overpulled:changed`, or an Echo `object:changed` from a
collaborator in a live session.

Not reachable from `select` in practice: `assignRaidMarker()` goes through AJAX, so the
`enemy:set_raid_marker` → `buildVisual(false)` that follows lands well after the fade and passes
`reopenCircleMenu = false` anyway.

The fix needs the visual to distinguish "menu open" from "menu closing" — `_circleMenu` alone cannot
express it. Nulling it eagerly when the fade starts is not enough on its own: `_openRaidMarkerMenu()`
and `_mouseOut()` both treat `_circleMenu === null` as "no menu", so that would let a second radial be
appended on top of the one still animating out.

Wotuu:
:robot: Closes #3730

## Summary
`EnemyVisual#_cleanupCircleMenu()` only nulls `_circleMenu` once its queued 500ms fade-out actually
runs (started from the plugin's `close`/`select` callbacks). During that window `_circleMenu` is
still non-null, so a rebuild landing inside it - `killzone:attached`, `overpulled:changed`, an Echo
`object:changed` - has `buildVisual()` call `_cleanupCircleMenu(false)` to tear down before
rebuilding. That call saw the menu the user just dismissed as one that got interrupted mid-rebuild
(`hadCircleMenu === true`), and `_canReopenCircleMenu()` didn't help since the map state hadn't ended
yet either - so the menu popped back open and restarted `RaidMarkerSelectMapState`.

Fix: a new `_circleMenuClosing` flag, set for the duration of the queued fade-out. While it's set,
`_cleanupCircleMenu()` is a no-op and reports "no menu was cleaned up" - the already-queued cleanup
is left to finish on its own once its delay elapses. `_circleMenu` itself stays non-null throughout
(not nulled eagerly), so the existing `_circleMenu === null` guards in `_openRaidMarkerMenu()` and
`_mouseOut()` are untouched.

## Test plan
- [x] New regression test in `enemyvisual.circlemenu.test.js` reproducing the exact race: open the
  menu, close it (starts the fade), then call `_cleanupCircleMenu(false)` as `buildVisual()` would -
  asserts it's no longer reported as reopenable, and that the original fade-out still completes
  normally afterwards. Confirmed this test fails without the fix (reverted the source change locally
  and re-ran it).
- [x] Second regression test covering the rebuild also detaching the still-fading radial from the
  document (the enemy's icon HTML gets replaced mid-fade) - the queued cleanup must still end the map
  state once its delay elapses rather than being silently skipped.
- [x] Full JS suite green (`npx vitest run`, 330 tests).
- [x] Independent fresh-context review of the diff: no issues found.

## Visual verification
No static visual delta exists between master and this branch - the change is purely to internal
event-timing/state-machine logic (a flag guarding a callback re-entrancy window), not to markup,
styling, or layout, so a before/after screenshot pair would be pixel-identical and wouldn't show
anything meaningful. Loaded an existing route in headless Chrome as a smoke test instead (page loads
cleanly, map and enemies render normally, no new console/page errors) to confirm no regression from
the change. The actual race-condition fix - which requires precise sub-500ms timing against the real
raid-marker plugin - is exercised directly against real jQuery and the real `zikes-circlemenu` plugin
in `enemyvisual.circlemenu.test.js`, which is a stronger check than a manual UI reproduction would be.

### DIFF
--- a/resources/assets/js/custom/enemyvisuals/enemyvisual.js
+++ b/resources/assets/js/custom/enemyvisuals/enemyvisual.js
@@ -30,6 +30,10 @@ class EnemyVisual extends Signalable {
         this.cachedRadius = 0;
 
         this._circleMenu = null;
+        // True for the entire 500ms fade-out queued by _cleanupCircleMenu(true) - the window in which
+        // _circleMenu is still non-null (nothing else marks a menu that is on its way out rather than
+        // open) but must not be treated as reopenable by buildVisual() (#3730).
+        this._circleMenuClosing = false;
 
         // Can be set to force the building of a visual when it's shown again
         this._forceBuildVisualOnShow = false;
@@ -325,6 +329,11 @@ class EnemyVisual extends Signalable {
 
     /**
      * Cleans up the circle menu, removing it from the object completely.
+     *
+     * A no-op while the menu is already mid fade-out (a previous call already queued cleanupFn below):
+     * without this, buildVisual() calling in here to tear down for a rebuild would see a menu the user
+     * is actively closing as one that just got interrupted, and put it right back afterwards (#3730).
+     * The already-queued cleanupFn is left alone to finish the teardown once its delay elapses.
      * @param fadeOut {boolean}
      * @return true if the menu was cleaned up, false if it was not
      * @private
@@ -334,7 +343,7 @@ class EnemyVisual extends Signalable {
 
         let self = this;
 
-        let shouldCleanUp = self._circleMenu !== null;
+        let shouldCleanUp = self._circleMenu !== null && !self._circleMenuClosing;
 
         if (shouldCleanUp) {
             // Remove any stray tooltip bubbles left over from hovering the circle menu items.
@@ -371,6 +380,7 @@ class EnemyVisual extends Signalable {
                 // nulled here, nothing could ever end that state again (#3723).
                 $radial.remove().dequeue();
                 self._circleMenu = null;
+                self._circleMenuClosing = false;
 
                 // Only stop the map state at this point - and only if it is still ours. The menu can
                 // only be opened while no map state is active, but the user can start one while it is
@@ -396,6 +406,7 @@ class EnemyVisual extends Signalable {
             // length check used to stand in for), and jQuery's queue drains on a detached element
             // just fine - it is keyed off the element's data, not off the document tree.
             if (fadeOut && $radial.length > 0 && document.contains($radial[0])) {
+                self._circleMenuClosing = true;
                 $radial.delay(500).queue(cleanupFn);
             } else {
                 cleanupFn();
--- a/resources/assets/js/custom/enemyvisuals/enemyvisual.circlemenu.test.js
+++ b/resources/assets/js/custom/enemyvisuals/enemyvisual.circlemenu.test.js
@@ -183,4 +183,33 @@ describe('EnemyVisual circle menu teardown (#3723)', () => {
         expect(self._circleMenu).toBeNull();
         expect(setMapStateCalls[1]).toBeNull();
     });
+
+    test('_cleanupCircleMenu_givenARebuildLandsWhileTheMenuIsFadingOut_doesNotReportAReopenableMenu', () => {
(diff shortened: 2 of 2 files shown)


Wotuu (commented):
Ok

resources/assets/js/custom/enemyvisuals/enemyvisual.js
```diff
@@ -334,7 +343,7 @@ class EnemyVisual extends Signalable {
 
         let self = this;
 
-        let shouldCleanUp = self._circleMenu !== null;
+        let shouldCleanUp = self._circleMenu !== null && !self._circleMenuClosing;
```
Wotuu, line 346: :robot: **The no-op guards a broader axis than the bug needs — the other three `fadeOut = false` call sites silently became deferred.**

Only `buildVisual()` needs "don't report this as reopenable". The other callers pass `fadeOut = false` precisely because they need the teardown to happen *now*, and while `_circleMenuClosing` is set they instead leave the already-queued 500ms `cleanupFn` pending:

- `cleanup()` (~L829) — its own comment states exactly this must not happen: *"a queued cleanup would end up calling `setMapState(null)` half a second later on behalf of an object that no longer exists"*. That is now reachable: `MapObjectGroup#clear()` (floor switch, group reload) → `Enemy#cleanup()` → `EnemyVisual#cleanup()` landing inside the fade window leaves the queued teardown to run against a discarded visual, after `Enemy#cleanup()` has already nulled `this.visual`.
- the `shown`/`hidden` handler (~L61) — same deferral, minus the discard.

I traced what the deferred `setMapState(null)` actually runs and it is inert: `MapState#stop()` only drops a `beforeunload` listener, and `setMapState()` toggles `map_enemy_tooltips_suppressed` plus signals `map:mapstatechanged` (a no-op for enemy visuals — `shouldRebuildEnemyVisuals()` is false on both sides). Nothing dereferences the discarded visual, so no crash. The observable effect is that enemy tooltips stay suppressed map-wide for up to 500ms after a floor switch / enemy delete that lands inside the fade window, then self-heal once the queue drains. **So this is an invariant note, not a live bug — I would not block the MR on it.**

Flagging it because the PR body only notes that the `_openRaidMarkerMenu()` / `_mouseOut()` guards are untouched, not that these two call sites changed behaviour, so it reads as unintended rather than deliberate. If you would rather keep them synchronous, the axis the fix needs is the *return value*, not the teardown — `_circleMenu !== null && !_circleMenuClosing` for what is reported back, while still tearing down whenever `_circleMenu !== null` (cutting the fade short in `buildVisual()` costs nothing visually, since `layer.setIcon()` detaches the radial moments later anyway).

Wotuu:

Issue #3740, opened 2026-07-29
Flip MDT export to the !~MDT2~ format at Midnight Season 2 / 12.1 (follow-up to #3735)


Wotuu:
:robot:

## Background

#3735 added full support for MDT 6.2's `!~MDT2~` CBOR export-string format:

- **Import** auto-detects and accepts both the legacy `!` format and `!~MDT2~`.
- **Export** still emits the legacy format, because MDT 6.2 continues to import it and flipping early would break copy-to-MDT for every user still on MDT < 6.2 (i.e. everyone until WoW 12.1 ships 2026-08-11).
- The MDT2 **encoder is already built and fully tested** (`MDT2Codec::encode()`), reachable via `mdt:encode --mdt2` and `MDTBaseService::encode($contents, useMDT2Format: true)`.

Upstream MDT removes legacy import support entirely with WoW 12.2 (Nnoggie/MythicDungeonTools#737, "after the compatibility window promised in the 6.2.0 changelog"). Once that lands, Keystone.guru's exported strings would stop importing into MDT unless the default flips.

## Remaining work

1. Flip the export default: in `app/Service/MDT/MDTExportStringService.php::getEncodedString()` (the `$this->encode($mdtObject)` call, currently ~line 514) pass `useMDT2Format: true` — or flip the default parameter value on `MDTBaseService::encode()` and drop the call-site argument.
2. Update export/round-trip tests to expect the `!~MDT2~` prefix (they currently expect the legacy `!` format), including `MDTExportStringServiceExtractArrowsTest` / `MDTExportStringServiceExtractObjectsTest` and the round-trip helpers in `MDTExportStringServiceTestBase`.
3. Re-verify the admin tools "view as string" round-trip (`AdminToolsMdtController::mdtviewasstringsubmit`, exercises export → import in one request).
4. Consider making `--mdt2` the `mdt:encode` default with a `--legacy` escape hatch instead.
5. Optional cleanup while there: the `report($ex)` gate in `MDTImportController::import()` uses `isValidBase64($string) || MDT2Codec::appliesTo($string)` — note that *legacy* `!`-prefixed strings fail `isValidBase64` too (the leading `!` breaks the strict round-trip check), so legacy import exceptions have never been reported. Worth deciding whether to widen that gate properly.

## Timing

[...]

PR #3739, opened 2026-07-29
Midnight Season 2 import dry run (MDT 6.2.0-alpha3 / ptr12.1)

Wotuu:
:robot:

> ## ⚠️ Read before merging
> Per @Wotuu this may merge before 12.1 ships — the S2 dungeons can be disabled to keep them
> invisible until then. Two things still deserve a decision at merge time:
> - ~~Composer is pinned to a third-party personal fork~~ — **done.** Both packages are served from
>   the `Wotuu/` forks now: `Wotuu/MythicDungeonTools@ff4ef5fc` and `Wotuu/MDT_Legacy@5cfa26f1`, the
>   same commits as before, verified byte-identical by reinstalling from scratch. The constraint is
>   still `^6.0@alpha`, which stays until there is a stable 6.2.0 tag to point at.
> - The seeded season starts **2026-08-18** with 8 placeholder affix groups.
>
> The full checklist is at the bottom; everything on it is still owed before the real release.

Closes #3734

Dry run of the whole "new season" pass against MDT PTR data, so release day is a re-run rather than
a debugging session. Everything found broken here is fixed here.

**All eight season 2 dungeons import with enemy forces matching MDT exactly, and the full suite is
green (1451 passed; the pre-change baseline was 1371).**

| dungeon | mv | MDT EF | KG EF | enemies | packs | on facade |
|---|---|---|---|---|---|---|
| murder_row | 812 | 690 | **690** | 221 | 56 | 0 |
| den_of_nalorakk | 813 | 729 | **729** | 116 | 0 ¹ | 0 |
| the_blinding_vale | 814 | 655 | **655** | 276 | 0 ¹ | 0 |
| voidscar_arena | 815 | 738 | **738** | 218 | 63 | 0 |
| altar_of_fangs | 816 | 836 | **836** | 160 | 46 | 0 |
| kingsrest | 829 | 654 | **654** | 101 | 29 | 0 |
| templeofsethraliss | 826 | 661 | **661** | 128 | 43 | 0 |
| rubylifepools | 822 | 531 | **531** | 166 | 34 | 0 |

The five expected totals from the issue (836 / 729 / 655 / 738 / 690) all match. Enemy forces are
compared against `MDTDungeon::getDungeonTotalCount()` rather than hardcoded numbers, and every pack
count equals MDT's own distinct group-id count for that dungeon.

¹ Den of Nalorakk and The Blinding Vale have **no group assignments in the PTR data at all** (zero
[...]

Issue #3734, opened 2026-07-28
Midnight Season 2 import dry run (MDT 6.2.0-alpha3 / ptr12.1)


Wotuu:
:robot:

## Context

Midnight Season 2 (patch 12.1) ships in ~3 weeks — patch on **2026-08-11**, season start
**2026-08-18**. MDT has S2 mapping on the unmerged `ptr12.1` branch
(`Nnoggie/MythicDungeonTools` PR #731, tagged `6.2.0-alpha3`, head `ff4ef5fc`).

Goal: do the whole "new season" pass now against PTR data — new Season row, MDT bump, mapping
import for the S2 pool — so that on release day it's a re-run against a final MDT tag rather than a
debugging session. Everything found broken here gets fixed here.

Reconnaissance is already done; it turned up three problems the naive runbook would have hit cold.

### S2 dungeon pool, resolved against MDT `ptr12.1`

| KG dungeon key | MDT lua (in `Midnight/`) | `dungeonIndex` → `mdt_id` | `mapID` → `challenge_mode_id` | today in `dungeons.json` |
|---|---|---|---|---|
| `murder_row` | `MurderRow.lua` | 160 | 587 | mdt_id ✓, **challenge_mode_id null** |
| `den_of_nalorakk` | `DenOfNalorakk.lua` | 161 | 586 | **both unset** |
| `the_blinding_vale` | `TheBlindingVale.lua` | 162 | 584 | **both unset** |
| `voidscar_arena` | `VoidscarArena.lua` | 163 | 585 | **both unset** |
| `altar_of_fangs` | `AltarOfFangs.lua` | 164 | 588 | both ✓, `mdt_supported: false` |
| `kings_rest` | `KingsRest.lua` | 17 | 249 | existing BFA row |
| `temple_of_sethraliss` | `TempleOfSethraliss.lua` | 20 | 250 | existing BFA row |
| `ruby_life_pools` | `RubyLifePools.lua` | 42 | 399 | existing DF row |

`dungeonIndex` is hardcoded per lua file, so the drastic reorder in `Midnight/load_midnight.xml`
does **not** churn existing `mdt_id`s. All five Midnight dungeons already have floors, tiles,
facades and floor unions; they have **zero enemies/packs** — that's what the import produces.

### Three problems found up front

1. **`composer.json` points at `Wotuu/MythicDungeonTools`, which has no `ptr12.1` branch.**
[...]

PR #3741, opened 2026-07-29
#3735 Support the MDT 6.2 '!~MDT2~' CBOR export-string format on import

### LINKED ISSUE #3735: Support the new MDT 6.2 '!~MDT2~' CBOR export-string format on import
:robot:

Split out of #3734 (Midnight S2 import dry run), where it was found during MDT `ptr12.1` recon.

## Problem

MDT 6.2 changes the export-string format. Strings produced by MDT 6.2+ will **hard-fail** in
`MDTImportStringService`, so on the day 12.1 ships every user pasting a fresh MDT string into
Keystone.guru gets an import error. This is the highest-impact day-one break of the 12.1 cycle, and
it is independent of the S2 mapping work.

Old `!`-prefixed strings still import fine, both in MDT and here — so this is purely about accepting
the *new* format. Export (KG → MDT) is unaffected: `MDTExportStringService` keeps emitting the
legacy `!` format, which MDT 6.2 still reads.

## The new format

`Modules/Transmission.lua` @ `Nnoggie/MythicDungeonTools` `ff4ef5fc` (branch `ptr12.1`, tagged
`6.2.0-alpha3`, upstream PR Nnoggie/MythicDungeonTools#731):

```lua
-- "~" cannot start either legacy LibDeflate encoding, so this marker cannot collide with existing exports.
local encodingPrefix = "!~MDT2~"

function MDT:TableToString(inTable)
  local serialized = C_EncodingUtil.SerializeCBOR(inTable)
  local compressed = C_EncodingUtil.CompressString(serialized, Enum.CompressionMethod.Deflate)
  local encoded    = C_EncodingUtil.EncodeBase64(compressed)
  return encodingPrefix..encoded
end
```

So the chain is:

```
!~MDT2~  +  Base64( Deflate( CBOR(table) ) )
```

Three things changed at once versus the legacy path:

| | legacy (`!…`) | new (`!~MDT2~…`) |
|---|---|---|
| serialization | AceSerializer-3.0 (custom text format) | **CBOR** (RFC 8949, binary) |
| compression | LibDeflate raw deflate | Blizzard `C_EncodingUtil.CompressString(..., Enum.CompressionMethod.Deflate)` |
| text encoding | LibDeflate `EncodeForPrint` (custom printable alphabet) | **standard Base64** |

The Base64 change is a simplification for us — `base64_decode()` replaces the custom-alphabet
decoder. The CBOR change is the substantial part.

## Why the current code fails

Decoding is done in Lua, not PHP: `app/Logic/MDT/Lua/MDTTransmission.lua` is our vendored copy of
MDT's transmission logic, driven through the PHP Lua bridge from
`MDTImportStringService::decode()`. Its `StringToTable` starts with:

```lua
local encoded, usesDeflate = inString:gsub("^%!", "")
```

That strips the leading `!` and leaves `~MDT2~<base64>`, which then goes into
`LibDeflate:DecodeForPrint`/`DecompressDeflate` and fails — surfacing as
`Unable to decode MDT import string` (`MDTImportStringService.php:111`) /
`services.mdt.io.import_string.unable_to_decode_mdt_import_string` (`:216`).

Note this Lua copy **cannot** simply mirror upstream: `C_EncodingUtil` is a Blizzard client API that
does not exist outside WoW. The decode for the new format has to happen PHP-side (base64 → inflate →
CBOR → array), with the resulting array handed across to the existing Lua `ValidateImportPreset` and
the rest of the pipeline unchanged.

## Open questions to settle before implementing

1. **Raw deflate or zlib-wrapped?** `Enum.CompressionMethod.Deflate` is not documented as to whether
   it emits a bare deflate stream (`gzinflate`) or a zlib container (`gzuncompress`). Settle this
   empirically against a **real 12.1 PTR export string** — trying both and seeing which one produces
   valid CBOR is a two-line check, but it needs a genuine sample string. Getting one is the first
   task on this issue.
2. **Which PHP CBOR package?** We have no CBOR decoder today, so this needs a new composer
   dependency, which per CLAUDE.md needs @Wotuu's approval. Candidates to evaluate: `spomky-labs/cbor-php`
   (well-maintained, PSR-ish, decode-only usage is small) vs `2tvenom/cborencode` (tiny, less active).
   A hand-rolled decoder is also viable — we only need decode, and MDT presets use a narrow subset of
   CBOR (maps, arrays, ints, floats, strings, bools, null) — but that is a maintenance liability
   against a spec with 8 major types and indefinite-length variants.
3. **Does MDT emit CBOR maps with integer keys?** Lua tables serialize to CBOR maps whose keys may be
   integers, and the existing pipeline downstream of `decode()` assumes PHP arrays with the same shape
   AceSerializer produced. A round-trip comparison of the same preset exported in both formats will
   confirm whether any shape adaptation is needed.

## Suggested scope

- Detect the `!~MDT2~` prefix in PHP before touching the Lua bridge; keep the legacy path byte-for-byte
  unchanged so nothing regresses for old strings.
- Decode base64 → inflate → CBOR → PHP array, then feed that array into the existing flow at the same
  point `decode()` returns today.
- Tests: add a real 12.1 PTR string as a fixture alongside the existing legacy fixtures in the
  `mdt-import` test suite, and keep at least one legacy-format test to prove no regression.
- A clear, translated error if a `!~MDT2~` string arrives on a build without the decoder, rather than
  the generic "unable to decode".

## Out of scope

- Export in the new format. MDT 6.2 still imports legacy `!` strings, so
  `MDTExportStringService` stays as-is until MDT drops legacy import support.
- The S2 mapping import itself — that is #3734.

Wotuu:
:robot:

Closes #3735

MDT 6.2 (ships with WoW 12.1 on 2026-08-11) switches export strings from the legacy `!` +
AceSerializer/LibDeflate format to `!~MDT2~` + Base64(RawDeflate(CBOR(preset))). Without this MR,
every string exported by MDT 6.2 fails import with "Unable to decode MDT import string" on day one
of 12.1.

**The preset _content_ is unchanged** — only the packaging (serialization/compression/text encoding)
differs, now confirmed against real client output (see below).

## What this does

- **`MDT2Codec`** (`app/Logic/MDT/IO/`): PHP-native decode **and** encode of the new format. The
  legacy path (`cli_weakauras_parser` Rust binary) is untouched — MDT2 strings never reach it.
  Decode produces the exact array shape the legacy path produces (CBOR arrays → 0-based PHP lists,
  CBOR int-keyed maps → 1-based keys, mirroring `json_decode` of the binary's JSON), so everything
  downstream of `MDTBaseService::decode()` — including the `isset($d[0])` dense/sparse discriminator
  in `ObjectImporter` — is format-agnostic and unchanged.
- **Import auto-detects** the format in `MDTBaseService::decode()`; `mdt:decode` handles both for
  CLI parity.
- **Export still emits the legacy format** — MDT 6.2 still imports it, and flipping now would break
  every user on MDT < 6.2. The encoder is fully built + tested and reachable via
  `mdt:encode --mdt2` / `MDTBaseService::encode($contents, useMDT2Format: true)`; the default flips
  in the 12.2 cycle when MDT removes legacy import support (upstream
  Nnoggie/MythicDungeonTools#737) — follow-up issue: #3740.
- **Telemetry**: failed `!~MDT2~` decodes are error-logged with the string (mirroring the legacy
  `shouldErrorLog` behavior), and `MDTImportController`'s `report()` gate now recognizes MDT2 strings
  (they fail `isValidBase64`, so real import errors were previously silently swallowed).
- **Drive-by**: the hardcoded English decode/validate exception messages in `MDTImportStringService`
  now use the existing translated keys.
- New dep (approved): `spomky-labs/cbor-php ^3.3` (+ `ext-zlib`, already present in all images).
  `brick/math` was already locked on master.

## Verified against real 12.1 PTR strings

@Wotuu supplied two genuine `!~MDT2~` strings straight out of in-game MDT (King's Rest, Ruby Life
Pools). They are committed as fixtures and drive
`MDTImportStringServiceMDT2AuthoritativeTest`, which settles all three open questions from the
issue:

1. **Raw deflate, not zlib.** `Enum.CompressionMethod.Deflate` emits a bare deflate stream — the
   test asserts `gzinflate()` succeeds *and* `gzuncompress()` fails on the real payload.
2. **`MDT2Codec::encode()` reproduces a real client string byte for byte** from its own decode
   output. That single assertion proves every type choice at once: Lua strings really are CBOR byte
   strings (major type 2) and not text strings, the float/int widths match, and map key order is
   preserved. It also **found a bug**: the encoder emitted an empty Lua table as an empty CBOR map
   (`0xa0`) on a documented-but-unverified assumption — Blizzard emits an empty CBOR *array*
   (`0x80`). Import was never affected (both decoders round-trip either form back to an empty
   table), but the export flip (#3740) rests on matching the client exactly, so it is fixed here.
3. **No shape adaptation is needed.** Re-encoding the decoded real preset back into the legacy
   format and running *both* through the full pipeline yields identical `getDetails()` output and
   identical pulls/enemies — so the import warnings a real string produces come from the mapping,
   not from the format.

Two **addon-side** changes surfaced that are not format changes (a legacy-format string exported by
MDT 6.2 would lack them too), both of which already fall back gracefully and are pinned by tests:

- MDT 6.2 explicitly nils `preset.week` (`Modules/Presets.lua`) → `parseAffixes()` falls back to the
  season's current affix group.
- `addonVersion` is `tonumber(version:gsub("%.", ""))`, which is **nil on an alpha build** like
  `6.2.0-alpha3` — a released `6.2.0` will yield `620`. With no `addonVersion`,
  `MappingService::getMappingVersionForMdtAddonVersion()` returns the dungeon's current mapping
  version, which is the right answer for a brand-new string anyway. (An unknown `620` with no
  `mdt_addon_versions` row takes the same fallback, so release day is covered either way.)

## Notable implementation constraints

- cbor-php's `normalize()` returns integers as base-10 **strings** — the codec walks the CBOR object
  tree manually to preserve the int/string distinction.
- Blizzard's `DeserializeCBOR` rejects tags + indefinite-length items — the encoder mirrors that, and
  the decoder rejects anything the client cannot produce.
- Decode is hardened against malicious payloads on the public endpoint: an iterative,
  allocation-free pre-scan of the raw CBOR enforces **128 nesting levels / 262,144 items / 1 MiB
  decompressed** and rejects trailing bytes, before cbor-php's *recursive* decoder ever runs
  (unbounded nesting segfaults it from a ~300-byte request — found by the pre-review agent,
  PoC-verified blocked). Out-of-int-range CBOR integers throw instead of silently saturating to
  `PHP_INT_MAX`.

## Tests

- `MDT2CodecTest` (unit, 27 tests): byte-level encode assertions (list/map/dense/sparse/mixed-key/
  empty/float/negint), decode shape guarantees (0-based lists, preserved sparse keys, text-string
  leniency, duplicate-key last-wins), rejection paths (tags, indefinite lengths, corrupt base64/
  deflate/CBOR, non-map root, >uint32 encode), full `assertSame` round trip.
- `MDTImportStringServiceMDT2AuthoritativeTest` (feature): the two **real client** strings above —
  decode, raw-deflate assertions, byte-for-byte encoder round trip, absent `week`/`addonVersion` +
  mapping-version fallback, full `getDungeonRoute` import, and MDT2-vs-legacy pipeline equivalence.
  Assertions are equivalence-only, never absolute pull/enemy-forces counts, so mapping changes cannot
  make them flaky.
- `MDTImportStringServiceMDT2Test` (feature): a self-generated fixture with the same preset content
  as the legacy v507 fixture decodes identically through the service; runtime route → legacy export →
  MDT2 re-encode → full `getDetails` + `getDungeonRoute` equivalence; corrupt string → null;
  `mdt:decode` CLI parity. That fixture pair's equivalence stops at decode level because its dungeon
  (Mists of Tirna Scithe) is no longer MDT-supported — the real-string test above covers the full
  pipeline on committed fixtures.

Two fresh-context review agents have run over this branch. The first, over the original
implementation, found two ship-blockers (the CBOR nesting segfault and the inflate-cap OOM) plus
lower findings — all fixed and covered by tests. The second, over the rebase + real-fixture delta,
found that both MDT2 feature tests leaked the `DungeonRoute` rows their imports persist
(`getDungeonRoute()` creates them regardless of the `save` flag) — fixed, and verified at zero net
rows across a run — and that the byte-string claim was only incidentally proven, which is what
prompted the byte-for-byte assertion above.

No visual verification: there is no visual change. The only user-facing deltas are on the import
page's error path — the `report()` gate now recognises MDT2 strings, and two hardcoded English
exception strings are routed through existing translation keys.

## Not stacked any more

This MR was originally stacked on #3739 (`3734-midnight-s2-dry-run`) for the MDT 6.2.0-alpha3
composer bump. It has been **rebased straight onto `master`**, so the two are now fully independent
and can merge in either order. Nothing here needed the bump — the codec is pure PHP — and unstacking
also strengthens the MR: the real-string fixtures import against the mappings that are live today
rather than against unmerged Season 2 data.

Whichever of the two merges second will hit one `composer.lock` conflict, since both touch it: the
resolution is **keep both** — the MDT 6.2.0-alpha3 bump from #3739 and `spomky-labs/cbor-php` from
this one. Nothing else overlaps.

## Out of scope

- Export in the new format. MDT 6.2 still imports legacy `!` strings, so `MDTExportStringService`
  stays as-is until MDT drops legacy import support (#3740).
- The Midnight S2 mapping import itself — that is #3734 / #3739.


app/Logic/MDT/IO/MDT2Codec.php
```diff
@@ -0,0 +1,384 @@
...
+            is_bool($value)  => $value ? TrueObject::create() : FalseObject::create(),
+            $value === null  => NullObject::create(),
+            default          => throw new MDT2EncodeException(sprintf('Unsupported value type: %s', get_debug_type($value))),
+        };
+    }
+
+    /**
+     * @throws MDT2EncodeException
+     */
+    private static function intToCborObject(int $value): CBORObject
```
Wotuu, line 368: :robot: **The encoder emits non-minimal CBOR for four exact integer values, so the byte-for-byte fidelity claim quietly does not hold there.**

`UnsignedIntegerObject::create()` / `NegativeIntegerObject::create()` pick the header width with `isLessThan(BigInteger::fromBase('FF', 16))` (and `'FFFF'`, `'FFFFFFFF'`) where it should be `isLessThanOrEqualTo` — an off-by-one in cbor-php that this encoder inherits. Measured against this branch in its worktree container (bytes are the encoded int, map/key headers stripped):

```
23         => (inline)
24         => 18          canonical
254        => fe          (0x18 fe) canonical
255        => 00ff        (0x19 00ff) — should be 0x18 ff
256        => 0100        canonical
65534      => fffe        canonical
65535      => 0000ffff    (0x1a 0000ffff) — should be 0x19 ffff
65536      => 00010000    canonical
4294967294 => fffffffe    canonical
4294967295 => THROWS MDT2EncodeException: Out of range. Please use PositiveBigIntegerTag...
-255       => fe          canonical
-256       => 00ff        — should be 0x38 ff
```

So `255`, `65535`, `-256` and `-65536` encode one width too wide, and `4294967295` / `-4294967296` throw outright rather than encoding.

Bounding the severity, because this is not a ship blocker: **decode is unaffected** — cbor-php reads `19 00ff` back as 255 and `assertSafeCborPayload` accepts the wider header — and the encoder is not on the production export path yet. What it undermines is precisely the guarantee #3740 will rest on. Note also that `intToCborObject` is used for **map keys** as well (`is_int($key) ? self::intToCborObject($key) : ByteStringObject::create($key)`), so enemy-index keys are exposed alongside values, and `currentDungeonIdx` is a monotonically growing MDT index that will pass through exactly 255 on some future patch.

The current tests cannot see it: `encode_givenIntBeyondUint32_throwsMDT2EncodeException` picks `0x1_0000_0000`, one value past the boundary that actually breaks, and I walked the CBOR of both real fixtures — 145 and 141 integers, max 161244, **zero** occurrences of 255 / 65535 / -256 / -65536 — which is why the byte-for-byte round trip passes today.

Suggested fix: bypass `create()` for integers and emit the minimal header directly via `UnsignedIntegerObject::createObjectForValue($additionalInformation, $data)` (and the `NegativeIntegerObject` equivalent — both are public), plus a boundary-value test at 255 / 65535 / -256 / -65536 and 4294967295.

app/Logic/MDT/IO/MDT2Codec.php
```diff
@@ -0,0 +1,384 @@
...
+     * levels; more importantly, cbor-php's decoder recurses per level and would exhaust the C
+     * stack (segfault, uncatchable) on a deeply nested crafted payload without this pre-check.
+     */
+    private const MAX_DEPTH = 128;
+
+    /**
+     * Cap on the total CBOR item count - cbor-php materializes an object per item, so an
+     * unbounded flat payload would exhaust PHP's memory_limit from a KB-scale request.
+     */
+    private const MAX_ITEMS = 262144;
```
Wotuu, line 64: :robot: **Not a blocker — the hardening holds up under measurement. Posting the numbers because `MAX_ITEMS` is the one cap not pinned to an observed value.**

Everything I could throw at `decode()` in this branch's worktree container behaved:

- **Deflate bomb**: 400 MiB of zeros (407 KB compressed, 543 KB request body) is rejected in **3 ms at 8 MiB peak** — `gzinflate($binary, MAX_DECOMPRESSED_BYTES)` returns `false` rather than truncating, so the cap is a real early-out and not a post-hoc length check.
- **Nesting**: 127 nested maps decode cleanly, 200 throws on the pre-scan — the recursive cbor-php decoder never sees an unbounded depth.
- **Item explosion**: the worst payload I could construct at the cap — an **895-byte** body, a root array of 87,380 single-pair byte-string maps, 262,140 items — decodes at **120 MiB peak in 0.4 s**. That is well under the 512 MiB per-request fpm cap (`docker-compose/app/usr/local/etc/php-fpm.d/zz.memory.conf`), so no single request can OOM a worker.

For scale on the legitimate side: the largest of the 100,718 rows in `mdt_imports` is a 99,417-char legacy string; re-encoded through `MDT2Codec` it is 143,560 bytes of CBOR, **22,540 items, depth 4**, and decodes with a ~2 MiB delta. So the allowed worst case sits roughly 60x above the largest real input ever observed, and `MAX_DECOMPRESSED_BYTES` about 7x above it.

I am deliberately not proposing a replacement number — 65,536 would be only ~3x over the historical max, and a genuinely huge route could trip it, which trades a measured non-vulnerability for a possible false rejection. Purely your call whether the headroom is worth trimming; the caps as written are sound.

Wotuu:
:robot: Cold review (fable): 2 findings posted (both non-blocking).

app/Logic/MDT/IO/MDT2Codec.php
Wotuu: :robot: Acknowledged, no code change — the measured headroom (~60x over the largest real import) is fine as-is. Leaving `MAX_ITEMS` unchanged.

app/Logic/MDT/IO/MDT2Codec.php
Wotuu: :robot: Fixed in 385407b4d.

Root cause confirmed: `UnsignedIntegerObject::create()`/`NegativeIntegerObject::create()` pick the header width with `isLessThan(0xFF/0xFFFF/0xFFFFFFFF)` where it should be `isLessThanOrEqualTo` - an off-by-one in cbor-php. `intToCborObject()` now computes the minimal width itself (unsigned magnitude for positive values, `-1 - value` for negatives, matching `NegativeIntegerObject::getValue()`'s own inverse) and calls `createObjectForValue()` directly instead of going through `create()`, so it's no longer subject to that boundary bug.

Verified against the exact values from the finding:
- `255` -> `18ff` (was `1900ff`)
- `65535` -> `19ffff` (was `1a0000ffff`)
- `-256` -> `38ff` (was widened the same way)
- `-65536` -> `39ffff`
- `4294967295` -> `1affffffff` (was throwing `MDT2EncodeException`)
- `-4294967296` -> `3affffffff` (was throwing)

`4294967296`/`-4294967296` beyond the 32-bit magnitude still throw as intended (the existing `encode_givenIntBeyondUint32_throwsMDT2EncodeException` test covers `0x1_0000_0000`, unchanged).

Also added a map-key boundary test (`encode([255 => 'x'])` -> key header `18ff`), since `intToCborObject` is used for map keys too, per the finding's `currentDungeonIdx` note.

Tests: `php artisan test --filter=MDT2` - 60 passed (106 assertions), including the real-fixture `MDTImportStringServiceMDT2AuthoritativeTest` byte-for-byte round trip (48 assertions, still green - confirms real presets don't hit these boundary values, so this closes a latent gap rather than a currently-observed regression). PHPStan clean. php-cs-fixer clean on both changed files.

app/Console/Commands/MDT/Decode.php
```diff
@@ -30,6 +32,30 @@ class Decode extends Command
...
+            } catch (MDT2DecodeException $mdt2DecodeException) {
+                $this->error($mdt2DecodeException->getMessage());
+
+                logger()->error($mdt2DecodeException->getMessage(), [
+                    'string' => substr($string, 0, 2048),
+                ]);
+            }
+
+            return;
+        }
```
Wotuu: This code should not go in here. This command changed, all changes should go into the converts mdt strings trait instead. This trait is used in other places and I want this new code to work there too.

app/Console/Commands/MDT/Encode.php
```diff
@@ -30,6 +31,22 @@ class Encode extends Command
...
+            if (!is_array($decoded)) {
+                $this->error('The string argument must be valid JSON describing a preset');
+
+                return;
+            }
+
+            $this->info(MDT2Codec::encode($decoded));
+
+            return;
+        }
```
Wotuu: Same thing goes for here, move it into the trait.

app/Http/Controllers/MDTImportController.php
```diff
@@ -101,7 +102,7 @@ public function import(
                 return abort(StatusCode::BAD_REQUEST, __('controller.mdtimport.error.mdt_string_format_not_recognized'));
             } catch (Exception $ex) {
                 // We're not interested if the string was 100% not an MDT string - it will never work then
-                if (isValidBase64($string)) {
+                if (isValidBase64($string) || MDT2Codec::appliesTo($string)) {
```
Wotuu: I saw in the trait that there is a regex that checks for the validity of this MDT string. So I think is valid base 64 is not a thorough check. As a result, I think we can maybe use a static function in the converts MDT strings trait and have it handle both legacy MDT strings and the new MDT2 strings.

app/Logic/MDT/IO/MDT2Codec.php
```diff
@@ -0,0 +1,384 @@
...
+        }
+
+        return sprintf('%s%s', self::PREFIX, base64_encode($deflated));
+    }
+
+    /**
+     * @return array<int|string, mixed> The preset in the same shape the legacy decode path produces
+     * @throws MDT2DecodeException
+     */
+    public static function decode(string $string): array
```
Wotuu: Seeing that this file has the same structure as the converts mdt strings trait, we can maybe create an interface with two class implementations that have the applies to, encode and decode functions in it. Maybe that will help generalize the implementation a bit and make it easier later on when we fully move over to the new implementation to just delete the old implementation.

Now it feels like we have the converts mdt strings trait and this static class together. Doing the same thing but not following an interface. And they have the same behavior so they should be modelled to an interface if you ask me.

app/Service/MDT/MDTBaseService.php
```diff
@@ -26,10 +29,18 @@ protected function getLua(): Lua
...
     /**
-     * @param array<string, mixed> $contents
+     * @param  array<string, mixed> $contents
+     * @param  bool                 $useMDT2Format True to encode into the MDT 6.2+ `!~MDT2~` format instead of the
+     *                                             legacy `!` format. Kept off by default until MDT drops legacy
+     *                                             import support with WoW 12.2.
+     * @throws MDT2EncodeException
      */
-    protected function encode(array $contents): string
+    protected function encode(array $contents, bool $useMDT2Format = false): string
```
Wotuu: Create a new enum that exposes the mdt1 format and the mdt2 format and then accept the enum as a parameter here. And based on that enum call the correct implementation to decode.

It does feel like this service should just also implement the converts mdt strings trait. It doesn't, so I thought that the trait was used in multiple places, but apparently it is not, and it's just used in the php artisan call, which is being called in this service. However, I do think that calling a random artisan command as part of application logic is wrong and it's a bit of a hack. So I do want to get rid of that. The artisan commands are for external usage and not for internal usage, if we can prevent it.

app/Service/MDT/MDTBaseService.php
```diff
@@ -46,6 +57,22 @@ protected function encode(array $contents): string
...
+                // The prefix matched, so this string genuinely claimed to be an MDT export - always worth
+                // reporting, mirroring what ConvertsMDTStrings::transform() does for plausible legacy strings.
+                // Truncated: the input is unvalidated user input of arbitrary size
+                logger()->error($mdt2DecodeException->getMessage(), [
+                    'string' => substr($string, 0, 2048),
+                ]);
+
+                return null;
+            }
+        }
```
Wotuu: Same goes with this thing, yeah

Wotuu (commented):
Some reworking is in order to have the two existing implementations work side-by-side better.

app/Console/Commands/MDT/Decode.php
Wotuu: :robot: Fixed in 29de6a1cf. Moved the MDT2 decoding out of this command entirely: `Decode.php` now just calls `$this->decode($string)` on `ConvertsMDTStrings`, which resolves the format via a new `MDTStringFormat` enum and dispatches to the matching `MDTStringCodecInterface` implementation (`MDT2Codec` or the new `LegacyMDTCodec`, which now holds the `cli_weakauras_parser` shell-out logic that used to live directly in the trait). The trait also now owns the CLI-specific error reporting (console output + structured logging on decode failure), so this same code path works anywhere the trait is used - not just here. See `app/Console/Commands/Traits/ConvertsMDTStrings.php` and `app/Logic/MDT/IO/MDTStringFormat.php`.

app/Console/Commands/MDT/Encode.php
Wotuu: :robot: Fixed in 29de6a1cf, same fix as the sibling comment on `Decode.php`. `Encode.php` no longer branches on `--mdt2` to call `MDT2Codec` directly - it now calls `$this->encode($decoded, $format)` on `ConvertsMDTStrings`, passing an `MDTStringFormat` enum value (`MDT2` when `--mdt2` is set, `Legacy` otherwise). The trait dispatches to the right `MDTStringCodecInterface` implementation, so this logic now lives in one place shared by every trait consumer.

app/Http/Controllers/MDTImportController.php
Wotuu: :robot: Fixed in 29de6a1cf, but with one deliberate deviation from "a static function in the trait" - explaining why below.

Added `MDTStringFormat::isValid(string $string): bool` (in the new `app/Logic/MDT/IO/MDTStringFormat.php` enum) that handles both formats properly:
- For MDT2, it does not just check the `!~MDT2~` prefix - it actually attempts a full decode (base64 -> raw deflate -> CBOR) and returns false if that throws. A string that merely starts with the prefix but is not valid Base64, not a valid deflate stream, or not valid CBOR now correctly reports as invalid (covered by a new unit test in `MDTStringFormatTest`).
- For the legacy format, it reuses the same character-class regex you spotted in the trait (now living on `LegacyMDTCodec::appliesTo()`, since that's where the rest of the legacy-format logic moved to per the other two comments) - going further and actually shelling out to `cli_weakauras_parser` just to validate would be far too expensive for what is ultimately a "is this worth `report()`-ing" heuristic.

`MDTImportController` now calls `MDTStringFormat::isValid($string)` in place of `isValidBase64($string) || MDT2Codec::appliesTo($string)`.

On "a static function in the trait": I put the real logic on the `MDTStringFormat` enum instead, and left `ConvertsMDTStrings::isValidMdtString()` as a one-line static delegate to it. Reason: the trait's other two methods (`encode()`/`decode()`) call `$this->error()` and are documented `@mixin Command`, so pulling the trait into a Controller for the sake of one static method would hand the controller two methods that would fatal if ever invoked outside a Command context. The enum has no such constraint and is where the format-detection logic already lives, so `MDTImportController` calls it directly.

app/Logic/MDT/IO/MDT2Codec.php
Wotuu: :robot: Fixed in 29de6a1cf. Added `App\Logic\MDT\IO\MDTStringCodecInterface` (`appliesTo(string): bool`, `encode(array): string`, `decode(string): array`) and made both formats implement it as fully symmetric implementations:

- `MDT2Codec` now implements the interface directly - its three public methods dropped `static` and became instance methods (internal helpers like `toCborObject()`/`fromCborObject()` stayed private static, they're pure algorithm code with no need to be part of the public contract). Updated the ~35 existing call sites (mostly `MDT2CodecTest`) from `MDT2Codec::encode(...)` to an instantiated `$this->codec->encode(...)` - internals and behavior are untouched, this was purely mechanical.
- The `cli_weakauras_parser` shell-out logic that used to live in `ConvertsMDTStrings` moved out into a new `LegacyMDTCodec`, implementing the same interface.
- A new `MDTStringFormat` enum (`Legacy` / `MDT2`) resolves to the right implementation via `codec()`, and centralizes format detection (`detect()`) and validity checking (`isValid()`, see the reply on the `MDTImportController` comment) in one place instead of every caller re-deriving "is this MDT2 or legacy" itself.

`ConvertsMDTStrings` is now a thin dispatcher over this - it resolves the format via the enum and layers Command-specific error reporting (console output + logging) on top, rather than owning any encode/decode logic itself.

This gets you the "trivial to delete later" property you were after: retiring the legacy path is deleting `LegacyMDTCodec`, its two exceptions, and the `Legacy` case (with `detect()`/`codec()` collapsing to always return MDT2) - nothing else references it directly, everything goes through the interface/enum.

app/Service/MDT/MDTBaseService.php
Wotuu: :robot: Fixed in 29de6a1cf. Added the `MDTStringFormat` enum (`Legacy` / `MDT2`, see the reply on the `MDT2Codec.php` interface comment for the full design) and `encode()` now takes `MDTStringFormat $format = MDTStringFormat::Legacy` instead of `bool $useMDT2Format`, calling `$format->codec()->encode($contents)` directly.

Also removed the internal `Artisan::call('mdt:encode', ...)` here entirely - agreed that calling an artisan command as application logic isn't right. `encode()`/`decode()` now go straight through `MDTStringCodecInterface` (`LegacyMDTCodec`/`MDT2Codec`), the same implementations the artisan commands themselves now call through `ConvertsMDTStrings`. The `mdt:encode`/`mdt:decode` commands are unaffected externally - they're still there for CLI use, just no longer the only way to invoke this logic.

I didn't have this service also `use ConvertsMDTStrings`, since that trait is `@mixin Command` (its methods call `$this->error()`) - `MDTBaseService` isn't a Command, so it talks to the codec/enum directly instead and does its own logging (mirroring what `ConvertsMDTStrings::decode()` does for CLI reporting, minus the console output).

app/Service/MDT/MDTBaseService.php
Wotuu: :robot: Fixed in 29de6a1cf, same fix as the sibling comment on `encode()`. `decode()` no longer calls `Artisan::call('mdt:decode', ...)` - it now resolves the format via `MDTStringFormat::detect($string)` and calls `->codec()->decode($string)` directly, re-throwing `CliWeakaurasParserNotFoundException` (still part of the documented `@throws` on `MDTImportStringServiceInterface`) and otherwise logging + returning null on any other decode failure, same as before.

app/Logic/MDT/IO/MDTStringFormat.php
```diff
@@ -0,0 +1,64 @@
...
+     * shelling out to cli_weakauras_parser, which is too expensive for what is ultimately just a
+     * "is this worth reporting" heuristic.
+     */
+    public static function isValid(string $string): bool
+    {
+        $mdt2 = new MDT2Codec();
+
+        if ($mdt2->appliesTo($string)) {
+            try {
+                $mdt2->decode($string);
```
Wotuu: I think decoding the string entirely is a bit much. Do we really need to completely verify the string's contents? Can we not do an educated guess lik ethe LegacyMDTCodec does? I feel like this would end up as a `isValid() ? decode() : throw` thing, now we're decoding it twice always which sucks for performance.

Can we not just check the start tag of the string and call it a day?

app/Console/Commands/Traits/ConvertsMDTStrings.php
```diff
@@ -2,91 +2,67 @@
...
+use App\Logic\MDT\IO\MDTStringFormat;
 use Illuminate\Console\Command;
-use Symfony\Component\Process\Process;
+use Throwable;
 
 /**
+ * Dispatches MDT export-string encoding/decoding to the correct format implementation (see
+ * MDTStringFormat / MDTStringCodecInterface) and wires up the CLI-specific error reporting -
+ * console output plus, for decode failures, structured logging - both artisan commands need.
+ *
```
Wotuu, line 13: Is this entire trait even necessary anymore? It should only be used in the decode/encode commands, so its guts can be split up in those commands instead?

If it's used in multiple places and the functions don't cleanly split between the two commands we can leave it.

Wotuu (commented):
Almost there.

app/Logic/MDT/IO/MDTStringFormat.php
Wotuu: :robot: Simplified — `MDTStringFormat::isValid()` no longer decodes anything for the MDT2 branch. It now just calls each format's own `appliesTo()` (the MDT2 prefix check / the legacy character-class check), matching how `LegacyMDTCodec` already does its own detection. That does mean a string carrying the `!~MDT2~` prefix with a garbage payload now counts as "valid" here, where before it did not — but the only caller (`MDTImportController`'s catch block) only uses this to decide whether a decode failure is worth a `report()` call, so the cost of that is a bit more Sentry noise on garbage-prefixed strings, nothing functional. Updated the accompanying unit test cases and docblocks (on `MDTStringFormat` and `MDTStringCodecInterface`) to match. Pushed in 70b8dc02b.

app/Console/Commands/Traits/ConvertsMDTStrings.php
Wotuu: :robot: Checked every usage (`grep -r ConvertsMDTStrings app/ tests/`) — the trait has three methods and they don't split cleanly one-per-command:

- `encode()` — only used by `mdt:encode` (`Encode.php`).
- `decode()` — used by `mdt:decode` (`Decode.php`), **and** directly by `MDTExportStringServiceExtractObjectsTest` / `MDTExportStringServiceExtractArrowsTest` via the shared `MDTExportStringServiceTestBase` (`$this->decode($encodedString)`), to decode an export string back down for assertions. That usage predates this PR (already present on master), so it's a real, existing consumer outside the two commands.
- `isValidMdtString()` — turned out to have zero callers anywhere in `app/` or `tests/`. Deleted it as dead code (see the thread-1 reply/commit 70b8dc02b) rather than moving it anywhere.

Since `decode()` is shared by a command and two test files, splitting the trait into the commands would mean duplicating it (or reaching into a command class from a test), so I left the trait in place with just the dead method removed.

Wotuu (commented):
Ok

Wotuu:
Create a follow-up ticket to ensure we switch the MDT string exporter to use the new CBOR format whenever Midnight Season 2 hits.

Wotuu:
:robot: #3740 already tracks the MDT2 export-flip follow-up from #3735 — updated its timing to flip at Midnight Season 2 / WoW 12.1 (2026-08-11) instead of waiting for the 12.2 PTR cycle, per this request. No duplicate ticket filed.

PR #3745, opened 2026-07-29
Heatmap: apply the event type's side effects when filters are restored from the URL

### LINKED ISSUE #3744: Heatmap: filters restored from the URL never apply their side effects (phantom dataType, hidden spell filter, spurious abort error)
:robot: Reported on Discord by Jods.

## Repro

Build a spell-cast filter in the heatmap UI. The address bar shows:

```
/heatmap/retail/algethar-academy/1?type=player_spell&includePlayerSpellIds=264667,2825,381301,390386,80353,32182,466904&region=world
```

Load (or reload) that exact URL and you get:

```
/heatmap/retail/algethar-academy/1?type=player_spell&dataType=player_position&includePlayerSpellIds=…&region=world
```

- an extra `dataType=player_position` appeared that was not in the link;
- the Spells filter section is gone from the sidebar (so the selection looks lost);
- clicking "Player death" then "Spell casts" restores the filter section, re-adds the spells to the URL and drops `dataType` again;
- sometimes a notification appears: `0: An error occurred while performing your request. Please try again.`

Three independent client-side defects. Nothing server-side is involved.

## 1. Restoring the event type from the URL never fires its `change` handler

`SearchFilterRadio.setValue()` (`resources/assets/js/custom/inline/common/search/filters/filterradio.js:36-47`) restores a radio with `$radio.prop('checked', true)`, which does **not** fire a `change` event.

The `change` handler on the event-type radio (`resources/assets/js/custom/inline/common/maps/heatmapsearchsidebar.js:86-106`) is the *only* place that shows/hides the dependent containers and enables/disables the dependent filters. Skipping it means, after a URL restore:

- `dataType` stays enabled (`SearchFilter.enabled = true` by default — `filters/filter.js:6`), so its blade-default `checked` value (`resources/views/common/maps/controls/heatmapsearch.blade.php:263-266`) is written into the URL by `_updateUrl()` (`inline/base/searchinlinebase.js:115-137`) → the phantom `dataType=player_position`;
- `#filter_player_spells_container` keeps the blade's `style="display:none"` (`heatmapsearch.blade.php:279-283` + `common/forms/labelinput.blade.php:13`) → the spell filter is not on screen at all. Same for the data-type and player-death containers.

The param order in the rewritten URL (`type`, `dataType`, `includePlayerSpellIds`, `region`) is exactly the `this.filters` declaration order, confirming the URL was regenerated client-side by `_updateUrl()`.

## 2. Tom Select never adopts a value restored onto the underlying `<select>`

`SearchFilterSelect.setValue()` (`filters/filterselect.js:19-27`) only does `$(sel).val(split)`. By then `refreshSelectPickers()` has already run (`inline/layouts/app.js:25`, in `<head>`, before the map scripts), so the Tom Select control is initialized and stays visually empty.

`searchWithFilters()` — the embed path, `heatmapsearchsidebar.js:258-265` — already calls `refreshSelectPickers()` for exactly this reason, with the comment "external changes don't cause a UI refresh". The normal `activate()` restore path does not.

## 3. Aborted XHRs are reported to the user as errors

`SearchHandler.search()` (`inline/common/search/searchhandler.js:36-39`) deliberately aborts the in-flight request when a new search starts. `complete` guards on `textStatus !== 'abort'` (line 67), but the global `$.ajaxSetup({error: defaultAjaxErrorFn})` (`inline/layouts/app.js:14-16, 116-151`) does not — so every superseded search pops `0: An error occurred…`. The `0` is `xhr.status`, i.e. no HTTP response was received.

## Not a factor

Nothing server-side injects `dataType`. `HeatmapUrlFormRequest.php:39-40` validates `type` and `dataType` as two independent nullable enums, and the controller redirects only carry `+ $request->validated()` forward. The heatmap data is also unaffected: `HeatmapDataFilter` defaults `dataType` to `player_position` anyway (`app/Service/RaiderIO/Dtos/HeatmapDataFilter.php:548-551`), so `type=player_spell` returns the same result with or without the param. This is a URL/UI correctness bug, not a wrong-data bug.

## Fix

1. Extract the event-type `onChange` body into a method and call it in `activate()` after the URL restore (guarded on `!passThroughEverything` so embeds keep receiving whatever `dataType` the parent frame supplies). No search storm: `_search()` returns early while `this.initializing` is true.
2. Call `refreshSelectPickers()` in `SearchInlineBase.activate()` after `_restoreFiltersFromQueryParams()`, mirroring `searchWithFilters()`. This also fixes the same latent bug on the dungeon-route search sidebars.
3. Return early from `defaultAjaxErrorFn` on `textStatus === 'abort'`, matching the existing `complete` guard. Keep it narrow — genuine network failures should still surface.

Canonical URL rule: `dataType` is emitted **only when it applies** (`type=npc_death`), which is the state the UI already converges to after a tab click. Consequence beyond the reported symptom: params that don't apply to the selected event type are now dropped from the URL on load rather than only after a tab click. Intended.

Tests: a vitest suite driving the real sidebar/filter classes over a restored URL (param set in == param set out, container visibility, Tom Select adoption), plus an abort case for `defaultAjaxErrorFn`.

Note: JS is baked into the release assets, so this **cannot** be hotfixed — it ships with the next release.

## Follow-ups found while investigating (not in scope here)

- `DungeonHeatmapController::viewDungeon()` (`app/Http/Controllers/Dungeon/DungeonHeatmapController.php:93-97`) redirects `/heatmap/{gv}/{dungeon}` → `/…/{floor}` **without** `+ $request->validated()`, so every filter is dropped on that hop — unlike every other redirect in that controller.
- `CombatLogEventFilter::fromHeatmapDataFilter()` (`app/Service/CombatLogEvent/Dtos/CombatLogEventFilter.php:485-525`) never copies `includePlayerSpellIds`, so the spell filter is a silent no-op on the local/OpenSearch service path (production goes through `HeatmapDataFilter::toArray()` and is fine).

Wotuu:
:robot: Closes #3744

Reported on Discord by Jods: a shared heatmap link doesn't survive a reload.

```
# what the UI produces
/heatmap/retail/algethar-academy/1?type=player_spell&includePlayerSpellIds=264667,2825,…&region=world
# what you got after loading that exact link
/heatmap/retail/algethar-academy/1?type=player_spell&dataType=player_position&includePlayerSpellIds=…&region=world
```

…with the Spells filter section missing from the sidebar, so the selection looked lost. Clicking through "Player death" → "Spell casts" fixed both. Sometimes a `0: An error occurred while performing your request. Please try again.` notification appeared.

Two client-side defects.

### 1. Restoring the event type never fired its `change` handler

`SearchFilterRadio.setValue()` restores a radio with `.prop('checked', true)`, which does **not** fire `change`. That handler is the only place that enables/disables the filters which only apply to a specific event type and shows/hides their containers (the blade renders them hidden). So after a restore, `dataType` stayed enabled and its blade-default value was written back into the URL, while `#filter_player_spells_container` stayed `display: none` — which is the whole of "the spells are no longer selected": the section simply wasn't on screen.

Fixed by extracting those side effects into `_applyEventTypeVisibility()` and calling it after the restore — in `activate()` **and** in `searchWithFilters()`, the embed's `postMessage` entry point, which restores filters through the same path and had the same defect (found by the pre-review; `showSidebar` defaults to `true`, so the default embed is *not* in pass-through mode and does have a real sidebar to keep in sync). No search storm — `_search()` returns early while `initializing` is still true.

Genuinely pass-through embeds (`showSidebar=0`) early-return: there is no reachable sidebar there (the trigger handle isn't rendered and the nav is parked off-screen), so they keep receiving whatever `dataType` the parent frame passes. Worth stating plainly, since it isn't a no-op: `dataType` is honoured for `player_spell` server-side (`CombatLogEventService` only forces `player_position` for `player_death`), so an embed URL with `type=player_spell&dataType=enemy_position` now aggregates differently under `showSidebar=0` than under `showSidebar=1`. That's the intended split — the embedder owns the filters when there's no UI, the UI owns them when there is.

### 2. Aborted requests were reported as errors

`SearchHandler.search()` deliberately aborts the in-flight request when a new search starts. `complete` guards on `textStatus !== 'abort'`, but the global `$.ajaxSetup({error: defaultAjaxErrorFn})` did not — so every superseded search popped `0: An error occurred…` (`0` is `xhr.status`, i.e. no HTTP response). The guard added is narrow (`textStatus === 'abort'` only) so genuine network failures still surface.

## Behaviour change worth calling out

`dataType` is now emitted **only when it applies** (`type=npc_death`) — the state the UI already converged to after a tab click. More generally, params that don't apply to the selected event type are dropped from the URL **on load** rather than only after a tab click. A first visit (npc deaths) still gets `dataType=player_position`, unchanged.

The mirror image of that is visible on the **default** heatmap page: the Data type control was hidden there until you clicked through the event-type tabs, even though `dataType=player_position` was already in the URL and in the request. It's now shown on load, so the control matches what's being filtered on. Screenshots in the second comment below.

## Considered and rejected

Restoring a value onto a `<select>` with `$select.val()` doesn't make Tom Select adopt it, and `searchWithFilters()` (the embed path) calls `refreshSelectPickers()` for exactly that reason while `activate()` doesn't — so a `refreshSelectPickers()` call in `SearchInlineBase.activate()` looked like a second fix. Measured in headless Chrome against master instead: the select pickers *do* end up in sync there (`#filter_player_spells` and `#filter_classes` both adopt their restored values). Dropped rather than shipped — no demonstrable defect, and `SearchInlineBase` is shared with the route-search and discover sidebars. (The pre-review independently checked that the call *would* have been safe — always a re-sync, never a first init, for all four consumers — so it's a cheap addition if a real case ever turns up.)

## Not a factor

Nothing server-side injects `dataType`: `HeatmapUrlFormRequest` validates `type` and `dataType` as two independent nullable enums, and the redirects only carry `+ $request->validated()` forward. The heatmap data is unaffected either way — `HeatmapDataFilter` defaults `dataType` to `player_position`, so `type=player_spell` returned the same result with or without the param (the before/after screenshots below show the same 284906 runs and the same heat). This was a URL/UI correctness bug, not a wrong-data bug.

## Tests

- `heatmapsearchsidebar.urlrestore.test.js` — drives the real sidebar with real jQuery and the real filter/search classes over a restored URL. Asserts container visibility, the param set written back (compared as parsed params, not raw strings — `_updateUrl()` rebuilds the query in filter declaration order and re-encodes everything), the params that actually reach the backend, the `searchWithFilters()` path, the pass-through early-return, and that clicking a tab afterwards still behaves as before. 8 of its 11 tests fail without this change; the 3 that pass either way are the unchanged-behaviour guards (default first visit's params, pass-through embeds, tab click).
- `app.ajaxerror.test.js` — abort shows nothing, 500 and 422 still show their message.

Several global-script files gained the guarded `module.exports` footer so the tests can require them (the `custom/util.js` pattern; no-op in the browser bundle).

The `heatmap-javascript` skill is updated with the restore-doesn't-fire-`change` rule, the `_applyEventTypeVisibility()` contract, and the `showSidebar` / `passThroughEverything` distinction that made me get the embed path wrong the first time.

## Note

JS is baked into the release assets, so this **cannot** be hotfixed — it ships with the next release.


Wotuu:
:robot: Visual verification — same URL, same dungeon, side by side. Local stack, real Raider.IO data (both show *Showing data from 284906 runs*, and the heat is identical — confirming the phantom `dataType` never changed what was queried).

`/heatmap/retail/algethar-academy/4?type=player_spell&includePlayerSpellIds=264667,2825,381301,390386,80353,32182,466904&region=world`

**Before (master)** — URL rewritten to `?type=player_spell&dataType=player_position&includePlayerSpellIds=…&region=world`, active filters show a phantom "Data type: Player position", and there is no Spells section at all between the event type buttons and the region flags:

master

**After (this branch)** — URL unchanged from the one that was opened, no "Data type" badge, and the Spells section is there with all 7 spells selected:

branch

Measured in the same run (`window.location.search` + container `display` after load):

| | master | this branch |
| --- | --- | --- |
| params written back | `type`, **`dataType`**, `includePlayerSpellIds`, `region` | `type`, `includePlayerSpellIds`, `region` |
| `#filter_player_spells_container` | `none` | visible |
| `#filter_data_type_container` | `none` | `none` |

Wotuu:
:robot: Second visual pass — the **default** heatmap page (`/heatmap/retail/algethar-academy/4`, no query params), because this change is visible there too and that's the busiest heatmap page.

The default event type is Enemy death, which is exactly the event type the Data type filter applies to — and `dataType=player_position` was already being written into the URL and sent to the backend on master. But the control itself was `display: none` until you clicked through the event-type tabs, so you could see "Data type: Player position" in the active-filters row with no way to change it. It's now shown, consistent with what's in the URL.

**Before (master)** — Data type badge in the filter row, no Data type control:

default master

**After (this branch)** — Data type control present, Player position selected:

default branch

The URL is unchanged on this page (`?type=npc_death&dataType=player_position&region=world` on both), and so is the data.

Wotuu:
:robot: Two UI fixes from Wotuu's review, on the same URL as above.

**Before** — the Spells badge is cut off at the sidebar edge (a native scrollbar appears in the header once a third filter is active), and the spell chips' text sits above the middle of their icons:

filters before

**After** — the badge wraps onto a second line inside the sidebar, no scrollbar, and the chip text is centred against the icons:

filters after

### 1. Active filters no longer scroll inside a fixed-height box

`map.css` gave `#heatmap_search_options_current_filters` a hardcoded `height` (71px, 106px embedded) plus `overflow-y: auto`; two rows of badges are already ~82px, so a third filter produced a scrollbar.

The constraint was that `.data_container.explore`'s `max-height: calc(100vh - 306px)` baked that 71px in, so simply removing the height would push the sidebar off-screen. Instead the heatmap sidebar's `.bg-header` is now a flex column bounded at what the old fixed layout worked out to (header 133px + `100vh - 306px` = `100vh - 173px`), with the filter list as `flex: 1 1 auto; min-height: 0`. The badges take what they need and the filter list gives up the space; the sidebar's footprint is unchanged.

Measured with 6 active filters at a 1000px viewport: badge area 71 → 176px, filter list 453 → 383px, sidebar bottom unmoved at 771px. simplebar still scrolls the filter list (1197px of content in 383px).

Long badges also wrap now — the spells badge is 682px wide in a 376px container, and with the scroll box gone it would otherwise have painted ~294px outside the sidebar.

### 2. Select picker option text vertically centred

Not really a Tom Select issue: the option content is a Bootstrap row of a fixed-size icon plus a label column, left at the default `align-items: stretch`, so the label column stretched to the icon's height and its text rendered at the top edge. `align-items-center` on that row in the shared `common/forms/select/imageoption.blade.php` fixes selected items and dropdown options alike (icon midpoint == label midpoint; was 4.75px out).

The same one-word fix went to the two handlebars templates with the identical shape — `composition_icon_option_template` and `weekly_affix_group_select_option_template` (the weekly affixes filter in this same sidebar) — so they don't look wrong next to the fixed ones. **Heads-up:** that partial is shared by the class/spec/affix/dungeon/npc selects site-wide. Happy to split it into its own MR if you'd rather keep this one to the URL-restore fix.

Wotuu (commented):
Ok

resources/assets/js/custom/inline/common/maps/heatmapsearchsidebar.js
```diff
@@ -251,19 +235,64 @@ class CommonMapsHeatmapsearchsidebar extends SearchInlineBase {
...
         this.initializing = false;
         this._search();
     }
 
     searchWithFilters(filters) {
         this._restoreFiltersFromQueryParams(filters);
 
+        // Same as in activate(): restoring the event type doesn't fire its change event, so its side effects
+        // must be applied before we build the search params from the filters (#3744).
+        this._applyEventTypeVisibility();
```
Wotuu, line 253: :robot: This changes the embed contract for a shape the repo itself ships, and I don't think the PR body's "that's the intended split" quite covers it.

The body frames the `dataType` drop as a `showSidebar=0` vs `showSidebar=1` split, but the concrete example on the changed side is our own reference harness: `resources/views/misc/embedheatmap.blade.php` posts

```js
{ function: 'setFilters', type: "player_spell", dataType: "enemy_position", region: "us", ... }
```

and its iframe passes no `showSidebar`, so `DungeonHeatmapController::…($request->get('showSidebar', true))` → `passThroughEverything === false` → `_applyEventTypeVisibility()` runs → `dataType` is disabled and never reaches the request. Server-side that is not a no-op: `CombatLogEventService` only forces `PlayerPosition` for `PlayerDeath`, so `player_spell + enemy_position` previously aggregated enemy positions and now falls back to `HeatmapDataFilter`'s `player_position` default. Different heat, silently.

The same applies to the plain embed **URL** path, not just `postMessage`: `HeatmapEmbedUrlFormRequest` validates `type` and `dataType` independently, so an iframe `src` with `type=player_spell&dataType=enemy_position` gets restored in `activate()` and dropped there too.

Not asking to revert — the UI-side behaviour is clearly right. But since Raider.IO is a cross-origin consumer we can't grep, could this be confirmed with them (or the drop scoped to values the *UI* would have produced, i.e. only clear `dataType` when the embedder didn't explicitly supply one) before it ships? Worth at least calling out the `embedheatmap.blade.php` example in the PR body so the blast radius is on the record.

Wotuu:

resources/assets/js/custom/inline/common/maps/heatmapsearchsidebar.js
Wotuu: :robot: Confirmed with Wotuu this is not a regression: `dataType` (specifically `enemy_position`) is only ever meaningful for the `npc_death` event type. Verified server-side in `CombatLogEventService.php` — the `context` field indexed alongside each `CombatLogEvent` only carries `pos_enemy_grid_x`/`pos_enemy_grid_y` for `NpcDeath` events; `PlayerSpell` events' context is just `spell_id`, and `PlayerDeath`'s is just `npc_id`. So requesting `dataType=enemy_position` against a `player_spell` event (as `embedheatmap.blade.php`'s example does) executes the nested-aggregation query but against a field that never exists on those documents — it returns zero buckets and always has, both before and after this PR. Hiding the `dataType` control for non-`npc_death` event types matches the real data model rather than changing behavior. No code change needed.

PR #3747, opened 2026-07-29
Stop the site header crashing once a season is seeded before it starts

### LINKED ISSUE #3746: Seeding a season before it starts takes every page down via the site header
:robot:

`DungeonService::getDungeonsForGameVersion()` reads `$nextSeason->dungeons` lazily:

```php
$currentSeason = $this->seasonService->getCurrentSeason($gameVersion->expansion);
$nextSeason    = $currentSeason === null ? null : $this->seasonService->getNextSeason($currentSeason);

return ($nextSeason ?? $currentSeason)?->dungeons ?? $gameVersion->expansion->dungeons;
```

`getNextSeason()` returns a season out of `SeasonService`'s multi-row season cache, so
`preventLazyLoading` enforces on it — unlike the current season it falls back to, which is loaded as
a single model and is exempt from the guard.

**This is dormant only while no future season exists.** `getNextSeason()` returns `null` today, so
the lazy read never fires. The moment a season is seeded before its start date — which is normal
practice when preparing a new season — it throws.

`HeaderComposer` calls this to build the site header, so the failure is not scoped to one page:
**every page rendering `layouts/sitepage.blade.php` 500s in dev and logs a lazy-loading violation in
production.**

Found while seeding Midnight season 2 (#3734) ahead of its start date. It took the local dev site
down completely — every page including `/` returned 500 with
`Attempted to lazy load [dungeons] on model [App\Models\Season]` — until the fix was checked out.

### Fix

Query the relation instead of reading it lazily:

```php
return ($nextSeason ?? $currentSeason)?->dungeons()->get() ?? $gameVersion->expansion->dungeons;
```

### Why separate

The fix already exists inside #3739, but that PR carries the whole Midnight S2 dry run and is pinned
to an unmerged MDT PTR branch. Splitting this out so it can be merged on its own and unblock anyone
who needs a future season seeded locally.

Wotuu:
:robot:

Closes #3746

Extracted from #3739 so it can merge on its own — that PR carries the whole Midnight S2 dry run and
is pinned to an unmerged MDT PTR branch, and this fix blocks anyone who seeds a future season
locally.

`DungeonService::getDungeonsForGameVersion()` read `$nextSeason->dungeons` lazily. `getNextSeason()`
returns a season out of `SeasonService`'s multi-row season cache, so `preventLazyLoading` enforces on
it — unlike the current season it falls back to, which is loaded as a single model and is exempt.

That made the crash **dormant rather than absent**. While no future season exists `getNextSeason()`
returns `null` and the lazy read never fires. Seed a season before its start date — normal practice
when preparing one — and it throws. `HeaderComposer` builds the site header with this, so it is not
scoped to one page: **every page rendering `layouts/sitepage.blade.php` 500s in dev and logs a
violation in production.**

It did exactly that on the local dev site while #3739 was seeding Midnight S2 — every page including
`/` returned 500 with `Attempted to lazy load [dungeons] on model [App\Models\Season]`.

## The fix

```diff
-return ($nextSeason ?? $currentSeason)?->dungeons ?? $gameVersion->expansion->dungeons;
+return ($nextSeason ?? $currentSeason)?->dungeons()->get() ?? $gameVersion->expansion->dungeons;
```

`dungeons()->get()` and deliberately **not** `loadMissing()`: `SeasonService` keeps its season models
in a service-level cache, so memoising the relation onto them makes it outlive the request and leak a
stale dungeon list into later ones. That was measured in #3739 — `loadMissing()` turned the suite red
with a route being created against the wrong dungeon.

## Tests

Two, both in a new `GetDungeonsForGameVersionTest` alongside the existing `GetDungeonContextTest`:
the upcoming-season branch and the current-season fallback. **Both fail with a
`LazyLoadingViolationException` without the one-line change** — verified by stashing it and re-running.

They load the seasons as a collection on purpose. A single-model load is exempt from the guard, which
is precisely why the fallback path never tripped it, so a test that fetched one season with `first()`
would pass with or without the fix and prove nothing.

The file is identical to #3739's version of it, so that PR's diff collapses on rebase once this
merges.

### DIFF
--- a/app/Service/Dungeon/DungeonService.php
+++ b/app/Service/Dungeon/DungeonService.php
@@ -120,6 +120,16 @@ class DungeonService implements DungeonServiceInterface
         $currentSeason = $this->seasonService->getCurrentSeason($gameVersion->expansion);
         $nextSeason    = $currentSeason === null ? null : $this->seasonService->getNextSeason($currentSeason);
 
-        return ($nextSeason ?? $currentSeason)?->dungeons ?? $gameVersion->expansion->dungeons; // @phpstan-ignore nullsafe.neverNull
+        // Load the relation explicitly rather than reading it lazily. getNextSeason() returns a season out
+        // of a collection, so as soon as an upcoming season exists the lazy read trips preventLazyLoading
+        // and takes down every page that renders the header (HeaderComposer). This never fired before
+        // because getNextSeason() returned null while no future season was seeded, and the current season
+        // it fell back to is loaded as a single model, which the guard exempts.
+        //
+        // Deliberately dungeons()->get() and NOT loadMissing(): SeasonService keeps its season models in a
+        // service-level cache, so memoising the relation onto them makes it outlive the request and leak a
+        // stale dungeon list into later ones. That was measurable - loadMissing() turned the suite red with
+        // a route being created against the wrong dungeon.
+        return ($nextSeason ?? $currentSeason)?->dungeons()->get() ?? $gameVersion->expansion->dungeons;
     }
 }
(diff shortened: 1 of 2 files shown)


Wotuu (commented):
Ok

PR #3749, opened 2026-07-29
#3720 Scope MappingVersion clone-on-create to the new version's game_version_id

### LINKED ISSUE #3720: MappingVersion clone-on-create can pick the wrong previous version for dungeons on multiple game versions (follow-up to #3713)

Wotuu:
:robot:

Closes #3720

## Summary

`version` is only unique per `game_version_id` (a separate counter per game version, see
`MappingService::createNewBareMappingVersion()`), so a dungeon with mapping versions on more than
one game version (e.g. retail + Legion Remix) can have interleaved `version` numbers across game
versions. Several mapping-version actions silently ignored `game_version_id` when they should
have scoped by it, letting an unscoped lookup pick the wrong game version's data. Per the review
round on this PR, the audit was widened to find and fix every other instance of the same class of
bug rather than stopping at the one reported in #3720.

- `MappingVersion::boot()`'s `static::created` clone-on-create hook picked the "previous mapping
  version" to clone from by ordering the dungeon's mapping versions globally by `version` desc,
  with no `game_version_id` filter - could clone an unrelated game version's enemies/map
  icons/etc. into a new mapping version.
- `MappingService::copyMappingVersionToDungeon()` derived the clone's `game_version_id` and next
  `version` from the target dungeon's *ambient* current mapping version (resolved through the
  acting user's/default game version) instead of the source's own game version.
- `Console/Commands/Mapping/Copy.php` (`mapping:copy`) recomputed the moved mapping version's
  `version` from the target dungeon's ambient current mapping version instead of the command's own
  `gameVersion` argument.
- `CombatLogMappingVersionService::createMappingVersionFromCombatLog()` picked the "most recent
  mapping version" for a newly-bootstrapped dungeon without scoping by `game_version_id`.
- `DungeonRepository::getMappingVersionByVersion()` / `getByMappingVersion()` (and the
  Swoole-cached override) matched a raw `version` number with no `game_version_id` filter at all.
  Confirmed with Wotuu that the only real caller - combat-log-derived route creation (the Auto
  Route Creator) - only ever runs on retail, so these now take an explicit `GameVersion` and
  `CombatLogRouteRequestModel::createDungeonRoute()` passes retail explicitly.
  `getByMappingVersion()` also had a second, independent bug where it matched `version` and
  `game_version_id` via two separate `whereRelation()` calls (each its own EXISTS clause,
  satisfiable by two *different* rows) - fixed with a single `whereHas()` closure requiring both
  on the same row.

Each fix mirrors the existing scoping pattern already used elsewhere in this codebase
(`Dungeon::getCurrentMappingVersionForGameVersion()`).

## Test plan

- [x] `MappingVersionCloneGameVersionScopingTest` - reproduces the boot() hook interleaving via a
  decoy mapping version inserted so its creation doesn't itself trigger the clone hook. Includes a
  second test pinning down the early-return branch (first mapping version for a new game version).
- [x] `CopyMappingVersionToDungeonGameVersionScopingTest` - reproduces the ambient-vs-source
  mismatch for `copyMappingVersionToDungeon()`.
- [x] `CopyGameVersionScopingTest` - drives the real `mapping:copy` Artisan command via
  `Artisan::call()` and asserts the new version number isn't derived from an unrelated decoy.
- [x] `CombatLogMappingVersionServiceGameVersionScopingTest` - drives
  `createMappingVersionFromChallengeMode()` through a minimal synthetic combat log file.
- [x] `DungeonRepositoryTest` - two new cases covering the game-version ambiguity and the
  two-independent-EXISTS-clauses bug in `getByMappingVersion()`.
- Every new test was verified to fail against the pre-fix code and pass with the fix (reverted the
  relevant hunk, re-ran, restored) rather than trusting a single green run.
- [x] `php artisan test --compact --group=MappingVersion --group=DungeonRepository --group=MDT`,
  the full combat-log-route API suite
  (`tests/Feature/Controller/Api/V1/APICombatLogController/CombatLogRoute`), and a full
  `php artisan test --compact --exclude-group=MapTiles` run - all green except one pre-existing,
  environment-specific failure (`NpcCompendiumControllerActivityTest`) confirmed to reproduce
  identically with every change in this PR reverted (local dev-DB drift, not a regression - this
  branch's CI already passed all 9 checks before this failure was investigated).
- [x] `composer run fix` (no drift) and `composer run analyse` (no errors).

## Follow-ups filed

- #3754 originally tracked the `DungeonRepository` gap separately, since fixing it seemed to need a
  protocol-level decision (no game version signal exists in the combat log payload). Wotuu
  confirmed the Auto Route Creator is retail-only, which resolved the ambiguity - #3754 is fixed
  here and closed.
- #3757 tracks a related but larger gap found during independent review of this PR:
  `MappingService::createNewMappingVersionFromMDTMapping()` has the same unscoped-fallback bug, but
  its root cause is one level up in `MDTMappingImportService::importMappingVersionFromMDT()` - the
  SAME `$currentMappingVersion` also feeds checkpoint/enemy/patrol/POI cloning, so a real fix means
  touching ~5 method signatures in the primary, scheduled MDT import pipeline (`mdt:importmapping`)
  to make a null predecessor (genuinely first import for a new game version) a handled case rather
  than an unscoped fallback. That's materially more invasive than anything else in this PR (mostly
  rarely-used admin/dev tools) and higher-traffic/higher-risk to rush, so it's tracked separately
  rather than folded in here.

### DIFF
--- a/app/Models/Mapping/MappingVersion.php
+++ b/app/Models/Mapping/MappingVersion.php
@@ -550,8 +550,13 @@ class MappingVersion extends Model
             if ($newMappingVersion->dungeon === null) {
                 return;
             }
+            // Scope by game_version_id - `version` is only unique per game version, so a dungeon with
+            // mapping versions on multiple game versions would otherwise have their `version` numbers
+            // interleaved when ordered globally, picking the wrong previous mapping version to clone from.
             /** @var EloquentCollection<int, MappingVersion> $existingMappingVersions */
-            $existingMappingVersions = $newMappingVersion->dungeon->mappingVersions()->get();
+            $existingMappingVersions = $newMappingVersion->dungeon->mappingVersions()
+                ->where('game_version_id', $newMappingVersion->game_version_id)
+                ->get();
             // Nothing to do if we don't have an older mapping version
             if ($existingMappingVersions->count() < 2) {
                 return;
--- a/tests/Feature/App/Model/Mapping/MappingVersionCloneGameVersionScopingTest.php
+++ b/tests/Feature/App/Model/Mapping/MappingVersionCloneGameVersionScopingTest.php
@@ -5,7 +5,7 @@ namespace Tests\Feature\App\Model\Mapping;
 use App\Models\Dungeon;
 use App\Models\GameVersion\GameVersion;
 use App\Models\Mapping\MappingVersion;
-use Illuminate\Support\Facades\DB;
+use App\Service\Mapping\MappingServiceInterface;
 use PHPUnit\Framework\Attributes\Group;
 use PHPUnit\Framework\Attributes\Test;
 use Tests\TestCases\PublicTestCase;
@@ -37,12 +37,13 @@ final class MappingVersionCloneGameVersionScopingTest extends PublicTestCase
             // globally-ordered list of the dungeon's mapping versions can put this decoy where the
             // real same-game-version predecessor belongs - the exact interleaving #3720 describes.
             //
-            // Inserted via the query builder, not MappingVersion::create(): going through Eloquent
+            // Inserted via insertGetId(), not MappingVersion::create(): going through Eloquent
             // would fire this very clone-on-create hook for the decoy itself, and - since it's the
             // dungeon's globally-highest version at that point - the (buggy) hook would clone
             // $existingMappingVersion's real values into it, overwriting the 111111 sentinel this
-            // test relies on to detect a wrong pick.
-            $decoyMappingVersionId = DB::table('mapping_versions')->insertGetId([
+            // test relies on to detect a wrong pick. This needs to happen quietly as to not trigger
+            // MappingVersion events defined in its class, matching MappingService's own usage.
+            $decoyMappingVersionId = MappingVersion::insertGetId([
                 'game_version_id'                 => $otherGameVersion->id,
                 'dungeon_id'                      => $dungeon->id,
                 'version'                         => $existingMappingVersion->version + 500000,
@@ -87,7 +88,68 @@ final class MappingVersionCloneGameVersionScopingTest extends PublicTestCase
                 $newMappingVersion->delete();
             }
             if ($decoyMappingVersionId !== null) {
-                DB::table('mapping_versions')->where('id', $decoyMappingVersionId)->delete();
+                MappingVersion::query()->where('id', $decoyMappingVersionId)->delete();
+            }
+        }
+    }
+
+    /**
+     * #3720 review follow-up: scoping the clone-on-create lookup by game_version_id also scopes its
+     * `count() < 2` guard, so creating the FIRST mapping version for a game version the dungeon doesn't
+     * have yet now hits the early-return branch instead of the clone. This pins that branch down: the new
+     * mapping version must come out with raw column defaults (matching what
+     * MappingService::createNewBareMappingVersion() already produces for the same scenario) and no
+     * content cloned in from an unrelated game version - not the previous, pre-#3720 behaviour of
+     * grabbing whatever mapping version happened to sort second in the dungeon's globally-ordered list.
+     */
+    #[Test]
+    public function create_givenDungeonWithNoMappingVersionOnThisGameVersionYet_earlyReturnsWithRawDefaultsAndNoContent(): void
+    {
+        // Arrange
+        $mappingService = $this->app->make(MappingServiceInterface::class);
+
+        /** @var Dungeon $dungeon */
+        $dungeon = Dungeon::whereNotNull('challenge_mode_id')->firstOrFail();
+        /** @var MappingVersion $existingMappingVersion */
+        $existingMappingVersion = $dungeon->mappingVersions()->firstOrFail();
+
+        /** @var GameVersion $newGameVersion */
+        $newGameVersion = GameVersion::query()
+            ->whereNotIn('id', $dungeon->mappingVersions()->pluck('game_version_id'))
+            ->firstOrFail();
+
+        $newMappingVersion = null;
+
+        try {
+            // Act
+            $newMappingVersion = $mappingService->createNewMappingVersionFromPreviousMapping($dungeon, $newGameVersion);
+
+            // Assert
+            $freshNewMappingVersion = $newMappingVersion->fresh();
+            $this->assertNotNull($freshNewMappingVersion);
+            $this->assertSame(
+                0,
+                $freshNewMappingVersion->enemy_forces_required,
+                'A dungeon\'s first mapping version for a game version has nothing to clone scalars from and must land on the raw column default.',
+            );
+            $this->assertSame(
+                0,
+                $freshNewMappingVersion->timer_max_seconds,
+                'A dungeon\'s first mapping version for a game version has nothing to clone scalars from and must land on the raw column default.',
+            );
+            $this->assertNotSame(
+                $existingMappingVersion->enemy_forces_required,
+                $freshNewMappingVersion->enemy_forces_required,
+                'The early-return branch must not fall back to cloning scalars from a mapping version on a different game_version_id.',
+            );
+            $this->assertSame(
+                0,
+                $freshNewMappingVersion->enemies()->count(),
+                'The early-return branch must not clone content from a mapping version on a different game_version_id.',
+            );
+        } finally {
+            if ($newMappingVersion !== null) {
+                $newMappingVersion->delete();
             }
         }
     }
--- a/tests/Feature/App/Model/Mapping/MappingVersionCloneGameVersionScopingTest.php
+++ b/tests/Feature/App/Model/Mapping/MappingVersionCloneGameVersionScopingTest.php
@@ -1,0 +1,94 @@
+<?php
+
+namespace Tests\Feature\App\Model\Mapping;
+
+use App\Models\Dungeon;
+use App\Models\GameVersion\GameVersion;
+use App\Models\Mapping\MappingVersion;
+use Illuminate\Support\Facades\DB;
+use PHPUnit\Framework\Attributes\Group;
+use PHPUnit\Framework\Attributes\Test;
+use Tests\TestCases\PublicTestCase;
+
+#[Group('MappingVersion')]
+final class MappingVersionCloneGameVersionScopingTest extends PublicTestCase
+{
+    #[Test]
+    public function create_givenDungeonWithMappingVersionOnAnotherGameVersion_clonesFromSameGameVersionPredecessor(): void
+    {
+        // Arrange
+        /** @var Dungeon $dungeon */
+        $dungeon = Dungeon::whereNotNull('challenge_mode_id')->firstOrFail();
+        /** @var MappingVersion $existingMappingVersion */
+        $existingMappingVersion = $dungeon->mappingVersions()->firstOrFail();
+
+        /** @var GameVersion $otherGameVersion */
+        $otherGameVersion = GameVersion::query()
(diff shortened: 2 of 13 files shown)


app/Models/Mapping/MappingVersion.php
```diff
@@ -550,8 +550,13 @@ protected static function boot()
             if ($newMappingVersion->dungeon === null) {
                 return;
             }
+            // Scope by game_version_id - `version` is only unique per game version, so a dungeon with
+            // mapping versions on multiple game versions would otherwise have their `version` numbers
+            // interleaved when ordered globally, picking the wrong previous mapping version to clone from.
             /** @var EloquentCollection<int, MappingVersion> $existingMappingVersions */
-            $existingMappingVersions = $newMappingVersion->dungeon->mappingVersions()->get();
+            $existingMappingVersions = $newMappingVersion->dungeon->mappingVersions()
+                ->where('game_version_id', $newMappingVersion->game_version_id)
```
Wotuu, line 558: :robot: The content-clone scoping itself looks right (correct column, and the `count() < 2` guard is consistently scoped with the index into `[1]`). I also pulled the new test onto pre-fix `master` and ran it — it fails with `Failed asserting that 111111 is identical to 431`, so it genuinely covers the bug rather than asserting already-true behaviour.

One collateral consequence I'd like confirmed as intended, since the description doesn't mention it: scoping this list also scopes the `count() < 2` guard, so creating the **first** mapping version for a game version a dungeon doesn't have yet now returns early.

`MappingService::createNewMappingVersionFromPreviousMapping()` — the admin "Add mapping version" button (the game version dropdown in `admin/dungeon/mappingversions.blade.php` offers *all* game versions, not just ones the dungeon already has) and `mapping:copy <gameVersion> ...` — passes no `enemy_forces_required` / `enemy_forces_required_teeming` / `enemy_forces_shrouded*` / `timer_max_seconds` in its `create([...])`; it relies entirely on the `update([...])` a few lines below to seed them. With the early return those now land at raw DB defaults: `enemy_forces_required = 0` and `timer_max_seconds = 0` (both `NOT NULL DEFAULT '0'`), teeming/shrouded `NULL` — plus of course an empty mapping instead of a copy.

To be explicit: this is **not** a request to restore cross-game-version cloning — that is the bug. The pre-fix behaviour in this same scenario was also wrong: for a brand-new `version = 1` row, global index `[1]` isn't even the newest other-game-version mapping, it's whatever sorts second overall. It's just that the scalar seed is separable from the content clone, and "first mapping version for a new game version now comes out with `enemy_forces_required = 0` / `timer_max_seconds = 0`" is the one user-visible change in this diff. Also worth a second test case if you do want to pin the new early-return branch down. If the current outcome is what you want, no action needed.

tests/Feature/App/Model/Mapping/MappingVersionCloneGameVersionScopingTest.php
```diff
@@ -0,0 +1,94 @@
...
+            // game_version_id (see MappingService::createNewBareMappingVersion()), so an unscoped,
+            // globally-ordered list of the dungeon's mapping versions can put this decoy where the
+            // real same-game-version predecessor belongs - the exact interleaving #3720 describes.
+            //
+            // Inserted via the query builder, not MappingVersion::create(): going through Eloquent
+            // would fire this very clone-on-create hook for the decoy itself, and - since it's the
+            // dungeon's globally-highest version at that point - the (buggy) hook would clone
+            // $existingMappingVersion's real values into it, overwriting the 111111 sentinel this
+            // test relies on to detect a wrong pick.
+            $decoyMappingVersionId = DB::table('mapping_versions')->insertGetId([
```
Wotuu: :robot: Nit (non-blocking): `.claude/CLAUDE.md` says "Avoid `DB::`; prefer `Model::query()`". The reason for not using `MappingVersion::create()` is well explained, but the repo already has an idiom for exactly this "insert without firing MappingVersion's events" need — `MappingService::createNewMappingVersionFromMDTMapping()` and `copyMappingVersionToDungeon()` both do `MappingVersion::insertGetId([...])` under the comment *"This needs to happen quietly as to not trigger MappingVersion events defined in its class"*.

`MappingVersion::insertGetId([...])` here (and `MappingVersion::query()->where('id', $decoyMappingVersionId)->delete()` in the `finally`, matching `MappingVersionDeletionTest`'s guard cleanup) bypasses model events in exactly the same way while keeping the convention and dropping the `Illuminate\Support\Facades\DB` import.

Wotuu:

tests/Feature/App/Model/Mapping/MappingVersionCloneGameVersionScopingTest.php
Wotuu: :robot: Good catch, done in `2be8ee931`: swapped `DB::table('mapping_versions')->insertGetId([...])` / `DB::table('mapping_versions')->where(...)->delete()` for `MappingVersion::insertGetId([...])` / `MappingVersion::query()->where('id', $decoyMappingVersionId)->delete()`, matching the exact idiom `MappingService::createNewMappingVersionFromMDTMapping()` / `copyMappingVersionToDungeon()` already use under the "needs to happen quietly" comment.

Confirmed `insertGetId()` genuinely bypasses the `creating`/`created` boot hooks (it goes through the query builder, same as `DB::table()->insertGetId()` did) - the swap is behavior-neutral. Re-ran the test after the change: still passes (`php artisan test --filter=MappingVersionCloneGameVersionScopingTest`), and it still fails the same way against pre-#3720 `master` (wrong game-version pick), so it still covers what it did before.

app/Models/Mapping/MappingVersion.php
Wotuu: :robot: Investigated - confirmed this branch is reachable, and confirmed as intended: no code change made, resolving as a non-issue.

**Reachability**: yes. `resources/views/admin/dungeon/mappingversions.blade.php:11-12` builds the "Add mapping version" dropdown from `$allGameVersions` unfiltered - every game version is offered regardless of whether the dungeon already has a mapping version for it - and `MappingVersionController::saveNew()`'s `'Add mapping version'` branch (line 28-29) calls `createNewMappingVersionFromPreviousMapping()` unconditionally, with no null-guard (unlike the sibling `'Add bare mapping version'` branch just below it, which explicitly checks `getCurrentMappingVersion($gameVersion) === null` first). So an admin picking a game version the dungeon has zero mapping versions for, via "Add mapping version", does hit the new early-return branch.

**Why the outcome is correct, not a gap**: `createNewBareMappingVersion()` also goes through `MappingVersion::create()`, so it fires the exact same `boot()` clone-on-create hook as `createNewMappingVersionFromPreviousMapping()`. Pre-#3720, that hook was completely unscoped - `$newMappingVersion->dungeon->mappingVersions()->get()`, ordered globally by `version` desc with no `game_version_id` filter at all. That means **"Add bare mapping version" for a dungeon's first-ever mapping version on a new game version had the identical defect**, and a worse manifestation of it: if the dungeon already had a mapping version on any other game version, the global `count() < 2` guard would not have early-returned, and the hook would have cloned that unrelated game version's *entire content* (enemies, packs, patrols, everything) into what was supposed to be a bare mapping version - not just the four scalar fields the review flagged for "Add mapping version".

Post-#3720, both actions now correctly land on a genuinely empty mapping version with raw column defaults (`enemy_forces_required = 0`, `timer_max_seconds = 0`, teeming/shrouded `NULL`) for this exact scenario - which is exactly what "Add bare mapping version" was always supposed to produce, and the only sensible outcome when no same-game-version predecessor exists to clone from (there's nothing "correct" to seed from an unrelated game version's numbers). So this isn't a new asymmetry introduced by the fix - it's one latent cross-game-version bug removed from *both* actions, which now correctly converge on the same safe, empty result instead of "Add bare" silently importing unrelated content and "Add mapping version" silently importing unrelated scalars.

I also confirmed these scalar fields aren't a dead end afterward: `common/modal/mappingversion.blade.php` + `AjaxMappingVersionController`/`APIMappingVersionFormRequest` let an admin edit `enemy_forces_required`/`enemy_forces_required_teeming`/`enemy_forces_shrouded*`/`timer_max_seconds`/`facade_enabled` on a mapping version directly after creation - the same fix-up step already required for a genuinely bare mapping version.

Added a test pinning this branch down: `create_givenDungeonWithNoMappingVersionOnThisGameVersionYet_earlyReturnsWithRawDefaultsAndNoContent` in `2be8ee931` - asserts raw-default scalars, zero cloned enemies, and that the scalars don't match the existing (different-game-version) mapping version's values. It fails against pre-#3720 `master` for the same reason the first test does.

Wotuu (commented):
Ok

PR #3750, opened 2026-07-29
Report scheduler command exceptions so failures are diagnosable

### LINKED ISSUE #3748: patreon:refreshmembers fails hourly with no diagnosable log output
:robot: `patreon:refreshmembers` has been failing hourly in production with nothing but the generic scheduler line:

```
Scheduled command ['/usr/local/bin/php' 'artisan' patreon:refreshmembers] failed with exit code [1].
```

## Why there's no further information

Exit code 1 for this command can only come from `SchedulerCommand::trackTime()`:

```php
try {
    $callable();
} catch (Exception $ex) {
    $this->error($ex->getMessage());

    return 1;
}
```

The exception message is written to the command's **stdout** (`appendOutputTo` → container stdout), which is a different sink than the one that produced the alert above — that one comes from `App\Exceptions\Handler::report()` (visible in frame #4 of the stack trace). So the actual exception is caught, its message thrown at a channel nobody is watching, and only "exit code 1" survives. Nothing is ever `report()`ed.

This affects **all 18 scheduler commands**, not just this one.

## Secondary issue: `trackTime()` discards the closure's return value

```php
$callable();   // return value ignored
...
return 0;
```

`RefreshMembershipStatus` returns `-1` when any of `loadCampaignBenefits()` / `loadCampaignTiers()` / `loadCampaignMembers()` returns `null` — but that value is dropped and the command exits `0`. That entire class of failure (expired admin token, Patreon API errors, unexpected response shape) is **currently completely silent**.

## Most likely cause of the actual crash

`PatreonService::applyPaidBenefitsForMember()`:

```php
if (!$user->hasPatreonBenefit($benefit)) {
    PatreonUserBenefit::create([
        'patreon_user_link_id' => $user->patreon_user_link_id,
        'patreon_benefit_id'   => PatreonBenefit::ALL[$benefit],   // <-- unguarded
    ]);
```

`$benefit` is a benefit **title as configured on patreon.com**, and `PatreonBenefit::ALL` only maps the titles we know about — note `unlimited-dungeonroutes` is commented out of that map. `hasPatreonBenefit()` guards with `isset(PatreonBenefit::ALL[$key])` and returns `false` for an unknown title, so an unknown/renamed/added benefit falls straight through to the unguarded lookup. Undefined array key → Laravel converts the warning to `ErrorException` → caught by `trackTime` → exit 1, message discarded. A single patron on such a tier kills the whole hourly run, and every member after them is skipped.

## Proposed fix

1. `SchedulerCommand::trackTime()`: `report($ex)` so the exception reaches the same sink as the alert, with the command name as context. Widen to `catch (Throwable)` — only together with the `report()`, since `Error`/`TypeError` are currently reported by the console kernel and must not become less visible.
2. `trackTime()`: honour the closure's return value.
3. `RefreshMembershipStatus`: `FAILURE` instead of `-1` (Symfony clamps `-1` to 255), and split the combined null check so the log names *which* load failed.
4. Guard the `PatreonBenefit::ALL[$benefit]` lookup and log the unknown benefit title instead of crashing.
5. Wrap the per-member loop so one bad member doesn't abort the run — log the member id, keep going, and still exit non-zero at the end so the failure stays visible.

Note that (2) means previously-silent failures will start alerting after this ships.

Wotuu:
:robot: Closes #3748

`patreon:refreshmembers` has been failing hourly in production with nothing but `failed with exit code [1]`. The reason there was no usable information is `SchedulerCommand::trackTime()`:

```php
catch (Exception $ex) {
    $this->error($ex->getMessage());   // -> command stdout, no stack trace
    return 1;
}
```

The exception is never `report()`ed. Its message goes to the command's stdout (which `routes/console.php` does append to the container's stdout, so it is *somewhere* in CloudWatch) — but never to `App\Exceptions\Handler::report()`, which is what produces the alert that actually gets read, and never with a stack trace. So the alert carries an exit code and nothing else. This affects all 18 scheduler commands.

## Changes

**`SchedulerCommand::trackTime()`**
- Reports the throwable through the application's exception handler, so the message *and its stack trace* land in the same place as the `Scheduled command [...] failed with exit code [1]` line. The command name (and any extra context, e.g. the Patreon member id) is pushed onto Laravel's `Context` so it shows up next to the `trace_id`.
- Widened `catch (Exception)` to `catch (Throwable)`. Deliberately paired with the `report()`: `Error`/`TypeError` previously escaped to the console kernel which *does* report them, so catching them without reporting would have made things less visible.
- Honours the callable's return value, which it previously discarded.

**`RefreshMembershipStatus`**
- The three `null` load checks are separate, so the exit code is paired with the log line naming which load failed.
- Each member is applied in isolation: one that throws is reported with its member id and no longer aborts every member after it, while the command still exits non-zero. Only the first 3 failures are reported — a systemic failure hits every member and would otherwise write the same stack trace to the log once per member.
- Returns `FAILURE` instead of `-1` (Symfony clamps `-1` to 255).

**`RefreshOutdatedThumbnails`** — returned a hardcoded `0`, discarding its `trackTime()` result. It runs every 15 minutes, so it would have kept exiting 0 through any failure despite the change above.

**`PatreonService::applyPaidBenefitsForMember()`** — likely root cause
- `PatreonBenefit::ALL` is keyed by benefit **titles configured on patreon.com**, and `hasPatreonBenefit()` already returns `false` for a title we don't know — so a renamed or newly added campaign benefit fell straight through into `PatreonBenefit::ALL[$benefit]`, an undefined array key. Laravel converts that warning to an `ErrorException`, `trackTime` caught it, and the message was discarded. One patron on such a tier kills the whole hourly run.
- An unknown title now aborts that member *before* any add or revoke, logging the offending titles. Skipping only the unknown title would be worse than the crash: `$user->getPatreonBenefits()->diff($newBenefits)` would then revoke the benefit the title was renamed *from*, silently stripping a paying patron.
- The method now returns an `ApplyPaidBenefitsForMemberResult` enum (following the existing `LinkToUserIdResult` convention) so the command can tell "member has no linked account" — the common case, not a failure — from "we cannot compute this member's benefits", which fails the command.

Each guard was verified by reverting it: without the first, the test fails with an undefined array key at the exact production failure site; with a naive skip-this-title guard, the revoke test fails; without the enum handling, the silent-success test fails.

## Behavioural change worth knowing about

`trackTime()` now honours the return value, so failures that were previously **silent** start alerting: all three `null`-load paths of `patreon:refreshmembers`. (`metric:aggregate`'s failure branch is unreachable in practice — `DB::insert` throws under `PDO::ERRMODE_EXCEPTION` rather than returning `false`.)

## Review

Cold-reviewed by a fresh-context agent. It found one blocking defect — the unknown-benefit guard returned `false` into callers that discarded it, which would have converted the loud hourly failure into a silent success with benefits quietly no longer applied. Fixed by the enum above, with a regression test. It also corrected the premise that these reports reach Discord/Rollbar: `Handler::report()` gates that branch behind `!runningInConsole()`, so scheduled commands report to the default log channel only. The comments and this description have been corrected accordingly.

## Tests

- `SchedulerCommandTest` — throwable is reported and mapped to `FAILURE`; an `Error` is caught too; the callable's exit code is honoured; a `void` callable is success.
- `RefreshMembershipStatusTest` — each failing load returns `FAILURE` and names itself; a throwing member is reported while the rest still update; unknown benefits fail the command; an unlinked member does not; the report cap holds with 10 failing members.
- `PatreonServiceTest` — known titles apply; an unknown title skips the member instead of throwing; a renamed title does not revoke the benefit it was renamed from.


Wotuu (commented):
Ok

PR #3751, opened 2026-07-29
#3737 Fix crashed mdt:importmapping leaving a silent-no-op mapping version

### LINKED ISSUE #3737: A crashed mdt:importmapping leaves an empty mapping version whose hash makes every retry a silent no-op
:robot:

Found while doing the Midnight S2 import dry run (#3734), where it bit three times in a row and
took a while to recognise.

## Problem

`mdt:importmapping` records the new mapping version's `mdt_mapping_hash` **up front**, when the
mapping version row is created — before any enemies, packs, patrols or map POIs have been imported.
If the import then throws partway through, the half-built mapping version is left behind as the
dungeon's **current** one, already carrying the hash of the MDT data it failed to import.

The next run compares that stored hash against the freshly computed one, finds them equal, logs
`importDungeonMappingVersionFromMDTNoChangeDetected` and **exits 0 without doing anything**.

So a failed import degrades into a silent no-op on every subsequent attempt, while the dungeon is
left with a mapping version containing **zero enemies**. On the site the dungeon simply looks empty.

## Reproduction

Any import that throws after `createNewMappingVersionFromMDTMapping()` will do. In #3734 it was a
`LazyLoadingViolationException` on `npcHealths` and then an `InvalidArgumentException` from
`CoordinatesService` on a facade-floor coordinate:

```
php artisan mdt:importmapping kingsrest retail     # throws, exit 1, empty mapping version created
php artisan mdt:importmapping kingsrest retail     # "No change detected", exit 0, still empty
php artisan mdt:importmapping kingsrest retail     # ... forever
```

Recovery is manual and non-obvious: delete the empty mapping version(s), then re-run with `--force`.

```php
$d = Dungeon::where('key', 'kingsrest')->first();
$mv = $d->mappingVersions()->latest('id')->first();
$mv->delete();   // repeat while the latest version has 0 enemies
```

This is worse than it sounds on a **shared dev database** — a worktree running the import breaks
the dungeon for the main checkout's site too, with no obvious cause.

## Relevant code

- `app/Service/MDT/MDTMappingImportService.php:81-128` — `importMappingVersionFromMDT()`. The hash
  is passed into `createNewMappingVersionFromMDTMapping()` at `:89`, i.e. before the `try` block at
  `:94` that does all the actual importing. The `finally` at `:118` only logs; nothing rolls back.
- `app/Service/Mapping/MappingService.php:62-79` — `createNewMappingVersionFromMDTMapping()` writes
  the row (and the hash) immediately.
- The no-change branch is `MDTMappingImportService.php:123-127`.

## Suggested fix

Any of these would remove the trap; the first is the most robust:

1. **Wrap the import in a transaction** so a failure rolls the new mapping version back entirely.
   The dungeon then keeps its previous working mapping version and the next run genuinely retries.
2. **Write the hash only on success** — create the mapping version with a null
   `mdt_mapping_hash` and set it after `importMapPOIs()` returns. A crashed import then leaves a
   version whose hash does not match, so the next run re-imports instead of skipping.
3. At minimum, **delete the new mapping version in a `catch`** before rethrowing.

Option 2 is the smallest change and also makes the failure self-healing without `--force`.

Worth adding a regression test that forces a throw mid-import and asserts the dungeon's current
mapping version is unchanged afterwards.

Wotuu:
:robot:

Closes #3737

## Problem

`createNewMappingVersionFromMDTMapping()` recorded the new mapping version's `mdt_mapping_hash`
up front, before any enemies/packs/patrols/POIs were imported. A crash partway through then left
the half-built mapping version behind as the dungeon's **current** one, already carrying the hash
of the MDT data it failed to import - so every subsequent run compared hashes, found them equal,
logged `importDungeonMappingVersionFromMDTNoChangeDetected` and exited without doing anything,
while the dungeon rendered with zero enemies.

## Fix

Implements suggestion 2 + 3 from the issue (smallest change, self-healing without `--force`, and
doesn't leave dead mapping versions littered behind on repeated crashes):

- `createNewMappingVersionFromMDTMapping()` always creates the mapping version with a `null`
  `mdt_mapping_hash` now (the `$hash` parameter was removed - it was only ever passed by
  `importMappingVersionFromMDT()`, which no longer computes it up front either).
- `importMappingVersionFromMDT()` only stamps the real hash once every import step
  (`copyEnemyForcesCheckpointsToMappingVersion` through `importMapPOIs`) has succeeded.
- If anything throws, the partially-imported mapping version is deleted (its `deleting` hook
  cascades away whatever enemies/packs/patrols/POIs made it in) before the exception is
  rethrown, so the dungeon's current mapping version reverts to its previous, fully-working one
  and the next run genuinely retries.

I did **not** wrap the whole import in a DB transaction (the issue's suggestion 1, "most
robust"): `importNpcsDataFromMDT()`/`importSpellDataFromMDT()` also write shared,
dungeon-independent rows (`Npc`, `NpcHealth`, `NpcSpell`, `Spell`, `SpellDungeon`, ...) - a
transaction wrap would roll those back too on any failure, which is a bigger behavior change
than this issue calls for, on top of holding a long write transaction against the shared dev DB.

## Test plan

- `MDTMappingImportCrashRecoveryTest` (two tests):
  - Crash path: decorates `MappingServiceInterface` (a method parameter of
    `importMappingVersionFromMDT()`, not a constructor dep) to throw from
    `copyEnemyForcesCheckpointsToMappingVersion()` - the first call inside the try block, so no
    shared Npc/Spell data is touched - and asserts the dungeon's current mapping version and its
    full set of mapping version rows are unchanged afterward. Verified this fails without the
    `catch { delete; rethrow; }` fix (the stale mapping version stays "current").
  - Success path: runs the real import through to completion and asserts the new mapping
    version's `mdt_mapping_hash` matches the freshly computed one - added after independent
    review pointed out the crash-only test can't distinguish "hash written after success" from
    "hash never written at all".
- Full `MDT` + `MappingVersion` test groups green (87 tests).

## Independent review

Reviewed by a fresh-context agent, which verified the fix and cascade-delete mechanism by tracing
`MappingVersion`'s `deleting` boot hook and running the suite live, then raised one gap and a few
lower-severity edge cases:

- **Fixed**: no test proved the hash actually gets written on a *successful* import (see the
  second test added above).
- **Not addressed here, noted as out of scope**:
  - `createNewMappingVersionFromMDTMapping()` itself (called *before* `importMappingVersionFromMDT()`'s
    try block starts) can throw while cloning the previous mapping version's contents
    (`copyMappingVersionContentsToDungeon()`), leaving a dead partial version with no handle for
    the caller to clean up. The null-hash default still prevents a *permanent* no-op (the next run
    sees the mismatch and retries), it just doesn't self-clean until then. A real fix belongs
    inside `MappingService::createNewMappingVersionFromMDTMapping()` itself, not in widening this
    PR's try/catch - filed as #3752 rather than growing this change's scope.
  - `importEnemyPatrols()` creates two `Polyline` rows before creating the owning `EnemyPatrol`
    that links them; a throw in that narrow window leaks two rows the cascade delete can't reach
    (`Polyline` carries no `mapping_version_id`). Pre-existing gap, not a regression, narrow
    window.
  - If `$newMappingVersion->delete()` itself throws inside the catch block, the original
    exception is lost. Cheap hardening, not required for this fix to work correctly in practice.

### DIFF
--- a/app/Service/MDT/MDTMappingImportService.php
+++ b/app/Service/MDT/MDTMappingImportService.php
@@ -37,6 +37,7 @@ use Illuminate\Database\UniqueConstraintViolationException;
 use Illuminate\Support\Collection;
 use Psr\SimpleCache\InvalidArgumentException;
 use Str;
+use Throwable;
 
 class MDTMappingImportService implements MDTMappingImportServiceInterface
 {
@@ -86,7 +87,7 @@ class MDTMappingImportService implements MDTMappingImportServiceInterface
         if ($forceImport || $currentMappingVersion->mdt_mapping_hash !== $latestMdtMappingHash) {
             $this->log->importMappingVersionFromMDTMappingChanged($currentMappingVersion->mdt_mapping_hash, $latestMdtMappingHash);
 
-            $newMappingVersion = $mappingService->createNewMappingVersionFromMDTMapping($dungeon, $gameVersion, $this->getMDTMappingHash($dungeon));
+            $newMappingVersion = $mappingService->createNewMappingVersionFromMDTMapping($dungeon, $gameVersion);
             $this->log->importMappingVersionFromMDTCreateMappingVersion($newMappingVersion->version, $newMappingVersion->id);
 
             $mdtDungeon = new MDTDungeon($this->cacheService, $this->coordinatesService, $dungeon);
@@ -115,6 +116,19 @@ class MDTMappingImportService implements MDTMappingImportServiceInterface
                 $this->importEnemyPacks($newMappingVersion, $mdtDungeon, $dungeon, $enemies);
                 $this->importEnemyPatrols($currentMappingVersion, $newMappingVersion, $mdtDungeon, $dungeon, $enemies);
                 $this->importMapPOIs($currentMappingVersion, $newMappingVersion, $mdtDungeon, $dungeon);
+
+                // Only stamp the hash once every step above actually succeeded (#3737) - a mapping version
+                // recorded with the hash of MDT data it never finished importing would look up to date to
+                // every subsequent run and be skipped forever ("no change detected"), leaving the dungeon
+                // stuck on a half-built "current" mapping version.
+                $newMappingVersion->update(['mdt_mapping_hash' => $latestMdtMappingHash]);
+            } catch (Throwable $throwable) {
+                // The partially-imported mapping version is not safe to leave behind as the dungeon's
+                // current one - cascading delete removes whatever enemies/packs/etc. made it in before the
+                // failure, and the next run retries against the still-null hash instead of silently no-oping.
+                $newMappingVersion->delete();
+
+                throw $throwable;
             } finally {
                 $this->log->importMappingVersionFromMDTEnd();
             }
--- a/app/Service/MDT/MDTMappingImportService.php
+++ b/app/Service/MDT/MDTMappingImportService.php
@@ -126,6 +126,8 @@ class MDTMappingImportService implements MDTMappingImportServiceInterface
                 // The partially-imported mapping version is not safe to leave behind as the dungeon's
                 // current one - cascading delete removes whatever enemies/packs/etc. made it in before the
                 // failure, and the next run retries against the still-null hash instead of silently no-oping.
+                $this->log->importMappingVersionFromMDTDeletePartialMappingVersion($newMappingVersion->version, $newMappingVersion->id, $throwable);
+
                 $newMappingVersion->delete();
 
                 throw $throwable;
--- a/tests/Feature/App/Service/MDT/MDTMappingImportCrashRecoveryTest.php
+++ b/tests/Feature/App/Service/MDT/MDTMappingImportCrashRecoveryTest.php
@@ -1,0 +1,126 @@
+<?php
+
+namespace Tests\Feature\App\Service\MDT;
+
+use App\Models\Dungeon;
+use App\Models\GameVersion\GameVersion;
+use App\Models\Mapping\MappingVersion;
+use App\Service\Mapping\MappingServiceInterface;
+use App\Service\MDT\MDTMappingImportServiceInterface;
+use PHPUnit\Framework\Attributes\Group;
+use PHPUnit\Framework\Attributes\Test;
+use RuntimeException;
+use Tests\TestCases\PublicTestCase;
+
+/**
+ * #3737: `mdt:importmapping` used to record the new mapping version's `mdt_mapping_hash` up front, before any
+ * of the actual import steps ran. A crash partway through then left the half-built mapping version behind as
+ * the dungeon's current one, already carrying the hash of the MDT data it failed to import - so every
+ * subsequent run compared hashes, found them equal, and silently no-op'd forever while the dungeon rendered
+ * with zero enemies.
+ *
+ * The hash is now written only once every import step has succeeded, and a crash deletes the partial mapping
+ * version instead of leaving it behind - so the dungeon's current mapping version is left completely
+ * unchanged and the next run genuinely retries.
+ *
+ * The forced failure is injected via a decorator around MappingServiceInterface (a method parameter of
+ * importMappingVersionFromMDT(), not a constructor dependency) that throws from
+ * copyEnemyForcesCheckpointsToMappingVersion() - the first call inside the try block - so nothing downstream
+ * of it runs: no shared Npc/Spell rows are touched, and the only residue is the mapping version row itself.
+ */
+#[Group('MDT')]
+#[Group('MappingVersion')]
+final class MDTMappingImportCrashRecoveryTest extends PublicTestCase
+{
+    #[Test]
+    public function importMappingVersionFromMDT_givenImportThrowsPartway_leavesDungeonsCurrentMappingVersionUnchanged(): void
+    {
+        // Arrange
+        /** @var Dungeon $dungeon */
+        $dungeon     = Dungeon::query()->where('key', 'throne_of_the_tides')->firstOrFail();
+        $gameVersion = GameVersion::query()->findOrFail($dungeon->getCurrentMappingVersion()->game_version_id);
+
+        $mappingVersionIdsBefore       = MappingVersion::query()->where('dungeon_id', $dungeon->id)->pluck('id')->sort()->values()->all();
+        $currentMappingVersionIdBefore = $dungeon->getCurrentMappingVersion($gameVersion)->id;
+
+        $realMappingService   = $this->app->make(MappingServiceInterface::class);
+        $mappingImportService = $this->app->make(MDTMappingImportServiceInterface::class);
+
+        $throwingMappingService = new class ($realMappingService) implements MappingServiceInterface {
+            public function __construct(private readonly MappingServiceInterface $real)
+            {
+            }
+
+            public function createNewBareMappingVersion(Dungeon $dungeon, GameVersion $gameVersion): MappingVersion
+            {
+                return $this->real->createNewBareMappingVersion($dungeon, $gameVersion);
+            }
+
+            public function createNewMappingVersionFromPreviousMapping(Dungeon $dungeon, GameVersion $gameVersion): MappingVersion
+            {
+                return $this->real->createNewMappingVersionFromPreviousMapping($dungeon, $gameVersion);
+            }
+
+            public function createNewMappingVersionFromMDTMapping(Dungeon $dungeon, ?GameVersion $gameVersion): MappingVersion
+            {
+                return $this->real->createNewMappingVersionFromMDTMapping($dungeon, $gameVersion);
+            }
+
+            public function getMappingVersionForMdtAddonVersion(Dungeon $dungeon, ?int $addonVersion, ?GameVersion $gameVersion = null): ?MappingVersion
+            {
+                return $this->real->getMappingVersionForMdtAddonVersion($dungeon, $addonVersion, $gameVersion);
+            }
+
+            public function copyMappingVersionToDungeon(MappingVersion $sourceMappingVersion, Dungeon $dungeon): MappingVersion
+            {
+                return $this->real->copyMappingVersionToDungeon($sourceMappingVersion, $dungeon);
+            }
+
+            public function copyMappingVersionContentsToDungeon(
+                MappingVersion $sourceMappingVersion,
+                MappingVersion $targetMappingVersion,
+            ): MappingVersion {
+                return $this->real->copyMappingVersionContentsToDungeon($sourceMappingVersion, $targetMappingVersion);
+            }
+
+            public function copyEnemyForcesCheckpointsToMappingVersion(
+                MappingVersion $sourceMappingVersion,
+                MappingVersion $targetMappingVersion,
+            ): array {
+                throw new RuntimeException('Forced failure to test crash recovery (#3737)');
+            }
+        };
+
+        try {
+            // Act
+            try {
+                $mappingImportService->importMappingVersionFromMDT($throwingMappingService, $dungeon, $gameVersion, true);
+
+                $this->fail('Expected the import to throw.');
+            } catch (RuntimeException $exception) {
+                $this->assertSame('Forced failure to test crash recovery (#3737)', $exception->getMessage());
+            }
+
+            // Assert
+            /** @var Dungeon $freshDungeon */
+            $freshDungeon = Dungeon::query()->findOrFail($dungeon->id);
+
+            $this->assertSame(
+                $currentMappingVersionIdBefore,
+                $freshDungeon->getCurrentMappingVersion($gameVersion)->id,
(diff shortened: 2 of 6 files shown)


tests/Feature/App/Service/MDT/MDTMappingImportCrashRecoveryTest.php
```diff
@@ -0,0 +1,172 @@
...
+        // importNpcsDataFromMDT() does not eager-load Npc::npcHealths before reading it - a real,
+        // pre-existing bug unrelated to #3737 (it is one of the crash causes #3737 cites as an example
+        // trigger, not something this PR fixes). In production that only logs and lazy-loads anyway
+        // (AppServiceProvider::boot()); in dev/test it throws. Match production's tolerance here so this
+        // test can exercise a genuinely successful import instead of tripping over that unrelated bug.
+        Model::preventLazyLoading(false);
+
+        try {
+            // Act
+            $newMappingVersion = $mappingImportService->importMappingVersionFromMDT($mappingService, $dungeon, $gameVersion, true);
```
Wotuu, line 156: :robot: **The success-path test runs a real, full MDT import against the shared persistent test DB but only cleans up the mapping version.**

`importMappingVersionFromMDT()` does more than build the new mapping version. Before it ever reaches `importEnemies()` it calls:

- `importDungeon()` -> `$dungeon->update(['mdt_id' => ...])` (a `dungeons` row)
- `importNpcs()` -> `importNpcsDataFromMDT()`, which `save()`s every MDT NPC (`name`, `level`, `display_id`, `encounter_id`, `mdt_scale`, `npc_type_id`, `truesight`), upserts an `NpcHealth` row per NPC, and mass-inserts `NpcSpell` / `NpcDungeon` rows. For any genuinely new NPC it also calls `createNpcEnemyForcesForExistingMappingVersions()`, which writes `NpcEnemyForces` rows into **every historical mapping version of the dungeon**.

None of that is reverted - the `finally` only does `$newMappingVersion?->delete()`, and the cascade in `MappingVersion::deleting` reaches only rows carrying this mapping version's id. The `writing-tests` convention is explicit that the test DB is persistent, non-transactional and pre-seeded, and that every record a test creates must be cleaned up.

The sibling `MDTMappingImportEnemyForcesCheckpointTest` in this same directory documents avoiding exactly this:

> `importEnemies()` is exercised directly (with a stubbed `MDTDungeon`) rather than through `importMappingVersionFromMDT()`: a real import parses MDT's Lua and rewrites shared NPC/dungeon rows of the seeded test database [...]

Worth noting green CI does not clear this: CI seeds a fresh DB each run, so the leak is invisible there. On the shared dev DB it is not - the first run permanently shifts the baseline (new `NpcSpell` rows, NPC attributes overwritten with the vendored MDT package's values) and every run after that looks idempotent, which is precisely the failure mode #3737 itself calls out as "worse than it sounds on a shared dev database".

Two acceptable resolutions, either is fine:

1. Snapshot and restore what the import touches for this dungeon in the `finally` (dungeon `mdt_id`, the affected `Npc`/`NpcHealth` rows, and delete any `NpcSpell`/`NpcDungeon`/`NpcEnemyForces` rows created during the test).
2. Or follow the sibling test's approach and assert the hash write without driving the full pipeline - e.g. have the `MappingServiceInterface` decorator capture the mapping version at creation (asserting `mdt_mapping_hash === null` there), and drive the remaining steps directly rather than through `importMappingVersionFromMDT()`.

tests/Feature/App/Service/MDT/MDTMappingImportCrashRecoveryTest.php
```diff
@@ -0,0 +1,172 @@
...
+            $this->assertSame(
+                $mappingVersionIdsBefore,
+                MappingVersion::query()->where('dungeon_id', $dungeon->id)->pluck('id')->sort()->values()->all(),
+                'A crashed import must not leave a half-built mapping version behind.',
+            );
+        } finally {
+            MappingVersion::query()
+                ->where('dungeon_id', $dungeon->id)
+                ->whereNotIn('id', $mappingVersionIdsBefore)
+                ->delete();
```
Wotuu: :robot: This safety net is a query-builder mass delete, which **bypasses `MappingVersion`'s `deleting` model event** (Eloquent mass deletes never fire model events).

That matters more than it looks, because the net only ever does anything in the one case it exists for: the fix regressing so the `catch` no longer deletes the partial version. And at the moment `copyEnemyForcesCheckpointsToMappingVersion()` throws, the new mapping version is already far from empty - `createNewMappingVersionFromMDTMapping()` runs `copyMappingVersionContentsToDungeon()` *before* the try block, so it already holds cloned `dungeonFloorSwitchMarkers`, `mapIcons`, `mountableAreas`, `floorUnions` and `floorUnionAreas`.

So in the exact scenario this cleanup guards against, it deletes the `mapping_versions` row and orphans all of those child rows in the shared DB, pointing at a `mapping_version_id` that no longer exists.

Fetching and deleting through the model instead makes the net actually do what it promises:

```php
MappingVersion::query()
    ->where('dungeon_id', $dungeon->id)
    ->whereNotIn('id', $mappingVersionIdsBefore)
    ->get()
    ->each(fn(MappingVersion $mappingVersion) => $mappingVersion->delete());
```

app/Service/MDT/MDTMappingImportService.php
```diff
@@ -115,6 +116,19 @@ public function importMappingVersionFromMDT(
...
+                // Only stamp the hash once every step above actually succeeded (#3737) - a mapping version
+                // recorded with the hash of MDT data it never finished importing would look up to date to
+                // every subsequent run and be skipped forever ("no change detected"), leaving the dungeon
+                // stuck on a half-built "current" mapping version.
+                $newMappingVersion->update(['mdt_mapping_hash' => $latestMdtMappingHash]);
+            } catch (Throwable $throwable) {
+                // The partially-imported mapping version is not safe to leave behind as the dungeon's
+                // current one - cascading delete removes whatever enemies/packs/etc. made it in before the
+                // failure, and the next run retries against the still-null hash instead of silently no-oping.
+                $newMappingVersion->delete();
```
Wotuu, line 131: :robot: Minor: this delete is silent. Every other step of this service reports through `$this->log` (the logging interface has ~66 methods, including `importMappingVersionFromMDTCreateMappingVersion($version, $id)` for the counterpart creation a few lines up).

Operationally, "the mapping version that was just logged as created has vanished" is exactly the kind of thing you want in the structured log when triaging a failed `mdt:importmapping` - especially since the whole point of #3737 is that this failure mode was hard to recognise. A matching `importMappingVersionFromMDTDeletePartialMappingVersion($newMappingVersion->version, $newMappingVersion->id)` line before the delete would keep the pairing symmetrical.

app/Service/MDT/MDTMappingImportService.php
```diff
@@ -115,6 +116,19 @@ public function importMappingVersionFromMDT(
                 $this->importEnemyPacks($newMappingVersion, $mdtDungeon, $dungeon, $enemies);
                 $this->importEnemyPatrols($currentMappingVersion, $newMappingVersion, $mdtDungeon, $dungeon, $enemies);
                 $this->importMapPOIs($currentMappingVersion, $newMappingVersion, $mdtDungeon, $dungeon);
+
+                // Only stamp the hash once every step above actually succeeded (#3737) - a mapping version
+                // recorded with the hash of MDT data it never finished importing would look up to date to
+                // every subsequent run and be skipped forever ("no change detected"), leaving the dungeon
+                // stuck on a half-built "current" mapping version.
+                $newMappingVersion->update(['mdt_mapping_hash' => $latestMdtMappingHash]);
```
Wotuu, line 124: :robot: Not a defect in this diff, and I would not widen the PR for it - but worth capturing as a follow-up next to #3752, because it leaves the door to #3737's exact symptom open.

Reaching this line means "no step threw", which is weaker than "the import completed". Several steps swallow their own failures and continue:

- `importEnemies()` wraps each `$mdtEnemy->save()` in `catch (Exception $exception) { $this->log->importEnemiesSaveNewEnemyException($exception); }` inside the loop, so individual enemies can silently fail to persist.
- `importNpcs()` similarly swallows per-NPC exceptions around `NpcEnemyForces::create()`.
- `importNpcsDataFromMDT()` does the same around each NPC `save()`.

So an import can lose an arbitrary number of enemies (or all of them, if the failure is systematic - a floor lookup, a coordinate conversion, a constraint), never throw, reach this `update()`, stamp the hash, and leave an under-populated mapping version that every subsequent run skips as "no change detected". That is the same trap #3737 describes, just entered through a different door.

A cheap guard that fits the current shape would be a sanity check before stamping - e.g. refuse the hash (or throw) when `$newMappingVersion->enemies()->count()` is 0 while MDT reported clones, or when the saved-enemy count is materially below what `getClonesAsEnemies()` returned.

Wotuu:

tests/Feature/App/Service/MDT/MDTMappingImportCrashRecoveryTest.php
Wotuu: :robot: Fixed. `importNpcsDataFromMDT()` is now stubbed out via a partial mock instead of run for real - it saves NPC/NpcHealth/NpcSpell/NpcDungeon rows and writes NpcEnemyForces into historical mapping versions, none of which a mapping-version delete can revert. `throne_of_the_tides` already has every NPC MDT reports for it seeded, so `importNpcs()` still finds a real NPC for every enemy and the rest of the pipeline (importEnemies/importEnemyPacks/importEnemyPatrols/importMapPOIs, and the hash write itself) still runs for real, unstubbed. The one remaining side effect outside the new mapping version, `dungeons.mdt_id` (written by `importDungeon()`), is now captured before the act and restored in `finally`. Verified: 83/83 tests in the MDT group still pass.

tests/Feature/App/Service/MDT/MDTMappingImportCrashRecoveryTest.php
Wotuu: :robot: Fixed. Switched to `MappingVersion::destroy()` (loads each matched row and calls `->delete()` on it individually) instead of the query-builder mass delete, so `MappingVersion`'s `deleting` model event still fires and cascades to any cloned child rows the half-built version holds if this safety net is ever actually hit.

app/Service/MDT/MDTMappingImportService.php
Wotuu: :robot: Fixed. Added `importMappingVersionFromMDTDeletePartialMappingVersion()` to the logging interface/implementation (logged at error level, matching the other failure-path entries like `importDungeonFailed()`) and call it right before the delete, passing the version, id, and the caught throwable.

app/Service/MDT/MDTMappingImportService.php
Wotuu: :robot: Agreed, no change needed in this PR. Filed as a follow-up: https://github.com/RaiderIO/keystone.guru/issues/3755 (correction: my first reply here mistakenly linked #3752, which is a different, unrelated follow-up about a crash-before-the-try-block gap - #3755 is the correct one for the swallowed-per-item-exception concern raised in this thread).

tests/Feature/App/Service/MDT/MDTMappingImportCrashRecoveryTest.php
```diff
@@ -0,0 +1,199 @@
...
+                return $this->real->copyMappingVersionContentsToDungeon($sourceMappingVersion, $targetMappingVersion);
+            }
+
+            public function copyEnemyForcesCheckpointsToMappingVersion(
+                MappingVersion $sourceMappingVersion,
+                MappingVersion $targetMappingVersion,
+            ): array {
+                throw new RuntimeException('Forced failure to test crash recovery (#3737)');
+            }
+        };
```
Wotuu: Is it not possible to use a PartialMock for this? Idk it's kinda strange to create an anonymous class like this and pass everything through to a real implementation except one throwing function. Partial mock should do this for you?

Wotuu (commented):
One thing

tests/Feature/App/Service/MDT/MDTMappingImportCrashRecoveryTest.php
Wotuu: :robot: Good call — this repo has no Mockery in the test suite, so PHPUnit's native partial mock (`createPartialMock`/`getMockBuilder(...)->onlyMethods([...])`) is the local equivalent, already used elsewhere in this file and in `GetDungeonContextTest`. Replaced the anonymous class with a partial mock of the concrete `MappingService`: only `copyEnemyForcesCheckpointsToMappingVersion()` is mocked to throw, every other method now falls through to the real implementation instead of being hand-wired.

Sanity-checked the refactor did not lose coverage by temporarily reverting the production delete-on-crash fix and confirming the test then fails; restored it afterwards. Pushed in 940904fb5.

PR #3762, opened 2026-07-30
Scope the MDT import pipeline's predecessor lookup to the target game version

### LINKED ISSUE #3757: MDT import pipeline can silently clone content/version from the wrong game version (follow-up to #3720)
:robot: Surfaced during independent review of #3749 (the #3720 game_version_id-scoping audit).

## Background

#3720/#3749 fixed several places where a mapping-version action computed a new mapping version's
`game_version_id`/`version`, or picked a "predecessor" to clone from, without scoping by
`game_version_id` - `version` is only unique per game version, so an unscoped lookup can silently
pick a completely different game version's data.

The review found the same bug class in the **primary, scheduled MDT import pipeline**, which is
more heavily used and higher-risk than anything else fixed in #3749 (mostly rarely-used admin/dev
tools), so it's being tracked separately rather than folded into that PR under time pressure.

## The bug

`MappingService::createNewMappingVersionFromMDTMapping()` (`app/Service/Mapping/MappingService.php:62-82`):

```php
$currentMappingVersion = $dungeon->getCurrentMappingVersion($gameVersion);
...
'game_version_id' => $currentMappingVersion?->game_version_id ?? GameVersion::ALL[GameVersion::GAME_VERSION_RETAIL],
'version'         => ($currentMappingVersion?->version ?? 0) + 1,
```

`Dungeon::getCurrentMappingVersion($gameVersion)` only scopes to `$gameVersion` on its *first*
attempt - if the dungeon has **zero** existing mapping versions for that specific `$gameVersion`
(e.g. the first-ever MDT import of a new game version, such as Legion Remix, for a dungeon that
already has retail mappings - this is exactly how the original #3720 issue's own example data
came to exist), it silently falls through to the ambient/session-based fallback chain and returns
a **different game version's** mapping version instead of `null`. Reproduced empirically against
the seeded DB:

```
$dungeon->getCurrentMappingVersion($gameVersionWithNoExistingMapping)
  => resolved to game_version_id=1 (retail), version=6   // wrong: should be null
```

So the new mapping version gets mislabeled with the wrong `game_version_id` and its `version` is
computed from an unrelated game version's counter.

## Why this is bigger than a one-line fix (unlike the other #3749 fixes)

The root cause actually sits one level up, in the caller:
`MDTMappingImportService::importMappingVersionFromMDT()` (`app/Service/MDT/MDTMappingImportService.php:75-128`):

```php
$currentMappingVersion = $dungeon->getCurrentMappingVersion($gameVersion);
if ($forceImport || $currentMappingVersion->mdt_mapping_hash !== $latestMdtMappingHash) {
    ...
    $newMappingVersion = $mappingService->createNewMappingVersionFromMDTMapping($dungeon, $gameVersion, ...);
    ...
    $enemyForcesCheckpointIdMapping = $mappingService->copyEnemyForcesCheckpointsToMappingVersion($currentMappingVersion, $newMappingVersion);
    ...
    $enemies = $this->importEnemies($currentMappingVersion, $newMappingVersion, ...);
    ...
    $this->importEnemyPatrols($currentMappingVersion, $newMappingVersion, ...);
    $this->importMapPOIs($currentMappingVersion, $newMappingVersion, ...);
```

This SAME `$currentMappingVersion` - resolved once at the top, with the identical unscoped-fallback
defect - is what checkpoints/enemies/patrols/map-icons get cloned *from* for the new mapping
version. A minimal fix only inside `createNewMappingVersionFromMDTMapping()` would correctly tag
the new mapping version's own `game_version_id`/`version`, but its *content* would still silently
be cloned from the wrong game version's predecessor - arguably a more confusing bug than today's
(correct label, wrong content) rather than less.

A real fix needs to:
1. Scope `$currentMappingVersion` in `importMappingVersionFromMDT()` via
   `$dungeon->getCurrentMappingVersionForGameVersion($gameVersion)` instead of the ambient-fallback
   `getCurrentMappingVersion($gameVersion)`.
2. Handle the resulting `null` (a genuinely first-ever import for that game version) as a real,
   valid case - not an error - across every downstream call site that currently assumes a non-null
   `MappingVersion` to clone from: `copyEnemyForcesCheckpointsToMappingVersion()`, `importEnemies()`,
   `importEnemyPatrols()`, `importMapPOIs()`, and `MappingService::createNewMappingVersionFromMDTMapping()`
   itself. In the null case, there is nothing to carry over - checkpoints/enemies/patrols/POIs
   should just be imported fresh from MDT with no predecessor-derived carryover, which is already
   what most of this code degrades to gracefully when its lookups find nothing (needs verification
   per call site, not assumed).

This touches ~5 method signatures in the codebase's primary, scheduled MDT import pipeline
(`mdt:importmapping`), which is materially higher-traffic and higher-risk than the admin/dev tools
already fixed in #3749 - hence tracking it separately rather than rushing it into that PR.

## Suggested approach

- Start from `MDTMappingImportService::importMappingVersionFromMDT()` and trace every use of
  `$currentMappingVersion` down through `createNewMappingVersionFromMDTMapping()`,
  `copyEnemyForcesCheckpointsToMappingVersion()`, `importEnemies()`, `importEnemyPatrols()`, and
  `importMapPOIs()`, converting each to tolerate (and correctly no-op on) a null predecessor.
- Add a regression test that imports MDT mapping for a dungeon + game version combination that has
  zero existing mapping versions, and asserts the new mapping version's `game_version_id` matches
  the requested game version (not an ambient/unrelated one) and that no enemies/checkpoints/patrols
  leaked in from a different game version's mapping.
- Given the real-world trigger condition (importing a dungeon's mapping for a NEW game version it
  doesn't have yet, e.g. bootstrapping Legion Remix/classic-era support for an existing retail
  dungeon) is a recognized, periodically-occurring operational task, this is worth prioritizing
  rather than leaving indefinitely - unlike #3754, which is dormant until the combat-log-streamer
  protocol changes.

Not confirmed as having caused live data corruption yet - same caveat as #3720 itself - but the
trigger condition is realistic and has plausibly already occurred historically (see the original
#3720 issue's own example data, which shows a dungeon with both retail and Legion Remix mapping
versions already coexisting).

Wotuu:
:robot: Closes #3757

## Summary

`MDTMappingImportService::importMappingVersionFromMDT()` resolved its "predecessor"
mapping version via `Dungeon::getCurrentMappingVersion()`, which silently falls back to
an ambient/unrelated game version's mapping version whenever the dungeon has zero
mapping versions for the game version actually being imported (e.g. a dungeon's
first-ever import for a newly-added game version, such as Legion Remix). That mistagged
the new mapping version's `game_version_id`/`version` **and** cloned
checkpoints/enemies/patrols/POIs from the wrong game version's data - a follow-up to
#3720/#3749, deliberately deferred out of that MR because it's the primary, scheduled
import pipeline (`mdt:importmapping`) rather than an admin/dev tool.

- Switched to the already-scoped `Dungeon::getCurrentMappingVersionForGameVersion()`.
- Treated a `null` predecessor as a real, valid "nothing to carry over from" case across
  every downstream call site: `MappingService::createNewMappingVersionFromMDTMapping()`
  (now takes the `GameVersion` and predecessor as separate params - `game_version_id`
  always comes from the former, never derived from the latter),
  `copyEnemyForcesCheckpointsToMappingVersion()`, `importEnemies()`,
  `importEnemyPatrols()`, and `importMapPOIs()`.
- **Facade geometry fix (found by independent review):** an earlier draft of this PR left
  `facade_enabled = false` and zero `FloorUnion`s on the null-predecessor path, since
  those were only ever cloned via `copyMappingVersionContentsToDungeon()`, which that
  path skips. `CoordinatesService::convertFacadeMapLocationToMapLocation()` degrades to
  the identity transform for a non-facade mapping version, so every enemy/patrol/POI
  imported for a facade dungeon's first-ever game-version import would have been silently
  persisted with the facade pseudo-floor's `floor_id` instead of the real one - confirmed
  empirically against a real dungeon before the fix. `facade_enabled`/`FloorUnion`s are
  the dungeon's *physical floor layout*, not game-version-specific curated content, so
  they're now inherited from the dungeon's current mapping version in **any** game
  version even with no same-game-version predecessor to clone the rest of the content
  from.
- Verified `enemy_forces_required`/`_teeming`/`_shrouded`/`_shrouded_zul_gamux` are
  **not** affected despite initially looking like another gap: `importDungeon()` and
  `importNpcs()` unconditionally recompute all four from the current MDT source data on
  every import, regardless of predecessor state.
- Remaining known, non-blocking gap: `MountableArea`s (mount-speed-override zones) and
  `timer_max_seconds` are still not inherited on the null-predecessor path - MDT carries
  no source data for either, and only `copyMappingVersionContentsToDungeon()` clones
  them, same root cause as the facade issue above but lower severity (missing
  nice-to-have data, not silently-wrong persisted data) - left for a manual follow-up
  re-add/re-import rather than expanding this fix's scope further.

## Test plan

- [x] Regression test (`MDTMappingImportGameVersionScopingTest`) runs the real import
  pipeline (real Lua parsing, like `MDTNpcMappingCoverageTest`) against a **facade**
  dungeon that already has a retail mapping version with a checkpoint (members included)
  but zero mapping versions for a second game version, and asserts: the new mapping
  version is tagged with the requested game version and starts at `version = 1`, has
  zero cloned checkpoints, the unrelated retail mapping version is left untouched,
  `facade_enabled`/`FloorUnion`s are inherited, and no enemy lands on the facade
  pseudo-floor.
- [x] Verified the test actually guards both fixes: reverted just the
  `getCurrentMappingVersionForGameVersion()` call back to the old
  `getCurrentMappingVersion()` and confirmed the test fails (`version` resolved to `9`,
  continuing retail's counter, instead of `1`); separately reverted just the facade-
  geometry inheritance and confirmed the test fails on the `facade_enabled` assertion;
  restored both fixes and confirmed green again.
- [x] `tests/Feature/App/Service/MDT/` (23 tests) and
  `tests/Feature/App/Service/Mapping/` (11 tests) all pass (34 total, 84 assertions).
- [x] Independent review completed; the one substantive finding (facade geometry) is
  fixed above, and the two smaller disclosure gaps are documented above rather than
  expanded into this fix.
- [ ] No UI surface - visual verification not applicable.

### DIFF
--- a/app/Service/Mapping/MappingService.php
+++ b/app/Service/Mapping/MappingService.php
@@ -110,13 +110,8 @@ class MappingService implements MappingServiceInterface
             // part of the same import) already (re)creates the physical subset it recognizes itself from MDT's
             // own map POI data whenever $currentMappingVersion is null - see its "(#3757)" comments. Cloning
             // them here too would just create duplicate rows a few lines later in the same import. Mountable
-            // areas and dungeon floor switch markers have no such backfill to rely on: MDT has no
-            // mount-speed-zone concept at all, and per Wotuu (#3762 review) modern MDT exports only generate a
-            // combined/facade view and no longer supply the per-floor MapLink POI data importMapPOIs() used to
-            // recreate floor switch markers from - so both are PHYSICAL data (fixed dungeon terrain/geometry,
-            // like floor unions) with no other source, and need the same explicit clone floor unions get.
-            // importMapPOIs() still guards against duplicating whatever gets cloned in here (it only creates a
-            // fresh floor switch marker when the new mapping version doesn't already have one).
+            // areas have no such backfill to rely on: MDT has no mount-speed-zone concept at all, so they need
+            // the same explicit clone floor unions get.
             // The physical facade geometry resolved above still needs cloning in, though, along with
             // timer_max_seconds - it's a NOT NULL DEFAULT 0 column that several call sites divide by unguarded
             // (CombatLogEventFilter::fromHeatmapDataFilter(), CombatLogEvent::setTimeInterval()), so leaving it
@@ -124,12 +119,30 @@ class MappingService implements MappingServiceInterface
             if ($facadeSourceMappingVersion !== null) {
                 $this->cloneFloorUnionsToMappingVersion($facadeSourceMappingVersion, $newMappingVersion);
                 $this->cloneMountableAreasToMappingVersion($facadeSourceMappingVersion, $newMappingVersion);
-                $this->cloneDungeonFloorSwitchMarkersToMappingVersion($facadeSourceMappingVersion, $newMappingVersion);
 
                 $newMappingVersion->update([
                     'timer_max_seconds' => $facadeSourceMappingVersion->timer_max_seconds,
                 ]);
             }
+
+            // Dungeon floor switch markers have no MDT backfill either, but unlike FloorUnions they are NOT
+            // facade-specific geometry - every dungeon has staircases between floors, facade-enabled or not
+            // (per Wotuu, #3762 review round 2: modern MDT exports only generate a combined/facade view and no
+            // longer supply the per-floor MapLink POI data importMapPOIs() used to recreate them from, and this
+            // app still needs them for both facade mode and the per-floor "visit style" mode). So their source
+            // must NOT be gated behind $facadeSourceMappingVersion the way FloorUnion/MountableArea legitimately
+            // are - that would silently drop them for every non-facade dungeon's first-ever import, which is
+            // most of them. Fall back to the newest mapping version of ANY game version, deterministically,
+            // same rationale as the facade source above.
+            $physicalGeometrySourceMappingVersion = $facadeSourceMappingVersion
+                ?? $dungeon->mappingVersions()->orderByDesc('id')->first();
+
+            if ($physicalGeometrySourceMappingVersion !== null) {
+                $this->cloneDungeonFloorSwitchMarkersToMappingVersion($physicalGeometrySourceMappingVersion, $newMappingVersion);
+            }
+
+            // importMapPOIs() still guards against duplicating whatever gets cloned in here (it only creates a
+            // fresh floor switch marker when the new mapping version doesn't already have one).
 
             return $newMappingVersion->load([
                 'dungeonFloorSwitchMarkers',
--- a/app/Service/Mapping/MappingService.php
+++ b/app/Service/Mapping/MappingService.php
@@ -59,25 +59,81 @@ class MappingService implements MappingServiceInterface
         ]);
     }
 
-    public function createNewMappingVersionFromMDTMapping(Dungeon $dungeon, ?GameVersion $gameVersion): MappingVersion
-    {
-        /** @var MappingVersion|null $currentMappingVersion */
-        $currentMappingVersion = $dungeon->getCurrentMappingVersion($gameVersion);
-        $now                   = Carbon::now()->toDateTimeString();
+    public function createNewMappingVersionFromMDTMapping(
+        Dungeon         $dungeon,
+        GameVersion     $gameVersion,
+        ?MappingVersion $currentMappingVersion,
+    ): MappingVersion {
+        // facade_enabled is a PHYSICAL fact about the dungeon (does it have a facade floor at all), not
+        // game-version-specific curated content, and it does not need to be inherited from any mapping
+        // version - Floor::facade is the authoritative source (#3757). Resolving it from a mapping version
+        // instead risks picking up one whose facade_enabled happens to be false for that particular game
+        // version (seen in production: e.g. dungeon 12's legion-remix/dungeon 49's wotlk mapping versions),
+        // even though the dungeon itself is a facade dungeon.
+        $facadeEnabled = $dungeon->getFacadeFloor() !== null;
+
+        // FloorUnions (and areas) ARE actual rows, not a boolean, so they still need a concrete source
+        // mapping version to clone from for a genuinely first-ever import of a game version. Pin that source
+        // explicitly and deterministically to the newest facade-enabled mapping version of ANY game version -
+        // never resolve it ambiently via Dungeon::getCurrentMappingVersion(), which falls back through
+        // GameVersion::getUserOrDefaultGameVersion()/authenticated-user context that doesn't exist for this
+        // pipeline's real caller (the mdt:importmapping CLI command), and can silently source the clone from
+        // the wrong game version's (facade-disabled) mapping version (#3757).
+        $facadeSourceMappingVersion = $currentMappingVersion
+            ?? ($facadeEnabled
+                ? $dungeon->mappingVersions()->where('facade_enabled', true)->orderByDesc('id')->first()
+                : null);
+
+        $now = Carbon::now()->toDateTimeString();
         // This needs to happen quietly as to not trigger MappingVersion events defined in its class
         $id = MappingVersion::insertGetId([
-            'dungeon_id'      => $dungeon->id,
-            'game_version_id' => $currentMappingVersion?->game_version_id ?? GameVersion::ALL[GameVersion::GAME_VERSION_RETAIL], // @phpstan-ignore nullsafe.neverNull
+            'dungeon_id' => $dungeon->id,
+            // $gameVersion is the source of truth, NOT $currentMappingVersion->game_version_id - the caller is
+            // responsible for scoping $currentMappingVersion to $gameVersion, and it may legitimately be null
+            // (the dungeon's first-ever mapping version for this game version) (#3757).
+            'game_version_id' => $gameVersion->id,
             // Set only once the import actually succeeds, see importMappingVersionFromMDT() (#3737)
             'mdt_mapping_hash'  => null,
             'mdt_addon_version' => $this->mdtAddonVersionService->getCurrentAddonVersion(),
             'version'           => ($currentMappingVersion?->version ?? 0) + 1, // @phpstan-ignore nullsafe.neverNull
-            'facade_enabled'    => $currentMappingVersion?->facade_enabled ?? false, // @phpstan-ignore nullsafe.neverNull
+            'facade_enabled'    => $facadeEnabled,
             'created_at'        => $now,
             'updated_at'        => $now,
         ]);
 
         $newMappingVersion = MappingVersion::find($id);
+
+        if ($currentMappingVersion === null) {
+            // Checkpoints/enemies stay empty for a genuinely first-ever import of this game version - there is
+            // no game-version-agnostic source for curated content like that. Map icons and dungeon floor switch
+            // markers ALSO stay empty here, but not because they're curated: importMapPOIs() (called right
+            // after this method returns, as part of the same import) already (re)creates them itself from
+            // MDT's own map POI/MapLink data whenever $currentMappingVersion is null - see its "(#3757)"
+            // comments. Cloning them here too would just create duplicate rows a few lines later in the same
+            // import. Mountable areas have no such backfill - MDT has no mount-speed-zone concept at all, so
+            // nothing ever (re)creates them - which makes them PHYSICAL data (fixed dungeon terrain, like floor
+            // unions) with no other source, and they need the same explicit clone floor unions get.
+            // The physical facade geometry resolved above still needs cloning in, though, along with
+            // timer_max_seconds - it's a NOT NULL DEFAULT 0 column that several call sites divide by unguarded
+            // (CombatLogEventFilter::fromHeatmapDataFilter(), CombatLogEvent::setTimeInterval()), so leaving it
+            // at its column default here is a live DivisionByZeroError, not just missing data.
+            if ($facadeSourceMappingVersion !== null) {
+                $this->cloneFloorUnionsToMappingVersion($facadeSourceMappingVersion, $newMappingVersion);
+                $this->cloneMountableAreasToMappingVersion($facadeSourceMappingVersion, $newMappingVersion);
+
+                $newMappingVersion->update([
+                    'timer_max_seconds' => $facadeSourceMappingVersion->timer_max_seconds,
+                ]);
+            }
+
+            return $newMappingVersion->load([
+                'dungeonFloorSwitchMarkers',
+                'mapIcons',
+                'mountableAreas',
+                'floorUnions',
+                'floorUnionAreas',
+            ]);
+        }
 
         return $this->copyMappingVersionContentsToDungeon($currentMappingVersion, $newMappingVersion);
     }
@@ -210,22 +266,14 @@ class MappingService implements MappingServiceInterface
         }
 
         // Mountable Areas
-        foreach ($sourceMappingVersion->mountableAreas as $mountableArea) {
-            $mountableArea->cloneForNewMappingVersion($targetMappingVersion);
-        }
+        $this->cloneMountableAreasToMappingVersion($sourceMappingVersion, $targetMappingVersion);
 
         // Enemy Forces Checkpoints are cloned by copyEnemyForcesCheckpointsToMappingVersion() instead - the
         // caller must invoke it, because only the caller knows how to re-link the membership of the enemies
         // it clones or re-imports, and this method copies no enemies at all (#3702).
 
         // Floor Unions (and areas)
-        foreach ($sourceMappingVersion->floorUnions as $floorUnion) {
-            /** @var FloorUnion $newFloorUnion */
-            $newFloorUnion = $floorUnion->cloneForNewMappingVersion($targetMappingVersion);
-            foreach ($floorUnion->floorUnionAreas as $floorUnionArea) {
-                $floorUnionArea->cloneForNewMappingVersion($targetMappingVersion, $newFloorUnion);
-            }
-        }
+        $this->cloneFloorUnionsToMappingVersion($sourceMappingVersion, $targetMappingVersion);
 
         // Copy these properties over only if the dungeons match - doesn't make sense otherwise
         if ($sourceMappingVersion->dungeon_id === $targetMappingVersion->dungeon_id) {
@@ -252,20 +300,42 @@ class MappingService implements MappingServiceInterface
     }
 
     public function copyEnemyForcesCheckpointsToMappingVersion(
-        MappingVersion $sourceMappingVersion,
-        MappingVersion $targetMappingVersion,
+        ?MappingVersion $sourceMappingVersion,
+        MappingVersion  $targetMappingVersion,
     ): array {
         $enemyForcesCheckpointIdMapping = [];
 
-        foreach ($sourceMappingVersion->enemyForcesCheckpoints as $enemyForcesCheckpoint) {
-            /** @var EnemyForcesCheckpoint $newEnemyForcesCheckpoint */
-            $newEnemyForcesCheckpoint = $enemyForcesCheckpoint->cloneForNewMappingVersion($targetMappingVersion);
+        // No predecessor to clone from (e.g. a dungeon's first-ever mapping version for a game version) - the
+        // target simply starts with none, same as a bare mapping version (#3757).
+        if ($sourceMappingVersion !== null) {
+            foreach ($sourceMappingVersion->enemyForcesCheckpoints as $enemyForcesCheckpoint) {
+                /** @var EnemyForcesCheckpoint $newEnemyForcesCheckpoint */
+                $newEnemyForcesCheckpoint = $enemyForcesCheckpoint->cloneForNewMappingVersion($targetMappingVersion);
 
-            $enemyForcesCheckpointIdMapping[$enemyForcesCheckpoint->id] = $newEnemyForcesCheckpoint->id;
+                $enemyForcesCheckpointIdMapping[$enemyForcesCheckpoint->id] = $newEnemyForcesCheckpoint->id;
+            }
         }
 
         $targetMappingVersion->load('enemyForcesCheckpoints');
 
         return $enemyForcesCheckpointIdMapping;
+    }
+
+    private function cloneFloorUnionsToMappingVersion(MappingVersion $sourceMappingVersion, MappingVersion $targetMappingVersion): void
+    {
+        foreach ($sourceMappingVersion->floorUnions as $floorUnion) {
+            /** @var FloorUnion $newFloorUnion */
+            $newFloorUnion = $floorUnion->cloneForNewMappingVersion($targetMappingVersion);
+            foreach ($floorUnion->floorUnionAreas as $floorUnionArea) {
+                $floorUnionArea->cloneForNewMappingVersion($targetMappingVersion, $newFloorUnion);
+            }
+        }
+    }
+
+    private function cloneMountableAreasToMappingVersion(MappingVersion $sourceMappingVersion, MappingVersion $targetMappingVersion): void
+    {
+        foreach ($sourceMappingVersion->mountableAreas as $mountableArea) {
+            $mountableArea->cloneForNewMappingVersion($targetMappingVersion);
+        }
     }
 }
--- a/app/Service/Mapping/MappingService.php
+++ b/app/Service/Mapping/MappingService.php
@@ -105,14 +105,18 @@ class MappingService implements MappingServiceInterface
 
         if ($currentMappingVersion === null) {
             // Checkpoints/enemies stay empty for a genuinely first-ever import of this game version - there is
-            // no game-version-agnostic source for curated content like that. Map icons and dungeon floor switch
-            // markers ALSO stay empty here, but not because they're curated: importMapPOIs() (called right
-            // after this method returns, as part of the same import) already (re)creates them itself from
-            // MDT's own map POI/MapLink data whenever $currentMappingVersion is null - see its "(#3757)"
-            // comments. Cloning them here too would just create duplicate rows a few lines later in the same
-            // import. Mountable areas have no such backfill - MDT has no mount-speed-zone concept at all, so
-            // nothing ever (re)creates them - which makes them PHYSICAL data (fixed dungeon terrain, like floor
-            // unions) with no other source, and they need the same explicit clone floor unions get.
+            // no game-version-agnostic source for curated content like that. Map icons ALSO stay empty here,
+            // but not because they're curated: importMapPOIs() (called right after this method returns, as
+            // part of the same import) already (re)creates the physical subset it recognizes itself from MDT's
+            // own map POI data whenever $currentMappingVersion is null - see its "(#3757)" comments. Cloning
+            // them here too would just create duplicate rows a few lines later in the same import. Mountable
+            // areas and dungeon floor switch markers have no such backfill to rely on: MDT has no
+            // mount-speed-zone concept at all, and per Wotuu (#3762 review) modern MDT exports only generate a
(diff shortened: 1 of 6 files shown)


app/Service/Mapping/MappingService.php
```diff
@@ -59,25 +59,55 @@ public function createNewMappingVersionFromPreviousMapping(
...
+        ?MappingVersion $currentMappingVersion,
+        ?string         $hash,
+    ): MappingVersion {
+        // Facade geometry (facade_enabled + FloorUnions) is the dungeon's PHYSICAL floor layout, not
+        // game-version-specific curated content - it does not vary between game versions. A genuinely
+        // first-ever import for a game version must still inherit it from whatever mapping version the
+        // dungeon already has (in ANY game version), otherwise
+        // CoordinatesService::convertFacadeMapLocationToMapLocation() silently degrades to the identity
+        // transform and every object this import creates is persisted with the wrong floor_id (#3757).
+        $facadeSourceMappingVersion = $currentMappingVersion ?? $dungeon->getCurrentMappingVersion();
```
Wotuu: :robot: **The facade source is itself an ambient, unscoped lookup - and `facade_enabled` is empirically *not* uniform across game versions.**

Two separate points:

**1. The premise in the comment above doesn't hold in the current mapping data.** Per-dungeon, per-game-version newest mapping versions:

| dungeon | game version | `facade_enabled` | `FloorUnion`s |
|---|---|---|---|
| 12 `theseatofthetriumvirate` | retail (mv 798) | 1 | 1 |
| 12 `theseatofthetriumvirate` | legion-remix (mv 637) | 0 | 0 |
| 49 `pitofsaron` | retail (mv 797) | 1 | 1 |
| 49 `pitofsaron` | wotlk (mv 49) | 0 | 0 |

Both of those dungeons *do* have a `Floor` with `facade = 1`. So `facade_enabled` varies per game version today, which means *which* mapping version `getCurrentMappingVersion()` lands on materially changes the geometry this clones in - and that resolution runs through `GameVersionService::getGameVersion(Auth::user())` → `GameVersion::getUserOrDefaultGameVersion()` → the bare `$this->mappingVersions()->orderByDesc('id')->first()` catch-all (`app/Models/Dungeon.php:255-279`). In the only caller (`mdt:importmapping`, CLI, no authenticated user) that means "retail's newest, or else whatever has the highest id" - i.e. precisely the ambient resolution the rest of this PR is removing, reintroduced one layer down.

**2. `facade_enabled` doesn't need to be inherited from another mapping version at all.** The authoritative "is this a facade dungeon" signal is `Floor::facade`: `MDTDungeon::getClonesAsEnemies()` (`app/Logic/MDT/Data/MDTDungeon.php:157-161`) switches to parsing only the facade floor based on `$floor->facade`, while `CoordinatesService::convertFacadeMapLocationToMapLocation()` (`app/Service/Coordinates/CoordinatesService.php:92-94`) decides whether to convert *back* based on `MappingVersion::facade_enabled`. That asymmetry - physical floor flag on the way in, mapping-version flag on the way out - is the actual root cause of the wrong-`floor_id` bug this commit fixes. Deriving `facade_enabled` from `$dungeon->getFacadeFloor() !== null` removes the inheritance question entirely for that column.

The `FloorUnion`s still need a source mapping version, but pinning it explicitly (e.g. `$dungeon->mappingVersions()->where('facade_enabled', true)->orderByDesc('id')->first()`) makes it deterministic and states the intent, instead of depending on ambient game-version resolution. As written, the legion-remix row above is a live example of a mapping version this code would silently source geometry *from* if it were the resolved "current" one.

app/Service/Mapping/MappingService.php
```diff
@@ -59,25 +59,55 @@ public function createNewMappingVersionFromPreviousMapping(
...
             'version'           => ($currentMappingVersion?->version ?? 0) + 1, // @phpstan-ignore nullsafe.neverNull
-            'facade_enabled'    => $currentMappingVersion?->facade_enabled ?? false, // @phpstan-ignore nullsafe.neverNull
+            'facade_enabled'    => $facadeSourceMappingVersion?->facade_enabled ?? false, // @phpstan-ignore nullsafe.neverNull
             'created_at'        => $now,
             'updated_at'        => $now,
         ]);
 
         $newMappingVersion = MappingVersion::find($id);
 
+        if ($currentMappingVersion === null) {
```
Wotuu, line 106: :robot: **This branch leaves `timer_max_seconds` at its column default of `0`, and two call sites divide by it.**

`mapping_versions.timer_max_seconds` is `int NOT NULL DEFAULT 0` (verified against the schema - not nullable). Before this PR the MDT import *always* went through `copyMappingVersionContentsToDungeon()`, which carries the value over (line 260 of this file). The new null-predecessor branch does not, so `mdt:importmapping <season> <new-game-version>` now produces mapping versions with `timer_max_seconds = 0` - the first time the primary scheduled pipeline can do that.

Downstream, `0` is not just "missing data":

- `app/Service/CombatLogEvent/Dtos/CombatLogEventFilter.php:515-521` guards with `if ($timerSeconds === null)` - already carrying `@phpstan-ignore identical.alwaysFalse` precisely because the column is non-nullable - and then does `(… * 60) / $timerSeconds`. That's a `DivisionByZeroError` on any heatmap request with a timer-fraction filter for a dungeon whose current mapping version (for the viewer's game version) came from a first-ever import.
- `app/Models/CombatLog/CombatLogEvent.php:390`: `$durationMs / ($dungeon->getCurrentMappingVersion()->timer_max_seconds * 1000)` - same failure mode, and line 384 additionally makes every run report `success = false`.

So I'd push back on the PR body's "missing nice-to-have data, not silently-wrong persisted data" classification for this one: a hard `DivisionByZeroError` is worse than a wrong value, and it's reachable through a normal user-facing page rather than only through mapping tooling.

`timer_max_seconds` is also a per-dungeon property in exactly the sense the comment above uses to justify inheriting facade geometry, so carrying it over from `$facadeSourceMappingVersion` here is a one-line, internally consistent fix. The same argument applies to `mountableAreas` (mount-speed zones are physical geometry too, cloned at line 242) - that one only degrades data rather than throwing, so it's genuinely lower priority.

tests/Feature/App/Service/MDT/MDTMappingImportGameVersionScopingTest.php
```diff
@@ -0,0 +1,216 @@
...
+        $retailGameVersion = GameVersion::query()->where('key', GameVersion::GAME_VERSION_RETAIL)->firstOrFail();
+
+        $dungeon = $this->getMdtDungeonWithRetailCheckpointAndNoMappingForGameVersion($targetGameVersion, $retailGameVersion);
+
+        $retailMappingVersion  = $dungeon->getCurrentMappingVersionForGameVersion($retailGameVersion);
+        $retailCheckpointCount = $retailMappingVersion->enemyForcesCheckpoints()->count();
+        $retailFloorUnionCount = $retailMappingVersion->floorUnions()->count();
+        $facadeFloor           = $dungeon->getFacadeFloor();
+
+        // importNpcsDataFromMDT() saves an NpcHealth row keyed by (npc_id, game_version_id) for every MDT NPC -
```
Wotuu, line 64: :robot: **This test reverses a decision the sibling test documents, and the cleanup covers only one of the shared-row classes the real pipeline writes.**

`tests/Feature/App/Service/MDT/MDTMappingImportEnemyForcesCheckpointTest.php:26-28` spells out why it drives `importEnemies()` via reflection with a stubbed `MDTDungeon` rather than the public entry point: *"a real import parses MDT's Lua and rewrites shared NPC/dungeon rows of the seeded test database"*. Going through `importMappingVersionFromMDT()` here is defensible (the resolution step under test only exists there), and this comment correctly identifies the problem class for `NpcHealth` - but the very same `importNpcsDataFromMDT()` call writes several other rows that are **not** mapping-version-scoped and are neither snapshotted nor restored:

- `importDungeon()` → `$dungeon->update(['mdt_id' => …])` (`MDTMappingImportService.php:362`)
- `importNpcsDataFromMDT()` → `$npc->save()`/`$npc->update()` overwriting `display_id` / `encounter_id` / `name` / `level` / `mdt_scale` / `npc_type_id` / `truesight` (lines 181-188), plus `NpcSpell::insert()` / `NpcDungeon::insert()` (lines 272-273)
- `Npc::createNpcEnemyForcesForExistingMappingVersions()` (`app/Models/Npc/Npc.php:372-397`, reached at `MDTMappingImportService.php:258`) for any MDT NPC not yet in the DB - that writes `NpcEnemyForces` into **every** mapping version of the dungeon, including the retail one this test asserts is left untouched.

I checked the currently seeded data rather than just asserting the risk: the selector picks `pitofsaron`, all 23 of its MDT NPCs already exist, and none of the columns above actually differ from the MDT values - so today every one of these is a no-op and the test is green. But that equality only holds while the seeded `dungeondata` matches the vendored MDT snapshot, and it stops holding right after an `update-mdt-package` bump - which is exactly when a mapping-import regression test matters most. The breakage would then surface as *other* tests failing against mutated shared rows, not as this test failing.

Cheapest way to shrink the exposure: the facade-geometry half needs no Lua at all. `createNewMappingVersionFromMDTMapping()` is public and can be called directly with `$currentMappingVersion = null` against an existing facade dungeon, which covers the `facade_enabled` / `FloorUnion` assertions without touching the NPC pipeline. Only the `game_version_id` / `version` assertions genuinely need the real entry point.

Wotuu:

The core scoping fix reads correct to me. I traced every `?->` / null branch the diff introduces (`importEnemies`, `importEnemyPatrols`, `importMapPOIs`, `deleteEmptyEnemyForcesCheckpoints`, `copyEnemyForcesCheckpointsToMappingVersion`) and none of them dereference a null predecessor - including `$currentMappingVersion->dungeonFloorSwitchMarkers->count()` at `MDTMappingImportService.php:1034`, which is only reachable in the `else` branch. `createNewMappingVersionFromMDTMapping()` has exactly one caller, so the signature change is complete. I also confirmed `$facadeSourceMappingVersion` is resolved *before* `insertGetId()` - had it been after, `getCurrentMappingVersion()`'s `orderByDesc('id')->first()` fallback would have resolved to the bare row this method just inserted.

I independently verified the facade-geometry mechanism rather than taking the PR body's word for it: `MDTDungeon::getClonesAsEnemies()` (`app/Logic/MDT/Data/MDTDungeon.php:157-161`) switches to the facade floor based on `Floor::facade`, and `convertFacadeMapLocationToMapLocation()` (`CoordinatesService.php:92-94`) returns the coordinate unchanged when `facade_enabled` is false - so enemies really would be persisted on the facade pseudo-floor. That half of the fix is real, and finding 1 is about how its source is chosen, not whether it's needed.

Findings posted inline:
1. `MappingService.php:74` - `$facadeSourceMappingVersion` falls back to ambient `getCurrentMappingVersion()`; `facade_enabled` is empirically *not* uniform across game versions in the current data (dungeon 12 retail=1 / legion-remix=0; dungeon 49 retail=1 / wotlk=0), and `Floor::facade` is the authoritative signal that removes the inheritance question for that column.
2. `MappingService.php:94` - the null-predecessor branch leaves `timer_max_seconds` at its `NOT NULL DEFAULT 0`, and `CombatLogEventFilter.php:515-521` / `CombatLogEvent.php:390` divide by it (`DivisionByZeroError`). Newly reachable from the scheduled pipeline; I don't think "nice-to-have data" covers it.
3. `MDTMappingImportGameVersionScopingTest.php:66` - the test runs the real write-side pipeline, which the sibling `MDTMappingImportEnemyForcesCheckpointTest.php:26-28` explicitly avoids for documented reasons; cleanup covers `NpcHealth` but not `dungeons.mdt_id`, the `Npc` column overwrites, or `NpcSpell`/`NpcDungeon`/`NpcEnemyForces` inserts. No-ops on today's seeded data (verified), fragile after the next MDT package bump.

Two notes not worth an inline thread:

- **Small factual correction to the PR body.** The bullet claiming `enemy_forces_required`/`_teeming`/`_shrouded`/`_shrouded_zul_gamux` are "unconditionally recompute[d] … regardless of predecessor state" holds for the first two (`importDungeon()`) but not the two shrouded columns: `importNpcs()` writes them only inside `if ($npc->isShrouded())` / `elseif ($npc->isShroudedZulGamux())` (`MDTMappingImportService.php:419-423`). On the null-predecessor path they stay `NULL` where they used to be inherited, and `MappingVersion` declares both as `@property int`. Reachability is thin enough that I wouldn't block on it (it needs an `Enemy` with `SEASONAL_TYPE_SHROUDED`, which `MDTDungeon.php:266` sets from MDT's `disguised` clone flag, plus an MDT route import against that mapping version, where `ImportStringPulls::addEnemyForces(int $amount)` would get `null`) - but the "verified not affected" wording overstates it.
- **Pre-existing, not a diff defect, but it affects how this fix behaves in practice.** `MDTDungeon::getClonesAsEnemies()` memoizes under `mdt_enemies_<dungeonKey>` with no mapping-version or game-version component. Running `mdt:importmapping <season> retail` and then `… <season> legion-remix` back-to-back within the MDT cache TTL - the natural sequence when onboarding a game version - reuses enemy rows whose facade→real-floor conversion was computed against the *first* mapping version's `facade_enabled`. In that ordering the cache happens to mask the bug this PR fixes; in the reverse ordering it would defeat the fix. Worth an issue rather than scope creep here, and it may explain why dungeon 12's legion-remix mapping version (`facade_enabled=0`, 0 `FloorUnion`s) nonetheless has all 274 of its enemies on the real floor.

app/Service/Mapping/MappingService.php
Wotuu: :robot: Fixed. `facade_enabled` is no longer derived from any mapping version - it now comes from `$dungeon->getFacadeFloor() !== null`, a physical fact about the dungeon that already had a helper (`Dungeon::getFacadeFloor()`, `app/Models/Dungeon.php:475-478`). For the `FloorUnion`/`FloorUnionArea` clone source (which does need a concrete row to clone from), the ambient `$dungeon->getCurrentMappingVersion()` fallback is gone - it's now pinned explicitly and deterministically to `$dungeon->mappingVersions()->where("facade_enabled", true)->orderByDesc("id")->first()` when the dungeon has a facade floor, never touching `GameVersion::getUserOrDefaultGameVersion()`/authenticated-user context.

Verified this actually matters: a new test (`MappingServiceCreateNewMappingVersionFromMDTMappingTest`) authenticates a user whose profile `game_version_id` points at a facade-DISABLED mapping version (legion-remix, matching the dungeon-12/49 data you found) and confirms `createNewMappingVersionFromMDTMapping()` still resolves `facade_enabled=true` and clones FloorUnions from the facade-enabled (retail) mapping version, not the ambient one. Temporarily reverted just this change and confirmed the test fails (`facade_enabled` resolves to `0`) before restoring the fix.

See `app/Service/Mapping/MappingService.php:68-83` and the new test at `tests/Feature/App/Service/Mapping/MappingServiceCreateNewMappingVersionFromMDTMappingTest.php`.

app/Service/Mapping/MappingService.php
Wotuu: :robot: Fixed - agreed this is a live bug, not just missing data. `timer_max_seconds` is now carried over from the resolved facade-source mapping version on the null-predecessor path (`app/Service/Mapping/MappingService.php:106-119`), alongside the FloorUnion clone. Confirmed via the new `MappingServiceCreateNewMappingVersionFromMDTMappingTest` that the new mapping version's `timer_max_seconds` matches the source's (a non-zero fixture value), rather than sitting at the column default.

Left `mountableAreas` untouched/undocumented-as-is per your note - out of scope for this fix.

tests/Feature/App/Service/MDT/MDTMappingImportGameVersionScopingTest.php
Wotuu: :robot: Implemented as suggested. Split the facade_enabled/FloorUnion/timer_max_seconds assertions out into a new `tests/Feature/App/Service/Mapping/MappingServiceCreateNewMappingVersionFromMDTMappingTest.php`, which calls `createNewMappingVersionFromMDTMapping()` directly - no Lua, no MDT string, no touching the shared seeded NPC/dungeon rows at all.

`MDTMappingImportGameVersionScopingTest` now keeps only what genuinely needs the real `importMappingVersionFromMDT()` pipeline: the `game_version_id`/`version` scoping and the checkpoint-count assertions (4 assertions, down from 6+the removed facade ones), with its docblock updated to explain the split and point at the new test.

`createNewMappingVersionFromMDTMapping()` was already public with `$currentMappingVersion` as a separate param from #3720/#3749, so no signature change was needed - the new test just calls it with `null` directly against a real facade dungeon that has zero mapping versions for the target game version, which is exactly the scenario this whole issue is about.

app/Service/Mapping/MappingService.php
```diff
@@ -59,25 +59,74 @@ public function createNewMappingVersionFromPreviousMapping(
...
+            // along with timer_max_seconds - it's a NOT NULL DEFAULT 0 column that several call sites divide
+            // by unguarded (CombatLogEventFilter::fromHeatmapDataFilter(), CombatLogEvent::setTimeInterval()),
+            // so leaving it at its column default here is a live DivisionByZeroError, not just missing data.
+            if ($facadeSourceMappingVersion !== null) {
+                $this->cloneFloorUnionsToMappingVersion($facadeSourceMappingVersion, $newMappingVersion);
+
+                $newMappingVersion->update([
+                    'timer_max_seconds' => $facadeSourceMappingVersion->timer_max_seconds,
+                ]);
+            }
```
Wotuu, line 126: Why are you only copying over floor unions? If a current mapping version does not exist, but there IS one from a different game version, shouldn't all physical properties copy over? Floor switches, map icons, mountable areas, floor unions? Why just floor unions?

Wotuu (commented):
Question

app/Service/Mapping/MappingService.php
Wotuu: :robot: Good question - checked each of the three you named against real evidence rather than guessing:

**MountableArea** - physical, now cloned. Same class of data as FloorUnion (fixed dungeon terrain, e.g. a mount-speed zone tied to `floor_id`/`vertices_json`), and critically: MDT has zero concept of mount-speed zones, so nothing else in the import pipeline ever (re)creates them. Leaving this path empty meant a first-ever import for a game version silently lost all mount-speed data forever. Fixed in `fe36d5b6f` by extracting `cloneMountableAreasToMappingVersion()` (shared with `copyMappingVersionContentsToDungeon()`) and calling it alongside `cloneFloorUnionsToMappingVersion()`.

**DungeonFloorSwitchMarker** - stays empty, but NOT because it's curated. `MDTMappingImportService::importMapPOIs()` - called right after `createNewMappingVersionFromMDTMapping()` returns, as part of the very same import - already recreates these itself from MDT's own `MapLink` POI data whenever `$currentMappingVersion` is null (see the `#3757` comment on that branch). Cloning them here too would just create duplicate staircase markers a few lines later in the same request.

**MapIcon** - also stays empty, same mechanical reason: `importMapPOIs()` recreates the physical subset it recognizes (graveyard, dungeon entrance, priory blessing, floodgate, eco-dome items, etc.) from MDT's map POI data on this exact path. On top of that, `MapIcon`'s type catalog is a genuine mixed bag that makes "just clone it" the wrong instinct even where the importer doesn't already cover it - it includes clearly season/mechanic-specific curated content (`awakened_obelisk_*`, `priory_blessing_of_the_sacred_flame`, `eco_dome_al_dani_*`) alongside purely decorative, non-geometric icons (raid markers, class/role icons) that carry no physical information at all. `seasonal_index` also isn't a "which season is this icon" flag as I first assumed - it's a disambiguator between multiple icons of the same type on the same floor (see `HasLinkedAwakenedObelisk`) - so it didn't end up mattering for this decision either way.

Added a second test (`...clonesMountableAreasButLeavesMapIconsAndDungeonFloorSwitchMarkersEmpty`) alongside the existing one, on a fixture (Halls of Atonement) that actually has MountableAreas/MapIcons/DungeonFloorSwitchMarkers on its source mapping version, so the assertions on all three are non-vacuous.

Wotuu: Thanks, I disagree on the dungeon floor switch markers, I do think they should be copied over from the existing mapping versions, since I don't think MDT really supplies them anymore, since all of the maps it generates have a combined view, so they don't need a floor switch marker. Well, I do need them extensively, since I need to support both facade mode and visit style mode with all the different dungeon floors. So I do think we should copy over the dungeon floor switch markers.

Wotuu: :robot: Agreed, and fixed - `21064a2f6`. `DungeonFloorSwitchMarker` is now cloned from the facade-source mapping version on the null-predecessor path, the same way `FloorUnion`/`MountableArea` already are (`MappingService::cloneDungeonFloorSwitchMarkersToMappingVersion()`, called from `createNewMappingVersionFromMDTMapping()` and shared with `copyMappingVersionContentsToDungeon()`).

That does reopen the duplicate-creation risk you'd expect: `importMapPOIs()` runs right after, in the same import, and used to unconditionally (re)create a floor switch marker from MDT's `MapLink` POI data whenever `$currentMappingVersion === null` - exactly the condition under which the clone above now happens. Fixed that too: the create-vs-skip decision in `importMapPOIs()` now keys off whether `$newMappingVersion` already has floor switch markers of its own (from *any* source - the predecessor clone or the new facade-source clone), rather than inspecting `$currentMappingVersion`, which is null in precisely the case that now pre-populates markers. So a first-ever import for a game version clones the markers once and `importMapPOIs()` no longer touches them; a dungeon with no facade source to clone from (nothing cloned) still falls back to creating them from MDT's data exactly as before.

Added `MDTMappingImportMapPOIsFloorSwitchMarkerTest` covering both branches directly (stubbed `MDTDungeon`, no real Lua) - "already cloned → no duplicate created" and "nothing cloned → still created from MDT data" - and updated the sibling `MappingServiceCreateNewMappingVersionFromMDTMappingTest` test (renamed to `...clonesMountableAreasAndDungeonFloorSwitchMarkersButLeavesMapIconsEmpty`) to assert the clone instead of emptiness. All affected tests green, `composer run fix`/`analyse` clean.

Wotuu: :robot: Follow-up: the previous commit only actually fixed this for facade dungeons - `cloneDungeonFloorSwitchMarkersToMappingVersion()` was called from inside the same `if ($facadeSourceMappingVersion !== null)` block as `FloorUnion`/`MountableArea`, and that variable is only non-null when the dungeon physically has a facade floor. Since staircases exist on non-facade dungeons too (most of the roster, by a DB check - `ataldazar`, `scholomance`, `naxxramas`, `icecrowncitadel`, etc. all have `DungeonFloorSwitchMarker`s with no facade floor at all), a first-ever import for one of those would still have silently dropped them, missing your actual point.

Fixed in `2fcb82017`: `DungeonFloorSwitchMarker`'s clone source is now resolved independently of `$facadeSourceMappingVersion` - it falls back to the newest mapping version of *any* game version when there's no facade source, same deterministic-pin rationale as the facade geometry above, rather than piggybacking on facade resolution. Added a test on a dungeon with no facade floor at all (`createNewMappingVersionFromMDTMapping_givenNoPredecessorAndDungeonWithoutFacadeFloor_stillClonesDungeonFloorSwitchMarkersFromNewestMappingVersionOfAnyGameVersion`) to prove it - it would have failed against the previous commit. All tests/PHPStan/CS green, CI passing.

Wotuu (commented):
Ok

PR #3764, opened 2026-07-30
#3755 Fail the whole MDT mapping import when a per-item save is swallowed

### LINKED ISSUE #3755: mdt:importmapping can silently lose enemies via swallowed per-item exceptions (follow-up to #3737)
:robot:

Follow-up to #3737/#3751, flagged during independent review of #3751 (thread on
`app/Service/MDT/MDTMappingImportService.php:124`, reply-only, no code change made in that PR).

## Background

#3737/#3751 fixed the main failure mode of `mdt:importmapping`: an exception propagating out of
`MDTMappingImportService::importMappingVersionFromMDT()`'s try block now deletes the
partially-imported mapping version and rethrows, instead of leaving a half-built version behind
that silently no-ops forever.

## Remaining gap

Several of the per-item loops called from inside that same try block catch and swallow
per-item exceptions instead of letting them propagate:

- `importNpcsDataFromMDT()` (around `app/Service/MDT/MDTMappingImportService.php:272-276`) -
  catches `UniqueConstraintViolationException` and `Exception` per NPC and just logs
  (`importNpcsDataFromMDTNpcNotMarkedForAllDungeons()` / `importNpcsDataFromMDTSaveNpcException()`).
- `importNpcs()` (around line 440) - catches `Exception` per NPC and logs
  (`importNpcsDataFromMDTSaveNpcException()`).
- `importEnemies()` (around line 578) and the nested loop around line 794 - same pattern, catch
  and log per enemy.

Because these exceptions never reach `importMappingVersionFromMDT()`'s outer catch, the mapping
version is still stamped with the fresh `mdt_mapping_hash` and treated as a fully successful
import - even though individual NPCs/enemies were silently dropped. The dungeon ends up missing
enemies, and because the hash now matches, the next run sees "no change detected" and skips
re-importing - reproducing #3737's user-visible symptom (dungeon missing enemies, stuck that way)
through a different door than the one #3737/#3751 closed.

## Recommended approach

Not obviously a place to blanket-rethrow (some of these failures - e.g.
`UniqueConstraintViolationException` for an NPC not marked for all dungeons - may be expected and
recoverable). Worth first auditing which of these swallowed cases are truly benign vs. which
should cause the whole import to fail loudly (delete the partial mapping version, same as the
outer catch does) rather than being logged and ignored. A plausible shape: accumulate the
swallowed failures during the run and, if any occurred, treat the import as failed after all
steps complete (delete the new mapping version, throw a summary exception) instead of stamping
the hash.

## How to verify

A regression test analogous to `MDTMappingImportCrashRecoveryTest` (added in #3751), asserting
that when a per-item exception is forced inside one of these loops (e.g. via a decorator or a
deliberately-malformed MDT fixture), the import either surfaces the failure or the resulting
mapping version does not silently claim success while missing data.

## Key files

- `app/Service/MDT/MDTMappingImportService.php` - `importNpcsDataFromMDT()`, `importNpcs()`,
  `importEnemies()` (the swallowed catches), and `importMappingVersionFromMDT()` (for the
  existing outer catch/delete pattern to potentially reuse)
- `app/Service/MDT/Logging/MDTMappingImportServiceLogging.php` - the existing log entries for
  these swallowed exceptions

Wotuu:
:robot:

Closes #3755

## Summary

Follow-up to #3737/#3751. `importNpcsDataFromMDT()`, `importNpcs()`, and `importEnemies()` each
catch a genuine per-NPC/per-enemy save exception, log it, and move on to the next item - which is
good for letting the rest of the import proceed, but it also meant that exception never reached
`importMappingVersionFromMDT()`'s outer catch (added by #3751). The new mapping version still got
stamped with the fresh `mdt_mapping_hash` even though data was silently dropped, so the next run
saw "no change detected" and skipped retrying forever - #3737's exact symptom, through a different
door.

Each of the three swallowing catches now also appends the exception to a `$failures` accumulator
threaded through the call chain by reference. Once every import step in
`importMappingVersionFromMDT()` has run, it checks the accumulator and, if anything failed, throws
`MDTMappingImportPartialFailureException` instead of stamping the hash - which lets #3751's
existing outer catch delete the partial mapping version and retry cleanly next run.

**#3751 has since merged** - `importMappingVersionFromMDT()`'s "throw before stamping" behavior this
PR relies on comes from that merge. This branch was rebased onto current master after #3751/#3753
landed; the diff now shows only this PR's own changes.

## Audit: which catches changed, which didn't

- `importNpcsDataFromMDT()`'s `UniqueConstraintViolationException` case (another dungeon's import
  already inserted this NPC) is left alone, unchanged - the issue itself flagged this as expected
  and recoverable, not a real failure.
- `importNpcsDataFromMDT()`'s and `importNpcs()`'s generic `catch (Exception)`, and
  `importEnemies()`'s per-enemy `catch (Exception)`, now accumulate.
- `importEnemyPatrols()`'s existing `catch (Exception)` (unable to find the attached enemy) was
  already rethrowing - no change needed there.
- `importNpcs()`'s "NPC not found" `continue` (no matching NPC after `importNpcsDataFromMDT()`) is
  deliberately **not** wired into the failure set - NPCs in `IGNORE_NPC_DATA_NPC_IDS` legitimately
  skip that lookup by design, so treating it as a failure would misclassify an intentional skip.

## Known trade-off

This does change behavior for a genuinely non-recoverable per-item exception (e.g. a malformed DB
write on one specific NPC/enemy): instead of losing just that one item, the **whole dungeon's**
import now fails every run until the underlying data problem is fixed, rather than quietly limping
along missing that one item forever. That matches the issue's recommended shape - a dungeon stuck
on its previous, still-working mapping version is preferable to one silently missing enemies while
looking up to date - but it is a real behavior change worth flagging explicitly.

The partial-mapping-version-delete log (`importMappingVersionFromMDTDeletePartialMappingVersion`,
reused from #3751) is already logged at `error`, so this reaches the same visibility channel as
before.

## Test plan

- [x] New regression test: `MDTMappingImportCrashRecoveryTest::importMappingVersionFromMDT_givenPerNpcSaveThrows_failsTheWholeImportInsteadOfSilentlyStampingTheHash`
      - forces a per-NPC exception via a mocked logger (throws from `importNpcsUpdateExistingNpc()`,
        a log call inside `importNpcs()`'s per-NPC try block) and asserts the import throws
        `MDTMappingImportPartialFailureException`, the dungeon's current mapping version is
        unchanged, and no half-built mapping version is left behind.
- [x] `MDTMappingImportEnemyForcesCheckpointTest` updated for `importEnemies()`'s new `$failures`
      parameter (it invokes the private method via Reflection).
- [x] `--group=MDT --group=MappingVersion --group=EnemyForcesCheckpoint` green, run repeatedly to
      rule out flakiness.

## Cold-review follow-up

Addressed the three inline findings from the cold review:
- `MDTMappingImportPartialFailureException` now passes the first accumulated failure as `$previous`
  (so its stack trace/class survive the rethrow) and caps the concatenated message to the first 10
  failures; covered by a new assertion in `MDTMappingImportCrashRecoveryTest`.
- Corrected the `UniqueConstraintViolationException` comment: the collision is deterministic (an
  NPC that exists globally but isn't yet linked to this dungeon), not a race - behavior unchanged.
- Left the `array &$failures` parameter spacing in `importNpcs()`/`importEnemies()` as-is: it's what
  `php-cs-fixer`'s `ErickSkrauch/align_multiline_parameters` fixer produces for this multi-line
  parameter list (`--dry-run` is clean on the current form; hand-editing it to single-sided padding
  gets reverted by `composer run fix` and would fail the `php-cs-fixer` CI job).

### DIFF
--- a/app/Service/MDT/Exceptions/MDTMappingImportPartialFailureException.php
+++ b/app/Service/MDT/Exceptions/MDTMappingImportPartialFailureException.php
@@ -1,0 +1,36 @@
+<?php
+
+namespace App\Service\MDT\Exceptions;
+
+use Exception;
+
+/**
+ * Thrown by {@see \App\Service\MDT\MDTMappingImportService::importMappingVersionFromMDT()} when one or more
+ * per-item saves (NPCs, enemies) inside the import failed. Those failures are individually caught and
+ * logged where they occur - so the loop keeps going and the rest of the items still get imported - but the
+ * run as a whole must still be treated as failed: stamping the new mapping version's mdt_mapping_hash despite
+ * the losses would make every subsequent run compare hashes, see no change, and silently skip retrying
+ * forever while the dungeon stays missing whatever failed to import (#3755).
+ */
+class MDTMappingImportPartialFailureException extends Exception
+{
+    /**
+     * @param list<Exception> $failures
+     */
+    public function __construct(private readonly array $failures)
+    {
+        parent::__construct(sprintf(
+            '%d item(s) failed to import: %s',
+            count($failures),
+            implode('; ', array_map(static fn(Exception $failure) => $failure->getMessage(), $failures)),
+        ));
+    }
+
+    /**
+     * @return list<Exception>
+     */
+    public function getFailures(): array
+    {
+        return $this->failures;
+    }
+}
--- a/app/Service/MDT/Exceptions/MDTMappingImportPartialFailureException.php
+++ b/app/Service/MDT/Exceptions/MDTMappingImportPartialFailureException.php
@@ -19,11 +19,17 @@ class MDTMappingImportPartialFailureException extends Exception
      */
     public function __construct(private readonly array $failures)
     {
-        parent::__construct(sprintf(
-            '%d item(s) failed to import: %s',
-            count($failures),
-            implode('; ', array_map(static fn(Exception $failure) => $failure->getMessage(), $failures)),
-        ));
+        $messages = array_map(static fn(Exception $failure) => $failure->getMessage(), $failures);
+        $shown    = array_slice($messages, 0, 10);
+        if (count($messages) > 10) {
+            $shown[] = sprintf('and %d more', count($messages) - 10);
+        }
+
+        parent::__construct(
+            sprintf('%d item(s) failed to import: %s', count($failures), implode('; ', $shown)),
+            0,
+            $failures[0] ?? null,
+        );
     }
 
     /**
--- a/app/Service/MDT/MDTMappingImportService.php
+++ b/app/Service/MDT/MDTMappingImportService.php
@@ -31,6 +31,7 @@ use App\Models\Spell\SpellDungeon;
 use App\Service\Cache\CacheServiceInterface;
 use App\Service\Coordinates\CoordinatesServiceInterface;
 use App\Service\Mapping\MappingServiceInterface;
+use App\Service\MDT\Exceptions\MDTMappingImportPartialFailureException;
 use App\Service\MDT\Logging\MDTMappingImportServiceLoggingInterface;
 use Exception;
 use Illuminate\Database\UniqueConstraintViolationException;
@@ -92,6 +93,10 @@ class MDTMappingImportService implements MDTMappingImportServiceInterface
 
             $mdtDungeon = new MDTDungeon($this->cacheService, $this->coordinatesService, $dungeon);
 
+            /** @var array<int, Exception> $importFailures NPC/enemy saves that were individually caught and
+             *  logged rather than propagated, so the loop kept going for the remaining items (#3755) */
+            $importFailures = [];
+
             try {
                 $this->log->importMappingVersionFromMDTStart($dungeon->id);
 
@@ -104,7 +109,7 @@ class MDTMappingImportService implements MDTMappingImportServiceInterface
                 );
 
                 $this->importDungeon($mdtDungeon, $dungeon, $newMappingVersion);
-                $this->importNpcs($newMappingVersion, $mdtDungeon, $dungeon, $gameVersion);
+                $this->importNpcs($newMappingVersion, $mdtDungeon, $dungeon, $gameVersion, $importFailures);
                 $enemies = $this->importEnemies(
                     $currentMappingVersion,
                     $newMappingVersion,
@@ -112,10 +117,18 @@ class MDTMappingImportService implements MDTMappingImportServiceInterface
                     $dungeon,
                     $enemyForcesCheckpointIdMapping,
                     $forceImport,
+                    $importFailures,
                 );
                 $this->importEnemyPacks($newMappingVersion, $mdtDungeon, $dungeon, $enemies);
                 $this->importEnemyPatrols($currentMappingVersion, $newMappingVersion, $mdtDungeon, $dungeon, $enemies);
                 $this->importMapPOIs($currentMappingVersion, $newMappingVersion, $mdtDungeon, $dungeon);
+
+                if (!empty($importFailures)) {
+                    // Every step above ran, but one or more individual NPCs/enemies were silently dropped
+                    // along the way - the mapping version is incomplete even though nothing propagated to
+                    // this try block on its own, so it must not be treated as a success (#3755).
+                    throw new MDTMappingImportPartialFailureException($importFailures);
+                }
 
                 // Only stamp the hash once every step above actually succeeded (#3737) - a mapping version
                 // recorded with the hash of MDT data it never finished importing would look up to date to
@@ -161,7 +174,7 @@ class MDTMappingImportService implements MDTMappingImportServiceInterface
         );
     }
 
-    public function importNpcsDataFromMDT(MDTDungeon $mdtDungeon, Dungeon $dungeon, GameVersion $gameVersion): void
+    public function importNpcsDataFromMDT(MDTDungeon $mdtDungeon, Dungeon $dungeon, GameVersion $gameVersion, array &$failures = []): void
     {
         try {
             $this->log->importNpcsDataFromMDTStart($dungeon->key);
@@ -270,9 +283,12 @@ class MDTMappingImportService implements MDTMappingImportServiceInterface
                         $npc->createNpcEnemyForcesForExistingMappingVersions($mdtNpc->getCount());
                     }
                 } catch (UniqueConstraintViolationException) {
+                    // Expected and recoverable: another dungeon's import already inserted this NPC (or its
+                    // historical NpcEnemyForces rows) between our lookup and our write - not a real failure.
                     $this->log->importNpcsDataFromMDTNpcNotMarkedForAllDungeons($npc->id);
                 } catch (Exception $exception) {
                     $this->log->importNpcsDataFromMDTSaveNpcException($exception);
+                    $failures[] = $exception;
                 }
             }
 
@@ -388,6 +404,8 @@ class MDTMappingImportService implements MDTMappingImportServiceInterface
     }
 
     /**
+     * @param  array<int, Exception> $failures Appended with NPC saves that failed - see
+     *                                         importNpcsDataFromMDT()'s $failures parameter.
      * @throws Exception
      */
     private function importNpcs(
@@ -395,11 +413,12 @@ class MDTMappingImportService implements MDTMappingImportServiceInterface
         MDTDungeon     $mdtDungeon,
         Dungeon        $dungeon,
         GameVersion    $gameVersion,
+        array          &          $failures,
     ): void {
         try {
             $this->log->importNpcsStart();
 
-            $this->importNpcsDataFromMDT($mdtDungeon, $dungeon, $gameVersion);
+            $this->importNpcsDataFromMDT($mdtDungeon, $dungeon, $gameVersion, $failures);
 
             // Get a list of NPCs and update/save them (re-fetch the list to include any new NPCs)
             $npcs = $dungeon->npcs()->get()->keyBy('id');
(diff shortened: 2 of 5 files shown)


Wotuu (commented):
Ok

app/Service/MDT/Exceptions/MDTMappingImportPartialFailureException.php
```diff
@@ -0,0 +1,36 @@
...
+     */
+    public function __construct(private readonly array $failures)
+    {
+        parent::__construct(sprintf(
+            '%d item(s) failed to import: %s',
+            count($failures),
+            implode('; ', array_map(static fn(Exception $failure) => $failure->getMessage(), $failures)),
+        ));
+    }
+
```
Wotuu, line 34: :robot: **The partial-failure exception drops every original stack trace — pass the first failure as `$previous`.**

`parent::__construct()` is called with only a message, so `$previous` stays `null`. By the time this reaches
`importMappingVersionFromMDT()`'s `catch (Throwable)` (MDTMappingImportService.php:138) and gets rethrown out of
`mdt:importmapping`, the only trace attached points at `MDTMappingImportService.php:130` — the `throw` statement
itself — and every accumulated `$failures[n]` object is reachable *only* via `getFailures()`, which nothing outside
this class calls. The exception class of each failure is lost too: `getMessage()` alone does not distinguish a
`QueryException` from the hand-rolled `new Exception(sprintf('Unable to save npc %d!', ...))` at line 276.

That matters here because the other channel is thin: `importMappingVersionFromMDTDeletePartialMappingVersion(...,
Throwable $throwable)` logs via `$this->error(__METHOD__, get_defined_vars())`, i.e. the `Throwable` goes into the
structured-log context as an object rather than a rendered trace.

One-line fix, no other change needed:

```php
parent::__construct(
    sprintf(...),
    0,
    $failures[0] ?? null,
);
```

`Exception::__construct(string $message = "", int $code = 0, ?Throwable $previous = null)` accepts it directly, and
`list<Exception>` satisfies `?Throwable`.

Secondary, same `sprintf`: the message concatenates *every* failure unbounded. An MDT change that breaks a whole
dungeon's NPC set produces one log line with hundreds of `; `-joined messages. Since each failure is already logged
individually at `error` where it happens, capping this (e.g. first 10 plus `and N more`) costs nothing in
diagnosability.

app/Service/MDT/MDTMappingImportService.php
```diff
@@ -270,9 +283,12 @@ public function importNpcsDataFromMDT(MDTDungeon $mdtDungeon, Dungeon $dungeon,
                         $npc->createNpcEnemyForcesForExistingMappingVersions($mdtNpc->getCount());
                     }
                 } catch (UniqueConstraintViolationException) {
+                    // Expected and recoverable: another dungeon's import already inserted this NPC (or its
+                    // historical NpcEnemyForces rows) between our lookup and our write - not a real failure.
```
Wotuu: :robot: **Comment accuracy (not a behaviour bug): this describes a race, but the case it actually covers is deterministic.**

"between our lookup and our write" implies a concurrent-insert race. The lookup this refers to is
`$existingNpcs = $dungeon->npcs()->with('npcSpells')->get()` (line 183), and `Dungeon::npcs()` is
`belongsToMany(Npc::class, 'npc_dungeons', ...)` — i.e. **scoped to this dungeon via the pivot**, not a global NPC
lookup. So an NPC row that already exists globally but has no `npc_dungeons` link to *this* dungeon is simply absent
from `$existingNpcs`, `$newlyCreated` is set to `true` at line 200, and `$npc->save()` at line 268 issues an INSERT
with an explicit primary key that collides **every single time** — no concurrency involved. The name of the log
method in this very branch, `importNpcsDataFromMDTNpcNotMarkedForAllDungeons()`, is describing exactly that
situation.

The conclusion ("expected and recoverable, not a real failure") still holds — `$npcHealth->save()` already ran at
line 243 and the missing pivot row is queued into `$npcDungeonsAttributes` and inserted at line 301, so the next run
finds the NPC through `$existingNpcs` and takes the update path. Only the stated mechanism is wrong, and since this
comment is the justification for *not* wiring this catch into `$failures`, it is worth being right about. Suggest
something like: "the NPC row already exists globally but is not linked to this dungeon yet, so the dungeon-scoped
lookup above missed it and we attempted an insert on an existing primary key — the npc_dungeons link is written
below, so the next run takes the update path."

app/Service/MDT/MDTMappingImportService.php
```diff
@@ -388,18 +404,21 @@ private function importDungeon(MDTDungeon $mdtDungeon, Dungeon $dungeon, Mapping
...
+     * @param  array<int, Exception> $failures Appended with NPC saves that failed - see
+     *                                         importNpcsDataFromMDT()'s $failures parameter.
      * @throws Exception
      */
     private function importNpcs(
         MappingVersion $newMappingVersion,
         MDTDungeon     $mdtDungeon,
         Dungeon        $dungeon,
         GameVersion    $gameVersion,
+        array          &          $failures,
```
Wotuu, line 418: :robot: **Nit: stray alignment padding between `&` and `$failures` reads like a typo.**

```php
array          &          $failures,
```

The parameter-alignment style in this file pads between the *type* and the `$name`; here the padding landed on both
sides of the `&`, so the reference marker ends up floating in the middle of whitespace. Same in `importEnemies()` at
line 484. It parses fine and php-cs-fixer is green (which is why this isn't linter-catchable), but `array &$failures,`
with the existing left padding — matching how `importNpcsDataFromMDT()` at line 177 and the interface write it — is
what a reader expects.

Wotuu:
:robot: Cold review (opus): 3 findings posted - all advisory, none blocking.

1. `MDTMappingImportPartialFailureException` never sets `$previous`, so every original stack trace and exception
   class is lost by the time this is rethrown out of `mdt:importmapping` (plus an unbounded concatenated message).
2. The newly added `UniqueConstraintViolationException` comment describes a race, but the case it covers is
   deterministic (`Dungeon::npcs()` is pivot-scoped) - comment accuracy only, the conclusion holds.
3. Nit: stray padding between `&` and `$failures` in the `importNpcs()`/`importEnemies()` signatures.

Verified independently rather than taking the PR body at its word: every `catch` and `continue` in the pipeline
methods at head was walked (`importMappingVersionFromMDT`, `importNpcsDataFromMDT`, `importNpcs`, `importEnemies`,
`importEnemyPacks`, `importEnemyPatrols`, `importMapPOIs`, `deleteEmptyEnemyForcesCheckpoints`). The audit claim is
accurate and complete: the three swallowing `catch (Exception)` blocks all accumulate, `importEnemyPatrols()` was
already rethrowing, `importEnemyPacks()`/`importMapPOIs()`/`importDungeon()` have no swallowing catch at all, and
every `continue` is a deliberate skip (ignore lists, teeming, empty patrol, group -1).

The `$failures` reference plumbing holds up: `$importFailures` is declared outside the `try`, every hop is `&`, there
is no recursion, and the only by-value/default entry point is the public `importNpcsDataFromMDT()` as used by
`mdt:importnpcs`, which is intentional. `MDTMappingImportPartialFailureException extends Exception`, so
`catch (Throwable)` at MDTMappingImportService.php:138 catches it, deletes the partial mapping version, and rethrows.
`ImportMapping`'s season loop catches per dungeon, so one bad dungeon does not abort the rest.

One behavioural note that is not a defect: on the *enemy* path the new exception will often be pre-empted. A failed
`$mdtEnemy->save()` leaves the object in the returned `$enemies` collection with `exists = false`, so
`importEnemyPacks()`'s `$savedEnemy->update([...])` returns false and throws `Unable to update enemy with enemy pack!`
before control reaches the `!empty($importFailures)` check. Same outcome (delete + rethrow + clean retry next run),
just a different exception in the alert.

Tests check out: the new regression test genuinely reproduces #3755 (without the accumulator the import completes and
stamps the hash, so `$this->fail('Expected the import to throw.')` fires), and the `invoke()` -> `invokeArgs()` change
in `MDTMappingImportEnemyForcesCheckpointTest` does not weaken anything - those assertions never depended on
`$failures`, and a swallowed enemy-save failure there would still surface as a null `$importedEnemy`.

app/Service/MDT/Exceptions/MDTMappingImportPartialFailureException.php
Wotuu: :robot: Fixed — `parent::__construct()` now passes `$failures[0] ?? null` as `$previous`, and the concatenated message is capped to the first 10 failures (plus "and N more") since each is already logged individually at `error` level. Also added an assertion in `MDTMappingImportCrashRecoveryTest` that `$exception->getPrevious()` is the first accumulated failure, so this is covered going forward.

app/Service/MDT/MDTMappingImportService.php
Wotuu: :robot: Fixed the comment wording — it now explains the actual mechanism: the NPC row exists globally but is not yet linked to this dungeon, so the dungeon-scoped lookup misses it and the insert collides on the existing primary key (deterministic, not a race). The "expected and recoverable" conclusion and behavior are unchanged.

app/Service/MDT/MDTMappingImportService.php
Wotuu: :robot: Checked but left as-is: this spacing is not a stray typo, it's what `php-cs-fixer` (specifically the `ErickSkrauch/align_multiline_parameters` custom fixer, configured with `variables => true`) produces and enforces for this multi-line parameter list — `--dry-run` reports 0 fixable files against the current form, and hand-editing it to pad only between the type and `&` gets rewritten right back to the symmetric padding by `composer run fix`. Since CI runs `php-cs-fixer fix --dry-run --diff`, the suggested single-sided padding would fail that check. The only way to get plain `array &$failures` is collapsing the whole signature onto one line, which `importEnemies()`'s 7 parameters can't reasonably do (and would leave it inconsistent with `importNpcs()`), so I'm leaving both sites as the formatter produces them.

Issue #3350, opened 2026-06-29
Drop legacy File-upload expansion icons in favor of asset-repo images (follow-up to #3299)


Wotuu:
## Background

Expansion icons are currently served through a legacy Laravel file-upload mechanism: `Expansion`
carries an `icon_file_id` and a `iconfile()` `HasOne` relation (via the `HasIconFile` trait), and the
admin Expansion edit form uploads an image into the `files` table (see `Expansion::saveExpansion()`-style
logic around `app/Models/Expansion.php:305-321`). This was the first use of Laravel-managed uploads on
the project and is now considered legacy — the icon images live in the assets repository instead and are
served via `ksgAssetImage(...)`.

While building the dungeon overview/landing page (#3299 / #3058) we needed an expansion icon next to the
dungeon name. Rather than use the legacy `$expansion->iconfile->getUrl()`, the landing page now uses:

```php
ksgAssetImage(sprintf('expansions/%s.png', $expansion->shortname))
```

This issue tracks removing the legacy `File` link for **expansions** so there's a single, consistent way
to render an expansion icon.

## Remaining work

1. Replace remaining reads of `$expansion->iconfile` with `ksgAssetImage('expansions/<shortname>.png')`:
   - `resources/views/admin/expansion/edit.blade.php:41` (current-image preview)
   - `resources/views/misc/table/affixrowtable.blade.php:43` (currently commented out — clean up or migrate)
2. Remove the upload UI/handling from the admin Expansion edit flow (`resources/views/admin/expansion/edit.blade.php`
   and the controller/model code that saves `icon_file_id`, `app/Models/Expansion.php:305-321`).
3. Drop the `iconfile` relation usage from `Expansion` (it inherits `HasIconFile`). Decide whether to
   stop using the trait on `Expansion` specifically, or keep the trait for the other models that still
   use it.
4. Add a migration to drop the `expansions.icon_file_id` column, and clean up any now-orphaned rows in the
   `files` table that were expansion icons.
5. Confirm every expansion has a matching `expansions/<shortname>.png` in the assets repo before removing
   the DB fallback.

## Risks / notes

- `HasIconFile` (`app/Models/Traits/HasIconFile.php`) and `icon_file_id` are **also** used by
  `CharacterClass`, `Affix`, and `CharacterClassSpecialization`. This ticket is scoped to **Expansion only** —
  do not remove the trait or touch those models unless a separate decision is made.
[...]

PR #3766, opened 2026-07-30
Add admin pages for affixes, seasons and affix groups

### LINKED ISSUE #3756: Rework AffixSeeder's ever-growing hardcoded affix-group data (follow-up to #3739)

Wotuu:
:robot: Part 2 of 2 for #3756. See the [proposal on the issue](https://github.com/RaiderIO/keystone.guru/issues/3756#issuecomment-5132205170) for the full plan.

Closes #3756 — this is the part that actually removes the hand-editing. Depends on #3765 (part 1, the DB→JSON export via `mapping:save`) only conceptually; this MR branches from `master` directly and can be reviewed independently.

## What

Full CRUD admin pages for `Affix`, `Season`, and `AffixGroup` (rotation weeks, nested under a season), following the existing `DungeonController`-style pattern rather than `ExpansionController::store()`'s `upsert()` (which bypasses `CacheModel` invalidation — all three models are cached).

- **Affixes** (`/admin/affixes`): list + new/edit over `key`, `affix_id`, `name`, `description`.
- **Seasons** (`/admin/seasons`): list + new/edit over the denormalized columns, plus a season↔dungeon multiselect (synced via `Season::syncDungeons()`).
- **Affix groups**: nested under a season's edit page. There's no repeatable-form-row widget in admin today, so each week is edited as **6 fixed numbered slots** (1 required + 5 optional, each an affix + key_level pair), mirroring `admin/dungeonspeedrunrequirednpc/new.blade.php`. Save recreates the coupling rows in submitted slot order via `AffixGroup::syncAffixGroupCouplings()`.

## Notable decisions

- **Sync via individual model deletes, not mass query deletes.** `Season::syncDungeons()` and `AffixGroup::syncAffixGroupCouplings()` both delete existing rows one at a time (`$model->delete()` in a loop) rather than `$relation()->delete()`, so each row's `CacheModel` cache is properly invalidated. A mass query delete bypasses Eloquent events entirely.
- **New affixes get `icon_file_id = -1`.** `mapping:save` (#3765) doesn't export `icon_file_id`, and the real seeder always creates its own `File` row pointing at a static `images/affixes/*.jpg` asset. `-1` is the same "no icon yet" sentinel `Team` already uses; there's no upload widget since the icon asset itself is a repo file, not admin-editable data.
- **`affix_group_count` is shown as a hint, not auto-derived.** Auto-deriving it from the actual group count on every save would silently rewrite the live rotation modulus for existing seasons (e.g. `SEASON_LEGION_TW_S1` intentionally declares 12 but seeds 2). The edit form shows "Currently has N affix group(s)" next to the (still manually-set) field instead.
- **`confirmed` defaults to `0` on save**, mirroring `DungeonController::store()`'s `??=` pattern, since an unchecked checkbox submits no key at all and `AffixGroup` has no cast for it.
- **Affix group ↔ season scoping is checked manually** (`abort(404)` if `$affixGroup->season_id !== $season->id`) since Laravel doesn't verify nested route-model bindings belong together.

## Known follow-up (not blocking this MR)

Admin-creating a rotation week on an *older* season appends a high `id`, which will eventually break MR1's "`affix_groups.id` contiguous and ascending by `season_id`" seeder round-trip assertion once both land and someone uses the admin page that way. Flagging for whoever picks that up next; not fixed here since MR1 hasn't merged yet and the two branches don't currently conflict.

## Test plan

- [x] New feature tests: `AdminAffixControllerTest`, `AdminSeasonControllerTest`, `AdminAffixGroupControllerTest` (list/new/edit, savenew/update round-trips, dungeon sync, coupling order, `confirmed` unchecked→false, season-scoping 404). All pass.
- [x] `tests/Feature/App/Service/Season/` (63 tests) still green.
- [x] `composer run fix` / `composer run analyse` — only pre-existing, unrelated `MDT2Codec.php`/CBOR vendor-skew errors remain (documented, not a CI failure).
- [x] Verified all three pages (list/edit/new) in headless Chrome — screenshots below.
- [ ] Cold independent review
- [ ] Green CI

## Screenshots

Affixes list, affix edit, seasons list, season edit (with rotation weeks), and affix group edit (fixed slots, prefilled from an existing week) all verified working end-to-end against a real admin session.


Wotuu:
:robot: Verification screenshots (headless Chrome, admin session).

**Affixes list** (`/admin/affixes`)
affixes list

**Season edit with rotation weeks** (`/admin/season/17`)
season edit

**Affix group edit — fixed slots, prefilled from an existing week** (`/admin/season/17/affixgroup/163`)
affixgroup edit

Wotuu:
:robot: Cold review: 3 findings.

1. **[Fixed]** \`seasonal_affix_id\` \`<select>\` had no bound value and no "none" option — editing any existing season and submitting, even for an unrelated field, silently overwrote it with the alphabetically-first affix. Added a \`-1\` "none" sentinel option (matching the affix-group slot pattern), bound the current value explicitly, normalize \`-1\` → \`null\` on save. Also corrected \`Season::$seasonal_affix_id\`'s PHPDoc (was wrongly non-nullable), which in turn let a stale \`@phpstan-ignore\` in \`AffixGroupEaseTierService\` be removed. Two new regression tests assert the "None" option is preselected for a null season and that a real value round-trips.
2. **[Fixed]** Two season tests cleaned up via a mass query-builder delete (\`$season->seasonDungeons()->delete()\`) instead of \`Season::syncDungeons([])\` — bypasses the same cache-invalidation-on-delete rule this PR's own sync methods exist to satisfy. Switched both.
3. **[Investigated, not changed]** Affix-group fixed slots renumber when a gap is left mid-list (e.g. slots 1,2,4 filled → reopening shows 1,2,3). Verified empirically: resubmitting an *unmodified* form is idempotent — no data loss, since only the order of couplings is persisted, not a slot number. The DB has never had a concept of "slot identity," so a schema change would be needed to actually preserve gaps, which is out of scope here. Added a hint under the affix slots explaining that empty slots are removed and remaining entries renumber on save, to head off the confusion risk the review correctly flagged.

Commit with the fixes: 7be0d8ac0.

Wotuu:
:robot: Cold review: 6 findings (2 blocking/should-fix bugs fixed, 2 test-hygiene fixed, 2 non-blocking nits)

A fresh-context agent (no shared context with the implementing session) reviewed the full diff via `gh pr diff 3766`, including files outside the diff (models, DB schema, other callers of the affected models) to check for side effects.

**Fixed:**
1. **Blocking, confirmed bug** — `AffixGroup::$fillable` was missing `expansion_id`, so every affix group created/edited through the new admin pages silently persisted `expansion_id = 0` instead of the season's real expansion (Laravel drops non-fillable mass-assigned attributes without erroring since this app doesn't call `preventSilentlyDiscardingAttributes()`). Verified empirically via tinker before fixing. Added `expansion_id` to `$fillable` and a regression assertion in `AdminAffixGroupControllerTest::savenew_givenValidData_createsAffixGroupWithCouplingsInSlotOrder`.
2. **Should-fix** — `AffixController::store()`'s update branch didn't strip `id` from validated data, and `Affix::$fillable` includes `id`. A PATCH request explicitly sending an `id` matching a currently-unused declared `Affix::ALL` value would pass validation and reassign an existing affix's primary key (admin-only, and only exploitable during the narrow window where a new `AFFIX_*` constant is declared but not yet seeded — but real, and this PR is exactly what creates that window). Defensively `unset($validated['id'])` before `update()`, mirroring `SeasonController` (whose `id` isn't fillable, so it was never affected). Not covered by an HTTP-level regression test: with the fixture DB fully seeded (all 40 declared affix ids in use), no `id` value can currently pass the `Rule::in()` gate to reach the vulnerable code path — same permanent constraint documented for the id-availability tests below.
3. Two `get_asNonAdmin_returnsForbidden` tests (Season + Affix) created a `User::factory()` row with no cleanup — this app's test DB is real, persistent, and not transaction-wrapped, so every CI run leaked a user + role-pivot row. Wrapped in `try/finally`.
4. `Season::getAvailableIds()` re-derived the declared-ids set via `ReflectionClass` instead of reading the existing `Season::ALL_SEASONS` constant directly (unlike `Affix::getAvailableIds()`, which already used `Affix::ALL`). Simplified for consistency.

**Not changed (non-blocking):**
5. `lang/en_US/view_admin.php`'s `affixgroup` key isn't adjacent to `affix` alphabetically — pre-existing from an earlier part of this PR, cosmetic only.
6. Changing a season's expansion via the edit form doesn't cascade to its existing `AffixGroup` rows' `expansion_id` — narrow/unlikely scenario (expansion is normally set once), not worth blocking on.

**Confirmed correct** (reviewer explicitly checked, no changes needed): the `seasonal_affix_id` `Rule::in` tightening, affix-group coupling slot-skip logic including closure-capture semantics, `common.dungeon.select` reuse matches the existing `admin/npc/edit.blade.php` pattern, no other production callers of `Season::create()`/`Affix::create()` rely on auto-increment id assignment, and no unique-affix_id regression risk (seed data already has a legitimate duplicate `affix_id` between `AFFIX_UNKNOWN`/`AFFIX_RELENTLESS`).

Visual verification: season list (sorted by id desc), season create page, affix create page (both showing the "no available ids" warning, the only reachable state with the current fully-seeded fixture DB), and the season edit page's dungeon picker (now the shared Tom Select widget) were all verified in headless Chrome this session — screenshots available on request, not embedded here since they didn't change since the prior review round.

app/Http/Controllers/AffixController.php
```diff
@@ -0,0 +1,115 @@
...
+        $validated = $request->validated();
+
+        if ($affix === null) {
+            // No icon is uploaded through the admin panel - mapping:save does not export icon_file_id,
+            // and the real seeder always creates its own File row pointing at a static images/affixes/*.jpg asset.
+            $validated['icon_file_id'] = -1;
+            // The id must be one of Affix::ALL's values - explicit rather than auto-incrementing so
+            // that a new affix's identity is a deliberate code change, matching how the rest of the
+            // codebase references affixes by their Affix::AFFIX_* constant (see AffixFormRequest).
+            $affix = Affix::create($validated);
```
Wotuu, line 30: You are right. But I think we should remove the has icon file trait from the affix model. We no longer use it. Any models that use files or have assets have them statically provided through the assets project. I don't want the admin pages to have the ability to change the icons. Yeah, so this was a leftover from when that was possible. We should get rid of it. Same goes for factions, by the way. See that they too also have the icon file trait and character classes and character class specializations. Those four all need to not have the has icon file trait. If you can file up if you can create a new github issue as a follow-up, that would be nice.

app/Http/Controllers/AffixGroupController.php
```diff
@@ -0,0 +1,146 @@
...
+            'expansion_id'   => $season->expansion_id,
+            'seasonal_index' => $validated['seasonal_index'] ?? null,
+            'confirmed'      => $validated['confirmed'] ?? 0,
+        ];
+
+        if ($affixGroup === null) {
+            $affixGroup = AffixGroup::create($attributes);
+        } else {
+            $affixGroup->update($attributes);
+        }
```
Wotuu, line 35: Can we upsert?

app/Http/Controllers/SeasonController.php
```diff
@@ -0,0 +1,165 @@
...
+            // matching the constants the rest of the codebase references by name (e.g.
+            // Season::SEASON_TWW_S1). MySQL's LAST_INSERT_ID() is not updated by an explicit
+            // AUTO_INCREMENT value on a bulk query-builder insert(), but a normal Eloquent save()
+            // does correctly report it back - verified empirically before relying on this.
+            $id = (int)$validated['id'];
+            unset($validated['id']);
+
+            $season     = new Season($validated);
+            $season->id = $id;
+            $season->save();
```
Wotuu: I still don't understand why the ID cannot just be part of the validated array. Why does it need to be plucked out and then reassigned? Can't we just use Season::create?

app/Service/AffixGroup/AffixGroupEaseTierService.php
```diff
@@ -246,7 +246,7 @@ public function getAffixGroupByString(string $affixString): ?AffixGroup
         $affixes = collect(explode(', ', $affixString));
 
         // Filter out properties that don't have the correct amount of affixes
-        if ($affixes->count() === 3 + (int)($currentSeason->seasonal_affix_id !== null)) { // @phpstan-ignore notIdentical.alwaysTrue
+        if ($affixes->count() === 3 + (int)($currentSeason->seasonal_affix_id !== null)) {
```
Wotuu, line 249: I wonder if this check is incorrect. You'd need to file a follow-up issue for this, but the affix count of an affix group is no longer necessary three or four, which this if applies implies. In the current season it has five. It has Lindormi's Guidance, Fortified, Tyrannical, Xal'Atath's Bargain and Xal'Atath's Guile. So something in this check is not going right and it needs a follow-up issue to figure out what and to figure out how to fix this code.

resources/views/admin/season/list.blade.php
```diff
@@ -0,0 +1,72 @@
...
+
+<?php
+
+use App\Models\Season;
+use Illuminate\Support\Collection;
+
+/**
+ * @var Collection<int, Season> $models
+ */
+?>
```
Wotuu: Can this be moved to the very top of the file?

resources/views/admin/affix/list.blade.php
```diff
@@ -0,0 +1,71 @@
...
+
+<?php
+
+use App\Models\Affix;
+use Illuminate\Support\Collection;
+
+/**
+ * @var Collection<int, Affix> $models
+ */
+?>
```
Wotuu: Can this be moved to the very top of the file?

Wotuu (commented):
.

app/Http/Controllers/AffixController.php
Wotuu: :robot: Removed `use HasIconFile;` from `Affix` in `e811fd86f` (`app/Models/Affix.php`) - the trait's `iconfile()` relation and `saveUploadedFile()` upload method are gone, so there's now no code path on `Affix` that could ever change an icon through the admin panel.

One thing I kept, deliberately: `affixes.icon_file_id` itself. `SHOW CREATE TABLE affixes` shows it's `int NOT NULL` with no default and no unique business key besides `id`, so dropping the column needs its own migration (mirroring the `Expansion` cleanup in #3350/`13650b395`) rather than something I want to do as an unattended change here. The column stays as a vestigial `-1` sentinel for now (PHPDoc on `Affix::$icon_file_id` explains this), and `AffixController::store()`/`AffixSeeder` are unchanged.

Filed the follow-up for `Faction`, `CharacterClass`, and `CharacterClassSpecialization` as #3775, and included a note there that the `affixes.icon_file_id` column-drop is fair game to bundle in at the same time if whoever picks it up wants to finish the job properly.

Tests (`AffixTest`, `AdminAffixControllerTest`, `AdminAffixGroupControllerTest`, plus the full `Affix`/`Controller`/`Admin` groups - 406 tests) all still pass unchanged, since nothing outside the trait itself referenced `iconfile`/`saveUploadedFile` for `Affix`.

app/Http/Controllers/AffixGroupController.php
Wotuu: :robot: Looked closely at this - the answer is no, not without giving up either the return value or cache correctness.

`AffixGroup extends AffixGroupBase extends CacheModel`, same as `Expansion`. `upsert()` is a raw query-builder statement - it never fires Eloquent's `saving`/`saved` events, so `laravel-model-caching`'s cache invalidation never runs. That's the exact bug category the PR description calls out for `ExpansionController::store()`'s `upsert()` usage; I didn't want to reintroduce it here.

But there's also a more basic, mechanical problem specific to this method: `store()` needs the actual `AffixGroup` instance right after the write - line 37 immediately calls `$affixGroup->syncAffixGroupCouplings(...)`, and `savenew()` redirects to `route('admin.affixgroup.edit', ['affixGroup' => $affixGroup])`. `ExpansionController::store()` gets away with `upsert()` only because it can re-query the row afterwards by a natural unique key (`Expansion::where('shortname', ...)->firstOrFail()`). `AffixGroup` has no such key - `SHOW CREATE TABLE affix_groups` shows no unique index besides `id`, and `AffixSeeder` seeds plenty of rows sharing the same `(season_id, seasonal_index)` (e.g. the BFA S1 groups). So after an `upsert()` there'd be no reliable way to fetch back the row that was just created/updated, on top of the cache-invalidation gap.

Keeping the plain `create()`/`update()` here is intentional, not an oversight.

app/Http/Controllers/SeasonController.php
Wotuu: :robot: Traced this all the way through rather than just restating the code shape - the short answer is: making that possible would reopen a bug we just fixed for `Affix`.

`Season::$fillable` deliberately does **not** include `id`. This app never calls `preventSilentlyDiscardingAttributeFill()`, so a non-fillable key inside a mass-assigned array is silently dropped, not thrown. That means `Season::create($this->validated())` (with `id` still in the array) would just drop `id` and let MySQL's `AUTO_INCREMENT` on `seasons.id` hand out the next free value instead of the intended `Season::SEASON_*` constant - breaking the invariant the rest of the codebase relies on (`Season::SEASON_TWW_S1` etc. referencing specific ids by name).

I also checked whether the alternative - making `id` fillable so `create()` works - is safe. It isn't: `SeasonFormRequest::rules()` only *requires* `id` on create (`Rule::requiredIf($isCreate)`), it doesn't forbid it on update. If `id` were fillable, a PATCH request carrying an `id` from `Season::getAvailableIds()` would pass validation and mass-assign a new primary key onto an existing season through `$season->update($validated)` - exactly the same class of bug already fixed for `Affix` in `21ba580f5` (`AffixController::store()` now explicitly `unset()`s `id` on update for this reason). Keeping `id` out of `Season::$fillable` entirely is a stronger version of that same guard.

One more thing worth calling out: I originally suspected the existing comment's `LAST_INSERT_ID()`/"bulk insert" reasoning might actually be the real blocker, but empirically it isn't - I verified in a rolled-back transaction that Eloquent's `insertGetId()` correctly reports back an *explicit* `AUTO_INCREMENT` id, so that comment was answering a different question than the one you asked. I rewrote it in `574ba8d5c` to state the actual reason (fillable/mass-assignment, as above), and added a regression test (`savenew_givenAvailableDeclaredId_persistsThatExactIdRatherThanAnAutoIncrementedOne`) that frees up a declared season id inside a transaction (always rolled back) and asserts the created season gets that exact id rather than an auto-incremented one.

So: left the code as-is (unset + direct attribute assignment), but fixed the misleading comment and added a test that pins down the actual invariant.

app/Service/AffixGroup/AffixGroupEaseTierService.php
Wotuu: :robot: You're right, and I confirmed it against live data rather than just filing on your say-so. For season 17 (Midnight S1, the current season), `seasons.seasonal_affix_id` is `NULL` (so the check requires exactly 3 affixes), but every one of season 17's `affix_groups` rows actually has 5 rows in `affix_group_couplings` - e.g. affix_group_id 163: Lindormi's Guidance, Fortified, Tyrannical, Xal'atath's Bargain: Devour, Xal'atath's Guile. So the `if ($affixes->count() === 3 + (int)($currentSeason->seasonal_affix_id !== null))` check at line 249 is always false for the current season, and `getAffixGroupByString()` silently returns `null` for every call instead of finding a match.

Filed the follow-up as **#3776**: https://github.com/RaiderIO/keystone.guru/issues/3776 - it has the verified counts, the quoted code, and what still needs figuring out (whether the fix should derive the expected count from real affix-group membership instead of a hardcoded 3-or-4, and whether `seasonal_affix_id` is even still a meaningful "extra affixes" signal now that a season can carry more than one bonus-style affix while leaving it `null`).

Did not touch `AffixGroupEaseTierService.php` in this PR per your instruction. I also double-checked the `key`/translated-`name` matching in the loop right below the count check (`$affixes->search($affix->key)` vs `__($affix->name, ...)` a few lines up) in case that was part of the bug too - it isn't: verified for every `Affix` row that `__($affix->name, [], 'en_US') === $affix->key`, so that part of the method is fine and the count check is the whole problem.

resources/views/admin/season/list.blade.php
Wotuu: :robot: Done in `574ba8d5c` - moved the `<?php use App\Models\Season; use Illuminate\Support\Collection; /** @var Collection<int, Season> $models */ ?>` block above `@extends(...)`, matching the convention already used by `admin/npc/list.blade.php`, `admin/dungeonroute/list.blade.php`, `admin/spell/list.blade.php` and `admin/user/list.blade.php`. Verified the page still renders correctly (`AdminSeasonControllerTest` - 11 tests, all green).

resources/views/admin/affix/list.blade.php
Wotuu: :robot: Done in `574ba8d5c` - moved the `<?php use App\Models\Affix; use Illuminate\Support\Collection; /** @var Collection<int, Affix> $models */ ?>` block above `@extends(...)`, same convention/commit as the season list view. Verified the page still renders correctly (`AdminAffixControllerTest` - 11 tests, all green).

app/Service/AffixGroup/AffixGroupEaseTierService.php
Wotuu: :robot: Correction to my reply above: I said "Did not touch `AffixGroupEaseTierService.php` in this PR", which is wrong - `git diff origin/master...HEAD -- app/Service/AffixGroup/AffixGroupEaseTierService.php` shows this PR *did* touch that exact line, in an earlier commit in this branch (`d2e0aa130`, "Fix cold-review findings: seasonal_affix_id had no 'none' option"). That commit removed a `// @phpstan-ignore notIdentical.alwaysTrue` annotation from line 249 - not the count logic itself, just the ignore comment - because fixing `seasonal_affix_id`'s nullability elsewhere in that same commit made the `!== null` check genuinely reachable both ways, so PHPStan stopped flagging it as always-true.

That's actually useful corroborating context for the bug you found: before `d2e0aa130`, PHPStan considered `$currentSeason->seasonal_affix_id !== null` **always true**, meaning the count check effectively hardcoded "always expect exactly 4 affixes" and needed an explicit ignore to suppress the resulting warning. I've added this history to #3776 so whoever picks it up has the full picture. The count logic itself is unchanged by this PR - what I said about not touching *that* still stands, I just phrased "touched the file" too broadly.

PR #3768, opened 2026-07-30
Handle a Patreon member missing the email attribute entirely

### LINKED ISSUE #3767: patreon:refreshmembers crashes on undefined array key "email"
:robot: Production error surfaced now that #3748/#3750 properly reports scheduler-command exceptions instead of swallowing them:

```
Undefined array key "email"
On /var/www/html/app/Service/Patreon/PatreonService.php:152
command: patreon:refreshmembers
memberId: 5160988d-9a4c-4c97-8909-31b28d4c91a1
```

`PatreonService::applyPaidBenefitsForMember()` reads `$member['attributes']['email']` directly. The
code already handles an *empty* email (`empty($memberEmail)` → `MemberNotLinked`), but not a
*missing* key — for at least this one member, the `attributes` array had no `email` key at all
(rather than an `email` key set to `null`), which triggered an `ErrorException`. `RefreshMembershipStatus::handle()`
already wraps each member in its own per-member try/catch (from #3750), so this didn't abort the
whole run — it reported the failure and moved on to the next member, then exited the command as
FAILURE with this member named. Still worth fixing so this member (and any other with the same
attribute shape) is handled cleanly instead of showing up as a reported failure every run.

Fix: use `$member['attributes']['email'] ?? null` so a missing key falls through to the existing
empty-email handling.

Wotuu:
:robot: Closes #3767

## Summary
- `patreon:refreshmembers` crashed on `Undefined array key "email"` at `PatreonService.php:152` for
  one member. `RefreshMembershipStatus::handle()` already wraps each member in its own
  `try { applyPaidBenefitsForMember(...) } catch (Throwable)` (from #3750), so the run itself kept
  going for every other member — the affected member's failure was reported (per #3750's exception
  reporting) and the command exited FAILURE with that member named, rather than the whole run
  aborting.
- `applyPaidBenefitsForMember()` already handles an *empty* email (`empty($memberEmail)` →
  `MemberNotLinked`), but read `$member['attributes']['email']` unguarded, so a member whose
  `attributes` array had no `email` key at all (rather than a null value) threw instead of
  falling into that existing handling.
- Fix: `$member['attributes']['email'] ?? null`. Confirmed a missing key is still logged with the
  member's id attached - `applyPaidBenefitsForMemberStart($member['id'])` registers it in the
  structured-logging grouped context, which stays attached to every log line (including the
  empty-email log) until `end()` is called - so this doesn't repeat the #3748 trap of turning a
  loud failure into a silent, unidentifiable one.

## Test plan
- [x] Added `applyPaidBenefitsForMember_givenMemberAttributesWithoutEmailKey_returnsMemberNotLinked`,
      verified it reproduces the exact production error before the fix and passes after
- [x] `php artisan test --compact --filter=PatreonServiceTest` (4 passed)
- [x] `composer run analyse` - no new PHPStan errors (pre-existing CBOR/vendor-lock skew noise only)
- [x] `composer run fix` - no changes needed

### DIFF
--- a/app/Service/Patreon/PatreonService.php
+++ b/app/Service/Patreon/PatreonService.php
@@ -149,7 +149,9 @@ class PatreonService implements PatreonServiceInterface
         try {
             $this->log->applyPaidBenefitsForMemberStart($member['id']);
 
-            $memberEmail = $member['attributes']['email'];
+            // The email attribute can be entirely absent for a member instead of present-but-null, so this can't
+            // rely on the key always being set (#3767) - a missing key is handled the same as an empty value below
+            $memberEmail = $member['attributes']['email'] ?? null;
 
             if (empty($memberEmail)) {
                 $this->log->applyPaidBenefitsForMemberEmptyMemberEmail();
--- a/tests/Feature/App/Service/Patreon/PatreonServiceTest.php
+++ b/tests/Feature/App/Service/Patreon/PatreonServiceTest.php
@@ -113,6 +113,27 @@ final class PatreonServiceTest extends PublicTestCase
         $this->assertDatabaseHas('patreon_user_benefits', ['id' => $patreonUserBenefit->id]);
     }
 
+    #[Test]
+    public function applyPaidBenefitsForMember_givenMemberAttributesWithoutEmailKey_returnsMemberNotLinked(): void
+    {
+        // Arrange
+        $member = [
+            'id'            => 'member-1',
+            'type'          => 'member',
+            'attributes'    => [],
+            'relationships' => ['currently_entitled_tiers' => ['data' => []]],
+        ];
+
+        /** @var PatreonServiceInterface $patreonService */
+        $patreonService = $this->app->make(PatreonServiceInterface::class);
+
+        // Act
+        $result = $patreonService->applyPaidBenefitsForMember([], [], $member);
+
+        // Assert
+        $this->assertSame(ApplyPaidBenefitsForMemberResult::MemberNotLinked, $result);
+    }
+
     private function createPatron(): void
     {
         $this->user = User::factory()->create();


app/Service/Patreon/PatreonService.php
```diff
@@ -149,7 +149,9 @@ public function applyPaidBenefitsForMember(array $campaignBenefits, array $campa
         try {
             $this->log->applyPaidBenefitsForMemberStart($member['id']);
 
-            $memberEmail = $member['attributes']['email'];
+            // The email attribute can be entirely absent for a member instead of present-but-null, so this can't
+            // rely on the key always being set (#3767) - a missing key is handled the same as an empty value below
+            $memberEmail = $member['attributes']['email'] ?? null;
```
Wotuu, line 154: :robot: The fix itself is correct and worth having — but the stated blast radius in the PR body (and in #3767) is wrong, and since `(#3767)` is now a durable rationale marker in this file, it's worth correcting before it misleads a future reader.

> aborting the entire run for every member, not just the affected one

That can't be true of the code this PR is based on. `RefreshMembershipStatus::handle()` already wraps every member in a per-member `try { $patreonService->applyPaidBenefitsForMember(...) } catch (Throwable $throwable)` (app/Console/Commands/Scheduler/Patreon/RefreshMembershipStatus.php:66-91), which reports the first few failures and keeps iterating, then fails the command at the end via `$failedMemberIds`. Those exact lines came from `57d4f14ec` (#3750) — the same commit whose exception reporting surfaced this error in the first place.

The production report in #3767 proves it rather than just implying it: `memberId: 5160988d-…` can only enter the report context through `reportThrowable($throwable, ['memberId' => $memberId])`, which lives *inside* that per-member catch block (`SchedulerCommand::reportThrowable()` merges `command` + the passed context into `Context`). So the `ErrorException` was caught per member, the loop continued for everyone else, and the command exited FAILURE with that one member named in the summary.

So the real effect of this change is narrower but still worth shipping: it stops one member from producing a reported exception + a FAILURE exit, and routes it into the existing `MemberNotLinked` handling instead. Suggest rewording that bullet so #3750's containment isn't recorded as having failed.

Wotuu:

The code change itself is correct and nothing here blocks the PR. The one finding is about the stated blast radius in the PR body / #3767 ("aborting the entire run for every member"), which #3750's per-member `try/catch` in `RefreshMembershipStatus` already contains - worth rewording so the `(#3767)` rationale doesn't record #3750's containment as having failed.

Checked and deliberately not raised:
- **Does the new test actually fail before the fix?** Yes - verified that with the framework bootstrapped, reading `$member['attributes']['email']` from an `attributes` array without that key throws `ErrorException: Undefined array key "email"` via Laravel's `HandleExceptions` error handler, so the test does reproduce the production error pre-fix. Naming, AAA structure, the class-level `#[Group('Patreon')]` and the no-DB-records/teardown path all match the repo conventions.
- **The structured-logging claim holds.** `applyPaidBenefitsForMemberStart($member['id'])` -> `start()` -> `addContext()` keeps `memberId` in the grouped context (and mirrors it into Laravel's `Context`) until the `finally { ...End() }`, so the `applyPaidBenefitsForMemberEmptyMemberEmail` line carries it. That line is logged at `info` while the "no link found" case is `debug`, so the missing-email case stays distinguishable in the logs.
- **Other unguarded reads nearby** (`$member['id']`, `$member['relationships']['currently_entitled_tiers']['data']`) - not flagged: the production error named `email`, not `attributes`, so the JSON:API envelope was intact and only a sparse-fieldset *field* went missing, and `relationships` is an explicit `include` on both request paths. `Stopwatch::stop()` is also safe on an unstarted timer, so even a hypothetical missing `id` stays contained by the caller's per-member catch rather than masking itself in the `finally`.
- **`$member` can be `null` on the `linkToUserAccount` path** (`collect(...)->first()` passed into an `array`-typed parameter) - real, but pre-existing and untouched by this diff.

Wotuu (commented):
Ok

app/Service/Patreon/PatreonService.php
Wotuu: :robot: Fixed — reworded the PR body and issue #3767 body so neither claims this aborted the whole run. Both now describe the actual effect: `RefreshMembershipStatus::handle()`'s per-member catch (from #3750) already contained it, this member was reported and the command exited FAILURE with them named, and the fix stops that member from being reported as a failure at all going forward.

PR #3769, opened 2026-07-30
#3761 Keep the dungeon selector on the current season until the next one starts

### LINKED ISSUE #3761: Dungeon selector fully switches to next season's dungeons as soon as it's seeded, not when it starts
:robot:

## Background

Raised while triaging PR #3739's (Midnight Season 2 dry run) pre-merge safety question — Wotuu
asked whether merging #3739 now, before WoW 12.1 ships, would "throw the site in disarray" and
specifically wants current Season 1 dungeons to stay selectable.

`DungeonService::getDungeonsForGameVersion()` (`app/Service/Dungeon/DungeonService.php:115`) is:

```php
$currentSeason = $this->seasonService->getCurrentSeason($gameVersion->expansion);
$nextSeason    = $currentSeason === null ? null : $this->seasonService->getNextSeason($currentSeason);
return ($nextSeason ?? $currentSeason)?->dungeons()->get() ?? $gameVersion->expansion->dungeons;
```

`SeasonService::getNextSeason()` returns the first season with a `start` date after the current
season's — it does **not** check whether that season has actually started yet, only that its row
exists and is chronologically next. So the dungeon selector fully switches to `$nextSeason`'s
dungeon list the moment a next-season row exists in the DB, replacing (not merging with) the
current season's dungeon list.

## Why this matters for #3739

#3739 seeds Season 18 (Midnight Season 2) with 8 dungeons and a placeholder start date currently
set weeks out (2026-10-19, per the seeder). As soon as that PR merges to master and the seeder
runs, `getNextSeason()` finds Season 18 and `getDungeonsForGameVersion()` switches the **entire
site's dungeon selector** to those 8 (not-yet-live) dungeons, dropping every current Season 1
dungeon from view — weeks before 12.1 actually ships on 2026-08-11, and regardless of any
per-dungeon "disable new dungeons" logic, since this operates one level higher (season-wide, not
per-dungeon).

This is a different, more precise mechanism than the "disable the 8 S2 dungeons" checklist item in
#3739's body assumed — disabling individual dungeon rows wouldn't fix this, because the selector
swap happens at the season level before any per-dungeon check runs.

## Suggested fix directions (not decided — needs design call)

- Only let `getNextSeason()`'s result take over the selector once its `start` date has actually
  passed (or is within some short pre-launch window), rather than as soon as the row exists.
- Or: keep seeding/preparing next-season data ahead of time (as #3739 already needs to, for review
  and QA) but gate `getDungeonsForGameVersion()`'s season selection on `Carbon::now()` being past
  `$nextSeason->start`, falling back to `$currentSeason` otherwise.
- Whichever direction is chosen, add a regression test seeding a future season and asserting the
  dungeon selector still returns the current season's dungeons until that season's start date
  arrives.

## Scope note

This bug is not introduced by #3739 — the same swap-over-look-ahead behavior would fire for any
future season seeded this far ahead of its start date. #3739 is what surfaces it in practice
because it's the first time a season has been seeded this many weeks before its start.

Wotuu:
:robot:

Closes #3761

## What was wrong

`DungeonService::getDungeonsForGameVersion()` — which feeds the header dungeon context bar and the
dungeon context links on discover, explore, heatmap and route search — did this:

```php
$nextSeason = $currentSeason === null ? null : $this->seasonService->getNextSeason($currentSeason);

return ($nextSeason ?? $currentSeason)?->dungeons()->get() ?? $gameVersion->expansion->dungeons;
```

`getNextSeason()` only checks that a season is chronologically next, not that it has started. So the
selector **replaced** the current season's dungeons with the next season's the moment a next-season row
existed, however far out its start was. Since a season is seeded weeks ahead so its mapping can be
reviewed, that would drop every current season dungeon off the site long before the new ones are
playable — which is what blocks #3739 (Midnight Season 2 dry run).

The look-ahead was added in `881240700` (#3118) without a stated rationale, and master has had no future
season seeded since, so that branch has never actually fired in production.

## What changed

- **`DungeonService`** follows the current season only.
- **A "Next season" card** is added to the dungeon context bar, next to the existing "More" card. It
  shows once the next season starts within `keystoneguru.season.upcoming_visible_days` (default 14), so
  a season can be seeded months ahead without changing anything on the live site. `HeaderComposer` makes
  that decision; the blades only render what it hands them.
- **The dungeon selection page** (`common/dungeon/gridtabs`, where "More" and the new card lead) had the
  same problem one level down: it made the next season's tab the active one as soon as a season was
  seeded. It now opens on the current season, and opens the next season's tab only when asked for it
  (`?season=<id>`, which is what the new card links to).

## Deliberate choices

- **The next-season tab itself is not gated** on the window — only its auto-selection. It is a tab you
  have to click, it has been there for a long time, and it is how an upcoming season's mapping gets
  reviewed on staging before it starts.
- **The card links to the explore selection page from every page**, including heatmap and route search.
  Those pages have no `nextSeason` of their own to build a link from; the existing `dungeonContextLinks`
  mechanism can carry a per-page override later if this reads wrong.
- **`getDungeonsForGameVersionTest_givenAnUpcomingSeason_...` had its expectation flipped and renamed** —
  it asserted the exact behaviour being removed. Its "load both seasons as a collection" arrangement is
  kept, since that is what covers the #3746 lazy-loading regression.
- `getDungeonContext()` resolved the default dungeon from the current season only, so the header could
  list dungeons that the selected dungeon was not among. Pre-existing; this change removes it.

## Tests

- `GetDungeonsForGameVersionTest` — upcoming season, no upcoming season, no current season, plus an
  end-to-end case that seeds a real future season through the container-resolved services.
- `HeaderComposerTest` — the card is set inside the window and absent outside it.
- `DungeonListTest` — the card renders (and does not) in `common/dungeon/list`.
- `DungeonExploreControllerTest` — `/explore/retail/select` opens on the current season's tab with a
  future season seeded, and on the requested season's tab with `?season=<id>`.

While writing those: `SeederModel` silently refuses `$model->delete()` for non-admins, so the usual
`try/finally` cleanup leaves rows behind and poisons the shared test database. That, and the two caches
that hide freshly created rows from tests, are now written down in the `writing-tests` skill.

### DIFF
--- a/.claude/skills/writing-tests/SKILL.md
+++ b/.claude/skills/writing-tests/SKILL.md
@@ -59,11 +59,13 @@ try {
 
 ### `SeederModel` models silently refuse `$model->delete()`
 
-Models using the `SeederModel` trait (`Season`, `SeasonDungeon`, `Dungeon`, `Expansion`, …) register a
-`deleting` hook that returns `false` unless the **authenticated** user is an admin. `$model->delete()`
-then does nothing and returns `false` — no exception — so a `finally` that looks correct leaves the row
-behind and poisons the shared DB for every later test (a leftover `Season` with no dungeons breaks any
-test doing `Season::orderByDesc('id')->first()->dungeons()`).
+~51 models use the `SeederModel` trait (`Season`, `SeasonDungeon`, `Expansion`, `EnemyPack`,
+`GameServerRegion`, … — check with `grep -rln "use SeederModel;" app/Models/`; `Dungeon` is **not** one of
+them). It registers a `deleting` hook that returns `false` unless the **authenticated** user is an admin.
+`$model->delete()` then does nothing and returns `false` — no exception — so a `finally` that looks
+correct leaves the row behind and poisons the shared DB for every later test (a leftover `Season` with no
+dungeons breaks any test doing `Season::orderByDesc('id')->first()->dungeons()`). The hook exempts
+`MappingVersion`, `Floor` and `Enemy`, which delete normally.
 
 Clean these up through the query builder, which does not fire model events:
 
@@ -80,7 +82,22 @@ Two caches make freshly created/deleted rows invisible in tests, and neither is 
 - **laravel-model-caching** (every `CacheModel`) — flush with `new Model()->flushCache()`.
 - **`RemembersToFile`** (`ViewService::cachedGlobal`, map context, …) writes to `Cache::store('tmp_file')`,
   a **file** store that survives between test runs. If your test touches `ViewService`, flush it:
-  `Cache::store('tmp_file')->flush()`.
+  `Cache::store('tmp_file')->flush()`. `phpunit.xml` points `CACHE_FILE_PATH` at `/tmp/phpunit_cache` so
+  that flush cannot reach the cache of the app running out of the same checkout.
+
+Create fixtures **inside** the `try`, not before it, so a failure halfway through setup still cleans up:
+
+```php
+$season = null;
+
+try {
+    $season = Season::create([...]);
+    SeasonDungeon::create([...]);   // this throwing must not leave $season behind
+    // Act + Assert
+} finally {
+    if ($season !== null) { /* delete */ }
+}
+```
 
 ## Creating users & roles (Laratrust)
 
--- a/.claude/skills/writing-tests/SKILL.md
+++ b/.claude/skills/writing-tests/SKILL.md
@@ -57,6 +57,31 @@ try {
 
 `DungeonRoute` has no soft-deletes, so `delete()` truly removes the row.
 
+### `SeederModel` models silently refuse `$model->delete()`
+
+Models using the `SeederModel` trait (`Season`, `SeasonDungeon`, `Dungeon`, `Expansion`, …) register a
+`deleting` hook that returns `false` unless the **authenticated** user is an admin. `$model->delete()`
+then does nothing and returns `false` — no exception — so a `finally` that looks correct leaves the row
+behind and poisons the shared DB for every later test (a leftover `Season` with no dungeons breaks any
+test doing `Season::orderByDesc('id')->first()->dungeons()`).
+
+Clean these up through the query builder, which does not fire model events:
+
+```php
+} finally {
+    Season::query()->whereKey($season->id)->delete();
+    // No model events fired, so laravel-model-caching does not invalidate on its own
+    new Season()->flushCache();
+}
+```
+
+Two caches make freshly created/deleted rows invisible in tests, and neither is the `array` store
+`phpunit.xml` configures:
+- **laravel-model-caching** (every `CacheModel`) — flush with `new Model()->flushCache()`.
+- **`RemembersToFile`** (`ViewService::cachedGlobal`, map context, …) writes to `Cache::store('tmp_file')`,
+  a **file** store that survives between test runs. If your test touches `ViewService`, flush it:
+  `Cache::store('tmp_file')->flush()`.
+
 ## Creating users & roles (Laratrust)
 
 - **Admin**: `User::findOrFail(1)` — the seeded admin. Assert it defensively:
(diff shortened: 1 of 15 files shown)


Wotuu:
:robot:

## Visual verification

Driven in headless Chrome against this branch's stack. A season starting in 5 days (inside the
14-day window) plus 4 dungeons was seeded temporarily, then removed again — the shared database is
back to exactly what it was, and master's header is unchanged.

**Before** — no upcoming season seeded, 8 current-season dungeons:

before

**After** — an upcoming season inside the window. Every current dungeon is still there; the new card
sits at the end:

after

**On a map page**, where the "More" card also renders — 10 cards, still readable:

map

**Where the card leads** — the dungeon selection opened on the upcoming season's tab, showing just
its dungeons. Without the `?season=` parameter the same page opens on Season 1 instead:

select

Seeding the same season 60 days out (outside the window) leaves the header exactly like the "before"
shot — no card, no other change.

Wotuu:
:robot:

## Review round (fresh-context review, findings resolved in `e70c75ddb`)

**Fixed — three real defects:**

1. **No `has_seasons` guard.** The upcoming season is resolved per *region*, never per game version,
   while the link was built with the visitor's game version. On the six non-retail game versions that
   produced a card advertising the retail season, linking to a selection page that hides every season
   tab for that game version — a card that does nothing. `HeaderComposer` now only advertises a season
   when the visitor's game version has them.
2. **Stray ease-tier badge.** Blade's `@include` inherits the enclosing scope, so `$thisWeekTier` — set
   per dungeon in the loop — leaked into the cards rendered after it: the "Next season" card came out
   carrying the last dungeon's "what's easy this week" badge and its archon.gg tooltip. Both that card
   and the pre-existing "More" card now pass `thisWeekTier => null` explicitly. `DungeonListTest` only
   missed it because it rendered with an empty dungeon collection, so the loop never ran; it now renders
   with real dungeons + ease tiers and asserts one badge per dungeon card and none on the other two.
   Confirmed the assertion fails without the fix.
3. **Fixtures created before the `try`.** In all three new test classes a failure between
   `Season::create()` and `SeasonDungeon::create()` would have left a season behind in the shared test
   database — the exact poisoning the new SKILL.md section warns about. Creation moved inside the `try`,
   cleanup null-guarded, and season_dungeons now deleted by `season_id` so partial fixtures clean up too.

**Also addressed:** `phpunit.xml` now points `CACHE_FILE_PATH` at `/tmp/phpunit_cache` — the `tmp_file`
store is a file cache `CACHE_STORE=array` does not cover, so the tests were reading, writing and flushing
the cache of the app running out of the same checkout. Added a test pinning that the window is really
config-driven (a hardcoded `14` would have passed the existing pair), and one for the game-version guard.
Corrected the SKILL.md claim that `Dungeon` uses `SeederModel` — it does not (51 models do; the trait
exempts `MappingVersion`, `Floor` and `Enemy`).

**Not changed, deliberately:**

- *Move the window rule out of the view composer into `SeasonService`.* Fair point that "upcoming" now
  means two different things (`getUpcomingSeasonForDungeon()` is un-windowed), but the rule is a
  presentation decision about one specific card, and moving it invites the next reader to apply it to
  `gridtabs`, which must **not** be windowed. Happy to move it if you disagree.
- *`$nextSeason->expansion` is one lazy query per request.* True, but only while the window is open, and
  the model comes out of a cache — priming the relation risks it being memoised into the cached copy,
  which is the failure mode `DungeonService`'s comment already warns about. Not worth trading.
- *`DungeonGridTabsComposer` injecting `Request` instead of going through `RequestViewContextInterface`.*
  Verified safe (a composer is constructed per render, and `integer()` casts junk to `0`); adding a
  `season` accessor to `RequestViewContext` for one optional int felt like more surface, not less.

CI: all 9 checks were green on the previous commit; re-running now.

Wotuu (commented):
Ok

PR #3771, opened 2026-07-30
#3770 Drop SeederModel's deleting listener, it guarded nothing and broke things

### LINKED ISSUE #3770: SeederModel's deleting listener guards nothing and stops Npc's cleanup from running
:robot:

## Background

Raised out of PR #3769's review round. `App\Models\Traits\SeederModel` registers a `deleting` listener
that returns `false` unless an admin is authenticated:

```php
static::deleting(function (Model $model) {
    $user = Auth::getUser();

    return $user?->hasRole(Role::ROLE_ADMIN) || $model instanceof MappingVersion || $model instanceof Floor || $model instanceof Enemy;
});
```

51 models use the trait.

## Why it should go

**It guards nothing.** Every route that deletes one of these models already sits behind
`Route::middleware(['auth', 'role:admin'])` — the admin panel routes and the ajax mapping-editor routes
alike. Any request that reaches a delete is already an admin, in which case the listener returns `true`
and allows it. It is a second lock on the inside of a door that is already locked from the outside.

**It silently no-ops legitimate deletes.** With no authenticated user — every Artisan command, every
queued job, every test — `$model->delete()` does nothing and returns `false`, without raising anything.
`DetectStaleCombatLogDataCommand:121` already works around it with `->toBase()`. In tests it is worse:
a textbook `try/finally` cleanup leaves rows behind in the shared seeded database, and the damage shows
up later in an unrelated test (a leftover `Season` with no dungeons breaks anything doing
`Season::orderByDesc('id')->first()->dungeons()`).

**It swallows other `deleting` listeners.** Eloquent fires `deleting` through the dispatcher's `until()`,
which halts on the first non-null result. SeederModel's listener always returns a boolean, and it is
registered from `boot()` — before anything a model registers in `booted()`. So `Npc`'s cleanup listener,
which deletes the NPC's spells, characteristics, bolstering whitelists, enemy forces and dungeon
couplings, **never runs**. For an admin the listener returns `true`, which halts just as well, so this is
not theoretical: deleting an NPC through the admin panel today leaves all of those rows orphaned.

Verified against the live dev database:

```
Auth::login(User::findOrFail(1));   // admin
Event::until('eloquent.deleting: App\Models\Npc\Npc', [$npc]);
// => true   (SeederModel's listener; halts here)
// npc spells before: 1, after: 1   — Npc's own listener never ran
```

`EnemyForcesCheckpoint` already documents the hazard in a comment and registers on `deleted` to dodge it.
Two more models (`MappingVersion`, `EnemyPatrol`) define `boot()` themselves, which overrides the trait's
`boot()` entirely, so the guard never applied to them at all.

## What to do

- Remove the listener; keep the trait as the marker `DatabaseSeeder::getTempTableName()` uses to stage a
  table as `<table>_temp`.
- Write the convention it half-enforced into CLAUDE.md instead: these models hold seeded data and nothing
  outside the admin panel, the mapping editor and the seeders should delete their rows. Note that mapping
  data is recoverable through a `mapping:save` diff but non-mapping data (seasons, expansions, published
  states, …) is not.
- Update `EnemyForcesCheckpoint`'s comment, which describes a constraint that no longer exists.

Wotuu:
:robot:

Closes #3770

Spun out of #3769's review round — the commit was written there, but that PR merged before this landed, so
it gets its own issue and branch.

## What this removes

`SeederModel` registered a `deleting` listener refusing to delete any of the 51 models using the trait
unless an admin was authenticated. It goes; the trait stays as the marker
`DatabaseSeeder::getTempTableName()` uses to stage a table as `<table>_temp`.

## Why

- **It guarded nothing.** Every route deleting one of these models already sits behind
  `['auth', 'role:admin']`, so any request reaching a delete is an admin — for whom the listener returned
  `true` anyway.
- **It broke deletes with no authenticated user** — Artisan commands, queued jobs, tests. `$model->delete()`
  returned `false` and did nothing, silently. `DetectStaleCombatLogDataCommand:121` already worked around it
  with `->toBase()`, and in tests it meant a correct-looking `try/finally` left rows in the shared seeded
  database, breaking unrelated tests later.
- **It swallowed other `deleting` listeners.** Eloquent fires `deleting` through the dispatcher's `until()`,
  which halts on the first non-null result, and this listener always returned a boolean from `boot()` —
  before anything a model registers in `booted()`. So `Npc`'s cleanup listener never ran, and since `true`
  halts as readily as `false`, **deleting an NPC through the admin panel left its spells, characteristics,
  bolstering whitelists, enemy forces and dungeon couplings orphaned**. Verified on the dev database:

  ```
  Auth::login(User::findOrFail(1));   // admin
  Event::until('eloquent.deleting: App\Models\Npc\Npc', [$npc]);   // => true, halts here
  // npc spells before: 1, after: 1   — Npc's own listener never ran
  ```

`EnemyForcesCheckpoint` already documented this hazard and registered on `deleted` to dodge it; its comment
is trimmed down to just the still-true reason for using `deleted` over `deleting`, since the hazard itself
is gone. `MappingVersion` and `EnemyPatrol` define `boot()` themselves, which overrides the trait's, so
the guard never applied to them at all.

## In its place

CLAUDE.md gains a "Seeded models" section carrying the convention the listener half-enforced: these models
hold seeded data and nothing outside the admin panel, the mapping editor and the seeders should delete their
rows. A delete there is recoverable directly from `database/seeders`, since that's where these models' rows
are authored.

**Edit:** per Wotuu's review, cleaned up several comments/docs that were left narrating the now-removed
listener instead of documenting current behaviour — `.claude/CLAUDE.md`'s "Seeded models" section (including
the recoverability wording above, corrected per his note), the `SeederModel` trait docblock, and
`EnemyForcesCheckpoint`'s `booted()` docblock.

## Tests

`SeederModelTest` asserts a seeded model deletes with nobody authenticated — it fails with the listener
restored, which I checked. The three test classes that came out of #3769 drop their query-builder cleanup
workarounds and use `$model->delete()` again, which also restores proper model-cache invalidation. The
`writing-tests` skill section that documented the trap is rewritten.

I did not ship a test for the NPC cascade itself: creating an `Npc` also spawns 23 `translations` rows, and
cleaning that up reliably in the shared database costs more than the test is worth. The behaviour is covered
by the verification above and documented in CLAUDE.md.

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -204,6 +204,24 @@ is simply not among the methods it overrides.)
 ### Model Creation
 - Every new model must also have a repository. Create the interface at `app/Repositories/Interfaces/{Domain}/{ModelName}RepositoryInterface.php`, the implementation at `app/Repositories/Database/{Domain}/{ModelName}Repository.php`, and register the binding in `app/Providers/RepositoryServiceProvider.php`. See the `repository-pattern` skill for the full convention.
 
+### Seeded models (`SeederModel`)
+Models using the `App\Models\Traits\SeederModel` trait own **seeded data**: their rows come from
+`database/seeders`, not from users. The trait is a marker only — `DatabaseSeeder::getTempTableName()` uses it
+to stage the table as `<table>_temp` while seeding.
+
+- **Nothing outside the admin panel, the mapping editor and the seeders should delete these rows.** That is a
+  convention, not an enforced rule: the routes that delete them already sit behind `role:admin`, and there is
+  no model-level guard. Deleting mapping data is recoverable — it shows up as a diff the next time
+  `mapping:save` runs — but anything not exported to `database/seeders/dungeondata` (seasons, expansions,
+  published states, character classes, …) has no such safety net, so treat a delete there as permanent.
+- The trait used to register a `deleting` listener returning `false` for non-admins. It was removed because it
+  guarded nothing while breaking two things: `$model->delete()` silently became a no-op wherever no user is
+  authenticated (Artisan commands, queued jobs, tests), and because `deleting` is fired through the
+  dispatcher's `until()` — which halts on the first non-null result — it swallowed any `deleting` listener a
+  model registered in `booted()`. `Npc`'s cleanup listener was one of those, so deleting an NPC as an admin
+  left its spells, characteristics and couplings orphaned. Don't reintroduce it; authorization belongs on the
+  route.
+
 ### Controllers & Validation
 - Always create Form Request classes for validation rather than inline validation in controllers. Include both validation rules and custom error messages.
 - Check sibling Form Requests to see if the application uses array or string based validation rules.
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -211,16 +211,9 @@ to stage the table as `<table>_temp` while seeding.
 
 - **Nothing outside the admin panel, the mapping editor and the seeders should delete these rows.** That is a
   convention, not an enforced rule: the routes that delete them already sit behind `role:admin`, and there is
-  no model-level guard. Deleting mapping data is recoverable — it shows up as a diff the next time
-  `mapping:save` runs — but anything not exported to `database/seeders/dungeondata` (seasons, expansions,
-  published states, character classes, …) has no such safety net, so treat a delete there as permanent.
-- The trait used to register a `deleting` listener returning `false` for non-admins. It was removed because it
-  guarded nothing while breaking two things: `$model->delete()` silently became a no-op wherever no user is
-  authenticated (Artisan commands, queued jobs, tests), and because `deleting` is fired through the
-  dispatcher's `until()` — which halts on the first non-null result — it swallowed any `deleting` listener a
-  model registered in `booted()`. `Npc`'s cleanup listener was one of those, so deleting an NPC as an admin
-  left its spells, characteristics and couplings orphaned. Don't reintroduce it; authorization belongs on the
-  route.
+  no model-level guard. A delete is recoverable directly from `database/seeders` — that's where these models'
+  rows are authored (some are being migrated to `.json` files under there instead, but the recovery path is
+  the same: restore from the seeder files).
 
 ### Controllers & Validation
 - Always create Form Request classes for validation rather than inline validation in controllers. Include both validation rules and custom error messages.
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -205,15 +205,23 @@ is simply not among the methods it overrides.)
 - Every new model must also have a repository. Create the interface at `app/Repositories/Interfaces/{Domain}/{ModelName}RepositoryInterface.php`, the implementation at `app/Repositories/Database/{Domain}/{ModelName}Repository.php`, and register the binding in `app/Providers/RepositoryServiceProvider.php`. See the `repository-pattern` skill for the full convention.
 
 ### Seeded models (`SeederModel`)
-Models using the `App\Models\Traits\SeederModel` trait own **seeded data**: their rows come from
-`database/seeders`, not from users. The trait is a marker only — `DatabaseSeeder::getTempTableName()` uses it
-to stage the table as `<table>_temp` while seeding.
+Models using the `App\Models\Traits\SeederModel` trait use `DatabaseSeeder::getTempTableName()` to stage
+their table as `<table>_temp` while seeding — the trait is a marker only, not a guarantee about where a
+model's rows come from.
 
-- **Nothing outside the admin panel, the mapping editor and the seeders should delete these rows.** That is a
-  convention, not an enforced rule: the routes that delete them already sit behind `role:admin`, and there is
-  no model-level guard. A delete is recoverable directly from `database/seeders` — that's where these models'
-  rows are authored (some are being migrated to `.json` files under there instead, but the recovery path is
-  the same: restore from the seeder files).
+For the mapping/season/expansion-style models (dungeons, seasons, expansions, mapping objects, …), rows are
+authored in `database/seeders` (some are being migrated to `.json` files under there instead), so a delete
+is recoverable directly from those files.
+
+- **A subset of trait users are combat-log-derived instead and are *not* recoverable from seeders:**
+  `SpellDungeon`, `NpcCharacteristic` and `NpcSpell` are intentionally omitted from the seeder export (see
+  `DungeonDataSeeder::getAffectedModelClasses()`), and `CombatLogNpcEvent`, `CombatLogSpellEvent` and
+  `ParsedCombatLog` (declared via grouped `use` statements, so `grep "use SeederModel;"` misses them) hold
+  pure runtime/audit data that was never seeder-sourced to begin with. A delete of one of these rows is
+  permanent unless fresh combat log data re-derives it.
+- **Nothing outside the admin panel, the mapping editor, the seeders and the hourly
+  `combatlog:detectstaledata` sweep should delete these rows.** That is a convention, not an enforced rule:
+  the routes that delete them already sit behind `role:admin`, and there is no model-level guard.
 
 ### Controllers & Validation
 - Always create Form Request classes for validation rather than inline validation in controllers. Include both validation rules and custom error messages.
--- a/app/Models/EnemyForcesCheckpoint.php
+++ b/app/Models/EnemyForcesCheckpoint.php
@@ -100,10 +100,11 @@ class EnemyForcesCheckpoint extends CacheModel implements HasLatLngInterface, Ma
     }
 
     /**
-     * Registered as `deleted` rather than `deleting` on purpose: SeederModel's `deleting` listener
-     * always returns a boolean, and Eloquent fires `deleting` through the dispatcher's `until()`,
-     * which halts on the first non-null result. A second `deleting` listener would therefore never
-     * run at all.
+     * Registered as `deleted` rather than `deleting`: the members only need detaching once the checkpoint is
+     * really gone. It also used to be the only option that worked at all - SeederModel registered a
+     * `deleting` listener returning a boolean, and Eloquent fires `deleting` through the dispatcher's
+     * `until()`, which halts on the first non-null result, so a second `deleting` listener never ran. That
+     * listener has since been removed.
      */
     #[Override]
     protected static function booted(): void
--- a/app/Models/EnemyForcesCheckpoint.php
+++ b/app/Models/EnemyForcesCheckpoint.php
@@ -101,10 +101,7 @@ class EnemyForcesCheckpoint extends CacheModel implements HasLatLngInterface, Ma
 
     /**
      * Registered as `deleted` rather than `deleting`: the members only need detaching once the checkpoint is
-     * really gone. It also used to be the only option that worked at all - SeederModel registered a
-     * `deleting` listener returning a boolean, and Eloquent fires `deleting` through the dispatcher's
-     * `until()`, which halts on the first non-null result, so a second `deleting` listener never ran. That
-     * listener has since been removed.
+     * really gone.
      */
     #[Override]
     protected static function booted(): void
--- a/app/Models/Traits/SeederModel.php
+++ b/app/Models/Traits/SeederModel.php
@@ -2,29 +2,23 @@
 
 namespace App\Models\Traits;
 
-use App\Models\Enemy;
-use App\Models\Floor\Floor;
-use App\Models\Laratrust\Role;
-use App\Models\Mapping\MappingVersion;
-use App\Models\User;
-use Auth;
 use Illuminate\Database\Eloquent\Model;
 
 /**
+ * Marks a model whose rows come from the seeders in database/seeders rather than from users, so that
+ * DatabaseSeeder::getTempTableName() knows to stage it in a `_temp` table while seeding. These models are
+ * read-mostly - see "Seeded models" in CLAUDE.md.
+ *
+ * This used to also register a `deleting` listener returning false for everyone but an admin. It guarded
+ * nothing - every route that deletes one of these models already sits behind `role:admin` - while doing two
+ * kinds of damage: it silently turned `$model->delete()` into a no-op wherever there is no authenticated
+ * user (Artisan commands, queued jobs, tests), and because Eloquent fires `deleting` through the
+ * dispatcher's `until()`, which halts on the first non-null result, it swallowed every `deleting` listener
+ * a model registered in `booted()` - including Npc's, so deleting an NPC as an admin left its spells,
+ * characteristics, bolstering whitelists, enemy forces and dungeon couplings behind.
+ *
  * @mixin Model
  */
 trait SeederModel
 {
-    public static function boot(): void
-    {
-        parent::boot();
-
-        // This model may NOT be deleted, it's read only! But if you're an admin, sure you can delete everything.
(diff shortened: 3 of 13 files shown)


.claude/CLAUDE.md
```diff
@@ -167,6 +167,24 @@ under Git worktrees above for why it must stay a draft until then.
...
+### Seeded models (`SeederModel`)
+Models using the `App\Models\Traits\SeederModel` trait own **seeded data**: their rows come from
+`database/seeders`, not from users. The trait is a marker only — `DatabaseSeeder::getTempTableName()` uses it
+to stage the table as `<table>_temp` while seeding.
+
+- **Nothing outside the admin panel, the mapping editor and the seeders should delete these rows.** That is a
+  convention, not an enforced rule: the routes that delete them already sit behind `role:admin`, and there is
+  no model-level guard. Deleting mapping data is recoverable — it shows up as a diff the next time
+  `mapping:save` runs — but anything not exported to `database/seeders/dungeondata` (seasons, expansions,
+  published states, character classes, …) has no such safety net, so treat a delete there as permanent.
```
Wotuu: This is not entirely true, those models you listed are recovered from the Seeder files directly, though I'm working on moving that data to .json files as well. Any delete there will be recoverable from Seeder files.

.claude/CLAUDE.md
```diff
@@ -167,6 +167,24 @@ under Git worktrees above for why it must stay a draft until then.
...
+- The trait used to register a `deleting` listener returning `false` for non-admins. It was removed because it
+  guarded nothing while breaking two things: `$model->delete()` silently became a no-op wherever no user is
+  authenticated (Artisan commands, queued jobs, tests), and because `deleting` is fired through the
+  dispatcher's `until()` — which halts on the first non-null result — it swallowed any `deleting` listener a
+  model registered in `booted()`. `Npc`'s cleanup listener was one of those, so deleting an NPC as an admin
+  left its spells, characteristics and couplings orphaned. Don't reintroduce it; authorization belongs on the
+  route.
```
Wotuu: This is gone, so I don't see a reason to mention it.

app/Models/Traits/SeederModel.php
```diff
@@ -2,29 +2,23 @@
...
+ * DatabaseSeeder::getTempTableName() knows to stage it in a `_temp` table while seeding. These models are
+ * read-mostly - see "Seeded models" in CLAUDE.md.
+ *
+ * This used to also register a `deleting` listener returning false for everyone but an admin. It guarded
+ * nothing - every route that deletes one of these models already sits behind `role:admin` - while doing two
+ * kinds of damage: it silently turned `$model->delete()` into a no-op wherever there is no authenticated
+ * user (Artisan commands, queued jobs, tests), and because Eloquent fires `deleting` through the
+ * dispatcher's `until()`, which halts on the first non-null result, it swallowed every `deleting` listener
+ * a model registered in `booted()` - including Npc's, so deleting an NPC as an admin left its spells,
+ * characteristics, bolstering whitelists, enemy forces and dungeon couplings behind.
```
Wotuu: Is this really necessary? Like, do you need this for your functioning? Does it make you better? If so, you can keep it. Otherwise, I don't see a reason to keep it.

app/Models/EnemyForcesCheckpoint.php
```diff
@@ -100,10 +100,11 @@ public function getDungeonId(): ?int
...
     /**
-     * Registered as `deleted` rather than `deleting` on purpose: SeederModel's `deleting` listener
-     * always returns a boolean, and Eloquent fires `deleting` through the dispatcher's `until()`,
-     * which halts on the first non-null result. A second `deleting` listener would therefore never
-     * run at all.
+     * Registered as `deleted` rather than `deleting`: the members only need detaching once the checkpoint is
+     * really gone. It also used to be the only option that worked at all - SeederModel registered a
+     * `deleting` listener returning a boolean, and Eloquent fires `deleting` through the dispatcher's
+     * `until()`, which halts on the first non-null result, so a second `deleting` listener never ran. That
+     * listener has since been removed.
```
Wotuu: again, same thing. it's been removed so no need to mention it anymore. Only if it somehow helps you write better code you can keep it. Otherwise just get rid of it.

tests/Feature/App/Models/Trait/SeederModelTest.php
```diff
@@ -0,0 +1,48 @@
...
+use PHPUnit\Framework\Attributes\Test;
+use Tests\TestCases\PublicTestCase;
+
+/**
+ * SeederModel used to register a `deleting` listener returning false for everyone but an admin. It guarded
+ * nothing - every route deleting one of these models sits behind `role:admin` - while silently turning
+ * `$model->delete()` into a no-op wherever nobody is authenticated (Artisan commands, queued jobs, tests),
+ * and swallowing any `deleting` listener a model registered in `booted()`, because Eloquent fires `deleting`
+ * through the dispatcher's `until()`, which halts on the first non-null result.
+ */
```
Wotuu, line 18: Now this is the only place where I'd see that this is a decent comment. You explain why the test is there and the reason for it being there. There was something there, but it no longer is. So this comment is valid.

Wotuu (commented):
Some changes

.claude/CLAUDE.md
Wotuu: :robot: Reworded — the section now says a delete is recoverable directly from `database/seeders` (rows are authored there, or moving to `.json` files there), rather than the incorrect claim that only mapping data was recoverable via a `mapping:save` diff. Pushed in `9207c4d03`.

.claude/CLAUDE.md
Wotuu: :robot: Removed — that whole bullet only narrated the removed `deleting` listener and its dodges, which is gone now. Pushed in `9207c4d03`.

app/Models/Traits/SeederModel.php
Wotuu: :robot: Agreed, dropped it — it only narrated the removed listener and didn't document anything still true. Kept the first paragraph, which still documents the marker/temp-table usage. Pushed in `9207c4d03`.

app/Models/EnemyForcesCheckpoint.php
Wotuu: :robot: Trimmed the docblock to just the still-true reason for hooking `deleted` instead of `deleting` (members only need detaching once the checkpoint is really gone); dropped the rest, which only explained the now-removed `SeederModel` listener dodge. Pushed in `9207c4d03`.

tests/Feature/App/Models/Trait/SeederModelTest.php
Wotuu: :robot: Agreed, no change needed here — left as-is.

app/Models/Traits/SeederModel.php
```diff
@@ -2,29 +2,15 @@
...
-use App\Models\Laratrust\Role;
-use App\Models\Mapping\MappingVersion;
-use App\Models\User;
-use Auth;
 use Illuminate\Database\Eloquent\Model;
 
 /**
+ * Marks a model whose rows come from the seeders in database/seeders rather than from users, so that
+ * DatabaseSeeder::getTempTableName() knows to stage it in a `_temp` table while seeding. These models are
+ * read-mostly - see "Seeded models" in CLAUDE.md.
```
Wotuu, line 10: :robot: The cleanup commit missed four places that still describe this listener as active:

- `app/Console/Commands/CombatLog/DetectStaleCombatLogDataCommand.php:120` — "Use toBase() to bypass SeederModel's deleting observer, which only allows admin users to delete" (the PR body cites this exact workaround as one of the listener's victims)
- `app/Service/MDT/MDTMappingImportService.php:677-678` — "SeederModel's `deleting` listener refuses the delete unless an admin is authenticated, and this import runs from the scheduler with nobody logged in." The second half of that comment (the skipped `EnemyForcesCheckpoint` `deleted` hook) is still true; the first sentence is not.
- `.claude/skills/mapping-versioned-models/SKILL.md:161-164` — this one actively matters: step 2 of the new-model checklist still instructs "hook `static::deleted()`, **not** `deleting`: `SeederModel` already registers a `deleting` listener that returns a bool ... a second `deleting` listener would silently never run." That guidance is now wrong and will steer the next mapping model away from `deleting` for a reason that no longer exists.
- `.claude/skills/combatlog-data-pipeline/SKILL.md:136` — "(`->toBase()->delete()` to bypass the `SeederModel` admin-only observer)"

Since Wotuu's round explicitly asked for stale references to the removed listener to go, these four look like misses of the same request rather than intentional keeps.

.claude/CLAUDE.md
```diff
@@ -167,6 +167,17 @@ under Git worktrees above for why it must stay a draft until then.
...
 - Every new model must also have a repository. Create the interface at `app/Repositories/Interfaces/{Domain}/{ModelName}RepositoryInterface.php`, the implementation at `app/Repositories/Database/{Domain}/{ModelName}Repository.php`, and register the binding in `app/Providers/RepositoryServiceProvider.php`. See the `repository-pattern` skill for the full convention.
 
+### Seeded models (`SeederModel`)
+Models using the `App\Models\Traits\SeederModel` trait own **seeded data**: their rows come from
+`database/seeders`, not from users. The trait is a marker only — `DatabaseSeeder::getTempTableName()` uses it
+to stage the table as `<table>_temp` while seeding.
+
+- **Nothing outside the admin panel, the mapping editor and the seeders should delete these rows.** That is a
+  convention, not an enforced rule: the routes that delete them already sit behind `role:admin`, and there is
+  no model-level guard. A delete is recoverable directly from `database/seeders` — that's where these models'
```
Wotuu: :robot: This section overgeneralizes to all trait users, and both of its claims are contradicted by existing code for the combat-log-derived subset:

- **"their rows come from `database/seeders` ... A delete is recoverable directly from `database/seeders`"** — `DungeonDataSeeder::getAffectedModelClasses()` (database/seeders/DungeonDataSeeder.php:527) says the opposite for three of them: "SpellDungeon, NpcCharacteristic and NpcSpell are combat-log-derived behavior and are intentionally omitted: they are not exported to the seeders". A deleted row there is gone until fresh combat log observations re-derive it. `CombatLogNpcEvent`, `CombatLogSpellEvent` and `ParsedCombatLog` also use the trait (grouped `use` statements, so `grep "use SeederModel;"` misses them) and hold pure runtime audit data — not recoverable from seeders at all.
- **"Nothing outside the admin panel, the mapping editor and the seeders should delete these rows"** — the hourly `DetectStaleCombatLogDataCommand` deletes stale `NpcCharacteristic` rows by design (the PR body itself cites it), so the scheduler is a fourth legitimate deleter.

Suggest scoping the paragraph: the seeded-and-recoverable story holds for the mapping/season/expansion models, with an explicit carve-out for the combat-log-derived ones (`SpellDungeon`, `NpcCharacteristic`, `NpcSpell`, `CombatLogNpcEvent`, `CombatLogSpellEvent`, `ParsedCombatLog`), whose rows live only in the database.

tests/Feature/App/Model/Trait/SeederModelTest.php
```diff
@@ -0,0 +1,48 @@
+<?php
+
+namespace Tests\Feature\App\Model\Trait;
```
Wotuu: :robot: Minor: this creates a new directory `tests/Feature/App/Model/Trait/`, but the existing trait test sibling — `HasStartTest` — lives at `tests/Feature/App/Models/Trait/` (namespace `Tests\Feature\App\Models\Trait`). The `Model/` vs `Models/` split under `tests/Feature/App/` is pre-existing mess, but for `Trait` specifically there is exactly one established home, so this test starting a second parallel tree makes it worse. Moving it to `tests/Feature/App/Models/Trait/SeederModelTest.php` keeps the trait tests together.

Wotuu:

app/Models/Traits/SeederModel.php
Wotuu: :robot: Fixed all four - the previous cleanup commit missed them because it only touched files with an actual `SeederModel::deleting()` reference, not comments narrating the workaround around it.

- `DetectStaleCombatLogDataCommand.php:120` - kept `->toBase()->delete()` as-is (didn't remove it) and only rewrote the comment. I initially removed `toBase()` too, thinking it was now pointless, but verified that `NpcCharacteristic` extends `CacheModel` (`GeneaLabs\LaravelModelCaching`'s `Cachable` trait), whose `Buildable::delete()` invalidates cache tags on every delete with no `isCachable()` guard - so a plain Eloquent delete here would trigger cache invalidation that `toBase()` currently skips. Dropping `toBase()` would have been a real, unreviewed behavior change to an hourly production sweep, so I kept it and just fixed the comment to state the real remaining effect (skips Eloquent's caching builder layer) instead of the false admin-listener framing.
- `MDTMappingImportService.php:677-678` - kept the still-true second half about the `EnemyForcesCheckpoint` `deleted` hook being skipped (harmless here since pruned checkpoints have no members), replaced the false first sentence.
- `.claude/skills/mapping-versioned-models/SKILL.md:161-164` - replaced the false "second `deleting` listener would silently never run" hazard with the real, still-true reason to hook `deleted()` over `deleting()`: referencing rows should only be detached once the delete is confirmed, not beforehand (matches `EnemyForcesCheckpoint::booted()`).
- `.claude/skills/combatlog-data-pipeline/SKILL.md:136` - dropped the "bypass SeederModel admin-only observer" framing, kept the `toBase()` mention with a neutral description of what it actually does.

Commit `626e9c9bf`.

.claude/CLAUDE.md
Wotuu: :robot: Reworded per the suggestion. The "Seeded models" section now:

- Scopes the seeder-recoverable story to mapping/season/expansion-style models (rows authored in `database/seeders`, some migrating to `.json` files there).
- Adds an explicit carve-out for the six combat-log-derived trait users that are *not* recoverable from seeders: `SpellDungeon`, `NpcCharacteristic`, `NpcSpell` (per `DungeonDataSeeder::getAffectedModelClasses()`, which intentionally omits them) and `CombatLogNpcEvent`, `CombatLogSpellEvent`, `ParsedCombatLog` (grouped `use` statements, confirmed via `grep -rln "SeederModel" app/Models/`). I checked whether the two `CombatLog*Observation` models (deleted hourly by `pruneOldObservations()`) also needed listing - they don't use the `SeederModel` trait at all, so they're out of scope for this section.
- Adds the hourly `combatlog:detectstaledata` sweep as a fourth legitimate deleter alongside the admin panel, mapping editor and seeders.

Commit `626e9c9bf`.

tests/Feature/App/Model/Trait/SeederModelTest.php
Wotuu: :robot: Moved to `tests/Feature/App/Models/Trait/SeederModelTest.php` with the namespace updated to `Tests\Feature\App\Models\Trait`, matching `HasStartTest`. Deleted the now-empty `tests/Feature/App/Model/Trait/` directory (the sibling `tests/Feature/App/Model/` tree stays - it still has other unrelated tests, e.g. `Mapping/` and `CombatLog/`, so only the `Trait/` subdirectory was empty). Confirmed PHPUnit still discovers and runs it from the new location (`Tests: 1 passed`), and `composer.json`'s `autoload-dev`/`Tests\` PSR-4 mapping is `tests/` broadly, no hardcoded paths to update.

Commit `626e9c9bf`.

Wotuu (commented):
Ok

PR #3773, opened 2026-07-31
#3772 Clear a user's stored feature values when their roles change

### LINKED ISSUE #3772: Changing a user's roles should clear their stored feature flag values
:robot: When a user is granted (or loses) a role, Pennant keeps serving the value it resolved and stored *before* the role change.

Every feature in `app/Features` resolves off the users roles (e.g. `NpcCompendium` requires `ROLE_ADMIN` or `ROLE_INTERNAL_TEAM`), but Pennant stores the resolved value in the `features` table indefinitely. So a user who visited the site before being given the Internal Team role keeps a stored `App\Features\NpcCompendium = false` row and still cannot see the NPC Compendium after the role is granted.

Reproduced on Staging: user registered, was given the Internal Team role, and still got no NPC Compendium because the `features` table already had `NpcCompendium = false` rows for them.

Changing a users roles should drop that users stored feature values so they are resolved anew on the next request.

Note: `Feature::getAdminValue()` reads the row scoped to user 1 as the global on/off switch per feature, so the purge must leave that users rows alone.

Wotuu:
:robot: Closes #3772

## The problem

Every feature in `app/Features` resolves off the user's roles — `NpcCompendium` for instance requires `ROLE_ADMIN` or `ROLE_INTERNAL_TEAM`. Pennant stores the value it resolved *once* in the `features` table and keeps serving that forever after, so a user who browsed the site before being granted a role keeps their stale `false` and never gains access.

Hit for real on Staging: a user registered, was given Internal Team, and still could not see the NPC Compendium because their `App\Features\NpcCompendium = false` row was already there.

## The fix

`Feature::forgetAllForUser()` drops every stored value of a user, wired to Laratrust's `role.added` / `role.removed` / `role.synced` events on `User` in `AppServiceProvider`. Any role change — the admin panel, `RegisterController`, anything added later — now purges the user's rows, and they're resolved anew on that user's next request.

All three events are needed: `removeRoles([$roles])` loops `removeRole()` and fires `role.removed`, but `removeRoles()` with nothing to remove falls through to `syncRoles([])` and fires `role.synced` instead.

### The admin user is deliberately exempt

`Feature::getAdminValue()` reads the row scoped to **user 1** as the global on/off switch of each feature. Purging that user's rows would therefore take every feature down site-wide until an admin re-toggled them one by one — worse than the bug being fixed here. `forgetAllForUser()` returns early for that user, and the coupling now runs through a `Feature::ADMIN_USER_ID` constant instead of a hardcoded `1` in the `sprintf`, so the guard reads as intentional.

## Tests

`tests/Feature/App/Models/Feature/FeatureForgetAllForUserTest.php` — one test per event path (added / removed / synced-via-empty-`removeRoles`) plus one pinning the admin exemption. All four were verified to fail when the listeners / the guard are removed, so none of them pass vacuously.

The suite runs Pennant on its `array` store (`phpunit.xml`), so the test class swaps in the `database` store in `setUp()` — what's under test is precisely the rows in the `features` table. The admin test uses a synthetic feature name so it never touches the real switches.

## Note for after this deploys

The fix is forward-looking: it does not retroactively clean up users who are *already* stuck with a stale `false`. For the Staging user who prompted this, either delete their rows from `features` directly, or remove and re-add their role once this is live.

### DIFF
--- a/app/Models/Feature/Feature.php
+++ b/app/Models/Feature/Feature.php
@@ -6,6 +6,7 @@ use App\Models\User;
 use Eloquent;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Carbon;
+use Laravel\Pennant\Feature as PennantFeature;
 
 /**
  * @property int    $id
@@ -20,6 +21,11 @@ use Illuminate\Support\Carbon;
  */
 class Feature extends Model
 {
+    /**
+     * The user whose stored values double as the global on/off switch of each feature - see getAdminValue().
+     */
+    public const int ADMIN_USER_ID = 1;
+
     protected function casts(): array
     {
         return [
@@ -36,9 +42,25 @@ class Feature extends Model
     {
         /** @var Feature|null $feature */
         $feature = Feature::where('name', $name)
-            ->where('scope', sprintf('%s|%d', User::class, 1))
+            ->where('scope', sprintf('%s|%d', User::class, self::ADMIN_USER_ID))
             ->first();
 
         return $feature?->value === 'true';
+    }
+
+    /**
+     * Drop every stored feature value of the given user, so that they are resolved anew the next time the user
+     * requests them. Necessary whenever the user's roles change - every feature in App\Features resolves off the
+     * user's roles, but Pennant stores the value it resolved once and keeps serving it forever after.
+     */
+    public static function forgetAllForUser(User $user): void
+    {
+        // The admin user's stored values are the global on/off switch of each feature (see getAdminValue), so
+        // dropping them would disable every feature site-wide until an admin toggles them all back on by hand
+        if ($user->id === self::ADMIN_USER_ID) {
+            return;
+        }
+
+        PennantFeature::for($user)->forget(PennantFeature::stored());
     }
 }
--- a/app/Providers/AppServiceProvider.php
+++ b/app/Providers/AppServiceProvider.php
@@ -3,6 +3,7 @@
 namespace App\Providers;
 
 use App\Exceptions\Handler;
+use App\Models\Feature\Feature;
 use App\Models\Laratrust\Role;
 use App\Models\User;
 use App\Overrides\CustomRateLimiter;
@@ -62,6 +63,19 @@ class AppServiceProvider extends ServiceProvider
         Event::listen(CommandStarting::class, static function (): void {
             Context::addIf('trace_id', (string)Str::uuid());
         });
+
+        // Every feature in App\Features resolves off the user's roles, but Pennant stores the value it resolved
+        // once and keeps serving that forever after - so changing a user's roles has to drop their stored values.
+        // Laratrust passes the changed user as the first argument; the arguments following it differ per event
+        // (added/removed pass the role, synced passes the sync changes) so they're deliberately not accepted here.
+        $forgetFeatures = static function (User $user): void {
+            Feature::forgetAllForUser($user);
+        };
+
+        // removeRoles() loops removeRole() - unless it's handed an empty array, in which case it syncs instead
+        User::roleAdded($forgetFeatures);
+        User::roleRemoved($forgetFeatures);
+        User::roleSynced($forgetFeatures);
 
         $this->app->bind(ExceptionHandler::class, Handler::class);
 
--- a/tests/Feature/App/Models/Feature/FeatureForgetAllForUserTest.php
+++ b/tests/Feature/App/Models/Feature/FeatureForgetAllForUserTest.php
@@ -107,14 +107,17 @@ final class FeatureForgetAllForUserTest extends PublicTestCase
     public function forgetAllForUser_givenAdminUser_keepsStoredFeatureValues(): void
     {
         // Arrange - the admin's stored values are the global on/off switch of every feature, so dropping them
-        // would disable the lot site-wide. A synthetic name keeps the real switches out of this test entirely.
-        $admin = User::findOrFail(Feature::ADMIN_USER_ID);
+        // would disable the lot site-wide. Asserting on a synthetic name keeps the real switches out of it.
+        $admin      = User::findOrFail(Feature::ADMIN_USER_ID);
+        $adminScope = $this->serializeScopeOf($admin);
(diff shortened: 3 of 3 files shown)


Wotuu (commented):
Ok

Wotuu:
:robot: Cold review: 5 findings

Handed the diff to a fresh-context agent with no knowledge of how it was built. It found **no material bug in the change itself**, and independently verified the parts most likely to be wrong: the Laratrust payload shapes are `[$user, $roleId, $team]` / `[$user, $syncResult, $team]` so the single-parameter listener is safe for all three events; the pivot is committed and Laratrust's own cache flushed before the listener runs; every role-mutating call site in the app (`UserController::makeRole`, `RegisterController`, `OAuthLoginController`, `LaratrustSeeder`) routes through `addRole`/`removeRole`/`removeRoles`; registration happens exactly once per Octane worker with no stale-dispatcher hazard, and Pennant flushes its in-memory cache on `RequestReceived`/`JobProcessed` anyway; `stored()` forwards correctly, is a no-op on an empty list, and handles stored names whose class no longer exists (`DungeonOverview` and `FrontPageRework` are both in that state today - the purge simply cleans them up).

Findings and what was done about each:

1. **Admin toggle grants features to everyone with a stored value** - real, user-affecting, and pre-existing. `activateForEveryone()` bottoms out in an `UPDATE ... WHERE name = ?` with no scope predicate. Verified independently against the dev database: ordinary non-internal-team users carry `NpcCompendium = true`. Out of scope here - filed as #3774, which is the mirror image of this PR and shares its root cause.
2. **Deleted users leave orphan `features` rows** - `User`'s `deleting` hook syncs roles via the relation, so no Laratrust event fires. Harmless (ids are not reused), unbounded table growth only. Noted in #3774.
3. **The admin-guard test could damage the shared test database if the guard regresses** - valid and fixed in 22d1d1740. The Act purges every stored name, not just the synthetic one asserted on, so a weakened guard wiped every `App\Models\User|1` row and the failure outlived the run. The test now snapshots the admin's rows and restores them in the `finally`. This one was not hypothetical: it happened during development and the rows had to be put back by hand. Verified by regressing the guard again - the test still fails, and the rows now survive.
4. **Per-`addRole` cost on registration** - one `SELECT DISTINCT name` plus an indexed DELETE per name, on a user provably holding zero rows. Negligible given the `(name, scope)` unique index; not changed.
5. **Missing a file-level `#[Group]`** - cosmetic, and the sibling tests in `tests/Feature/App/Models/` are inconsistent about it; not changed.

app/Models/Feature/Feature.php
```diff
@@ -36,9 +42,25 @@ public static function getAdminValue(string $name): bool
...
+     */
+    public static function forgetAllForUser(User $user): void
+    {
+        // The admin user's stored values are the global on/off switch of each feature (see getAdminValue), so
+        // dropping them would disable every feature site-wide until an admin toggles them all back on by hand
+        if ($user->id === self::ADMIN_USER_ID) {
+            return;
+        }
+
+        PennantFeature::for($user)->forget(PennantFeature::stored());
```
Wotuu, line 64: :robot: Verified the mechanics here against Pennant vendor code — `stored()` lists all distinct names and `forget()` deletes only this user's scope, and `Decorator::delete()` also evicts the in-request cache, so the query side is correct.

One behavioral interplay worth being aware of (may well be intended): `AdminToolsFeaturesController::toggleFeature()` uses `Feature::activateForEveryone()`, whose `DatabaseDriver::setForAllScopes()` flips **every existing row** of that feature to `true` — including rows of users who fail the `resolve()` role gate. Those users currently have the feature active. After this PR, the moment any of their roles change (any `addRole`/`removeRole`, e.g. an unrelated role grant), their rows purge and re-resolution re-applies the role check — so they silently lose a feature the toggle had given them, while identical users whose roles never change keep it. That aligns access with `resolve()`'s stated intent, so arguably a fix, but it makes the "activated for everyone" population inconsistent over time and isn't mentioned in the PR body.

app/Providers/AppServiceProvider.php
```diff
@@ -63,6 +64,19 @@ public function boot(): void
...
+        // Every feature in App\Features resolves off the user's roles, but Pennant stores the value it resolved
+        // once and keeps serving that forever after - so changing a user's roles has to drop their stored values.
+        // Laratrust passes the changed user as the first argument; the arguments following it differ per event
+        // (added/removed pass the role, synced passes the sync changes) so they're deliberately not accepted here.
+        $forgetFeatures = static function (User $user): void {
+            Feature::forgetAllForUser($user);
+        };
+
+        // removeRoles() loops removeRole() - unless it's handed an empty array, in which case it syncs instead
+        User::roleAdded($forgetFeatures);
```
Wotuu, line 76: :robot: Minor / dev-only: I checked every role-mutation call site for paths that bypass these events — all runtime paths (UserController, RegisterController, OAuthLoginController) go through `addRole`/`removeRole`/`removeRoles`, so the coverage claim holds in production. The one bypass is `LaratrustSeeder::truncateLaratrustTables()`, which truncates `role_user` via raw `DB::table()` — no Laratrust event fires, so a role re-seed in dev leaves everyone's stored feature values stale. Probably fine to leave, just noting the "anything added later" claim only covers event-firing paths, not raw pivot writes.

Wotuu:
:robot: Cold review (fable): 2 findings posted (both minor/informational — the `activateForEveryone` interplay and the `LaratrustSeeder` raw-truncate bypass). Everything specifically flagged for scrutiny checked out against vendor code: the `removeRoles([])` → `syncRoles([])` fallthrough is real (`HasRolesAndPermissions.php:413`), events fire after the pivot change with the `User` as arg 1, the admin exemption is keyed on the same constant `getAdminValue()` reads so the switch rows are protected regardless of user 1's roles (absent-row → `false` is pre-existing behavior, not worsened), `Decorator::delete()` also evicts Pennant's in-request cache, `ArrayDriver` implements `CanListStoredFeatures` so the listener is harmless under the suite's array store, and the test's per-test app rebuild means the `pennant.default` swap cannot leak into sibling tests; DB cleanup follows the try/finally convention.

PR #3778, opened 2026-07-31
#3777 Give each git worktree its own isolated seeded database

### LINKED ISSUE #3777: Give each git worktree its own isolated seeded database
:robot: Agents working in parallel worktrees all share the main stack's single MySQL schema — `sh/worktree.sh` copies the main `.env` verbatim and network-attaches the main DB container, and the phpunit connection points at the same schema. In practice agents bite each other: concurrent test runs poison each other's data, and one agent's mistake can corrupt data every other checkout sees. A whole discipline exists just to cope (try/finally test cleanup, the `migrate:fresh` ban, non-destructive-migrations-only in worktrees).

## Change

Make `sh/worktree.sh create` provision a **private schema per worktree on the existing shared MySQL servers** (no new containers), by default:

- Schema `ksg_wt_<branch>_<hash>` on **both** servers (main + combatlog), created as root via `docker exec` (the root password never touches the host — it is read from the DB container's own environment), with grants copied from the users that already have access to the main schema.
- Populated with the proven CI recipe: Laravel schema dumps + pending migrations + full `DatabaseSeeder` + `LaratrustSeeder` (users 1=admin, 2=internal_team, 3=user). Takes ~2–3 minutes.
- The worktree's `.env` copy gets `DB_DATABASE` / `DB_PHPUNIT_DATABASE` / `DB_COMBATLOG_DATABASE` / `REDIS_PREFIX` rewritten — so tests, the model cache, sessions and queues are isolated too, not just the DB.
- `--shared-db` opts back into today's sharing behavior for tasks that genuinely need production-like data (heatmaps, Path B thumbnails, shared-Horizon work); all the old rules then apply.
- Lifecycle: a `.worktree-db` marker records mode + schema name; `down` keeps the schema (a later `create` re-attaches and skips the seed), `remove` drops it on both servers. New subcommands `provision-db` (idempotent retry after a failed seed) and `prune-db [--force]` (drop schemas whose checkout is gone).

Payoff: parallel agents cannot touch each other's data anymore, and destructive operations (`migrate:fresh`, destructive migrations) become safe to *test* in an isolated worktree because the private schema has its own `migrations` table.

## Known accepted limitations (documented in the `worktree-docker` skill)

- Path B thumbnails, the app-swoole-proxied endpoints and OpenSearch-backed search still read main-stack data — `--shared-db` is the escape hatch.
- A fresh seed has no dungeon routes; the `seed-dev-routes` skill fills that per-task need.
- ~200–300 MB InnoDB per live worktree, bounded by drop-on-remove + `prune-db`.

## Definition of done

- `sh/worktree.sh` implements the above; docs updated (`worktree-docker`, `writing-tests`, `generating-thumbnails` skills + `.claude/CLAUDE.md` + `.gitignore`).
- Verified end-to-end: isolated create seeds and serves, tests inside the worktree hit the private schema, main-schema sentinel counts unchanged, isolation probe (row + cache key visible only in the worktree), down→create reuse skips the seed, remove drops both schemas, `--shared-db` regression behaves exactly as before, `provision-db` retries after an interrupted seed, `prune-db` detects and force-drops an orphan.

Wotuu:
Closes #3777

:robot: Worktrees now get a **private, freshly-seeded database by default** instead of sharing the main stack's schema — ending parallel agents corrupting each other's (and the main checkout's) data.

## What this does

- `sh/worktree.sh create` provisions schema `ksg_wt_<branch>_<hash>` on **both** existing MySQL servers (no new containers): created as root via `docker exec` (root password read from the DB container's own env — never touches the host or any `.env`), grants copied from the users that already have access to the main schema, then Laravel schema dumps + pending migrations + full `DatabaseSeeder` + `LaratrustSeeder` (~2–3 min, streamed as progress).
- The worktree's `.env` copy gets `DB_DATABASE` / `DB_PHPUNIT_DATABASE` / `DB_COMBATLOG_DATABASE` / `REDIS_PREFIX` rewritten — tests, model cache, sessions and queues are isolated too. `migrate:fresh` becomes safe to use in an isolated worktree (own `migrations` table).
- `create ... --shared-db` opts back into the old sharing behavior (for heatmaps, Path B thumbnails, shared-Horizon work); the old rules then apply and are still documented.
- Lifecycle: `.worktree-db` marker records mode + schema; `down` keeps the schema (re-`create` re-attaches and skips the seed), `remove` drops it on both servers; new `provision-db` (idempotent retry) and `prune-db [--force]` (orphan cleanup) subcommands.
- Schema-dump loading is done by piping the dumps into the DB containers as root rather than letting artisan shell out to the app image's MariaDB client, which rejects the MySQL 8 server's self-signed certificate (ERROR 2026).
- Docs updated: `worktree-docker` skill (rewritten shared-DB section, what is/isn't isolated, lifecycle), `.claude/CLAUDE.md` worktree bullet, `writing-tests` + `generating-thumbnails` notes, `.gitignore` entry for the marker.

## Verified end-to-end (live, against the real main stack)

- Isolated create: 114 tables, 155 dungeons, 81 804 enemies, users 1–3 seeded; page serves HTTP 200; `SiteControllerTest` green **against the private schema**.
- Isolation probe: a created user row and a cache key existed only in the private schema / private redis prefix; main-schema sentinel counts (74 994 users / 507 768 routes) byte-identical before, during and after.
- `down` → `create`: schema kept, seed skipped (users sentinel), near-instant re-attach.
- `remove`: schemas dropped on both servers; main untouched.
- `--shared-db` regression: no schema created, app reads `keystone.guru.prod`, marker says `shared`, remove drops nothing.
- Failure path: an interrupted first provisioning was completed by `provision-db` (idempotent); `prune-db` detected and `--force`-dropped a fabricated orphan schema.

No PHP code changed — this is shell + docs only, so php-tests/phpstan run on an unchanged codebase.

### DIFF
--- a/.claude/skills/worktree-docker/SKILL.md
+++ b/.claude/skills/worktree-docker/SKILL.md
@@ -1,12 +1,13 @@
 ---
 name: worktree-docker
-description: Use when starting a task that should run in an isolated git worktree with its own Docker app stack, or when the user mentions worktrees, parallel branches, or running multiple checkouts side by side. Covers sh/worktree.sh (create/down/remove/list), how the isolated stack shares the main DB/redis, and how to run artisan/tests inside it. Do not use for the main development stack or generic docker-compose questions.
+description: Use when starting a task that should run in an isolated git worktree with its own Docker app stack, or when the user mentions worktrees, parallel branches, or running multiple checkouts side by side. Covers sh/worktree.sh (create/down/remove/provision-db/prune-db/list), the per-worktree private database (and the --shared-db escape hatch), and how to run artisan/tests inside it. Do not use for the main development stack or generic docker-compose questions.
 ---
 
 # Worktree + Docker stack
 
-Run each task in its own git **worktree** with a minimal Docker stack (`app` + `nginx`) that shares
-the main stack's database, redis, reverb, etc. This lets parallel agents work without colliding.
+Run each task in its own git **worktree** with a minimal Docker stack (`app` + `nginx`). By default
+each worktree also gets its **own freshly-seeded database schemas** (on the shared MySQL servers)
+and its own redis prefix, so parallel agents cannot touch each other's data — or the main stack's.
 
 By default, **every task starts in a fresh worktree** created with `sh/worktree.sh` (see the "Git
 worktrees" rule in `.claude/CLAUDE.md`), unless the user says otherwise.
@@ -19,14 +20,21 @@ isn't, start it from the main repo: `docker compose up -d`.
 ## Create a worktree
 
 ```bash
-sh/worktree.sh create <issue>-<slug>            # branches off origin/master
-sh/worktree.sh create <issue>-<slug> <base-ref> # or off an explicit base
+sh/worktree.sh create <issue>-<slug>              # branches off origin/master
+sh/worktree.sh create <issue>-<slug> <base-ref>   # or off an explicit base
+sh/worktree.sh create <issue>-<slug> --shared-db  # share the main stack's DB data (see below)
 ```
 
 This creates the worktree at `../keystone.guru-worktrees/<issue>-<slug>`, copies the main `.env`
-(never read — a plain shell `cp`), rewrites only `APP_URL`/`URL_HOST` to the worktree's port,
-appends `COMPOSE_PROJECT_NAME` / `COMPOSE_FILE` / `WORKTREE_HTTP_PORT` to that copy, starts the
-stack, wires up the shared services, and prints the URL (e.g. `http://localhost:8100`).
+(never read — a plain shell `cp`), rewrites `APP_URL`/`URL_HOST` to the worktree's port (and, in the
+default isolated mode, `DB_DATABASE`/`DB_PHPUNIT_DATABASE`/`DB_COMBATLOG_DATABASE`/`REDIS_PREFIX`
+to per-worktree values), appends `COMPOSE_PROJECT_NAME` / `COMPOSE_FILE` / `WORKTREE_HTTP_PORT` to
+that copy, starts the stack, wires up the shared services, provisions the private DB (isolated mode:
+create schemas + migrate + full seed — **expect ~5–15 minutes**, the streamed seeder output is the
+progress bar), and prints the URL (e.g. `http://localhost:8100`).
+
+A `.worktree-db` marker file in the worktree root records the DB mode and schema name — `remove`
+and `prune-db` read it to know what to drop. Don't delete or edit it.
 
 It also marks issue `#<issue>` (the leading number of the branch) with the `in progress` label on
 GitHub, so you can see at a glance what's actively being worked on. `remove` clears it again (see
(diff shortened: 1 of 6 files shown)


Wotuu (commented):
Ok

.claude/skills/worktree-docker/SKILL.md
```diff
@@ -87,15 +95,53 @@ that first build.
...
+- Queued jobs dispatched in an isolated worktree are invisible to the shared Horizon (different
+  schema *and* different redis prefix). Run a worker in the worktree if needed, or use
+  `--shared-db`.
+- Search-driven pages query the shared OpenSearch index, whose ids reference main-DB rows.
+- Sessions are per-worktree, so switching between main-stack and worktree tabs re-logins each side.
+
+### Destructive operations are now OK (isolated mode only)
+
+The private schema has its own `migrations` table, so in an **isolated** worktree you may test
+destructive migrations and even run `migrate:fresh` — it only wipes *your* schema (re-seed after
```
Wotuu, line 129: :robot: **`migrate:fresh` in an isolated worktree wrecks the private schema, and `provision-db` cannot repair it.**

This section — plus the matching claims in `.claude/CLAUDE.md` ("destructive operations (`migrate:fresh`, destructive migrations) are safe to *test* there") and in `.claude/skills/writing-tests/SKILL.md` ("if it ever gets poisoned you can rebuild it wholesale there (`sh/worktree.sh provision-db`, or `migrate:fresh` + reseed — isolated worktrees only!)") — tells agents to do something that leaves the schema unusable.

**The migration history is squashed.** `database/schema/migrate-schema.dump` records **579** applied migrations, but `database/migrations/` holds only **47** files, and those 47 contain just 8 `Schema::create` calls (`arrows`, `banned_ip_addresses`, `combat_log_parsing_criteria`, `dungeon_route_scheduled_publishes`, `dungeon_speedrun_difficulties`, `dungeon_speedrun_required_npc_npcs`, `enemy_forces_checkpoints`, `mdt_addon_versions`). `dungeons`, `users`, `affix_group_couplings` and everything else exist **only** in the dump.

A bare `migrate:fresh` resolves to the **default** connection, which is `mysql`: `config/database.php` has no top-level `default` key, so Laravel 12 falls back to the framework default `env('DB_CONNECTION', 'sqlite')` → `mysql` (verified — `php artisan tinker --execute 'echo config("database.default");'` prints `mysql`). Laravel then looks for `database/schema/mysql-schema.dump`, which does **not** exist (only `migrate-schema.dump`, `combatlog-schema.dump`, `phpunit-schema.dump` do). So it wipes the schema and replays only the 47 leftovers, the first of which (`2026_01_17_204152_add_initialism_column_to_dungeons_table.php`, a `Schema::table('dungeons', ...)`) dies on a table that no longer exists.

Passing `--database=migrate` instead *finds* a dump, but then routes through `MySqlSchemaState::load()` — i.e. the app image's MariaDB client, the exact ERROR 2026 TLS failure `load_schema_dump()` was written to work around. (That second branch I reasoned from this script's own comment rather than observed; the bare-`migrate:fresh` path above is verified end to end.)

**The worse half is the recovery.** `load_schema_dump()` skips whenever the target schema already has a `migrations` table — and the failed `migrate:fresh` just created one. So `worktree.sh provision-db` skips the dump load, finds nothing pending to migrate, sees `users` empty and runs `db:seed` straight into the missing tables. The documented recovery is a no-op against the exact state it is advertised to fix; the only way out is `remove` + `create` (another 5–15 min seed) or a manual `DROP DATABASE`.

The same guard makes the "Idempotent ... Safe to re-run after a mid-seed failure" comment on `provision_worktree_db()` only half true: the `user_count` gate does cover a mid-*seed* failure, but a mid-*schema-load* failure (container OOM, disk full) leaves partial tables plus a partial `migrations` table and is equally unrecoverable.

Suggested fix: give `provision-db` a `--recreate` flag that `DROP DATABASE`s first (it already validates the name against `^ksg_wt_[a-z0-9_]+$`, so it is cheap and safe), and point all three docs at that instead of at `migrate:fresh`.

sh/worktree.sh
```diff
@@ -223,6 +356,21 @@ cmd_create() {
...
 
+    # 2d. Record the DB mode + schema name in a marker file BEFORE any DB work, so `remove` and
+    #     `prune-db` can always tell whether this worktree owns private schemas — even after a
+    #     mid-create failure. The marker (not the worktree .env, which is read-protected) is the
+    #     source of truth. A re-create of an existing dir must not silently flip modes.
+    local db_mode schema
+    if [ "$shared_db" -eq 1 ]; then db_mode=shared; else db_mode=isolated; fi
+    schema="$(schema_name "$branch")"
+    if [ -f "$wt_path/.worktree-db" ] && [ "$(marker_get "$wt_path" WORKTREE_DB_MODE)" != "$db_mode" ]; then
+        die "worktree '$branch' already exists with DB mode '$(marker_get "$wt_path" WORKTREE_DB_MODE)' — 'worktree.sh remove $branch' first to switch modes"
```
Wotuu, line 367: :robot: **This `die` fires *after* the worktree `.env` has already been overwritten, leaving the worktree pointed at the main stack's database and the main compose file.**

Step 2 (line 311) does `cp "$REPO_ROOT/.env" "$wt_path/.env"` unconditionally, and the DB/redis rewrites plus the `COMPOSE_PROJECT_NAME` / `COMPOSE_FILE` / `WORKTREE_HTTP_PORT` append only happen later, at step 4 (lines 385–409). So when this mode-mismatch check aborts, an existing worktree is left with a `.env` carrying main's `DB_DATABASE` / `DB_PHPUNIT_DATABASE` / `DB_COMBATLOG_DATABASE` / `REDIS_PREFIX` and **no** `COMPOSE_FILE` / `COMPOSE_PROJECT_NAME` — while its `.worktree-db` marker still says `isolated`.

The next `docker compose exec -T app php artisan ...` run from that directory therefore (a) resolves the **main** `docker-compose.yml` under a project name derived from the directory rather than `docker-compose.worktree.yml`, and (b) whatever it reaches talks to the shared dev database while the agent believes it is isolated — precisely the cross-contamination this PR exists to prevent, now with the marker actively asserting the opposite.

This is reachable through a flow the same diff documents: `.claude/skills/generating-thumbnails/SKILL.md` now says to "create the worktree with `sh/worktree.sh create ... --shared-db` when real thumbnails are the point of the task", so an agent that discovers mid-task that it needs Path B will re-run `create` with `--shared-db` on an existing isolated worktree and hit exactly this abort.

Moving the marker read + mode-mismatch check to the top of `cmd_create()` — before the `cp` at step 2 — makes the abort clean, since nothing has been touched yet.

sh/worktree.sh
```diff
@@ -223,6 +356,21 @@ cmd_create() {
         fi
     done
 
+    # 2d. Record the DB mode + schema name in a marker file BEFORE any DB work, so `remove` and
```
Wotuu, line 359: :robot: **Worktrees that predate this change carry no marker, so the guard below never fires and they get silently flipped to isolated mode.**

The mismatch check on line 366 only runs `if [ -f "$wt_path/.worktree-db" ]`. Every worktree created before this lands has no marker file, so re-running `create` on one takes the default (`db_mode=isolated`), writes the marker, rewrites its `.env` DB vars at step 4, and drops into a 5–15 minute provisioning run — with no prompt and no warning that the worktree is about to stop seeing the shared data it has been working against.

Not hypothetical: there are currently 11 marker-less worktrees under `keystone.guru-worktrees/` (`3734-midnight-s2-dry-run`, `3756-affix-season-*`, `3757-mdt-import-game-version-scoping`, `3770-*`, `1769-*`, `700-*`, ...), and the `worktree-docker` skill in this same diff documents `down` + a later `create` as the normal re-attach flow ("`down` + a later `create` re-attaches to the existing schema and **skips the seed** (fast)") — which for those worktrees is instead a mode flip plus a full seed.

Cheap fix: when the directory already exists and carries no marker, grandfather it to `shared` and print a note telling the user to `remove` + `create` if they actually want isolation.

Wotuu:

PR #3779, opened 2026-07-31
Match an affix string on the affixes an affix group actually has (#3776)

### LINKED ISSUE #3776: AffixGroupEaseTierService::getAffixGroupByString() affix-count check assumes 3 or 4 affixes (follow-up to #3756)
:robot: Follow-up from Wotuu's review on #3766 (part 2 of #3756), thread on `app/Service/AffixGroup/AffixGroupEaseTierService.php` line 249.

## The check

`AffixGroupEaseTierService::getAffixGroupByString()` filters an Archon.gg-style comma-separated
affix string before trying to match it to one of the current season's `AffixGroup`s:

```php
// Filter out properties that don't have the correct amount of affixes
if ($affixes->count() === 3 + (int)($currentSeason->seasonal_affix_id !== null)) {
```

This assumes an affix group always has 3 affixes, or 4 when the season also has a
`seasonal_affix_id` set (e.g. Thundering/Xal'atath's Guile-style bonus affixes in older seasons).

## Why it's wrong now

Verified against the live current-season data (season id 17, Midnight S1):

- `seasons.seasonal_affix_id` is `NULL` for season 17, so the check requires
  `$affixes->count() === 3`.
- Every one of season 17's `affix_groups` rows actually has **5** rows in
  `affix_group_couplings`, e.g. affix_group_id 163: Lindormi's Guidance, Fortified, Tyrannical,
  Xal'atath's Bargain: Devour, Xal'atath's Guile.

So for the current season this `if` is always false, and `getAffixGroupByString()` will never
match any affix group - it silently returns `null` for every call instead of finding a match.

Wotuu's original comment:

> I wonder if this check is incorrect. You'd need to file a follow-up issue for this, but the
> affix count of an affix group is no longer necessarily three or four, which this if applies
> implies. In the current season it has five. It has Lindormi's Guidance, Fortified, Tyrannical,
> Xal'Atath's Bargain and Xal'Atath's Guile. So something in this check is not going right and it
> needs a follow-up issue to figure out what and to figure out how to fix this code.

## What needs figuring out

- Whether the 3-or-4 assumption should become "3 base affixes + however many bonus/seasonal-style
  affixes the season actually has" (i.e. driven by real affix-group membership counts rather than
  a hardcoded formula), or whether the count check should be dropped in favor of just trying to
  match the affix names directly.
- Whether `Season::$seasonal_affix_id` is even still a meaningful signal for "does this season have
  extra affixes" now that seasons can carry more than one bonus-style affix (e.g. both
  `Xal'atath's Bargain: Devour` and `Xal'atath's Guile` in season 17, with `seasonal_affix_id` left
  `null`).
- Confirm what actually calls `getAffixGroupByString()` in production and what user-visible
  behavior currently silently breaks because of this (Archon.gg ease-tier lookups for the current
  season presumably never resolve).

Not fixing in this issue - `key`/translated-`name` matching in the loop right below the count
check was also double-checked and is *not* part of the bug (verified: for every `Affix` row,
`__($affix->name, [], 'en_US') === $affix->key)`.

## Prior corroborating history on this exact line

Before PR #3766's commit `d2e0aa130` ("Fix cold-review findings: seasonal_affix_id had no 'none'
option"), this line carried `// @phpstan-ignore notIdentical.alwaysTrue` - PHPStan considered
`$currentSeason->seasonal_affix_id !== null` **always true** at that point, because
`Season::$seasonal_affix_id`'s type wasn't genuinely nullable yet. That means the count check
effectively hardcoded "always expect exactly 4 affixes" for a long stretch of this codebase's
history, with the ignore comment masking the fact that the `!== null` branch could never be
false. `d2e0aa130` fixed the nullability (so a season can now genuinely have `seasonal_affix_id
== null`) and the ignore was correctly dropped as dead - but the underlying 3-or-4 assumption
itself was never revisited, which is how it drifted out of sync with seasons like 17 that have 5
affixes per group. Worth keeping in mind when redesigning the check: it's had at least two
different "always wrong in a different way" states already.

See #3756, PR #3766.

Wotuu:
:robot: Closes #3776

## The bug

`AffixGroupEaseTierService::getAffixGroupByString()` gated its matching on a hardcoded affix count:

```php
if ($affixes->count() === 3 + (int)($currentSeason->seasonal_affix_id !== null)) {
```

That formula ("three affixes, or four when the season has a seasonal affix") stopped describing
reality several seasons ago. Checked against the seeded data:

| season | `seasonal_affix_id` | affixes per affix group | check expects |
|---|---|---|---|
| 12 / 13 (DF S3/S4) | `null` | 3 | 3 ✅ |
| 14 (TWW S1) | `null` | 5 | 3 ❌ |
| 15 / 16 (TWW S2/S3) | `null` | 4 | 3 ❌ |
| 17 (Midnight S1) | `null` | 5 | 3 ❌ |

So for every season since DF S4 the `if` was always false: the method returned `null` for every
input, and silently, because the only log line it had (`getAffixGroupByStringUnknownAffixes`) sat
*inside* the dead branch.

## What is actually broken for users today: nothing

Confirming the third bullet of the issue: Archon.gg no longer sends affix information at all. The
live payload of
`https://www.archon.gg/wow/tier-list/dps-rankings/mythic-plus/10/all-dungeons/this-week/data.json`
has `encounterTierList.label: null` and an `affixes` object of just `{"slug": "this-week"}`.
`parseTierList()` therefore always takes the `label === null` branch and falls back to
`SeasonAffixGroupService::getCurrentAffixGroup()`, so ease tiers still land on the correct affix
group and the tier lists render fine.

`getAffixGroupByString()` is only reachable if Archon ever restores the label (and from the
already-run legacy backfill migration). This is a correctness fix, not an outage fix.

## The fix

Drop the count check entirely and match on the affixes an affix group *actually has*:

- resolve every name in the string to an `Affix` (via the English translation of `Affix::$name`),
  logging and bailing out on any name we don't know;
- return the first affix group of the current season that has all of those affixes, using the
  existing `AffixGroupBase::hasAffix()` helper.

That is order-insensitive, works for any amount of affixes a season may have, and also tolerates a
partial affix string should Archon come back with a shorter label than the full affix set.

Resolving by English name is safe: no two `Affix` rows share one (verified over the whole table),
and every affix's English name currently equals its `key`. Going through a resolved `Affix` rather
than comparing names on the input side and `key` on the affix group side means the matching no
longer silently depends on that second equality holding.

### Which affix group of the week

A season has **two affix groups per affix set** — one for the Fortified week, one for the Tyrannical
week (season 17: 163 ≡ 167, 164 ≡ 169, 165 ≡ 168, 166 ≡ 170). No affix string can tell those apart,
and the choice is not cosmetic: ease tiers are stored *and read* per `affix_group_id`
(`HeaderComposer::getEaseTiers()` and `ViewService::getAffixGroupEaseTiersByAffixGroup()` both look
them up by the id of the **current** affix group), so writing the pull under the wrong twin makes
the tiers vanish from the header and from route cards for that whole week.

Archon's tier list is always for the current week, so the current affix group is preferred whenever
it has the given affixes. Only when it does not (a label for another week) does it fall back to
scanning the season, and it then refuses to guess — logging and returning `null` — if affix groups
with *different* affixes match, which is the case when too few affixes were given to tell them
apart.

On that fallback path two affix groups with *identical* affixes are still picked arbitrarily. That
is deliberate: a label for a week other than the current one carries no signal at all about which of
the two weeks it was, so there is nothing better to pick.

This one is worth calling out: my first version returned the first matching affix group and a
comment claimed set-identical affix groups "share an ease tier". They do not. Caught in review.

Also fixed along the way, both reachable from the legacy migration which calls the method directly:

- **`$currentSeason` was dereferenced unguarded** while `SeasonService::getCurrentSeason()` returns
  `?Season`.
- **An affix string without any affixes** (`''`, whitespace, separators only) would match the
  *first* affix group under the new subset matching, since `Collection::every()` is true on an empty
  collection. It now returns `null`.
- Added `getAffixGroupByStringNoAffixes()`, `getAffixGroupByStringNoMatchingAffixGroup()` and
  `getAffixGroupByStringAmbiguousAffixes()` to the logging interface so a failure to match is no
  longer silent, and so the three ways it can fail are distinguishable in the alert.

`seasonal_affix_id` itself is untouched and not deprecated — it still drives
`ExpansionSeason::$isAwakened` and friends, `DungeonRoute:1021` and the affix tables. It is just no
longer a valid signal for *how many* affixes an affix group has, so it comes out of this one check.

## Tests

There was no direct test of `getAffixGroupByString()` at all — it was either mocked out or exercised
only through `Season::find(13)` (DF S4, 3 affixes), which is exactly why this stayed invisible.
Added coverage in `tests/Feature/App/Service/AffixGroup/AffixGroupEaseTierServiceTest.php` driven by
real seeded affix-group membership rather than any new hardcoded count:

- full affix strings for a 3-affix (DF S4), 4-affix (TWW S3) and 5-affix (Midnight S1) season,
  plus the same affixes in a different order;
- the affixes of the current affix group, which must return **that** affix group and not its twin
  (the test deliberately makes the *last* of the twins current, so it fails if the first match wins);
- the affixes of a different week than the current affix group, which must return that other affix
  group — Archon can still be serving last week's tier list shortly after a reset;
- a partial affix string;
- too few affixes to tell affix groups apart;
- affixes that no single affix group has;
- an unknown affix name;
- empty / whitespace-only / separators-only strings;
- no current season.

Verified these fail against the implementation they replace: the TWW S3 and Midnight S1 rows return
`null` against the original count check, and the current-affix-group test returns 163 instead of 167
against the first version of this MR. Also checked end-to-end against the live current season
through the real service:

```
Lindormi's Guidance, Fortified, Tyrannical, Xal'atath's Bargain: Devour, Xal'atath's Guile
  => affix group 163
```

24 passed locally (`--filter=AffixGroupEaseTierService`), `composer run fix` and
`composer run analyse` are clean. No UI surface, so no visual verification.

One side effect worth knowing about: because the label branch now resolves the current affix group,
the three pre-existing `parseTierList` tests that exercise the real `getAffixGroupByString()` now run
time-based affix resolution where they previously did not. It is harmless — DF S4's affix sets are
all distinct, so the current-week preference can never disagree with the scan there — but that is a
clock coupling those tests did not have before.

`database/migrations_legacy/2024_02_22_104800_...` is the only other caller; that folder is
referenced from nowhere in `app/`, `database/`, `bootstrap/` or `config/`, and the migration has long
since run, so its behaviour change is inert.

### DIFF
--- a/app/Service/AffixGroup/AffixGroupEaseTierService.php
+++ b/app/Service/AffixGroup/AffixGroupEaseTierService.php
@@ -235,7 +235,7 @@ class AffixGroupEaseTierService implements AffixGroupEaseTierServiceInterface
     /**
      * {@inheritdoc}
      */
-    public function getAffixGroupByString(string $affixString): ?AffixGroup
+    public function getAffixGroupByString(string $easeTierAffixString): ?AffixGroup
     {
         $currentSeason = $this->seasonService->getCurrentSeason();
         if ($currentSeason === null) {
@@ -245,13 +245,13 @@ class AffixGroupEaseTierService implements AffixGroupEaseTierServiceInterface
         /** @var Collection<string, Affix> $affixesByName */
         $affixesByName = Affix::all()->keyBy(static fn(Affix $affix): string => (string)__($affix->name, [], 'en_US'));
 
-        $affixNames = collect(explode(',', $affixString))
+        $affixNames = collect(explode(',', $easeTierAffixString))
             ->map(static fn(string $affixName): string => trim($affixName))
             ->filter(static fn(string $affixName): bool => $affixName !== '');
 
         // Without any affixes to match on every affix group would be a match - refuse to guess instead
         if ($affixNames->isEmpty()) {
-            $this->log->getAffixGroupByStringNoAffixes($affixString);
+            $this->log->getAffixGroupByStringNoAffixes($easeTierAffixString);
 
             return null;
         }
@@ -291,7 +291,7 @@ class AffixGroupEaseTierService implements AffixGroupEaseTierServiceInterface
             ->filter(fn(AffixGroup $affixGroup): bool => $this->hasAllAffixes($affixGroup, $affixKeys));
 
         if ($matchingAffixGroups->isEmpty()) {
-            $this->log->getAffixGroupByStringNoMatchingAffixGroup($affixString);
+            $this->log->getAffixGroupByStringNoMatchingAffixGroup($easeTierAffixString);
 
             return null;
         }
@@ -302,7 +302,7 @@ class AffixGroupEaseTierService implements AffixGroupEaseTierServiceInterface
         )->unique();
 
         if ($matchingAffixes->count() > 1) {
-            $this->log->getAffixGroupByStringAmbiguousAffixes($affixString, $matchingAffixes->join(' / '));
+            $this->log->getAffixGroupByStringAmbiguousAffixes($easeTierAffixString, $matchingAffixes->join(' / '));
 
             return null;
         }
--- a/app/Service/AffixGroup/AffixGroupEaseTierService.php
+++ b/app/Service/AffixGroup/AffixGroupEaseTierService.php
@@ -237,41 +237,53 @@ class AffixGroupEaseTierService implements AffixGroupEaseTierServiceInterface
      */
     public function getAffixGroupByString(string $affixString): ?AffixGroup
     {
-        $result = null;
+        $currentSeason = $this->seasonService->getCurrentSeason();
+        if ($currentSeason === null) {
+            return null;
+        }
 
-        $affixList                = Affix::all();
-        $currentSeason            = $this->seasonService->getCurrentSeason();
-        $currentSeasonAffixGroups = $currentSeason->affixGroups;
+        /** @var Collection<string, Affix> $affixesByName */
+        $affixesByName = Affix::all()->keyBy(static fn(Affix $affix): string => (string)__($affix->name, [], 'en_US'));
 
-        $affixes = collect(explode(', ', $affixString));
+        $affixNames = collect(explode(',', $affixString))
+            ->map(static fn(string $affixName): string => trim($affixName))
+            ->filter(static fn(string $affixName): bool => $affixName !== '');
 
-        // Filter out properties that don't have the correct amount of affixes
-        if ($affixes->count() === 3 + (int)($currentSeason->seasonal_affix_id !== null)) {
-            // Check if there's any affixes in the list that we cannot find in our own database
-            $invalidAffixes = $affixes->filter(static fn(string $affixName) => $affixList->filter(static fn(
-                Affix $affix,
-            ) => __($affix->name, [], 'en_US') === $affixName)->isEmpty());
+        // Without any affixes to match on every affix group would be a match - refuse to guess instead
+        if ($affixNames->isEmpty()) {
+            $this->log->getAffixGroupByStringNoMatchingAffixGroup($affixString);
 
-            // No invalid affixes found, great!
-            if ($invalidAffixes->isEmpty()) {
-                // Now we must find affix groups that correspond to the affix list
-                foreach ($currentSeasonAffixGroups as $affixGroup) {
-                    // Loop over the affixes of the affix group and empty the list
-                    $notFoundAffixes = $affixGroup->affixes->filter(static fn(
-                        Affix $affix,
-                    ) => $affixes->search($affix->key) === false);
+            return null;
+        }
 
-                    // If we have found the match, we're done
-                    if ($notFoundAffixes->isEmpty()) {
-                        $result = $affixGroup;
-                        break;
-                    }
-                }
-            } else {
-                $this->log->getAffixGroupByStringUnknownAffixes($invalidAffixes->join(', '));
+        // Check if there's any affixes in the list that we cannot find in our own database
+        $invalidAffixes = $affixNames->reject(static fn(string $affixName): bool => $affixesByName->has($affixName));
+
+        if ($invalidAffixes->isNotEmpty()) {
+            $this->log->getAffixGroupByStringUnknownAffixes($invalidAffixes->join(', '));
+
+            return null;
+        }
+
+        $affixKeys = $affixNames->map(static function (string $affixName) use ($affixesByName): string {
+            /** @var Affix $affix Guaranteed to exist - any name that could not be resolved was rejected above */
+            $affix = $affixesByName->get($affixName);
+
+            return $affix->key;
+        });
+
+        // The amount of affixes an affix group has varies per season (three in DF S4, four in TWW S3, five in
+        // Midnight S1), so match on the affixes an affix group actually has rather than on a hardcoded count.
+        // A season may contain multiple affix groups with the exact same set of affixes - they share an ease tier,
+        // so returning the first affix group that has all the given affixes is fine.
+        foreach ($currentSeason->affixGroups as $affixGroup) {
+            if ($affixKeys->every(static fn(string $affixKey): bool => $affixGroup->hasAffix($affixKey))) {
+                return $affixGroup;
             }
         }
 
-        return $result;
+        $this->log->getAffixGroupByStringNoMatchingAffixGroup($affixString);
+
+        return null;
     }
 }
--- a/app/Service/AffixGroup/AffixGroupEaseTierService.php
+++ b/app/Service/AffixGroup/AffixGroupEaseTierService.php
@@ -278,7 +278,10 @@ class AffixGroupEaseTierService implements AffixGroupEaseTierServiceInterface
         // group of the current week whenever it has the given affixes.
         $currentAffixGroup = $this->seasonAffixGroupService->getCurrentAffixGroup($currentSeason);
 
-        if ($currentAffixGroup !== null && $this->hasAllAffixes($currentAffixGroup, $affixKeys)) {
+        // The current affix group may be that of a timewalking event, which is not an affix group of this season
+        if ($currentAffixGroup !== null &&
+            $currentAffixGroup->season_id === $currentSeason->id &&
+            $this->hasAllAffixes($currentAffixGroup, $affixKeys)) {
             return $currentAffixGroup;
         }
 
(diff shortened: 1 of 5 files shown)


app/Service/AffixGroup/AffixGroupEaseTierService.php
```diff
@@ -237,41 +237,84 @@ public function getTiers(): Collection
...
+
+        // A season has multiple affix groups with the exact same affixes - one for each week that the affixes are
+        // active - which no affix string can tell apart. Ease tiers are stored and read per affix group, so picking
+        // the wrong one hides them entirely. The tier list is always that of the current week, so prefer the affix
+        // group of the current week whenever it has the given affixes.
+        $currentAffixGroup = $this->seasonAffixGroupService->getCurrentAffixGroup($currentSeason);
+
+        // The current affix group may be that of a timewalking event, which is not an affix group of this season
+        if ($currentAffixGroup !== null &&
+            $currentAffixGroup->season_id === $currentSeason->id &&
```
Wotuu, line 283: :robot: Cold review: this `season_id` guard is not covered by any test.

Every test that supplies a current affix group derives it from `getAffixGroupsWithTheSameAffixes(Season::SEASON_MIDNIGHT_S1, ...)`, i.e. always an affix group of the same season that `getSeasonService()` returns as the current one. The remaining tests mock `getCurrentAffixGroup()` to `null`. So no test ever passes a `$currentAffixGroup` whose `season_id` differs from the current season's, and nothing would fail if this condition were dropped again.

The guard is genuinely load-bearing: `SeasonAffixGroupService::getAffixGroupAt()` short-circuits to `TimewalkingEventService::getAffixGroupAt()` for an expansion with a timewalking event, and that returns an affix group of `getCurrentSeason($expansion)` — a different season entirely. Without the guard, during a timewalking window the tier list can be filed under an affix group that isn't in `$currentSeason->affixGroups` at all, which is exactly what the method's docblock promises can't happen.

Worth noting because commit `23a3a6f45`'s own message says the previous commit's problem was that "dropping the affix check from the current week's fast path passed every test" — the same blind spot now applies to the condition added right next to it. A test that hands `getCurrentAffixGroup()` an affix group from another season (with affixes matching the requested string) and asserts the result belongs to the current season would pin it down.

app/Service/AffixGroup/AffixGroupEaseTierServiceInterface.php
```diff
@@ -34,5 +35,11 @@ public function getTiersByAffixGroups(Collection $affixGroups): Collection;
      */
     public function getTiers(): Collection;
 
+    /**
+     * Finds the affix group of the current season that has all affixes in the given comma separated list of (English)
+     * affix names, or null if no such affix group exists or if multiple affix groups with different affixes have them.
```
Wotuu: :robot: Cold review: this docblock states the null-on-ambiguity contract unconditionally, but the implementation has a preceding fast path that overrides it.

If the given affixes are a subset of the current week's affix group, `getAffixGroupByString()` returns that affix group and never reaches the ambiguity check — so e.g. a bare `Tyrannical` in Midnight S1 (every affix group of that season has both Fortified and Tyrannical) resolves to the current week's affix group rather than `null`, even though multiple affix groups with different affixes have it. The behaviour itself looks intentional and defensible (the tier list being parsed is the current week's), it's just that the documented contract doesn't mention the current-week preference takes precedence. Suggest wording it as "prefers the affix group of the current week when it has the given affixes, otherwise ... or null if ...".

Same root observation on the test side: `getAffixGroupByString_givenTooFewAffixesToTellAffixGroupsApart_returnsNull` only reaches the ambiguity branch because it mocks `getCurrentAffixGroup()` to `null`. For a live season that's not a realistic state, so the branch it covers is one production will rarely hit — an ambiguous string that is *not* a subset of the current week's affixes would exercise it under realistic conditions.

Wotuu:

app/Service/AffixGroup/AffixGroupEaseTierService.php
```diff
@@ -237,41 +237,84 @@ public function getTiers(): Collection
      */
     public function getAffixGroupByString(string $affixString): ?AffixGroup
```
Wotuu: Can you rename the affixString variable to something like externalAffixString or easeTierAffixString, archonAffixString something a bit more different than affixesByName.

Wotuu (commented):
One small nit

app/Service/AffixGroup/AffixGroupEaseTierService.php
Wotuu: :robot: Added `getAffixGroupByString_givenCurrentAffixGroupIsOfADifferentSeason_doesNotShortCircuitToThatAffixGroup`: it sets the current season to Midnight S1 but mocks `getCurrentAffixGroup()` to return a DF S4 affix group whose affixes (Fortified, Entangling, Bolstering) would otherwise satisfy the requested string, then asserts the method falls through instead of returning that wrong-season group (it ends up returning null, since no Midnight S1 affix group has Entangling/Bolstering). I confirmed the test is load-bearing by temporarily dropping the `season_id` guard locally - the test failed as expected, then reverted.

app/Service/AffixGroup/AffixGroupEaseTierServiceInterface.php
Wotuu: :robot: Reworded the docblock to: "Finds the affix group of the current season that has all affixes in the given comma separated list of (English) affix names - the affix group may have more affixes than were given. Prefers the affix group of the current week when it has them; otherwise returns the matching affix group, or null when no affix group of the season has them, or when multiple affix groups with differing affixes do." This now states both the subset-match behavior (confirmed against `hasAllAffixes()`, and the existing `givenPartialAffixStringOfTheCurrentSeason` test) and the current-week-preference precedence.

Also added `getAffixGroupByString_givenTooFewAffixesToTellAffixGroupsApartAndACurrentAffixGroupThatDoesNotMatch_returnsNull` for your second point - it sets a real (non-null) current affix group that does not have the ambiguous affix, so the ambiguity branch is reached the way it would be in a live season, not only because `getCurrentAffixGroup()` was mocked to null.

app/Service/AffixGroup/AffixGroupEaseTierService.php
Wotuu: :robot: Renamed the parameter to `easeTierAffixString` (throughout the method - it's also passed to the log calls, so I updated those call sites too, and updated the interface signature to match).

Wotuu (commented):
Ok

PR #3781, opened 2026-07-31
#3774 Stop the feature-toggle admin action from blanket-granting stored user values

### LINKED ISSUE #3774: Toggling a feature in the admin panel grants it to every user who already has a stored value
:robot: Turning a feature on in Admin -> Tools -> Features grants it to **every user who already has a stored value for it**, regardless of whether that user's roles entitle them to it.

## Mechanism

`AdminToolsFeaturesController::toggleFeature()` calls Pennant's `Feature::activateForEveryone($feature)`, which bottoms out in `DatabaseDriver::setForAllScopes()`:

```php
$this->newQuery()->where('name', $feature)->update(['value' => ...]);
```

There is **no scope predicate** - it flips every row with that name to `true`, for every scope.

Meanwhile `resources/views/common/layout/header.blade.php` evaluates `Feature::active(NpcCompendium::class)` on every page render, so effectively every user who has ever loaded a page already has a stored row. Toggling the feature on therefore flips all of them to `true` at once, and nothing ever re-resolves them against the user's roles afterwards.

## Impact

`NpcCompendium` resolves to `ROLE_ADMIN || ROLE_INTERNAL_TEAM` only, but this hands it to the general user base. Same shape for `XalatathTheme`, `SearchPageRework` and `DungeonRouteListRework`.

Confirmed in the local development database - a spread of ordinary, non-internal-team users carry `App\Features\NpcCompendium = true`:

```
App\Models\User|75721 = true
App\Models\User|75942 = true
App\Models\User|76017 = true
App\Models\User|76441 = true
App\Models\User|78836 = true
...
```

Worth checking Production/Staging for the same before deciding urgency.

## Suggested fix

The admin toggle wants to move the *global* switch, not everybody's resolved value. Purging instead of blanket-updating would let each user's value be resolved anew from their roles:

```php
Feature::purge($feature);          // drop every stored value
Feature::for(null)->activate($feature);   // ... then set the global switch
```

Note `Feature::getAdminValue()` currently reads the row scoped to user 1 as that global switch (see `Feature::ADMIN_USER_ID`, added in #3772), so a fix needs to keep that row consistent - or move the switch somewhere that isn't a user's resolved value in the first place, which is arguably the real cleanup.

## Related

This is the mirror image of #3772 (stale `false` never re-resolved after a role is granted) and shares its root cause: resolved values are stored forever and nothing invalidates them when their inputs change. #3772 fixes the role-change direction; this issue is the admin-toggle direction.

Second, much smaller leak found in the same review: `User`'s `deleting` hook syncs roles via the relation rather than Laratrust's `syncModels()`, so no role event fires and a deleted user's `features` rows are never cleaned up. Harmless (ids are not reused) but the table grows unbounded.

Wotuu:
:robot: Closes #3774

## What
`AdminToolsFeaturesController::toggleFeature()` called Pennant's `activateForEveryone()`/
`deactivateForEveryone()`, which bottom out in `DatabaseDriver::setForAllScopes()` — an
`UPDATE features SET value WHERE name = ?` with **no scope predicate**. Toggling a feature
on/off in the admin panel therefore flipped *every* stored row for that feature name,
regardless of whether the user's roles actually entitled them to it.

Fix: purge every stored row for the feature, then set only the row scoped to the admin
user (`Feature::ADMIN_USER_ID`, i.e. the global on/off switch read by `getAdminValue()`).
Every other user's value is then resolved fresh — off their current roles — the next time
`Feature::active()` runs for them.

I deliberately deviated from the snippet in the issue body: `Feature::for(null)->activate(...)`
would store scope `__laravel_null`, which `getAdminValue()` (which queries scope
`App\Models\User|1`) would never find — the switch would never actually turn on. Using
`Feature::for($adminUser)` instead targets that exact scope.

`$wasActive` now reads `Feature::getAdminValue()` instead of `Laravel\Pennant\Feature::active()`
— the latter resolves for whichever admin happens to be logged in and clicking the button,
the former reads the literal global switch. Same outcome in practice (an admin always has
`ROLE_ADMIN`), but it's the more direct/correct read for what's actually being displayed
and toggled.

Also fixes the smaller leak flagged in the same review: Laratrust's own `deleting` hook on
`User` calls `roles()->sync([])` directly rather than going through `syncRoles()`/`removeRoles()`,
so no `role.removed`/`role.synced` event fires and #3773's `Feature::forgetAllForUser()` wiring
never runs for a deleted user. `User::boot()` now calls it explicitly in its own `deleting` hook.

## Not in scope here
- **`forgetFeature()`** is untouched. It has an adjacent, pre-existing sharp edge: if the
  logged-in admin happens to be user 1, pressing "Forget" clears that feature's global switch
  row (since it forgets for the *currently authenticated* user, unscoped). Not touching it —
  "what should Forget do" is a product/behavior question, not a bug this issue is about.
- **Already-leaked data.** The issue itself notes ordinary users in the dev DB already carry
  `NpcCompendium = true` from a prior toggle. This MR only stops it from happening again — it
  does not retroactively clean up rows already flipped. If prod/staging carry the same shape,
  the cleanup is a manual one-time step *after* this deploys: toggle each affected feature off
  then on from Admin → Tools → Features (which, with this fix, purges + re-sets it cleanly).

## Test plan
- [x] New `AdminToolsFeaturesControllerTest`: toggling on with an ordinary user's stored
  `false` row purges it (not blanket-flips it) and the feature still resolves `false` for
  that user; toggling off flips the global switch off.
- [x] New `FeatureForgetAllForUserTest::delete_givenUserWithStoredFeatureValues_forgetsThoseValues`
  covers the deletion-hook gap.
- [x] `php artisan test --compact` — 413 passed across the Feature/Controller/AdminTools groups.
- [x] `composer run analyse` (PHPStan) — no errors.
- [x] `composer run fix` (PHP CS Fixer) — no changes needed beyond the new test file.
- Visual verification: N/A, no rendering change to the features list page.

## Cold review
An independent agent reviewed this PR and confirmed the core logic (scope-string match
between `Feature::for($adminUser)` and `getAdminValue()`, and that `purge($feature)` scopes
to just that one feature name). It raised two real findings, both addressed in a follow-up
commit:
- The purge-then-set wasn't atomic - a request resolving the feature in that gap could read
  a momentarily-missing admin-switch row as "disabled" and persist a stale `false`,
  recreating the #3772 bug this issue is adjacent to. Now wrapped in `DB::transaction()`.
- The deactivate-direction test only asserted the admin switch flipped, which the old buggy
  `deactivateForEveryone()` code would have passed identically - it didn't actually prove
  purging happens. It now also arranges an entitled user's stored `true` row and asserts
  it's purged rather than updated to `false` in place.

### DIFF
--- a/app/Http/Controllers/AdminTools/AdminToolsFeaturesController.php
+++ b/app/Http/Controllers/AdminTools/AdminToolsFeaturesController.php
@@ -3,11 +3,13 @@
 namespace App\Http\Controllers\AdminTools;
 
 use App\Http\Controllers\Controller;
+use App\Models\Feature\Feature;
+use App\Models\User;
 use HaydenPierce\ClassFinder\ClassFinder;
 use Illuminate\Http\RedirectResponse;
 use Illuminate\Http\Request;
 use Illuminate\View\View;
-use Laravel\Pennant\Feature;
+use Laravel\Pennant\Feature as PennantFeature;
 use Session;
 
 class AdminToolsFeaturesController extends Controller
@@ -23,11 +25,17 @@ class AdminToolsFeaturesController extends Controller
     {
         $feature = (string)$request->get('feature');
 
-        $wasActive = Feature::active($feature);
+        $wasActive = Feature::getAdminValue($feature);
+
+        // Purge every stored value first, so that each user's feature re-resolves against their own roles the
+        // next time it's checked, instead of blanket-flipping their already-cached rows to the new switch value
+        PennantFeature::purge($feature);
+
+        $adminUser = User::findOrFail(Feature::ADMIN_USER_ID);
         if ($wasActive) {
-            Feature::deactivateForEveryone($feature);
+            PennantFeature::for($adminUser)->deactivate($feature);
         } else {
-            Feature::activateForEveryone($feature);
+            PennantFeature::for($adminUser)->activate($feature);
         }
 
         Session::flash('status', __(!$wasActive ?
@@ -43,8 +51,8 @@ class AdminToolsFeaturesController extends Controller
     {
         $feature = (string)$request->get('feature');
 
-        Feature::forget($feature);
-        Feature::for(null)->forget($feature);
+        PennantFeature::forget($feature);
+        PennantFeature::for(null)->forget($feature);
 
         Session::flash('status', __('controller.admintools.flash.feature_forgotten', ['feature' => $feature]));
 
--- a/app/Http/Controllers/AdminTools/AdminToolsFeaturesController.php
+++ b/app/Http/Controllers/AdminTools/AdminToolsFeaturesController.php
@@ -8,6 +8,7 @@ use App\Models\User;
 use HaydenPierce\ClassFinder\ClassFinder;
 use Illuminate\Http\RedirectResponse;
 use Illuminate\Http\Request;
+use Illuminate\Support\Facades\DB;
 use Illuminate\View\View;
 use Laravel\Pennant\Feature as PennantFeature;
 use Session;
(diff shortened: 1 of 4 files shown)


Wotuu (commented):
Ok

PR #3784, opened 2026-08-01
Stop a test's environment flip from redirecting structured logging to stderr

### LINKED ISSUE #3782: Structured logging still leaks into PHPUnit CI output (follow-up to #3722)
:robot: #3722 / PR #3724 (`Stop structured logging from leaking into PHPUnit output`, commit `b2c791202`, squash-merged as `d6955346d`) fixed `StructuredLogging::resolveChannel()` to resolve the log channel per call instead of caching it once in the constructor. It's still happening.

## Evidence

CI run https://github.com/RaiderIO/keystone.guru/actions/runs/30668384463/job/91280579449?pr=3781 (`php-tests (feature-other)`, 2026-07-31) has **965** ANSI-coloured `production.INFO`/`production.DEBUG` lines interleaved with the PHPUnit progress dots — worse than the 82 lines the original issue reported.

All of them come from `MDTMappingImportServiceLogging` and `CacheServiceLogging`, and they're confined to one contiguous window in the log (roughly test 252/718 → 315/718, i.e. exactly one MDT-import test's execution) — before and after that window, output is clean progress dots. This points at one of the MDT-import-related tests specifically, most likely one of:
- `tests/Feature/App/Service/MDT/MDTMappingImportCrashRecoveryTest.php`
- `tests/Feature/App/Service/MDT/MDTMappingImportEnemyForcesCheckpointTest.php`
- `tests/Feature/App/Service/Mapping/MappingServiceCreateNewMappingVersionFromMDTMappingTest.php`

(all three exercise `MDTMappingImportService::importMappingVersionFromMDT`, matching the leaked log lines).

## What's already ruled out

- The fix from #3722 (per-call `resolveChannel()` resolution) *is* present on `master` in `app/Logging/StructuredLogging.php` — verified by reading the file directly, not just checking history.
- `phpunit.xml` sets `APP_ENV=testing`; `.env.ci.example`/`.env.example` set `APP_TYPE=local`, so `resolveChannel()`'s stderr condition (`runningInConsole() && !runningUnitTests() && app.type === 'local'`) should be reachable only if `runningUnitTests()` briefly answers `false` mid-suite.
- No `StructuredLogging::setChannel()` calls exist outside the class's own test.
- None of the three candidate test files spawn a subprocess (`Process`, `exec`, `Artisan::call` to a shell) or use paratest/parallel workers — CI runs plain `./vendor/bin/phpunit` per shard, one process per matrix leg.

## Hypothesis to check first

Given the leak is scoped to exactly one test's execution and not the whole suite, something during that specific test likely causes `$app->runningUnitTests()` (or `app()` itself) to transiently answer differently than normal — e.g. a fresh `Application`/container swap, a faked/rebound `app('env')`, or something in `MDTMappingImportService`/`MappingService` that resolves a new app instance rather than using the one PHPUnit bootstrapped. Worth adding a temporary `dump(app()->runningUnitTests(), app()->runningInConsole(), config('app.type'))` inside `StructuredLogging::resolveChannel()` and running just that one test file locally to catch it in the act, then removing the dump before merging the real fix.

## Definition of done

- Root cause identified (not just channel-resolution tweaked again without understanding why it flips for this one test).
- The three candidate test files run clean with zero `production.*` lines in PHPUnit's own stdout.
- A regression test added if the existing `tests/Unit/App/Logging/StructuredLoggingTest.php` coverage doesn't already cover this specific trigger.
- Full `feature-other` CI shard clean of structured-log leakage (re-check a fresh CI run, not just local).

Wotuu:
:robot: Closes #3782

## Root cause

`php-tests (feature-other)` printed **965** ANSI-coloured `production.INFO`/`production.DEBUG` lines into the job output. #3722 / PR #3724 had already made `StructuredLogging::resolveChannel()` resolve per log call instead of latching the channel in the constructor — a real fix, but for a different cause.

Sorting the [raw job log](https://github.com/RaiderIO/keystone.guru/actions/runs/30668384463/job/91280579449) by its own per-line timestamps splits the output in two (the on-screen interleaving with PHPUnit's block-buffered progress dots is a rendering artifact, which is what made the leak look like it spanned ~63 tests):

| window | lines | source |
| --- | --- | --- |
| 22:05:49.10 | 6 × `CacheServiceLogging::deleteKeysByPattern*` | the `php artisan db:seed` CI step, which runs *before* the workflow appends `APP_ENV=testing` — **legitimate**, not a leak |
| 22:06:43.10 → 22:06:45.94 | 959 × `MDTMappingImportServiceLogging::*`, exactly one `importMappingVersionFromMDTStart`/`End` pair, `dungeonId 49`, `elapsedMS: 2807.34` | **the leak** — one test, not the whole run |

The culprit is `MDTMappingImportGameVersionScopingTest` (landed the day of that CI run, in `0d6220580` / #3762 for #3757), which flipped the whole application environment to `production` around its Act step so `AppServiceProvider`'s lazy-loading violation handler would log rather than throw on `Npc::$npcHealths`:

```php
$this->app->detectEnvironment(static fn() => 'production');
```

`Application::runningUnitTests()` is `$this->bound('env') && $this['env'] === 'testing'`, so the flip turns it **false**. The other two conditions in `resolveChannel()` are always true in CI (`runningInConsole()` under phpunit; `config('app.type') === 'local'` from `.env.ci.example`, and `phpunit.xml` never sets `APP_TYPE`) — so every line went to the `stderr` channel with its `ColoredLineFormatter` tap. It also explains the `production.` prefix specifically: `ParsesLogConfiguration::parseChannel()` names a channel after `$app->environment()` when the config carries no explicit name.

Nothing here is transient or ordering-dependent; it reproduces deterministically.

## The fix

**1. The test stops lying about the environment.** It now swaps the lazy-loading *violation handler* for a no-op around the Act step, restoring the original in a `finally`.

Worth calling out, because the first attempt at this got it wrong and CI caught it: `Model::preventLazyLoading(false)` is **not** equivalent and does not work. That flag is snapshotted onto each model instance as it is hydrated (`Builder::hydrate()`), so instances `laravel-model-caching` unserializes from cache still carry `preventsLazyLoading = true` and throw anyway — which is why the flag variant passed locally (`MODEL_CACHE_ENABLED=false`) and errored in CI (`MODEL_CACHE_ENABLED=true`). The handler, by contrast, is a static consulted at violation time, which is exactly the mechanism the original environment flip relied on.

**2. `resolveChannel()` no longer decides "am I running tests" from `app.env` alone.** Five other test files already flip it to exercise production-only behaviour (`ThumbnailServiceTest`, `FileTest`, `MigrateThumbnailDiskTest`, `RenderThumbnailTest`, `TrustProxiesTest`); they just don't happen to drive a chatty service today. Fixing only the one test would leave the trap armed for the next one. `PHPUNIT_COMPOSER_INSTALL` is defined by PHPUnit's own entrypoint (`vendor/bin/phpunit`, which is how both CI and `php artisan test` invoke it), and phpunit is `require-dev`, so the constant cannot exist in production.

**3. A regression test** in `StructuredLoggingTest` pinning the exact trigger: environment flipped to `production`, channel must still be `null`.

**4. CI now fails the build on any leak.** This guard has regressed twice (#3722 → #3782) and each time the protection lived in application state a test can rewrite, so `php-tests.yml` asserts the PHPUnit step's own output carries no log lines. Scoped to that step deliberately — the setup steps run real artisan commands and log to stderr legitimately. Verified all five shards were already clean by this check on the previous run, so it goes in green rather than papering over an existing leak.

## Verification

| check | result |
| --- | --- |
| `MDTMappingImportGameVersionScopingTest`, before the change | 230 leaked lines |
| same, after | **0**, test green |
| same, with the environment flip deliberately put **back** | **0** — part 2 closes the leak on its own, independently of part 1 |
| same, under the CI condition (`MODEL_CACHE_ENABLED=true`) | errors before part 1's correction, green after |
| full `feature-other` shard locally (725 tests) | **0** leaked lines |
| `tests/Unit/App/Logging/` | 19/19 green |
| `composer run fix` / `composer run analyse` | clean |

The leak grep used here matches `<channel>.<LEVEL>:` for **any** channel name, not just `production.` — `FileTest` and `ThumbnailServiceTest` flip to `local`, and a `production.`-only grep would have reported those clean regardless.

**Pre-existing local failures, not caused by this change:** 5 MDT fixture tests (`MDTImportStringServiceMDT2AuthoritativeTest` for Kings' Rest / Ruby Life Pools, `MDTNpcMappingCoverageTest`) fail identically with this branch's `app/` and `tests/` reverted to master, in the same worktree. They are worktree-DB drift for those two dungeons — the main checkout on master passes all 49, and CI is green on them.

## Adjacent findings, deliberately not addressed

- `config/logging.php` defines `stack_docker` / `stack_docker_local`, both `['stderr', 'daily']`. If `LOG_CHANNEL` ever resolved to one of those, `resolveChannel()` returning `null` would not help. Nothing sets it today, and change 4 would now catch it at build time. Cold review suggested `force="true"` on `phpunit.xml`'s `LOG_CHANNEL`; not doing that, because it would also block a deliberate local `LOG_CHANNEL=stdout vendor/bin/phpunit` override, and the CI assertion covers the case that actually matters.
- Five other sites answer "am I in a test" from `app.env` alone (`ApiAuthentication`, `ApiMetrics`, `DebugInfoContextLogger`, `KeystoneGuruServiceProvider`, `AppServiceProvider`) and are exposed to the same environment flips. Deliberately **not** converging them on `isRunningTests()` — widening "in a test" is fine for picking a log channel and materially worse for an auth-bypass guard. Worth a separate issue.
- `StructuredLogging::$ENABLED` is a process-wide static that `SaveToOpensearch` and `ConvertToEvents` `disable()` without re-enabling. No test invokes those commands today, but it is the same class of latch with the opposite symptom — it would silence logging and make a leak-detection test pass for the wrong reason.
- The underlying gap that made the test need a workaround at all (the import pipeline lazy-loads `Npc::$npcHealths`) stays out of scope, exactly as #3757 scoped it.

---

## Cold review (two rounds, fresh-context agents)

Round 1 found no defects in the change; its nits (a docblock enumeration that this MR makes stale, an arrange guard sitting in the Assert block, an asymmetric `app['env']` save/restore) are applied.

Round 2 found one real gap, now fixed in `480f4c62a`: `resolveChannel()` honours an explicit `setChannel()` **before** any of its test-detection checks, and nothing resets `$CHANNEL` between tests — `TestCase::setUp()` already flushes StructuredLogging's other statics for exactly this reason, but `flushConfigCache()` covers only `$LOG_LEVEL` and `$PRETTY_PRINT`. A `setChannel('stderr')` a future test failed to unwind would latch the leak on for the rest of the shard, through the one door part 2 cannot close. Reset in `setUp()`; deliberately *not* folded into `flushConfigCache()`, which `StructuredLogging::enable()` also calls, where clearing a deliberately-set channel would be wrong.

Two round-2 suggestions were considered and declined, both stated here so they can be overridden:
- Adding the same assertion to `php-coverage.yml`. It is `schedule`/`workflow_dispatch` and gates nothing, and the `$CHANNEL` reset above addresses the single-process risk it was raised for.
- `force="true"` on `phpunit.xml`'s `LOG_CHANNEL`. It would also block a deliberate local `LOG_CHANNEL=stdout vendor/bin/phpunit` override, and the CI assertion covers the case that matters.

## CI result

The `feature-other` shard that this issue is about now runs **725 tests with 0 leaked lines** (was 959), and the new assertion step ran and passed on all five shards.

### DIFF
--- a/app/Logging/StructuredLogging.php
+++ b/app/Logging/StructuredLogging.php
@@ -342,6 +342,22 @@ abstract class StructuredLogging implements StructuredLoggingInterface
     }
 
     /**
+     * Whether this process is a test run - deliberately not just runningUnitTests().
+     *
+     * runningUnitTests() answers "is app.env equal to testing", which is not the question being asked: tests are
+     * free to change it, and several do to exercise production-only behaviour (ThumbnailServiceTest, FileTest,
+     * MigrateThumbnailDiskTest, RenderThumbnailTest, MDTMappingImportGameVersionScopingTest). One of those flipping
+     * the environment around a chatty service is what dumped ~1000 ANSI-coloured log lines into the CI test output
+     * in #3782. PHPUNIT_COMPOSER_INSTALL is defined by PHPUnit's own entrypoint (vendor/bin/phpunit, which is how
+     * both CI and `php artisan test` invoke it) and phpunit is a require-dev package, so the constant answers
+     * "is PHPUnit running" without depending on any application state that a test can rewrite.
+     */
+    private static function isRunningTests(Application $app): bool
+    {
+        return $app->runningUnitTests() || defined('PHPUNIT_COMPOSER_INSTALL');
+    }
+
+    /**
      * The channel every structured log line is written to: an explicit setChannel() override if one
      * was made, otherwise stderr for local console usage so an artisan command shows its own output,
      * and otherwise the configured default.
@@ -361,7 +377,7 @@ abstract class StructuredLogging implements StructuredLoggingInterface
         /** @var Application $app */
         $app = app();
 
-        if ($app->runningInConsole() && !$app->runningUnitTests() && config('app.type') === 'local') {
+        if ($app->runningInConsole() && !self::isRunningTests($app) && config('app.type') === 'local') {
             return 'stderr';
         }
 
--- a/app/Logging/StructuredLogging.php
+++ b/app/Logging/StructuredLogging.php
@@ -376,7 +376,27 @@ abstract class StructuredLogging implements StructuredLoggingInterface
         /** @var Application $app */
         $app = app();
 
-        if ($app->runningInConsole() && !self::isRunningTests($app) && config('app.type') === 'local') {
+        // Named arguments here on purpose: all three parameters are bool, so a transposition (e.g.
+        // runningInConsole <-> isRunningTests) would type-check and pass PHPStan while silently
+        // inverting the decision. Naming them makes that mistake visible at the call site.
+        return self::resolveConsoleChannel(
+            runningInConsole: $app->runningInConsole(),
+            isRunningTests: self::isRunningTests($app),
+            isLocalAppType: config('app.type') === 'local',
+        );
+    }
+
+    /**
+     * Pure decision extracted from resolveChannel() so the stderr branch can be unit-tested directly.
+     * Under PHPUnit, defined('PHPUNIT_COMPOSER_INSTALL') is true for the entire lifetime of the process,
+     * so that branch is permanently unreachable from resolveChannel() itself - there is no way to make
+     * isRunningTests() return false from within a test. Public (rather than private) for the same reason
+     * the other resolve/get/set methods here are public: it is stateless and side-effect-free, and a
+     * subclass-based test double cannot expose a private static method the way it can a protected instance one.
+     */
+    public static function resolveConsoleChannel(bool $runningInConsole, bool $isRunningTests, bool $isLocalAppType): ?string
+    {
+        if ($runningInConsole && !$isRunningTests && $isLocalAppType) {
             return 'stderr';
         }
 
--- a/app/Logging/StructuredLogging.php
+++ b/app/Logging/StructuredLogging.php
@@ -345,10 +345,9 @@ abstract class StructuredLogging implements StructuredLoggingInterface
      * Whether this process is a test run - deliberately not just runningUnitTests().
      *
      * runningUnitTests() answers "is app.env equal to testing", which is not the question being asked: tests are
-     * free to change it, and several do to exercise production-only behaviour (ThumbnailServiceTest, FileTest,
-     * MigrateThumbnailDiskTest, RenderThumbnailTest, MDTMappingImportGameVersionScopingTest). One of those flipping
-     * the environment around a chatty service is what dumped ~1000 ANSI-coloured log lines into the CI test output
-     * in #3782. PHPUNIT_COMPOSER_INSTALL is defined by PHPUnit's own entrypoint (vendor/bin/phpunit, which is how
+     * free to change it, and several do to exercise production-only behaviour. One of those flipping the
+     * environment around a chatty service is what dumped ~1000 ANSI-coloured log lines into the CI test output in
+     * #3782. PHPUNIT_COMPOSER_INSTALL is defined by PHPUnit's own entrypoint (vendor/bin/phpunit, which is how
      * both CI and `php artisan test` invoke it) and phpunit is a require-dev package, so the constant answers
      * "is PHPUnit running" without depending on any application state that a test can rewrite.
      */
--- a/tests/Feature/App/Service/MDT/MDTMappingImportGameVersionScopingTest.php
+++ b/tests/Feature/App/Service/MDT/MDTMappingImportGameVersionScopingTest.php
@@ -13,9 +13,11 @@ use App\Service\Cache\CacheServiceInterface;
 use App\Service\Coordinates\CoordinatesServiceInterface;
 use App\Service\Mapping\MappingServiceInterface;
 use App\Service\MDT\MDTMappingImportServiceInterface;
+use Closure;
 use Illuminate\Database\Eloquent\Model;
 use PHPUnit\Framework\Attributes\Group;
 use PHPUnit\Framework\Attributes\Test;
+use ReflectionProperty;
 use Tests\TestCases\PublicTestCase;
 
 /**
@@ -97,8 +99,17 @@ final class MDTMappingImportGameVersionScopingTest extends PublicTestCase
             // mapping:save runbook does via `APP_ENV=production`). That also made runningUnitTests() answer
             // false, so StructuredLogging::resolveChannel() selected 'stderr' and the import dumped ~1000
             // ANSI-coloured log lines straight into the CI test output - see #3782.
-            $preventsLazyLoading = Model::preventsLazyLoading();
-            Model::preventLazyLoading(false);
+            //
+            // Swapping the violation handler is what replaces that flip, NOT Model::preventLazyLoading(false):
+            // the handler is a static consulted at violation time (Model::handleLazyLoadingViolation), which is
+            // exactly why flipping the environment worked, whereas the prevention flag is snapshotted onto each
+            // model instance as it is hydrated (Builder::hydrate()). Instances that laravel-model-caching
+            // unserializes from cache therefore carry whatever the flag was when they were cached and still
+            // throw - which is why the flag variant passed locally (MODEL_CACHE_ENABLED=false) and errored in
+            // CI (MODEL_CACHE_ENABLED=true).
+            /** @var Closure|null $originalViolationCallback */
+            $originalViolationCallback = new ReflectionProperty(Model::class, 'lazyLoadingViolationCallback')->getValue();
+            Model::handleLazyLoadingViolationUsing(static fn() => null);
 
             try {
                 $newMappingVersion = $mappingImportService->importMappingVersionFromMDT(
@@ -110,7 +121,7 @@ final class MDTMappingImportGameVersionScopingTest extends PublicTestCase
                     false,
                 );
             } finally {
-                Model::preventLazyLoading($preventsLazyLoading);
+                Model::handleLazyLoadingViolationUsing($originalViolationCallback);
             }
 
             // Assert
--- a/tests/Feature/App/Service/MDT/MDTMappingImportGameVersionScopingTest.php
+++ b/tests/Feature/App/Service/MDT/MDTMappingImportGameVersionScopingTest.php
@@ -87,30 +87,15 @@ final class MDTMappingImportGameVersionScopingTest extends PublicTestCase
         try {
             // Act
 
-            // importNpcsDataFromMDT() lazy-loads Npc::$npcHealths (App\Models\Npc.php:353,
-            // getHealthByGameVersion()) without eager-loading it first - the mapping-import pipeline is designed
-            // to run from the scheduler/CLI in production, where Model::preventLazyLoading()'s violation handler
-            // only logs (see [[project_autoeager_no_lazyviolation]]); it THROWS in dev/test. This is a
-            // pre-existing gap in the import pipeline, unrelated to #3757, not something to paper over with a
-            // production code change here - so switch off exactly the guard that gets in the way, for the Act
-            // step only.
-            //
-            // This used to flip the whole application environment to 'production' instead (the way the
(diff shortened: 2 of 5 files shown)


app/Logging/StructuredLogging.php
```diff
@@ -341,6 +341,21 @@ private static function shouldPrettyPrint(): bool
...
+     * runningUnitTests() answers "is app.env equal to testing", which is not the question being asked: tests are
+     * free to change it, and several do to exercise production-only behaviour. One of those flipping the
+     * environment around a chatty service is what dumped ~1000 ANSI-coloured log lines into the CI test output in
+     * #3782. PHPUNIT_COMPOSER_INSTALL is defined by PHPUnit's own entrypoint (vendor/bin/phpunit, which is how
+     * both CI and `php artisan test` invoke it) and phpunit is a require-dev package, so the constant answers
+     * "is PHPUnit running" without depending on any application state that a test can rewrite.
+     */
+    private static function isRunningTests(Application $app): bool
+    {
+        return $app->runningUnitTests() || defined('PHPUNIT_COMPOSER_INSTALL');
```
Wotuu, line 356: :robot: **Coverage, not a defect — but this change makes the `'stderr'` branch permanently untestable.**

`defined('PHPUNIT_COMPOSER_INSTALL')` is true for the entire lifetime of any PHPUnit process and, unlike `app['env']`, it cannot be undefined. That is precisely the point of the guard — but the consequence is that after this MR the `return 'stderr';` on line 380 is **unreachable from any test, forever**. Before, a test could have driven it by flipping `app['env']` (none did); now nothing can.

So the only behaviour `resolveChannel()` exists to provide — "a locally-run artisan command shows its own output on stderr" — has no test and can no longer acquire one. `StructuredLoggingTest` now pins three negative cases and zero positive ones; a future refactor that breaks console output for artisan commands would go through CI completely silent.

Suggested remedy that doesn't weaken the gate: keep `resolveChannel()` as the un-rewritable entry point, but move the decision into a pure helper it feeds the real values into, e.g.

```php
public static function resolveChannel(): ?string
{
    if (self::$CHANNEL !== null) {
        return self::$CHANNEL;
    }

    $app = app();

    return self::resolveConsoleChannel(
        $app->runningInConsole(),
        self::isRunningTests($app),
        config('app.type') === 'local',
    );
}
```

`resolveConsoleChannel()` is then drivable from a unit test in both directions, while `isRunningTests()` stays exactly as strict as it is here. Non-blocking.

tests/TestCase.php
```diff
@@ -57,6 +57,13 @@ protected function setUp(): void
         // previous test must not leak into this one
         StructuredLogging::flushConfigCache();
 
+        // An explicit channel is the one input resolveChannel() honours before any of its "am I running tests"
+        // checks, so a setChannel() a previous test failed to unwind would send every later test's log lines to
+        // that channel - the #3782 leak again, through the one door the guard there cannot close. Not folded into
+        // flushConfigCache() because that is also called from StructuredLogging::enable(), where clearing a
+        // deliberately-set channel would be wrong.
+        StructuredLogging::setChannel(null);
```
Wotuu, line 67: :robot: The MR body discloses the `StructuredLogging::$ENABLED` latch and declines it, so this is only about the cost of closing it, which is lower than the disclosure implies.

`StructuredLogging::enable()` sets `$ENABLED = true` **and** calls `flushConfigCache()` itself. So replacing the existing `StructuredLogging::flushConfigCache();` on line 58 with `StructuredLogging::enable();` is a strict superset of what `setUp()` already does — zero extra lines, one call removed — and it closes the second latch at the same time as this MR closes the `$CHANNEL` one.

Worth doing here specifically because the latch's symptom is the failure mode this MR's new CI guard cannot see: a `disable()` that never gets unwound silences structured logging for the rest of the shard, and the "Assert no log lines leaked" step then passes for the wrong reason. Non-blocking.

Wotuu:
:robot: Cold review (opus): 2 findings posted. Both non-blocking; no defects found in the fix itself.

Verified independently: `Model::handleLazyLoadingViolationUsing()` in Laravel 12 returns `void` (no `tap`), so the `ReflectionProperty` read in `MDTMappingImportGameVersionScopingTest` is genuinely necessary and its comment is accurate; `PHPUNIT_COMPOSER_INSTALL` is defined by `vendor/phpunit/phpunit/phpunit` and phpunit is `require-dev`, so it cannot exist in production; every test base class in `tests/` descends from `Tests\TestCase`, so the new `setChannel(null)` reset in `setUp()` really does cover the whole suite; `set -o pipefail` is required (the GH Actions default shell is `bash -e {0}`, not `-eo pipefail`) and is correctly present, matching `php-coverage.yml`.

tests/Feature/App/Service/MDT/MDTMappingImportGameVersionScopingTest.php
```diff
@@ -87,11 +90,30 @@ public function importMappingVersionFromMDT_givenGameVersionWithNoExistingMappin
...
+            // exactly why flipping the environment worked, whereas the prevention flag is snapshotted onto each
+            // model instance as it is hydrated (Builder::hydrate()). Instances that laravel-model-caching
+            // unserializes from cache therefore carry whatever the flag was when they were cached and still
+            // throw - which is why the flag variant passed locally (MODEL_CACHE_ENABLED=false) and errored in
+            // CI (MODEL_CACHE_ENABLED=true).
+            //
+            // Reflection rather than re-registering AppServiceProvider to restore it: Application::register()
+            // re-runs boot(), which would duplicate its Event::listen() registrations. A Laravel rename of the
+            // property fails loudly here with a ReflectionException rather than silently skipping the restore.
+            /** @var Closure|null $originalViolationCallback */
```
Wotuu: There's literally twenty lines of comments here. Can you reduce that a bit? Just summarize it. Or get rid of things that are old.

Wotuu (commented):
One thing

app/Logging/StructuredLogging.php
Wotuu: :robot: Extracted the decision into a pure `resolveConsoleChannel(bool $runningInConsole, bool $isRunningTests, bool $isLocalAppType): ?string` static helper. `resolveChannel()` now just gathers the three inputs (via named arguments, so a bool transposition is visible at the call site) and delegates to it. Added `resolveConsoleChannel_GivenInputCombination_ShouldReturnExpectedChannel` with a `#[DataProvider]` covering all 8 input combinations, including the stderr case that was previously unreachable from any test. Kept the three existing `resolveChannel_*` tests as-is since they cover the wiring/environment behavior, not the decision table.

tests/TestCase.php
Wotuu: :robot: Swapped `StructuredLogging::flushConfigCache();` for `StructuredLogging::enable();` in `setUp()`. Confirmed `enable()` sets `$ENABLED = true` and calls `flushConfigCache()` internally, so this is a strict superset at zero extra cost and now also unwinds a stray `disable()` between tests.

tests/Feature/App/Service/MDT/MDTMappingImportGameVersionScopingTest.php
Wotuu: :robot: Condensed the ~24-line comment block down to 9 lines. Kept the essential why (lazy-loading throws in dev/test but only logs in prod; swapping the violation handler rather than flipping the whole env avoids the stderr-logging leak this PR fixes; not `preventLazyLoading(false)` because that flag is snapshotted per-instance at hydration and survives model-cache unserialization; reflection restore avoids duplicating `AppServiceProvider::boot()` listeners), dropped the repeated/redundant explanation of the same points. Pushed as 5bfb74fc4.

Issue #3786, opened 2026-08-01
Drop the vestigial icon_file_id columns from affixes, factions, character_classes and character_class_specializations


Wotuu:
:robot: Follow-up to #3775.

## Background

#3350 (`13650b395`) removed the legacy icon-file upload from `Expansion` **completely**: trait,
`$fillable`/`$hidden`, PHPDoc, controller, form request, seeder — and a migration that dropped
`expansions.icon_file_id` and deleted the orphaned `files` rows.

#3766 and #3775 did the same cleanup for `Affix`, `Faction`, `CharacterClass` and
`CharacterClassSpecialization`, but deliberately stopped short of the column drop. What remains is
purely the schema half.

## Why this had to be a separate release

Before #3775, `resources/views/common/maps/map.blade.php` rendered
`{{ $faction->iconfile->icon_url }}` **unguarded** on every Siege of Boralus / The Nexus map page.
The Expansion-shaped migration starts with

```php
DB::table('files')->where('model_class', Faction::class)->delete();
```

Deploys are not atomic — a cron runs `migrate` independently of the ECS web rollout — so running
that in the same release that removed the code would have 500'd the map page on the still-running
old containers for the length of the deploy window. That is the #3497 failure mode, and the
expand/contract rule in `.claude/CLAUDE.md` exists to prevent it.

**As of #3775 nothing in the codebase reads `iconfile` or `icon_file_id` for any of these four
models**, so this issue is now safe to pick up in any release after #3775 has shipped.

## Work

One migration per table, mirroring
`database/migrations/2026_07_02_000001_drop_icon_file_id_from_expansions_table.php`:

```php
DB::table('files')->where('model_class', <Model>::class)->delete(); // bypasses File's deleting hook on purpose
Schema::table('<table>', static function (Blueprint $table): void {
    $table->dropIndex('<table>_icon_file_id_index');
    $table->dropColumn('icon_file_id');
});
```

| Table | Column type | Index | Orphan `files` rows |
| --- | --- | --- | --- |
| `factions` | `int` NOT NULL, no default, 2nd column | `factions_icon_file_id_index` | 3 |
| `affixes` | `int` NOT NULL, no default, 3rd column | `affixes_icon_file_id_index` | ~40 |
| `character_classes` | **`varchar(255)`** NOT NULL, no default, last column | `character_classes_icon_file_id_index` | 13 |
[...]

PR #3787, opened 2026-08-01
Remove the HasIconFile trait from Faction, CharacterClass and CharacterClassSpecialization

### LINKED ISSUE #3775: Remove HasIconFile trait from Faction, CharacterClass, CharacterClassSpecialization (follow-up to #3756)
:robot: Follow-up to #3756 / PR #3766.

## Background

#3766 added full admin CRUD pages for `Affix`, `Season`, and `AffixGroup`. During review,
[Wotuu pointed out](https://github.com/RaiderIO/keystone.guru/pull/3766#discussion_r3691020380)
that `Affix` still carried the `HasIconFile` trait (`app/Models/Traits/HasIconFile.php`) even
though it's no longer used: any model with an icon/asset now gets it statically from the separate
assets project (see `Affix::getImageUrlAttribute()`, which already reads from
`ksgAssetImage('affixes/....jpg')` rather than the `iconfile` File relation). The admin pages
deliberately don't offer icon upload, so the trait's `iconfile()` relation and
`saveUploadedFile()` method are dead weight - a leftover from when uploading an icon through the
admin panel was actually possible.

#3766 removed `use HasIconFile;` from `Affix` (PR #3766, commit history around
`app/Models/Affix.php`). Wotuu noted the same trait is a leftover on three more models:

- `App\Models\Faction` (`app/Models/Faction.php`)
- `App\Models\CharacterClass` (`app/Models/CharacterClass.php`)
- `App\Models\CharacterClassSpecialization` (`app/Models/CharacterClassSpecialization.php`)

These were intentionally **not** touched in #3766 to keep that PR scoped to the admin-pages work.

## Prior art to follow

`Expansion` went through the exact same cleanup earlier in #3350 (commit `13650b395`,
"Removed legacy expansions icon file upload") - that's the reference shape for a *complete*
removal (trait + fillable/hidden + PHPDoc + controller + form request + seeder + a migration
dropping the DB column and clearing orphaned `files` rows). Use it as the template.

Note `Affix` itself was **not** taken that far in #3766: its `icon_file_id` column on
`affixes` is `NOT NULL` with no default and no unique business key to drop it safely in an
unattended change, so #3766 only removed the trait and left the column/`-1` sentinel in place
(see the updated PHPDoc on `App\Models\Affix::$icon_file_id`). If you pick up the column-drop
for `factions`/`character_classes`/`character_class_specializations`, consider doing the same
for `affixes.icon_file_id` at the same time, following the #3350 migration shape:

```php
DB::table('files')->where('model_class', <Model>::class)->delete(); // bypasses File's deleting hook on purpose
Schema::table('<table>', function (Blueprint $table) {
    $table->dropIndex('<table>_icon_file_id_index');
    $table->dropColumn('icon_file_id');
});
```

## Remaining work

For each of `Faction`, `CharacterClass`, `CharacterClassSpecialization`:

1. Check whether the model's admin controller (if any) still offers an icon upload UI/route tied
   to `HasIconFile` - if so, remove that UI/route too (mirroring #3350's `ExpansionController`
   diff).
2. Remove `use HasIconFile;`, the `iconfile()`/`saveUploadedFile()` reliance, `icon_file_id` from
   `$fillable`/`$hidden`, and the PHPDoc `@property` line.
3. Decide per-model whether the `icon_file_id` column can be dropped in the same change (check
   `SHOW CREATE TABLE <table>` for NOT NULL/default and any admin seeder that still writes to it)
   or whether it should stay vestigial for now like `Affix`.
4. If dropping the column, add a migration mirroring `2026_07_02_000001_drop_icon_file_id_from_expansions_table.php` (or the affix follow-up above).
5. Update any seeders (`FactionsSeeder`, `CharacterClassesSeeder`,
   `CharacterClassSpecializationsSeeder`) that create `File` rows / set `icon_file_id`.
6. Search the codebase for `->iconfile` and `icon_file_id` scoped to each model (views, tests,
   controllers, services) and clean up anything left dangling.
7. Run/update the affected tests, `composer run fix`, `composer run analyse`.

## Definition of done

- All three models no longer `use HasIconFile;`.
- No admin page can upload/change an icon for any of these four models.
- Green CI; `composer run analyse` clean.

Wotuu:
:robot: Closes #3775

Removes `App\Models\Traits\HasIconFile` from `Faction`, `CharacterClass` and
`CharacterClassSpecialization`, finishing what #3350 did for `Expansion` and #3766 for `Affix`.
Icons for all of these come from the assets project via `ksgAssetImage()`, and none of the three has
an admin controller, route, view or form request — so the trait's `iconfile()` relation and
`saveUploadedFile()` were dead weight. The trait itself stays: `User` and `Team` still genuinely
upload avatars and team logos.

## Faction was the only one that still rendered from the `File` relation

`CharacterClass` and `CharacterClassSpecialization` already had `ksgAssetImage()`-backed `icon_url`
accessors and nothing anywhere read their `iconfile` — for them this is a pure deletion.

`Faction` needed a real replacement, so it gains an appended `icon_url` accessor (mirroring
`CharacterClass`) and loses `protected $with = ['iconfile']`:

| File | Change |
| --- | --- |
| `app/Models/Faction.php` | `getIconUrlAttribute()` + `$appends`, `$with` removed |
| `app/Logic/MapContext/MapContextStaticData.php` | dropped `->with('iconfile')` |
| `resources/assets/js/custom/mapcontrols/factiondisplaycontrols.js` | `faction.iconfile.icon_url` → `faction.icon_url` |
| `resources/views/common/maps/map.blade.php` | dead handlebars block deleted (see below) |

The appended accessor rather than a plain `getIconUrl()` method is deliberate: unlike `Expansion`
(Blade-only), this value is consumed by **JS** off the serialized map-context payload.

**The 23-line Blade deletion is not as reckless as it looks.** The
`<script id="map_faction_display_controls_template" type="text/x-handlebars-template">` block was a
pre-precompilation leftover — nothing compiles inline `x-handlebars-template` script tags any more
(it was the last one left in all of `resources/views/`). The live template is
`resources/assets/js/handlebars/map_faction_display_controls_template.handlebars`, generated into
`resources/assets/js/handlebars.js` by `scripts/build/handlebars.mjs`, and it already reads
`{{ icon_url }}` off the JS-built `factionsData`. So only the JS needed rewiring.

Removing `$with` changes the serialized shape of *every* `Faction`, not just the map context. A
repo-wide `iconfile` sweep over `app/ resources/ database/ tests/ routes/` found exactly one
non-`User`/non-`Team` consumer (the JS above). The API v1 contract is untouched — not because
`app/Http/Resources/` contains no "Faction" (it does reach one indirectly: `DungeonRoute::$appends`
includes `setup`, and `getSetupAttribute()` returns `'faction' => $this->faction`), but because
`DungeonRouteSummaryResource::toArray()` is an explicit field allowlist with no `setup` in it.
`setup.faction` *does* change shape on the non-API surfaces (ajax datatables, biglist, map inline);
its only consumer is `resources/views/common/handlebars/groupsetup.blade.php`, which reads
`data.faction.key` / `.name`. Each serialized faction loses a nested `{url, icon_url}` object and
gains one string, so payloads get slightly smaller.

**User-visible URL change**: the faction icon now comes from `{ASSETS_BASE_URL}/images/factions/*.png`
instead of `{APP_URL}/storage/images/factions/*.png`, the same move `Expansion` made in #3350.

## The `icon_file_id` columns are deliberately left in place

They stay vestigial (`-1`) on `factions`, `character_classes`, `character_class_specializations` and
`affixes` — so `icon_file_id` stays in `$fillable`/`$hidden`. Dropping them is **#3786**, and it must
land in a *later* release. (Per review, the `@property` docblocks for these columns no longer carry
an explanatory note pointing at #3786 — the issue itself tracks that, so the four models just declare
the bare `@property` line.)

Before this MR, `map.blade.php:267` rendered `$faction->iconfile->icon_url` unguarded, so the
`DELETE FROM files WHERE model_class = Faction::class` half of the Expansion-shaped migration would
have 500'd every The Nexus / Siege of Boralus map page on the still-running old containers during
the deploy window — the #3497 failure mode `.claude/CLAUDE.md`'s expand/contract rule exists to
prevent. As of this MR nothing reads those rows, so #3786 is trivially safe to ship next release.

The four seeders do stop creating the now-dead `File` rows, but keep `'icon_file_id' => -1` (all
four columns are `NOT NULL` with no default). Two freebies fell out of that:

- `CharacterClassSpecializationsSeeder` read `CharacterClassSpecialization::all()` — the **live**
  table, not the `_temp` one the seeder was writing to. That particular read is gone with the loop
  (the seeder still opens with a live-table `CharacterClass::all()`, which is pre-existing and out
  of scope here).
- `AffixSeeder`'s `File` rows have been garbage since #3766: `$affix->image_url` is now a full URL,
  so it was seeding paths like `images/affixes/http://.../afflicted.jpg.jpg`.

Also swept one dangling leftover flagged by step 6 of the issue: `CharacterRace` listed
`icon_file_id` in `$hidden` even though `character_races` has no such column and it never used the
trait.

## Also removed: `Dungeon::isFactionSelectionRequired()`

The deleted Blade block was its only consumer — after this MR a repo-wide sweep returns nothing but
the declaration itself. It reads as live gating logic, so leaving it behind invites the next person
adding faction UI to wire against a method nothing calls. It never agreed with the real gate anyway:
`dungeonmap.js` and `admindungeonmap.js` both check `DUNGEON_THE_NEXUS` only, not Siege of Boralus.

## Update: the `Unspecified` filter is now actually fixed

`MapContextStaticData`'s `Faction::where('name', '<>', 'Unspecified')` filter matched nothing —
`name` holds a translation key (`factions.unspecified`), not the literal string `'Unspecified'`, so
all three factions shipped in the payload both before and (originally) after this MR. Per review,
this is now `Faction::where('key', '<>', Faction::FACTION_UNSPECIFIED)`, matching the pattern the
deleted Blade block used and `MapContextMappingVersion.php:41` still uses. `FactionDisplayControls`
iterates the payload by raw index and assigns the active radio (`fas`) to index 0; with Unspecified
excluded, index 0 is now Horde, which also happens to line up with the "start as Horde visible"
default a few lines below — previously Unspecified rendered as the (visually) selected radio despite
Horde being what was actually toggled on. `MapContextStaticDataTest` gained an assertion pinning
Unspecified's absence from the payload.

Filed rather than fixed: **#3788** — the faction control's `data-faction` is built from the
*translated, lowercased* faction name but compared against a faction *key*, so it only works in
`en_US`. Pre-existing on the live JS path; the now-deleted Blade block happened to hold the correct
`strtolower($faction->key)`, which is how it surfaced.

## Verification

- New tests: `FactionTest` (accessor per faction key + `toArray()` shape), `CharacterClassTest`
  (both class models), and `MapContextStaticDataTest`. The last is the one that actually pins what
  changed: it asserts the **serialized** map-context factions payload carries `icon_url` and no
  `iconfile`, which is precisely the contract `factiondisplaycontrols.js` depends on. (The
  `assertArrayNotHasKey('icon_file_id', ...)` lines are belt-and-braces — `$hidden` already
  guaranteed those on `master`.)
- All four seeders re-run clean against the `NOT NULL` columns (`db:seedone`).
- `composer run fix` (0 files changed) and `composer run analyse` (no errors).
- Full suite green apart from three pre-existing failures that reproduce identically on `master` in
  this worktree, all MDT/Lua-package environment issues:
  `MDTImportStringServiceMDT2AuthoritativeTest`, `MDTNpcMappingCoverageTest`, and
  `ConversionTest` (`vendor/nnoggie/mythicdungeontools/Midnight/KingsRest.lua` missing).

### Visual verification

Admin floor mapping page for The Nexus, which is where `FactionDisplayControls` actually renders
(`dungeonmap.js` only adds it in edit mode for `DUNGEON_THE_NEXUS`). Before, the icons resolve to
`/storage/images/factions/*.png`, which 403s on both the main dev stack and the worktree — those
files only exist in the assets project now. After, they load from the assets host.

### Cold review

Reviewed by a fresh-context agent with no shared context. It independently verified the dead-Blade
claim against `scripts/build/handlebars.mjs`, swept the repo for missed `iconfile` consumers (none
outside `User`/`Team`), confirmed every seeder's remaining `-1` is still reachable given each one's
insert mechanism, and checked the PHPDoc column types against both schema dumps. Its one medium
finding (`isFactionSelectionRequired()` left dangling) and the PHPDoc-placement nit are fixed in
`4c57111cb`; the wording issues it raised are corrected in this body; the locale bug became #3788.

### DIFF
--- a/app/Logic/MapContext/MapContextStaticData.php
+++ b/app/Logic/MapContext/MapContextStaticData.php
@@ -63,7 +63,7 @@ class MapContextStaticData implements Arrayable
                 'characterClasses'                  => $characterClasses,
                 'characterClassSpecializations'     => CharacterClassSpecialization::all(),
                 'raidMarkers'                       => RaidMarker::all(),
-                'factions'                          => Faction::where('name', '<>', 'Unspecified')->get(),
+                'factions'                          => Faction::where('key', '<>', Faction::FACTION_UNSPECIFIED)->get(),
                 'publishStates'                     => PublishedState::all(),
                 'gameVersions'                      => GameVersion::all(),
                 'selectableSpells'                  => $selectableSpells,
--- a/app/Logic/MapContext/MapContextStaticData.php
+++ b/app/Logic/MapContext/MapContextStaticData.php
@@ -63,7 +63,7 @@ class MapContextStaticData implements Arrayable
                 'characterClasses'                  => $characterClasses,
                 'characterClassSpecializations'     => CharacterClassSpecialization::all(),
                 'raidMarkers'                       => RaidMarker::all(),
-                'factions'                          => Faction::where('name', '<>', 'Unspecified')->with('iconfile')->get(),
+                'factions'                          => Faction::where('name', '<>', 'Unspecified')->get(),
                 'publishStates'                     => PublishedState::all(),
                 'gameVersions'                      => GameVersion::all(),
                 'selectableSpells'                  => $selectableSpells,
--- a/app/Models/Affix.php
+++ b/app/Models/Affix.php
@@ -11,10 +11,7 @@ use Str;
 
 /**
  * @property int    $id           The ID of this Affix.
- * @property int    $icon_file_id Vestigial - always -1. The icon itself is a static asset from the
- *                                assets project (see image_url), not an admin-editable File upload.
- *                                Column kept for now (int NOT NULL, no default); dropping it needs
- *                                its own migration - see #3786.
+ * @property int    $icon_file_id
  * @property int    $affix_id     The ID of the affix in-game.
  * @property string $key          The identifying key of the Affix.
  * @property string $name         The name of the Affix.
--- a/app/Models/Affix.php
+++ b/app/Models/Affix.php
@@ -13,8 +13,8 @@ use Str;
  * @property int    $id           The ID of this Affix.
  * @property int    $icon_file_id Vestigial - always -1. The icon itself is a static asset from the
  *                                assets project (see image_url), not an admin-editable File upload.
- *                                Column kept for now (NOT NULL, no default); dropping it needs its
- *                                own migration - see #3775.
+ *                                Column kept for now (int NOT NULL, no default); dropping it needs
+ *                                its own migration - see #3786.
  * @property int    $affix_id     The ID of the affix in-game.
  * @property string $key          The identifying key of the Affix.
  * @property string $name         The name of the Affix.
--- a/app/Models/CharacterClass.php
+++ b/app/Models/CharacterClass.php
@@ -17,10 +17,7 @@ use Str;
  * @property string $key
  * @property string $name
  * @property string $color
- * @property string $icon_file_id Vestigial - always -1. The icon itself is a static asset from the
- *                                assets project (see icon_url), not an admin-editable File upload.
- *                                Column kept for now (varchar(255) NOT NULL, no default); dropping
- *                                it needs its own migration - see #3786.
+ * @property string $icon_file_id
  *
  * @property string $icon_url Appended
  *
--- a/app/Models/CharacterClass.php
+++ b/app/Models/CharacterClass.php
@@ -14,13 +14,13 @@ use Str;
 /**
  * @property int    $id
  * @property int    $class_id     Blizzard class ID
+ * @property string $key
+ * @property string $name
+ * @property string $color
  * @property string $icon_file_id Vestigial - always -1. The icon itself is a static asset from the
  *                                assets project (see icon_url), not an admin-editable File upload.
  *                                Column kept for now (varchar(255) NOT NULL, no default); dropping
  *                                it needs its own migration - see #3786.
- * @property string $key
- * @property string $name
- * @property string $color
  *
  * @property string $icon_url Appended
  *
--- a/app/Models/CharacterClass.php
+++ b/app/Models/CharacterClass.php
@@ -4,7 +4,6 @@ namespace App\Models;
 
 use App\Models\DungeonRoute\DungeonRoutePlayerClass;
 use App\Models\DungeonRoute\DungeonRoutePlayerRace;
-use App\Models\Traits\HasIconFile;
 use App\Models\Traits\SeederModel;
 use Eloquent;
 use Illuminate\Database\Eloquent\Collection as EloquentCollection;
@@ -14,7 +13,11 @@ use Str;
 
 /**
  * @property int    $id
- * @property int    $class_id Blizzard class ID
+ * @property int    $class_id     Blizzard class ID
+ * @property string $icon_file_id Vestigial - always -1. The icon itself is a static asset from the
+ *                                assets project (see icon_url), not an admin-editable File upload.
+ *                                Column kept for now (varchar(255) NOT NULL, no default); dropping
+ *                                it needs its own migration - see #3786.
  * @property string $key
  * @property string $name
  * @property string $color
@@ -30,7 +33,6 @@ use Str;
  */
 class CharacterClass extends CacheModel
 {
-    use HasIconFile;
     use SeederModel;
 
     public $timestamps = false;
--- a/resources/views/common/maps/map.blade.php
+++ b/resources/views/common/maps/map.blade.php
@@ -254,29 +254,6 @@ if ($isAdmin) {
         });
     </script>
 
-    @if($dungeon->isFactionSelectionRequired())
-        <script id="map_faction_display_controls_template" type="text/x-handlebars-template">
-            <div id="map_faction_display_controls" class="leaflet-draw-section">
-                <div class="leaflet-draw-toolbar leaflet-bar leaflet-draw-toolbar-top">
-            @foreach(Faction::where('key', '<>', Faction::FACTION_UNSPECIFIED)->get() as $faction)
-                <a class="map_faction_display_control map_controls_custom" href="#"
-                   data-faction="{{ strtolower($faction->key) }}"
-                           title="{{ __($faction->name) }}">
-                            <i class="{{ $loop->index === 0 ? 'fas' : 'far' }} fa-circle radiobutton"
-                               style="width: 15px"></i>
-                            <img src="{{ $faction->iconfile->icon_url }}" class="select_icon faction_icon"
-                                 data-bs-toggle="tooltip" title="{{ __($faction->name) }}"
-                                 alt="Faction"/>
-                        </a>
-
-            @endforeach
-            </div>
-            <ul class="leaflet-draw-actions"></ul>
-        </div>
-
-
-        </script>
-    @endif
 @endsection
 
 @if(!$noUI)
--- a/app/Models/CharacterClassSpecialization.php
+++ b/app/Models/CharacterClassSpecialization.php
@@ -4,7 +4,6 @@ namespace App\Models;
 
 use App\Models\Interfaces\CombatLogCriterionModelInterface;
 use App\Models\Traits\HasCombatLogCriterion;
-use App\Models\Traits\HasIconFile;
 use App\Models\Traits\SeederModel;
 use Eloquent;
 use Illuminate\Database\Eloquent\Relations\BelongsTo;
@@ -14,7 +13,10 @@ use Str;
  * @property int    $id
  * @property int    $character_class_id Internal ID - not a blizzard ID!
  * @property int    $specialization_id  Blizzard ID
- * @property int    $icon_file_id
+ * @property string $icon_file_id       Vestigial - always -1. The icon itself is a static asset from
+ *                                      the assets project (see icon_url), not an admin-editable File
+ *                                      upload. Column kept for now (varchar(255) NOT NULL, no
+ *                                      default); dropping it needs its own migration - see #3786.
  * @property string $key
  * @property string $name
  *
@@ -27,7 +29,6 @@ use Str;
 class CharacterClassSpecialization extends CacheModel implements CombatLogCriterionModelInterface
 {
     use HasCombatLogCriterion;
-    use HasIconFile;
     use SeederModel;
 
     public $timestamps = false;
--- a/app/Models/CharacterClassSpecialization.php
+++ b/app/Models/CharacterClassSpecialization.php
@@ -15,10 +15,7 @@ use Str;
  * @property int    $specialization_id  Blizzard ID
  * @property string $key
  * @property string $name
- * @property string $icon_file_id       Vestigial - always -1. The icon itself is a static asset from
- *                                      the assets project (see icon_url), not an admin-editable File
- *                                      upload. Column kept for now (varchar(255) NOT NULL, no
- *                                      default); dropping it needs its own migration - see #3786.
+ * @property string $icon_file_id
  *
  * @property string $icon_url Appended
  *
--- a/app/Models/CharacterClassSpecialization.php
+++ b/app/Models/CharacterClassSpecialization.php
@@ -13,12 +13,12 @@ use Str;
  * @property int    $id
  * @property int    $character_class_id Internal ID - not a blizzard ID!
  * @property int    $specialization_id  Blizzard ID
+ * @property string $key
+ * @property string $name
  * @property string $icon_file_id       Vestigial - always -1. The icon itself is a static asset from
  *                                      the assets project (see icon_url), not an admin-editable File
  *                                      upload. Column kept for now (varchar(255) NOT NULL, no
  *                                      default); dropping it needs its own migration - see #3786.
- * @property string $key
- * @property string $name
  *
  * @property string $icon_url Appended
  *
--- a/app/Models/CharacterRace.php
+++ b/app/Models/CharacterRace.php
@@ -96,7 +96,6 @@ class CharacterRace extends CacheModel
     public $timestamps = false;
 
     public $hidden = [
-        'icon_file_id',
         'pivot',
     ];
 
--- a/app/Models/Dungeon.php
+++ b/app/Models/Dungeon.php
@@ -554,14 +554,6 @@ class Dungeon extends CacheModel implements CombatLogCriterionModelInterface, Ma
         return $this->loadMappingVersions()->mappingVersions->map(static fn(MappingVersion $mappingVersion) => $mappingVersion->gameVersion)->unique('id');
     }
 
-    public function isFactionSelectionRequired(): bool
-    {
-        return in_array($this->key, [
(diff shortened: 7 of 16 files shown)


Wotuu:
:robot: Visual verification — admin floor mapping page for The Nexus (`/admin/dungeon/the-nexus/floor/125/mapping`), the faction display control.

| Before (`master`) | After |
| --- | --- |
| before | after |

Before, all three icons point at `{APP_URL}/storage/images/factions/*.png` and fail to load (`naturalWidth: 0`) — those files only live in the assets project now, and the path 403s on the main dev stack too, not just the worktree. After, all three load from `{ASSETS_BASE_URL}/images/factions/*.png`.

app/Logic/MapContext/MapContextStaticData.php
```diff
@@ -63,7 +63,7 @@ public function toArray(): array
                 'characterClasses'                  => $characterClasses,
                 'characterClassSpecializations'     => CharacterClassSpecialization::all(),
                 'raidMarkers'                       => RaidMarker::all(),
-                'factions'                          => Faction::where('name', '<>', 'Unspecified')->with('iconfile')->get(),
+                'factions'                          => Faction::where('name', '<>', 'Unspecified')->get(),
```
Wotuu: Can Unspecified be changed to a static reference

app/Models/Affix.php
```diff
@@ -13,8 +13,8 @@
  * @property int    $id           The ID of this Affix.
  * @property int    $icon_file_id Vestigial - always -1. The icon itself is a static asset from the
  *                                assets project (see image_url), not an admin-editable File upload.
- *                                Column kept for now (NOT NULL, no default); dropping it needs its
- *                                own migration - see #3775.
+ *                                Column kept for now (int NOT NULL, no default); dropping it needs
```
Wotuu: I say get rid of this comment and let the issue itself handle it

app/Models/CharacterClass.php
```diff
@@ -14,10 +13,14 @@
...
 /**
  * @property int    $id
- * @property int    $class_id Blizzard class ID
+ * @property int    $class_id     Blizzard class ID
  * @property string $key
  * @property string $name
  * @property string $color
+ * @property string $icon_file_id Vestigial - always -1. The icon itself is a static asset from the
+ *                                assets project (see icon_url), not an admin-editable File upload.
+ *                                Column kept for now (varchar(255) NOT NULL, no default); dropping
```
Wotuu: Same here really

resources/views/common/maps/map.blade.php
```diff
@@ -254,29 +254,6 @@
...
-                            <img src="{{ $faction->iconfile->icon_url }}" class="select_icon faction_icon"
-                                 data-bs-toggle="tooltip" title="{{ __($faction->name) }}"
-                                 alt="Faction"/>
-                        </a>
-
-            @endforeach
-            </div>
-            <ul class="leaflet-draw-actions"></ul>
-        </div>
-
```
Wotuu, line 276: So why is this entire blade file removed?

Wotuu (commented):
.

Wotuu:

Independently verified (fresh context, no shared session with the implementer):

- **Dead Blade block** — `Handlebars.compile` / `x-handlebars-template` appear nowhere in `resources/assets/js/` or `scripts/`, and `resources/assets/js/handlebars/map_faction_display_controls_template.handlebars` exists and already reads `{{ icon_url }}`. The deleted inline `<script type="text/x-handlebars-template">` was genuinely unreachable.
- **`iconfile` sweep** — repo-wide grep over `app/ resources/ database/ tests/ routes/ config/` leaves only `User`/`Team` consumers plus the guarded `isset($menuModel->iconfile)` in `sitepage.blade.php`, whose only caller is `team/edit.blade.php` (`$user->teams`).
- **`isFactionSelectionRequired()`** — sole consumer was the deleted Blade block; removal leaves nothing dangling.
- **PHPDoc column types** — checked against `database/schema/migrate-schema.dump`: `character_classes` / `character_class_specializations` are `varchar(255) NOT NULL`, `factions` / `affixes` are `int(11) NOT NULL`. `character_races` has no `icon_file_id` column, so the `$hidden` removal is correct.
- **Seeders** — all four still supply `icon_file_id => -1`, including the easy-to-miss `->map()` closure in `CharacterClassSpecializationsSeeder`, so the `NOT NULL`-no-default columns are still satisfied.
- **API contract** — `DungeonRouteSummaryResource::toArray()` carries no `setup`, so the v1 payload is unchanged. `setup.faction`'s only consumer, `common/handlebars/groupsetup.blade.php`, reads `data.faction.key` / `.name` only.
- **Asset path shape** — `ksgAssetImage()` resolves against `images_base_url` (`ASSETS_BASE_URL . "/images"`), so `factions/<key>.png` reproduces the old `images/factions/<key>.png` layout, and the faction keys need no `strtolower`/space stripping.
- **Deploy-window staleness** — checked whether the redis-cached `static_data` payload (TTL 1 day) could serve an old-shape `factions` array to the new JS after a deploy. It cannot: `map.blade.php` only hits `js.mapcontext.static` when `config("app.type") === "local"`; every other environment loads the release-versioned build-time artifact generated by `make:mapcontextstatic` (`release-deploy.yml:101`), so payload and JS ship in lockstep.
- **Test isolation** — `phpunit.xml` sets `CACHE_STORE=array` and `CACHE_FILE_PATH=/tmp/phpunit_cache`, and CI runs its phpunit matrix as separate jobs rather than in-process parallel, so the `Cache::store("tmp_file")->flush()` in `MapContextStaticDataTest::setUp()` is both necessary and safe.

Everything else I considered was either cosmetic or pre-existing and already acknowledged in the PR body (the no-op `where("name", "<>", "Unspecified")` filter, the locale-dependent `data-faction` comparison filed as #3788, `CharacterClass::all()` reading the live table in the spec seeder). Nothing worth blocking on.

app/Logic/MapContext/MapContextStaticData.php
Wotuu: :robot: Good catch — fixed properly rather than just swapping in a constant. Changed to `Faction::where('key', '<>', Faction::FACTION_UNSPECIFIED)->get()`, which also fixes the underlying bug: the old `name` comparison against the literal `'Unspecified'` never matched anything, since `name` holds a translation key (`factions.unspecified`), not that string — so Unspecified was shipping in the payload all along. `key` is the column that actually holds `'unspecified'`, matching what the now-deleted Blade block and `MapContextMappingVersion.php:41` already do.

Verified this is safe: `FactionDisplayControls` (the only consumer of `getStaticFactions()`) just iterates the array and assigns the active radio class by index — excluding Unspecified shifts index 0 to Horde, which now actually matches the "start as Horde visible" default a few lines below (previously Unspecified rendered as the visually-selected radio despite Horde being what was toggled on). The separate faction list used by the group-composition form (`composition.blade.php`, which explicitly needs Unspecified for `unspecifiedFactionId`) comes from a completely different source (`$allFactions`), so it is unaffected.

Added `MapContextStaticDataTest::toArray_givenFactions_excludesUnspecifiedFaction` to pin this. Updated the PR body/description accordingly.

app/Models/Affix.php
Wotuu: :robot: Removed the "Column kept for now (int NOT NULL, no default); dropping it needs its own migration - see #3786" sentence — `@property int $icon_file_id` is now bare, same shape it had on `master` before this PR touched it (it previously referenced #3775, which this PR closes).

Since `CharacterClass.php`, `CharacterClassSpecialization.php` and `Faction.php` carried the exact same sentence (the first two are new additions from this PR, `Faction.php` too), I removed it from all three as well for consistency rather than leaving two fixed and two not — see the reply on `CharacterClass.php:22`. While there, `CharacterClass`/`CharacterClassSpecialization`'s `icon_file_id` docblock type is now `string` (was wrongly `int`) — `SHOW COLUMNS` confirms those two tables are `varchar(255)`, unlike `factions`/`affixes` which are genuinely `int`.

app/Models/CharacterClass.php
Wotuu: :robot: Removed here too, same as `Affix.php:16`. `@property int $icon_file_id` is now bare. Also applied the same removal to `CharacterClassSpecialization.php` and `Faction.php`, which carried the identical sentence, so all four models end up consistent rather than two fixed and two left behind.

resources/views/common/maps/map.blade.php
Wotuu: :robot: Re-verified this independently rather than just trusting the PR description, and the removal is correct — the deleted block was dead.

What it was: a `<script id="map_faction_display_controls_template" type="text/x-handlebars-template">` tag, rendered by Blade into the page's DOM. That pattern is how templates worked *before* the project moved to precompiled Handlebars — a JS library would read the tag's `innerHTML` from the DOM at runtime and compile it on the fly.

What actually happens today: `factiondisplaycontrols.js` gets its template via `Handlebars.templates['map_faction_display_controls_template']` (a lookup into the precompiled template registry), not from the DOM. That registry — `resources/assets/js/handlebars.js` — is generated at build time by `scripts/build/handlebars.mjs`, which runs the `handlebars` CLI over every `.handlebars` file in `resources/assets/js/handlebars/`. The live source for this exact template is `resources/assets/js/handlebars/map_faction_display_controls_template.handlebars`, which already exists as a standalone file and is functionally identical to what the deleted Blade block rendered (same markup, same `{{ icon_url }}`/`{{ name_lc }}`/`{{ fa_class }}` variables that `factiondisplaycontrols.js` builds into `factionsData`).

I confirmed:
- `grep -rn "text/x-handlebars-template" resources/views/` returns nothing anywhere in the repo (post-deletion) — it was the last inline template tag left.
- `resources/assets/js/handlebars/map_faction_display_controls_template.handlebars` exists independently and its variable names line up exactly with what the JS supplies.
- `grep -rn "map_faction_display_controls_template"` across `resources/assets/js/` and `resources/views/` finds only the `Handlebars.templates[...]` lookup — nothing reads the DOM tag.
- `grep -rn "isFactionSelectionRequired"` across the worktree now returns nothing — the deleted block was its only caller, so removing the method alongside it (already done in an earlier commit) doesn't leave anything dangling.

So the Blade block was inert: it rendered a `<script>` tag into every map page's DOM that nothing ever read. The one thing it got right that the live JS path doesn't is `data-faction` — the old block used `strtolower($faction->key)` (locale-independent), while the live JS builds it from the translated, lowercased faction name (English-only). That's a real but pre-existing bug on the code path that was already live before this PR; it's filed as #3788 rather than folded into this cleanup.

Wotuu (commented):
Ok

Issue #3791, opened 2026-08-02
Combat log parse aborted by Valkey read error mid-run (partially ingested run)


Wotuu:
:robot: Filed automatically by `/combatlog-parse-failure-poll` (Staging sweep, 2026-08-02).

**Unresolved failures in this cluster: 1** (count = unresolved `CombatLogParseFailure` rows visible at poll time — a later *decrease* means rows were resolved in the admin panel, not a regression).

**This is not a parser defect.** The failure is a Valkey/ElastiCache connection drop that happened to occur mid-parse; the run was partially ingested and then aborted. The `raw_line` below is simply wherever the reader had got to when the connection died — it is *not* the failing input and should not be treated as one.

### Row (verbatim)

```
id=746  run_id=41910839  season_id=17  combat_log_version=22012000007  line_number=15667
exception: RedisException
message:   read error on connection to tcp://ksg-valkey.cniaai.0001.use1.cache.amazonaws.com:6379
raw_line:  7/31/2026 20:58:05.5509  SPELL_CAST_SUCCESS,Creature-0-3898-2874-13267-61245-00006C8DCB,"축전 토템",0x2112,0x80000000,0000000000000000,nil,0x80000000,0x80000000,118905,"축전 토템",0x8,Creature-0-3898-2874-13267-61245-00006C8DCB,Player-205-0A43C537,15044,15044,0,2860,1470,0,0,0,1,0,0,0,-634.24,346.25,2501,0.0000,292
```

### Useful question for a follow-up session

Not "why doesn't the parser handle this line", but: was run `41910839` retried and completed, or is it left half-ingested? If a Valkey read error aborts a parse with no retry, every such blip silently drops a run's data.

PR #3793, opened 2026-08-02
#3790 Fold surplus EMOTE parameters back into the emote text

### LINKED ISSUE #3790: Combat log parse failures: EMOTE with commas in its text breaks SpecialEvents\Emote parameter count

**Unresolved failures in this cluster: 1** (count = unresolved `CombatLogParseFailure` rows visible at poll time — a later *decrease* means rows were resolved in the admin panel, not a regression).

### Row (verbatim)

```
id=743  run_id=41076792  season_id=17  combat_log_version=22012000007  line_number=10802
message:   Invalid parameter count for App\Logic\CombatLog\SpecialEvents\Emote - wanted 6, got 7
raw_line:  7/23/2026 21:16:11.380-5  EMOTE,Player-1427-0E8EA11C,"Baifrosth",Player-1427-0E8EA11C,"Baifrosth",Esto es patético, antiguo maestro, ni siquiera ese sucio demonio eredar recibe tantos golpes.
```

### Hypothesis

Inferable from the raw line: the emote text is emitted **unquoted** and contains commas, so the comma split yields 7 fields where `SpecialEvents\Emote` expects 6. Note that 7 is a lower bound — this particular emote has two commas; one with three would give 8, so a fix needs to join the trailing fields back into the emote text rather than special-case a count of 7.

See the `combatlog-parsing-internals` skill for how `SpecialEvent` subclasses consume their parameters.

Wotuu:
:robot: Closes #3790

## Problem

`EMOTE` emits its emote text **unquoted**, so every comma inside the text is
indistinguishable from a field separator to the generic comma split. The count
was previously widened to `6` with `1` optional, which absorbs exactly one comma
and no more — the Staging failure has two:

```
Invalid parameter count for App\Logic\CombatLog\SpecialEvents\Emote - wanted 6, got 7
7/23/2026 21:16:11.380-5  EMOTE,Player-1427-0E8EA11C,"Baifrosth",Player-1427-0E8EA11C,"Baifrosth",Esto es patético, antiguo maestro, ni siquiera ese sucio demonio eredar recibe tantos golpes.
```

There is no upper bound on the amount of commas an emote may contain, so raising
the limit again would only move the failure rather than remove it.

## Fix

`Emote` now folds every parameter past the fourth field back into a single one
before the count is validated. Re-joining those parameters loses the whitespace
that directly followed each comma, so the text is preferably recovered from the
raw line instead — but that recovery has to find the field boundaries with a
regex, which (unlike `CombatLogStringParser`) cannot account for escaped quotes.
It is therefore trusted **only when both agree on everything but whitespace**;
otherwise the parameters win. Losing a space beats silently returning a
differently-delimited string, and it keeps the parameter path load-bearing
rather than unreachable.

The accepted parameter count goes `6` (1 optional) → `5` (0 optional). The
**minimum is unchanged at 5**, so no line that parsed before stops parsing.

Source/dest GUID and name are exposed as getters alongside `getEmoteText()` —
nothing consumed `Emote` before, and without them the fix is only assertable as
"did not throw". `CombatLogStringParser` is deliberately untouched: an unquoted
trailing-text mode there would change behaviour for every event type, and this
is EMOTE-specific.

## Verification against real Staging data

Segments pulled for run `41076792` (the run behind failure row `id=743`) per the
`combatlog-parse-failure-triage` skill and re-parsed with
`parseCombatLogToEvents()`:

| | before | after |
|---|---|---|
| segments parsed OK | 8 / 9 | **9 / 9** |
| events | 240,084 | **309,968** |
| failures | 1 (`01_trash.txt` line 10802) | **0** |

All 14 EMOTE lines in the run now parse, and each reconstructed
`getEmoteText()` matches its raw line verbatim — including the three
`|TInterface\...|Hspell:...|h[...]|h|r` creature emotes with a `nil` dest name.

The cluster is a single row, so **the skill's "reproduce against two distinct
failing runs" step could not be met** — no second failing run exists. As a
broader regression check the segments for run `41910839` (from #3791, a
different cluster) were parsed too: 446,111 events, 0 failures. That run
contains no EMOTE lines, so it exercises the change only as a no-regression
check, not as a second reproduction.

## Tests

New `EmoteTest` data provider covers the discriminating cases: the #3790 line
(2 commas), a 3-comma variant, a comma-free emote, the untargeted `nil`
dest-name form, the quoted source name containing a comma
(`"Chargath, Bane of Scales"`), a fully quoted emote text, and an escaped quote
in a name that forces the parameter fallback — asserting the reconstructed text,
not merely that parsing succeeded. A separate test pins the below-minimum
boundary still throwing.

- `EmoteTest` — 8 passed (50 assertions)
- `--group=CombatLog` — 221 passed (1077 assertions)
- `composer run fix` / `composer run analyse` — clean

Note this MR does **not** mark failure row `id=743` resolved; per the triage
skill that needs an explicit go-ahead and should follow the fix actually
shipping.

### DIFF
--- a/app/Logic/CombatLog/SpecialEvents/Emote.php
+++ b/app/Logic/CombatLog/SpecialEvents/Emote.php
@@ -23,9 +23,10 @@ class Emote extends SpecialEvent
 
     /**
      * Skips the four leading fields - each either a quoted string, which may itself contain commas, or an
-     * unquoted one - to recover the emote text from the raw line exactly as it was logged.
+     * unquoted one - to recover the emote text from the raw line exactly as it was logged. Deliberately
+     * without the /s modifier, to match CombatLogEntry's own line regex in never running past a newline.
      */
-    private const string EMOTE_TEXT_REGEX = '/(?:^|\s)EMOTE,(?:(?:"[^"]*"|[^,]*),){4}(.*)$/s';
+    private const string EMOTE_TEXT_REGEX = '/(?:^|\s)EMOTE,(?:(?:"[^"]*"|[^,]*),){4}(.*)$/';
 
     private string $sourceGuid;
 
@@ -79,7 +80,7 @@ class Emote extends SpecialEvent
         $this->sourceName = (string)$foldedParameters[1];
         $this->destGuid   = (string)$foldedParameters[2];
         $this->destName   = (string)$foldedParameters[3];
-        $this->emoteText  = $this->parseEmoteTextFromRawEvent() ?? (string)$foldedParameters[4];
+        $this->emoteText  = $this->resolveEmoteText((string)$foldedParameters[4]);
 
         return $this;
     }
@@ -109,18 +110,26 @@ class Emote extends SpecialEvent
     }
 
     /**
-     * @return string|null Null if the raw event did not have the expected shape, in which case the caller must
-     *                     fall back to the re-joined parameters, which lose any whitespace directly following
-     *                     a comma in the emote text.
+     * Re-joining the folded parameters loses any whitespace that directly followed a comma in the emote text,
+     * so the text is preferably recovered from the raw line instead. That recovery has to find the field
+     * boundaries with a regex, which - unlike CombatLogStringParser - cannot account for escaped quotes, so
+     * it is only trusted when both agree on everything but whitespace. Otherwise the parameters win: losing
+     * a space beats silently returning a differently-delimited string.
+     *
+     * @param string $foldedEmoteText The emote text as recovered from the folded parameters.
      */
-    private function parseEmoteTextFromRawEvent(): ?string
+    private function resolveEmoteText(string $foldedEmoteText): string
     {
         $matches = [];
 
         if (preg_match(self::EMOTE_TEXT_REGEX, $this->getRawEvent(), $matches) !== 1) {
-            return null;
+            return $foldedEmoteText;
         }
 
-        return trim($matches[1]);
+        $rawEmoteText = trim($matches[1]);
+
+        return str_replace(' ', '', $rawEmoteText) === str_replace(' ', '', $foldedEmoteText)
+            ? $rawEmoteText
+            : $foldedEmoteText;
     }
 }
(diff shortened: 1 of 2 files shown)


Wotuu (commented):
Ok

Wotuu:
:robot: Cold review: 5 findings

Handed the diff to a fresh-context agent with no shared context with the implementation. It found
**nothing blocking** and verified empirically that the parameter-count arithmetic is strictly more
permissive than what it replaces (old accepted `{5,6}`; new accepts `{5,6,7,…}` and still rejects
`<5`), that the regex is not ReDoS-prone (bounded at 4 reps, `[^,]*` cannot cross a comma — a
40-comma emote parses in 0.1 ms), and that a literal `EMOTE,` inside the emote text cannot
mis-anchor it.

All five findings are addressed in ce13dd4ed:

| # | Severity | Finding | Resolution |
|---|---|---|---|
| 1 | should-fix | The raw-line regex re-implements field splitting `CombatLogStringParser` already does correctly, and where they disagree (escaped quote in a name) the regex silently won and returned a differently-delimited string. The `?? $folded` fallback was therefore unreachable and untested. | The raw capture is now trusted **only** when it agrees with the folded parameters apart from whitespace; otherwise the parameters win. This also unquotes a fully quoted emote text like any other field, and makes the fallback load-bearing. |
| 2 | nit | `/s` let the pattern run past a newline, unlike `CombatLogEntry`'s own line regex. | Modifier dropped — behaviour-identical on single-line input, removes the asymmetry by construction. |
| 3 | nit | `#[Group]` on the method, not the class, contrary to `.claude/CLAUDE.md`. | Moved to class level. (Noting the sibling event tests all use method level, so the convention text and the codebase disagree here.) |
| 4 | nit | Type asserted on `getParsedEvent()` while getters were called on the `@var`-typed `$result`. Missing a below-minimum boundary test and a fallback-path test. | Assertion moved onto `$result` (plus an identity assert), and both tests added. |
| 5 | nit | `trim()` on the recovered text mildly contradicts the whitespace-preservation rationale. | Kept deliberately — it matches `CombatLogStringParser`'s own per-field trim. |

Two out-of-scope observations from the review, recorded rather than pulled into this MR:

- An **unpaired `"` inside emote text** (e.g. `it's 6" long, mate`) still fails one layer earlier, in
  `CombatLogStringParser`, with `Unbalanced quotes in string …`. Same root cause — unquoted free text
  — but a different failure, pre-existing and not a regression. It has not been observed in Staging;
  worth its own issue if it shows up in a future poll.
- The regex deliberately stays in **byte mode**: adding `/u` for the Spanish text would make
  `preg_match` return `false` on invalid UTF-8, silently routing every encoding-noisy line to the
  fallback.

Re-verified after the changes: both real runs reparsed (756,079 events, 14 emotes, 0 failures),
`--group=CombatLog` 221 passed (1077 assertions), `composer run fix` / `analyse` clean.

PR #3794, opened 2026-08-02
Wire structured error logs into Sentry

### LINKED ISSUE #3792: Wire structured error logs into Sentry
## Background

Production/staging errors are currently only visible as Discord messages: a flat, unqueryable feed
with no occurrence counts, grouping, or history. `sentry/sentry-laravel` has been a dependency since
#3017 and `bootstrap/app.php` wires `Integration::handles($exceptions)`, but that only captures
**uncaught exceptions**.

The most actionable errors are not uncaught exceptions. They are deliberate `Log::error()` calls from
`App\Logging\StructuredLogging` subclasses, e.g.:

```
[2026-08-02 17:36:15] production.ERROR
ProcessCombatLogSegmentsLogging::handleSegmentsNotAvailable
Context {"runId": 42015954, "combatLogVersion": 22012000007, "depth": 1}
Extra   {"trace_id": "f3776964-...", "structured:processcombatlogsegmentslogging::handle": {...}}
```

Those reach Discord (and Rollbar, for classes using `InteractsWithRollbar`) but never Sentry.

This issue is the app-side half of making them visible. It is inert until `sentry` is present in the
production log stack, so it is safe to merge on its own.

### 1. `DeduplicateHandlers` store argument

`app/Logging/Handlers/DeduplicateHandlers.php` hardcodes `storage_path('logs/discord-deduplication.log')`.
Applied to a second channel as-is, Discord and Sentry share one ledger keyed by level+message hash —
an error already sent to Discord within the 60s window would be **silently suppressed from Sentry**,
and vice versa.

`LogManager::parseTap()` splits a tap string on the first `:` and spreads the rest as arguments, so:

```php
public function __invoke(Logger $logger, string $store = 'discord'): void
```

with `storage_path(sprintf('logs/%s-deduplication.log', $store))`. The default keeps discord's
existing `'tap' => [DeduplicateHandlers::class]` working unchanged.

Note the store is a file under `storage_path()`, so on ECS it is per-task, not global.

### 2. `SkipsExceptionMirrors` tap

`app/Exceptions/Handler.php::report()` calls `HandlerLoggingInterface::uncaughtException()` at error
level **in addition to** Sentry's own capture. That method logs `get_defined_vars()` under a
**constant** message, and Sentry groups message events by their message — so without a filter,
*every uncaught exception in the app collapses into a single Sentry issue* titled
`HandlerLogging::uncaughtException`. That alone would make the feed useless.

Add `app/Logging/Handlers/SkipsExceptionMirrorsHandler.php` (extends Monolog `HandlerWrapper`) plus a
`SkipsExceptionMirrors` tap, dropping records whose message contains `HandlerLogging::uncaughtException`
or `HandlerLogging::tooManyRequests`. Match with `str_contains` — under `APP_TYPE=local` the message
is padded and dash-prefixed by `StructuredLogging`.

Two deliberate consequences:
- `Handler::report()` guards the `HandlerLogging` call with `!app()->runningInConsole()`, so the
  duplicate only ever existed on web requests; console/queue paths are unaffected.
- `Handler::$dontReport` is an exact-class `in_array` while the framework's `shouldntReport` is
  `instanceof`-based, so a *subclass* of e.g. `HttpException` is logged by `HandlerLogging` but never
  reported to Sentry as an exception. Filtering the mirror makes those invisible to Sentry. Acceptable
  (they are 4xx), but it is a real behaviour change.

**Why a tap and not `before_send`**: a `before_send` closure cannot survive `php artisan config:cache`,
and whether the ECS entrypoint runs it lives in the infra repo. The tap also sits next to the channel
it filters and applies to nothing else.

### 3. `sentry` channel in `config/logging.php`

```php
'sentry' => empty(env('SENTRY_LARAVEL_DSN', env('SENTRY_DSN'))) ? [
    'driver'  => 'monolog',
    'handler' => NullHandler::class,
] : [
    'driver'            => 'sentry',
    'level'             => 'error',
    'report_exceptions' => false,
    // Order matters: the last tap wraps outermost, so mirrors are dropped before the dedup store.
    'tap' => [DeduplicateHandlers::class . ':sentry', SkipsExceptionMirrors::class],
],
```

Add `sentry` to the three stacks in this file: `stack_docker`, `stack_docker_local`, `scheduler`.

- The empty-DSN ternary mirrors `discord`. `Log::extend('sentry', ...)` is registered unconditionally
  so the driver normally resolves regardless, but `ServiceProvider::registerFeatures()` wraps each
  `register()` in a `try/catch` that swallows `Throwable` — if that ever fires the driver disappears
  and every stack including it falls back to the emergency logger. That is the #3445 failure mode.
- `report_exceptions => false` keys on the literal context key `exception` (`SentryHandler::doWrite()`
  line 164), **not** on any `Throwable` in context. The six Logging classes taking a `Throwable` name
  the param `$e`/`$ex`/`$throwable`, so none are affected. The only in-app `'exception' =>` log context
  is `SeasonAffixGroupService` (4 sites) and all four rethrow, so nothing goes dark. This only stops
  framework-level duplicates such as `LogManager::createEmergencyLogger()`.

### 4. `release` / `environment` in `config/sentry.php`

Both unset today, so Sentry cannot do regression detection or distinguish prod from staging:

```php
'release'     => env('SENTRY_RELEASE') ?: (is_file(base_path('version'))
    ? trim((string)file_get_contents(base_path('version')))
    : null),
'environment' => env('SENTRY_ENVIRONMENT') ?: env('APP_TYPE', 'local'),
```

Use `env('APP_TYPE')`, **not** `config('app.type')` — config files load in filename order and a
cross-config read at load time is unsafe (`config/horizon.php:55` is the in-repo precedent).

### 5. `.env.*.example`

`SENTRY_TRACES_SAMPLE_RATE` is `1.0` in all three example files, and `SENTRY_TRACE_QUEUE_ENABLED`
defaults to `true` — so every combat-log job is currently a transaction. Set the sample rate to `0`
and add blank `SENTRY_RELEASE` / `SENTRY_ENVIRONMENT` keys. Span quota is metered separately from
error quota, so this does not affect error capture.

## Works already — do not break

- **Context survives for free**: `LogManager::get()` pushes `ContextLogProcessor` onto the Monolog
  logger `Sentry\Laravel\LogChannel` returns, merging Laravel's Context into `LogRecord->extra`;
  `SentryHandler::doWrite()` then `setExtra()`s every key. So `structured:*`, `trace_id` and
  `correlationId` arrive as Sentry extras, `depth` inside `log_context`.
- **Breadcrumbs already on** (`sentry.breadcrumbs.logs = true`) — the structured lines preceding an
  error are attached, so each issue carries the `start()`/`end()` narrative.
- **Grouping is correct by default**: `SentryHandler` calls `captureMessage($record['message'])` and
  Sentry groups message events by message. `StructuredLogging` guarantees a constant message with all
  variability in context, so this yields exactly one stable issue per log site. No custom fingerprint.
  - **Do not enable `attach_stacktrace`** — it would make Sentry group by stack trace, fragmenting one
    log site into one issue per call path.
  - `SentryHandler` supports `context['fingerprint']` as a per-call escape hatch. Do not use it yet.
- **Octane**: do not add `log` to `config/octane.php`'s `flush` list — `LogManager` is a per-worker
  singleton and the SDK's `EventHandler` pushes/pops the Sentry scope on `RequestReceived`/
  `RequestTerminated`, so a long-lived `SentryHandler` is correct by design. Nothing may call
  `StructuredLogging::setChannel('sentry')`; those statics are process-global under Swoole.

## Tests

Under `tests/Unit/App/Logging/`, following `SchedulerLogChannelTest`:

- `SentryLogChannelTest` — empty-DSN resolves to a valid channel; logging to it emits no PHP warnings
  (the #3445 guard); and every `driver === 'stack'` entry in `config('logging.channels')` contains
  `sentry`. That last assertion must be on **config, not a resolved logger** — `phpunit.xml:32` pins
  `LOG_CHANNEL=daily`.
- `Handlers/SkipsExceptionMirrorsHandlerTest` — both `HandlerLogging` messages are not forwarded,
  including the padded/dash-prefixed `APP_TYPE=local` shape; an ordinary structured message is.
- `Handlers/DeduplicateHandlersTest` — tapping with `'sentry'` vs the default yields different
  `DeduplicationHandler::$deduplicationStore` paths (via `ReflectionProperty`). This is what stops the
  Discord/Sentry cross-suppression from regressing.
- `SentryStructuredContextTest` — a real `Hub` over a `ClientBuilder` with a stub `TransportInterface`;
  push a `LogRecord` through `SentryHandler` and assert the extras carry `trace_id`/`structured:*` and
  that `$event->getMessage()` is the bare `FooLogging::bar`. Pins both the context and grouping claims.
- A config test for the `release`/`environment` fallbacks, skip-guarded when the env vars are set.

Suggested order: (1) dedup store argument → (2) mirror tap → (3) `config/logging.php` →
(4) `config/sentry.php` → context/grouping test → (5) `.env.*.example`.

## Key files

- `config/logging.php`
- `config/sentry.php`
- `app/Logging/Handlers/DeduplicateHandlers.php`
- `app/Logging/Handlers/SkipsExceptionMirrorsHandler.php` (new) + `SkipsExceptionMirrors.php` (new)
- `app/Exceptions/Logging/HandlerLogging.php` (reference only)
- `tests/Unit/App/Logging/SchedulerLogChannelTest.php` (reference)
- `.env.example`, `.env.docker.example`, `.env.ci.example`

## Definition of done

- New and existing tests green: `php artisan test --compact tests/Unit/App/Logging/`
- `composer run fix` and `composer run analyse` clean
- Manual smoke: `AdminToolsExceptionController`'s `DiscordException` branch fires a `Log::error`, a
  stacked log **and** `HandlerLogging->uncaughtException()` in one click — with a DSN pointed at a
  catcher, assert one event for the exception (not two) plus one for the plain structured error
- PR body states the infra follow-up explicitly: `sentry` must be added to `LOG_STACK` **only if**
  production uses `LOG_CHANNEL=stack` (if it uses `stack_docker`, this change already covers it), and
  `SENTRY_TRACES_SAMPLE_RATE` should go to `0`

Wotuu:
:robot: Closes #3792

Errors logged through `StructuredLogging` are not exceptions, so `Integration::handles()` never saw
them — they only ever reached Discord. This adds a `sentry` log channel at error level and includes
it in every stack this repository declares, so they become Sentry issues.

This is the app-side half only. It is **inert until `sentry` is in the deployed log stack**, so it is
safe to merge on its own.

## What is in here

| Change | Why |
|---|---|
| `sentry` channel in `config/logging.php`, added to `stack_docker`, `stack_docker_local`, `scheduler` | the actual feature |
| `SkipsExceptionMirrors` tap | stops every uncaught exception collapsing into one Sentry issue |
| `DeduplicateHandlers` store argument | stops Discord and Sentry suppressing each other's records |
| `release` / `environment` in `config/sentry.php` | ties issues to a deploy; distinguishes staging from production |
| `SENTRY_TRACES_SAMPLE_RATE=0` in the example env files | spans are metered separately from errors and that budget is being spent on dropped data |

## The two non-obvious bits

**Exception mirrors.** `Handler::report()` logs every uncaught exception through
`HandlerLogging::uncaughtException()` *in addition to* the SDK's own capture. That method logs
`get_defined_vars()` under a **constant** message, and Sentry groups message events by their message
— so without a filter, every uncaught exception in the application would land in a single issue
titled `HandlerLogging::uncaughtException`. `SkipsExceptionMirrors` drops those records, on the
`sentry` channel only; Discord and the file logs still get them.

Two deliberate consequences:
- `Handler::report()` guards that call with `!app()->runningInConsole()`, so the duplicate only ever
  existed on web requests — console and queue paths are unaffected.
- `Handler::$dontReport` is an exact-class `in_array` while the framework's `shouldntReport` is
  `instanceof`-based, so a *subclass* of e.g. `HttpException` is logged by `HandlerLogging` but never
  reported to Sentry as an exception. Filtering the mirror makes those invisible to Sentry. They are
  4xx, so this is acceptable, but it is a real behaviour change.

**Deduplication store.** `DeduplicateHandlers` hardcoded `logs/discord-deduplication.log`. The store
is keyed on level + message alone, so reusing it for a second channel would have meant an error
already sent to Discord within the 60s window silently never reaching Sentry, and vice versa. It now
takes the store name as a tap argument (`DeduplicateHandlers::class . ':sentry'`), defaulting to
`discord` so that channel is unchanged. Note the store is a file under `storage_path()`, so on ECS it
is per-task rather than global.

`report_exceptions => false` on the channel is *not* what handles the mirrors — it keys on the
literal context key `exception` (`SentryHandler::doWrite()` line 164), and `StructuredLogging` names
its parameters `$e`/`$ex`/`$throwable`. The six Logging classes taking a `Throwable` are therefore
unaffected and still reach Sentry; the flag only suppresses framework-level duplicates such as the
emergency logger.

## Verified, not assumed

`SentryStructuredContextTest` pushes a real `LogRecord` through `SentryHandler` against an in-memory
transport and asserts both properties the triage workflow depends on, because both would break
silently:
- `trace_id` and the `structured:*` groups arrive as Sentry extras (via Laravel's `ContextLogProcessor`)
- the event message stays the bare `ClassLogging::method`, which is what gives **one stable issue per
  log site**

Related: do not enable `attach_stacktrace` — it would make Sentry group by stack trace and fragment
one log site into one issue per call path.

`SentryLogChannelTest::sentryChannel_givenConfiguredDsn_wrapsHandlerInBothTaps` resolves the
DSN-present shape explicitly (never exercised locally or in CI) and asserts the full tap chain and the
dedup store path.

## Follow-up needed in `keystoneguru-infra`

Not blocking this PR, but this change does nothing in production until the infra side lands.
Both values below are now confirmed against the prod config rather than assumed:

- **`LOG_STACK` must gain `sentry`.** Production runs `LOG_CHANNEL=stack` with
  `LOG_STACK=stderr,daily,discord`, i.e. the framework's own `stack` channel — *not* one of the
  stacks defined in `config/logging.php`. So the stack edits in this PR do not cover production on
  their own; that variable is the only thing that switches this on.
- **`SENTRY_TRACES_SAMPLE_RATE=0`.**

`LOG_LEVEL=warning` in production is not a problem: the `sentry` channel sets `level => 'error'`
itself, so warnings continue to go to `stderr`/`daily` only.

The `release` fallback is safe: the `version` file is generated at build time and baked into the
image, and `php artisan config:cache` runs at container start (after hotfixes are overlaid) in
`cdk/lib/constructs/services/php-fpm/php-fpm-staged-service.ts`, so the cached release always matches
the running image.

## Testing

- `php artisan test --compact tests/Unit/App/Logging/` — 31 passed
- `php artisan test --compact tests/Unit` — 501 passed; the one failure is `MapTilesExistenceTest`,
  a known worktree-environment failure (assets are mounted on the main stack only) that CI excludes
- `composer run fix` and `composer run analyse` clean


Wotuu:

A fresh-context agent reviewed this and reproduced three defects inside the worktree's container.
All three were silent — the channel resolved, the logging call succeeded, and no event was ever
sent. All are fixed in 2427d38.

| # | Finding | Fix |
|---|---|---|
| 1 | **The channel sent nothing at all with a DSN configured.** `DeduplicationHandler::flush()` forwards through `handleBatch()`, and `SentryHandler::handleBatch()` still compares a Monolog 3 `Level` enum against an integer — `array_filter` emptied the batch and every record was discarded. | Dropped the dedup tap from the sentry channel; Sentry groups and rate-limits server-side anyway. That made the tap's `$store` argument unnecessary, so it and its test are reverted — `DeduplicateHandlers.php` is no longer in the diff. |
| 2 | **The discord no-op guard swallowed sentry in a stack.** `NullHandler::handle()` returns `true`, which ends Monolog's handler loop, so every channel listed after discord was skipped. Production runs `LOG_CHANNEL=stack` with discord listed first, so this would have hit immediately. | Both guards now use `NoopHandler`, which returns `false` and keeps the record bubbling. |
| 3 | **The mirror filter dropped records the SDK never captured.** `Handler::$dontReport` is an exact class match while `shouldReport()` matches with `instanceof`, so an `HttpException` subclass is logged yet never reported natively — those records were the only trace of the failure, and the filter binned them. | `Handler::report()` now passes `shouldReport()`'s own answer along in the log context and the filter keys on that instead of on the message, so only true duplicates are dropped. What it keeps gets a fingerprint derived from the exception class, so those no longer collapse into one issue either. |

**Why the tests missed all three.** They asserted on configuration shape, or called `SentryHandler::handle()` directly — the one path production never takes. Rewritten to resolve the *real* DSN-shaped channel (re-evaluating `config/logging.php` with a DSN present, rather than a hand-written copy that would stay green if the config changed) against a stub transport, and assert an event actually arrives — including through a stack with a no-op channel in front of it.

Worth recording for anyone who hits it later: `Log::build()` does **not** apply taps. `LogManager::tap()` looks them up by channel name in `logging.channels`, so an on-demand channel silently gets none. The first version of the end-to-end test passed for that reason.

Findings deliberately not addressed:
- The reviewer flagged `SENTRY_TRACES_SAMPLE_RATE=0` in the `.env.*.example` files as having no billing effect, since production env comes from `keystoneguru-infra`. Correct, and already called out under the infra follow-up — the example files are there so a fresh checkout doesn't start burning span quota.

Wotuu (commented):
Ok

Issue #3358, opened 2026-07-02
Remove the deprecated git-based mapping sync path (mapping:sync/commit/merge) (follow-up to #3354)


Wotuu:
## Background

The git-based mapping sync path auto-regenerates the seeder JSON from the live DB and commits/pushes it to Git on a schedule. This path is **deprecated** — mapping data is no longer managed by round-tripping the whole dev DB through auto-commits, and it actively produces destructive drift (see below). It was marked `@deprecated` in `app/Console/Commands/Mapping/Sync.php` as part of #3354.

The path is:

```
Schedule::command('mapping:sync')->everyFiveMinutes();   // routes/console.php:39
  -> mapping:save     (App\Console\Commands\Mapping\Save.php)     writes database/seeders/dungeondata/*.json from the live DB
  -> mapping:commit   (App\Console\Commands\Mapping\Commit.php)   git add database/seeders/dungeondata/* && git commit && checkout development && merge mapping && git push
  -> mapping:merge    (App\Console\Commands\Mapping\Merge.php)
```

gated by `App\Service\Mapping\MappingService::shouldSynchronizeMapping()`.

### Why it is harmful today
- `mapping:save` regenerates **all** seeder JSON from whatever is in the local/dev DB. When the DB lacks demo dungeon routes, `*/dungeonroutes.json` get emptied to `[]` and `mapping_commit_logs.json` churns — perpetual working-tree drift on every branch.
- `mapping:commit` does a blind `git add database/seeders/dungeondata/*` then commits/pushes, i.e. it can commit destructive drift automatically. (Locally it is currently only prevented by the in-container "dubious ownership /var/www" git guard — that is luck, not a safeguard.)
- `mapping:save` itself is still needed as a **manual** developer command; only the automated `sync -> commit -> merge` + the 5-minute schedule are being removed.

## Remaining work

1. Remove the schedule entry: `Schedule::command('mapping:sync')->everyFiveMinutes();` in `routes/console.php` (~line 39; also drop `'mapping'` from the relevant command group array if it becomes unused).
2. Delete the commands and their tests/wiring:
   - `app/Console/Commands/Mapping/Sync.php`
   - `app/Console/Commands/Mapping/Commit.php`
   - `app/Console/Commands/Mapping/Merge.php` (confirm it is only used by this path)
[...]

PR #3808, opened 2026-08-03
Add a new-machine-setup skill for bootstrapping a fresh dev environment

### LINKED ISSUE #3807: Add a new-machine-setup skill for bootstrapping a fresh dev environment
:robot: Setting up this project on a new machine is currently undocumented tribal knowledge. Everything needed lives in the repo somewhere — `docker-compose.yml`, `docker-compose/app/Dockerfile`, `sh/worktree.sh`, `config/database.php` — but nothing ties it together, and several steps are impossible to guess from a fresh clone.

## Background

Prompted by a laptop being retired. Reconstructing the setup from scratch surfaced a set of traps that are invisible until you hit them:

- **`docker-compose/app/.env`** (a single `GIT_TOKEN=` line) is gitignored, and the app `Dockerfile` does `COPY .env /tmp/.env` to envsubst it into composer's `auth.json`. Without it `docker compose build app` fails outright, with no hint as to why.
- **`keystone.guru.assets` must be cloned as a sibling directory** — `docker-compose.yml` bind-mounts `../keystone.guru.assets` into both `app` and `app-assets`. A missing sibling gives you an empty mount and confusingly broken images.
- **Map tiles are not in any repo and are not needed locally.** `config/keystoneguru.php` derives `tiles_base_url` from `ASSETS_BASE_URL`, so maps render from the production CDN. The consequence is that `tests/Unit/MapTiles/MapTilesExistenceTest.php` fails on every dev machine — expected, but nothing says so.
- **`PUID`/`PGID` are baked into the image** as build args. A wrong value produces a passwd-less container user, which surfaces as *every MDT encode/decode test failing* (`sudo: you do not exist in the passwd database`) rather than as a permissions error.
- **The seed sequence has two steps.** `db:seed --database=migrate` does not create users; `LaratrustSeeder` is a separate invocation. Skipping it yields a schema that looks fine and fails half the suite (tests assume user id 1 is the admin).
- **DB users**: a fresh MySQL container creates exactly one non-root user from `MYSQL_USER`/`MYSQL_PASSWORD`, while the `migrate` and `phpunit` connections authenticate as `DB_MIGRATION_*` / `DB_PHPUNIT_*`. These happen to be the same credentials today, which is the only reason `migrate` works out of the box. `sh/worktree.sh` sidesteps this by granting to every user that already has grants on the main schema — which presupposes those users exist.

## What this adds

`.claude/skills/new-machine-setup/SKILL.md` — eleven phases, from "Docker is installed" to "an agent can run tests, build assets, render maps and create worktrees". Each phase ends in an explicit pass/fail probe, so a half-configured stack can't be declared ready.

The final acceptance gate is `sh/worktree.sh create 9999-smoke` + `remove`, which exercises the deploy key, private DB provisioning, the baked `ksg` user and shared-service network attach in one shot.

It also carries a transfer list of the gitignored files that cannot be regenerated (both `.env` files, four SSH key pairs), and a troubleshooting table of the transferable traps: Rancher phantom bind-mounts and the `auto.cnf` diagnosis, `credsStore: wincred.exe` blocking pulls, the compose 5.0.1 `monitor.Start` panic, root-owned `storage/logs` causing a 504 loop, and `.wslconfig` sizing.

## Caveats for review

- The skill was **derived by reading the repo on a working machine, not executed against a clean one**. It says so up front and tells the reader to trust the source files over the doc, and to fix the doc in-session on any discrepancy. Someone actually running it end-to-end is the real validation.
- Machine-specific incidents from the retired laptop (GPU driver crashes, a one-off 4GB php-fpm request) were deliberately left out; only transferable traps are carried.
- The Docker section asks which backend is in use rather than assuming Rancher Desktop.

## Definition of done

- Skill file merged.
- Ideally: someone follows it end to end on a genuinely fresh machine and pushes back any corrections.

Docs-only change — no code paths touched, so `composer run analyse` / `composer run fix` / the test suite are unaffected.

Wotuu:
:robot: Closes #3807

Adds `.claude/skills/new-machine-setup/SKILL.md` — a runbook for bringing up a complete working environment on a fresh machine, from "Docker is installed" to "an agent can run tests, build assets, render maps and create worktrees".

Docs-only: one new file, no code paths touched.

## Why

Everything needed is already in the repo somewhere (`docker-compose.yml`, `docker-compose/app/Dockerfile`, `sh/worktree.sh`, `config/database.php`), but nothing ties it together and several steps cannot be guessed from a fresh clone. See #3807 for the full list; the sharpest ones:

- `docker-compose/app/.env` is gitignored and `COPY .env /tmp/.env` in the app Dockerfile needs it — **without it `docker compose build app` fails**, and nothing explains why.
- `keystone.guru.assets` must be a **sibling** clone; compose bind-mounts `../keystone.guru.assets`.
- Map tiles are in no repo and aren't needed — tiles resolve from the prod CDN via `ASSETS_BASE_URL`. Corollary: `MapTilesExistenceTest` fails on every dev machine, by design.
- `PUID`/`PGID` are baked into the image; a mismatch surfaces as *every MDT test failing*, not as a permissions error.
- `db:seed` does not create users — `LaratrustSeeder` is a separate step, and skipping it silently breaks half the suite.

## Shape

Eleven phases, each ending in an explicit **pass/fail probe** — a setup doc without probes is where an agent declares success on a half-broken stack. Final acceptance gate is `sh/worktree.sh create 9999-smoke` + `remove`, which exercises the deploy key, private DB provisioning, the baked `ksg` user and shared-service network attach in one shot.

Also includes a transfer list of the gitignored, unregenerable files (both `.env`s, four SSH key pairs) and a troubleshooting table of transferable traps only — Rancher phantom bind-mounts with the `auto.cnf` diagnosis, `credsStore: wincred.exe` blocking pulls, the compose 5.0.1 `monitor.Start` panic, root-owned `storage/logs` causing a 504 loop, `.wslconfig` sizing.

## Review notes

- **Not yet cold reviewed** — opened as a draft. Wotuu is taking review and merge from another machine.
- **Derived by reading the repo, not executed against a clean machine.** The skill says so up front and tells the reader to trust the source files over the doc, and to fix the doc in-session on any discrepancy. Running it end to end on a genuinely fresh machine is the real validation, and corrections are welcome as follow-ups.
- Machine-specific incidents from the retired laptop (GPU driver crashes, a one-off 4GB php-fpm request) were deliberately excluded.
- The Docker section asks which backend is in use rather than assuming Rancher Desktop.

### DIFF
--- a/.claude/skills/new-machine-setup/SKILL.md
+++ b/.claude/skills/new-machine-setup/SKILL.md
@@ -37,7 +37,10 @@ the same session; you are the last person who will have the context to.
 
 Six files are gitignored, secret, and **cannot be regenerated or reconstructed**. If the old machine
 is gone before these are copied off it, the environment cannot be rebuilt without reissuing every
-credential by hand.
+credential by hand. On top of these six, `~/Git/private/keystone.guru.assets/tiles/` (~36G,
+gitignored — see phase 2) is *also* unrecoverable: it isn't in any repo and neither is the software
+that generates it (`readme.md:59-61`). It doesn't need "reissuing" like a credential, but if the old
+machine is gone before it's copied or synced, it is simply gone.
 
 | File | Why it is unrecoverable |
 |---|---|
@@ -110,21 +113,50 @@ Takes effect only after `wsl --shutdown` — the user must do that themselves.
 
 ---
 
-## Phase 2 — Clone layout
+## Phase 2 — SSH keys, then clone
 
-Two repos, **siblings in the same parent directory**. This is not cosmetic: `docker-compose.yml`
-bind-mounts `../keystone.guru.assets` into both the `app` and `app-assets` services, and
-`create_host_path: false` is not set on that mount — a missing sibling gives you an empty dir and
-silently broken images.
+The clone below is over `git@github.com:` (a private repo), and phase-4 git config turns on
+`commit.gpgsign true` globally — so the SSH auth key and the signing key both have to exist
+**before** cloning, not after. Restore them here, first, rather than waiting for the "Restore
+secrets" phase:
+
+```bash
+mkdir -p ~/.ssh && chmod 700 ~/.ssh
+cp <bundle>/ssh/* ~/.ssh/
+chmod 600 ~/.ssh/wotuu_passwordless ~/.ssh/keystone_worktree_ed25519 ~/.ssh/git_signing_ed25519
+chmod 644 ~/.ssh/*.pub ~/.ssh/config ~/.ssh/allowed_signers
+```
+
+`~/.ssh/config` must contain:
+
+```
+Host github.com
+    HostName github.com
+    IdentityFile ~/.ssh/wotuu_passwordless
+    IdentitiesOnly yes
+```
+
+**Probe:** `ssh -T git@github.com` returns `Hi Wotuu! You've successfully authenticated…` with no
+passphrase prompt.
+
+Now clone. Two repos, **siblings in the same parent directory**. This is not cosmetic:
+`docker-compose.yml` bind-mounts `../keystone.guru.assets` into both the `app` and `app-assets`
+services, and `create_host_path: false` is not set on that mount — a missing sibling gives you an
+empty dir and silently broken images.
 
 ```bash
 mkdir -p ~/Git/private && cd ~/Git/private
 git clone git@github.com:RaiderIO/keystone.guru.git
-git clone git@github.com:RaiderIO/keystone.guru.assets.git   # ~342M, images only
+git clone git@github.com:RaiderIO/keystone.guru.assets.git   # ~342M tracked (images, webfonts)
 ```
 
 `sh/worktree.sh` also expects `~/Git/private/keystone.guru-worktrees/` as its worktree root; it
 creates that itself on first use.
+
+The clone above only brings `images/` and `webfonts/`. The checkout also has a `tiles/` directory
+(~36G, gitignored per `keystone.guru.assets/.gitignore:1`) that is **not** part of the git history —
+map tiles have to be transferred separately (rsync/external drive/etc. from the old machine). See
+the tiles note in phase 4 for what happens if you skip this.
 
 Git config (global — replicate from the old machine):
 
@@ -158,39 +190,27 @@ mkdir -p ~/.config/git && echo '**/.claude/settings.local.json' >> ~/.config/git
 
 ---
 
-## Phase 3 — Restore secrets
+## Phase 3 — Restore remaining secrets
+
+The SSH keys were already restored in phase 2 (they had to be, before the clone could happen). What's
+left is the two `.env` files, which do need the repo cloned first since they're copied into it:
 
 ```bash
 cd ~/Git/private/keystone.guru
 cp <bundle>/env                  .env
 cp <bundle>/app-env              docker-compose/app/.env
-
-mkdir -p ~/.ssh && chmod 700 ~/.ssh
-cp <bundle>/ssh/* ~/.ssh/
-chmod 600 ~/.ssh/wotuu_passwordless ~/.ssh/keystone_worktree_ed25519 ~/.ssh/git_signing_ed25519
-chmod 644 ~/.ssh/*.pub ~/.ssh/config ~/.ssh/allowed_signers
-```
-
-`~/.ssh/config` must contain:
-
-```
-Host github.com
-    HostName github.com
-    IdentityFile ~/.ssh/wotuu_passwordless
-    IdentitiesOnly yes
 ```
 
 Note `~/.ssh` is bind-mounted read-write into the `app` container (`~/.ssh:/home/ksg/.ssh`), which
 is how in-container composer reaches private VCS repos.
 
-**Probe:** `ssh -T git@github.com` returns `Hi Wotuu! You've successfully authenticated…` with no
-passphrase prompt.
+**Probe:** `.env` and `docker-compose/app/.env` both exist and are non-empty.
 
 ---
 
 ## Phase 4 — Per-machine `.env` edits
 
-The restored `.env` is correct except for three things that are machine-specific:
+The restored `.env` is correct except for four things that are machine-specific:
 
 1. **`PUID` / `PGID`** must match the new host user, or every file the container writes into the
    bind mount lands root-owned (breaks `git status`, and root-owned `storage/logs` puts php-fpm
@@ -204,6 +224,15 @@ The restored `.env` is correct except for three things that are machine-specific
    rather than mount an empty dir, so a wrong value here shows up as a compose error, not silent
    data loss.
 3. **`ASSETS_BASE_URL=https://assets.keystone.guru`** — see the tiles note below.
+4. **`ASSETS_BASE_URL_INTERNAL`** — changing `ASSETS_BASE_URL` alone is not enough.
+   `config/keystoneguru.php:21` derives a separate `tiles_base_url_internal` from
+   `ASSETS_BASE_URL_INTERNAL`, and `DungeonRouteController::preview()`
+   (`app/Http/Controllers/DungeonRoute/DungeonRouteController.php:301-304`) swaps to that internal
+   URL specifically for the server-side (puppeteer/thumbnail) render path when there's no
+   authenticated user. `.env.docker.example:64` ships `ASSETS_BASE_URL_INTERNAL=http://app-assets`,
+   which is correct for the default Docker Compose network — verify the restored `.env` has this
+   set, or the phase-9 browser-render probe will silently render against a tile-less/asset-less
+   container instead of erroring.
 
 Also create the empty data dirs compose refuses to create for you:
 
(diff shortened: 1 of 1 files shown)


.claude/skills/new-machine-setup/SKILL.md
```diff
@@ -0,0 +1,469 @@
...
+| `Unable to locate file in Vite manifest` / stale JS in browser | Assets not built, or `version` stale | `npm run production` (writes `version` from the git rev) |
+| PHPStan red locally, green in CI | Local `vendor/` drifted from `composer.lock` | `docker compose exec -T app composer install` |
+| WSL crashes / everything freezes | VM-wide OOM | `.wslconfig` (phase 1) + keep OpenSearch off + `wsl --shutdown` |
+| `worktree.sh push` dies on a missing key | `~/.ssh/keystone_worktree_ed25519` not restored | Phase 3 |
+
+## Known-good failures — do not chase
+
+- **`MapTilesExistenceTest`** fails on every dev machine (no local tiles; see phase 4).
+- **Cron drift in `database/seeders/dungeondata/`** — the `cron` container runs `mapping:sync` every
+  5 minutes and rewrites seeder JSON. Benign dev noise. Never `git add -A` while working on seeders.
```
Wotuu: :robot: **`mapping:sync` no longer exists — this "known-good failure" is now a real signal.**

`grep -rn "mapping:sync"` returns nothing in the repo, and `routes/console.php` has no such scheduled entry. The command was deleted in #3358; the `cron` container does not rewrite seeder JSON any more.

That inverts the advice: drift in `database/seeders/dungeondata/` is now a genuine, unexplained change that must be investigated, not benign dev noise to ignore. Suggest dropping this bullet entirely (the `git add -A` caution can stay elsewhere if wanted, but not justified by a cron that no longer runs).

.claude/skills/new-machine-setup/SKILL.md
```diff
@@ -0,0 +1,469 @@
...
+```bash
+mkdir -p docker-compose/mysql docker-compose/mysql-combatlog docker-compose/redis-data
+```
+
+### About map tiles
+
+Map tiles are **not** in any repo (`readme.md`: "Not included in this repository"). The
+`keystone.guru.assets` checkout contains `images/` only. `config/keystoneguru.php` derives
+`tiles_base_url` from `ASSETS_BASE_URL`, so with `ASSETS_BASE_URL=https://assets.keystone.guru` the
+maps load tiles straight from the production CDN and **everything renders correctly with no local
```
Wotuu: :robot: **The assets checkout is not "images only", and the transfer bundle silently drops 36G of unrecoverable tiles.**

On the reference machine this doc was derived from, `~/Git/private/keystone.guru.assets/` contains `images`, `webfonts` **and `tiles` — 36G, gitignored in that repo** (`.gitignore:1:tiles`). So:

1. "~342M, images only" (phase 2) and "contains `images/` only" (phase 4) are both wrong.
2. Phase 0 enumerates six unrecoverable gitignored files and omits the tiles tree. `readme.md:59-61` says both the tiles *and* the software that generates them are not in any repo — so if the old machine is wiped, that 36G is gone with no documented way to get it back. Whether or not it belongs in the "copy this" bundle (it is large, and the CDN serves the same content), phase 0 should at least state that it exists, is gitignored, and is not reconstructible from a clone.
3. Consequently the claim that `MapTilesExistenceTest` "fails on every dev machine, by design" is false on the very machine this was written from — it fails only because tiles were not transferred. Recommend rewording to "fails unless the `tiles/` tree is present locally; CI excludes it via `--exclude-group=MapTiles`".

.claude/skills/new-machine-setup/SKILL.md
```diff
@@ -0,0 +1,469 @@
...
+
+Two repos, **siblings in the same parent directory**. This is not cosmetic: `docker-compose.yml`
+bind-mounts `../keystone.guru.assets` into both the `app` and `app-assets` services, and
+`create_host_path: false` is not set on that mount — a missing sibling gives you an empty dir and
+silently broken images.
+
+```bash
+mkdir -p ~/Git/private && cd ~/Git/private
+git clone git@github.com:RaiderIO/keystone.guru.git
+git clone git@github.com:RaiderIO/keystone.guru.assets.git   # ~342M, images only
```
Wotuu: :robot: **Phase 2 depends on phase 3, so the strict top-to-bottom contract breaks at the second phase.**

This block does `git clone git@github.com:…` (SSH) and sets `commit.gpgsign=true` with `user.signingkey ~/.ssh/git_signing_ed25519.pub`, but `~/.ssh/config` (which pins `github.com` → `wotuu_passwordless`), the auth key and the signing key are only restored in **phase 3**. On a genuinely fresh machine the clone fails at this point, and the signing config points at a nonexistent file.

The header's bootstrap-paradox note also tells the reader to clone first and only then read this file, which reinforces the wrong order. Either move the secret restore (phase 3) ahead of the clone, or split phase 2 into "clone over HTTPS" + "git config, after phase 3".

.claude/skills/new-machine-setup/SKILL.md
```diff
@@ -0,0 +1,469 @@
...
+   into a 504 exception loop):
+   ```bash
+   sed -i "s/^PUID=.*/PUID=$(id -u)/; s/^PGID=.*/PGID=$(id -g)/" .env
+   ```
+2. **`DB_DATA_VARIANT=`** must be **empty**. It selects the MySQL bind-mount source
+   (`docker-compose/mysql${DB_DATA_VARIANT}`); `-prod` points at local restores of production dumps
+   which will not exist on a fresh machine. `create_host_path: false` makes compose fail loudly
+   rather than mount an empty dir, so a wrong value here shows up as a compose error, not silent
+   data loss.
+3. **`ASSETS_BASE_URL=https://assets.keystone.guru`** — see the tiles note below.
```
Wotuu, line 226: :robot: **`ASSETS_BASE_URL_INTERNAL` is a separate key and is not covered here.**

`config/keystoneguru.php:21` derives `tiles_base_url_internal` from `ASSETS_BASE_URL_INTERNAL` independently of `ASSETS_BASE_URL`, and `DungeonRouteController.php:303-304` swaps the config over to the `_internal` values for the server-side render path. `.env.docker.example:64` ships `ASSETS_BASE_URL_INTERNAL=http://app-assets`.

So following phase 4 as written (change `ASSETS_BASE_URL` only) gives working maps in the browser but a thumbnail/preview path still pointed at the local `app-assets` container, which has no `tiles/` — a failure the phase-9 map probe won't catch, since that probe looks at the browser render. Worth adding `ASSETS_BASE_URL_INTERNAL` to bullet 3 and to the phase-4 probe `grep`.

Wotuu:

.claude/skills/new-machine-setup/SKILL.md
Wotuu: :robot: Fixed — dropped the "cron drift is benign" framing entirely. The "Known-good failures" section now states plainly that `mapping:sync` was deleted in #3358, nothing schedules a rewrite of `database/seeders/dungeondata/`, and any unexpected diff there is a real signal to investigate (confirmed via `grep -rn "mapping:sync"` and `routes/console.php`).

.claude/skills/new-machine-setup/SKILL.md
Wotuu: :robot: Fixed. Phase 0 now calls out `tiles/` (~36G, gitignored, unrecoverable — neither the tiles nor the generating software live in any repo per `readme.md:59-61`) alongside the six secret files. Phase 2 clone comment corrected from "images only" to "images, webfonts", with a note that `tiles/` is gitignored and must be transferred separately. Phase 4's "About map tiles" section and the phase-9 acceptance check both now say `MapTilesExistenceTest` only fails when `tiles/` wasn't transferred, not unconditionally.

.claude/skills/new-machine-setup/SKILL.md
Wotuu: :robot: Reordered. Phase 2 is now "SSH keys, then clone" — it restores `~/.ssh` (auth key, signing key, `~/.ssh/config`) and probes `ssh -T git@github.com` before doing the `git@github.com:` clone. Phase 3 is now just the two `.env` file copies (which do need the repo cloned first). Updated the two other references to "phase 3" that pointed at the SSH key restore (troubleshooting table) to point at phase 2 instead.

.claude/skills/new-machine-setup/SKILL.md
Wotuu: :robot: Fixed. Phase 4 now has a 4th machine-specific `.env` item for `ASSETS_BASE_URL_INTERNAL`, explaining that `config/keystoneguru.php:21` derives `tiles_base_url_internal` from it independently, that `DungeonRouteController::preview()` (`app/Http/Controllers/DungeonRoute/DungeonRouteController.php:301-304`) uses it for the server-side render path, and that `.env.docker.example:64` ships `ASSETS_BASE_URL_INTERNAL=http://app-assets` as the correct default. Added it to the phase-4 probe `grep` as well.

PR #3809, opened 2026-08-03
Pin ext-lua and cli_weakauras_parser CI builds to commit SHAs

### LINKED ISSUE #3802: Pin the ext-lua build in php-ci-setup to a commit SHA instead of unpinned HEAD
Found incidentally while assessing package split-offs (#3797).

## Problem

`.github/actions/php-ci-setup/action.yml` builds the `ext-lua` PHP extension from
`https://github.com/Wotuu/php-lua` at **unpinned HEAD**. There is no tag, no commit SHA, no branch
pin.

This means **an upstream commit to that repository can turn CI red across every PR in this repo,
with no change on this side.** The failure would look like an unexplained build break in a step
nobody touched, on branches that were green an hour earlier — one of the more expensive classes of
CI failure to diagnose, because the natural instinct is to look at the diff.

The same setup action also `cargo install`s `cli_weakauras_parser` from codeberg; worth checking
whether that is version-pinned while you are in there.

## Why it has not bitten yet

The build result is cached (`~/lua-ext-cache`), so a fresh upstream commit only takes effect when
the cache key rotates or misses. That makes the eventual failure **delayed and uncorrelated with the
upstream change**, which makes it worse to diagnose, not better.

## Recommended approach

Pin to an explicit commit SHA, following the pattern already used elsewhere in the project — the
`nnoggie/mythicdungeontools` and `nnoggie/mdt-legacy` entries in `composer.json` both pin an exact
`reference` SHA rather than a branch, so this is consistent with existing practice rather than a new
convention.

Include the pinned SHA in the cache key so a deliberate bump busts the cache correctly.

## Relevance to the package split

Noted in #3797 but worth restating: this toolchain is paid on **every one of the five CI shards**,
and any package extracted from `app/Logic/MDT` (#3799) inherits it. Pinning it is a prerequisite for
that package's CI being reproducible.

## Key files

- `.github/actions/php-ci-setup/action.yml`

## Definition of done

- `Wotuu/php-lua` pinned to an explicit commit SHA (and `cli_weakauras_parser` too, if it is not
  already).
- Cache key includes the pin so a bump invalidates it.
- A PR's `php-tests` run goes green from a cold cache.

Wotuu:
:robot: Closes #3802

## Summary
- `.github/actions/php-ci-setup/action.yml` built `Wotuu/php-lua` and `codeberg.org/Velithris/cli-weakauras-parser` from unpinned HEAD — an upstream commit to either repo could turn CI red across every PR with no change on this side, and since both builds are cache-hidden the failure would surface late and look uncorrelated with the actual cause.
- Pin both to explicit commit SHAs (`a76f01149e6132810fe6b3095f2b9bc597d34f9e` for php-lua, `febb57faeaa805a79c9c2e0425c804e441a0e969` for cli_weakauras_parser), matching the existing `composer.json` convention of pinning `reference` SHAs for the `nnoggie/mythicdungeontools`/`mdt-legacy` packages.
- Fold both pins into their respective `actions/cache@v5` keys so a deliberate bump busts the cache and triggers a rebuild.

## Test plan
- [ ] `php-tests` CI run goes green from a cold cache (new cache keys guarantee this on first run of this PR)

### DIFF
--- a/.github/actions/php-ci-setup/action.yml
+++ b/.github/actions/php-ci-setup/action.yml
@@ -10,6 +10,15 @@ inputs:
 runs:
   using: "composite"
   steps:
+    # Pinned commits for third-party toolchain repos with no tagged releases. Bump these
+    # deliberately (not via unpinned HEAD) when an upstream fix/feature is needed - the pin is
+    # baked into the cache keys below so a bump busts the cache correctly.
+    - name: Pin toolchain commit SHAs
+      id: pins
+      run: |
+        echo "php-lua-sha=a76f01149e6132810fe6b3095f2b9bc597d34f9e" >> "$GITHUB_OUTPUT"
+        echo "cli-weakauras-parser-sha=febb57faeaa805a79c9c2e0425c804e441a0e969" >> "$GITHUB_OUTPUT"
+      shell: bash
     - name: Checkout
       uses: actions/checkout@v5
       with:
@@ -47,14 +56,14 @@ runs:
       shell: bash
 
     # Cache the compiled lua.so / bit.so so we skip the ~30-40s phpize/configure/make on every run
-    # (paid per shard once the test job is sharded). Bump the trailing "-v1" to bust the cache if the
-    # Wotuu/php-lua repo (built from HEAD, unpinned) changes and a rebuild is needed.
+    # (paid per shard once the test job is sharded). The pinned php-lua SHA is part of the key so
+    # bumping it above busts the cache and triggers a rebuild.
     - name: Restore Lua extension build cache
       id: cache-lua-ext
       uses: actions/cache@v5
       with:
         path: ~/lua-ext-cache
-        key: lua-ext-php8.4-${{ hashFiles('storage/lua/LuaBitOp-1.0.3.zip') }}-v1
+        key: lua-ext-php8.4-${{ hashFiles('storage/lua/LuaBitOp-1.0.3.zip') }}-${{ steps.pins.outputs.php-lua-sha }}
 
     - name: Build PHP Lua extension
       run: |
@@ -75,6 +84,7 @@ runs:
           echo "==> Building Lua extension from source"
           git clone https://github.com/Wotuu/php-lua.git /tmp/php-lua
           cd /tmp/php-lua
+          git checkout "${{ steps.pins.outputs.php-lua-sha }}"
           phpize
           ./configure
           make -j$(nproc)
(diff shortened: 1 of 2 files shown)


.github/actions/php-ci-setup/action.yml
```diff
@@ -75,6 +84,7 @@ runs:
           echo "==> Building Lua extension from source"
           git clone https://github.com/Wotuu/php-lua.git /tmp/php-lua
           cd /tmp/php-lua
+          git checkout "${{ steps.pins.outputs.php-lua-sha }}"
```
Wotuu, line 90: :robot: Non-blocking, follow-up: this pin creates a new divergence between CI and the app image. `docker-compose/app/Dockerfile` still builds both toolchain deps from unpinned HEAD (`RUN cargo install --git https://codeberg.org/Velithris/cli-weakauras-parser.git --locked` at line 9, `git clone https://github.com/Wotuu/php-lua.git .` at line 46). After this PR, CI tests against php-lua `a76f011` / parser `febb57f`, while the container builds whatever HEAD is at image-build time — so a green CI run no longer implies the container has the same extension/binary build, and the original "upstream commit breaks us with no change on this side" failure mode still applies to the Docker image (dev + prod). Worth a follow-up issue to pin the same two SHAs there.

.github/actions/php-ci-setup/action.yml
```diff
@@ -10,6 +10,15 @@ inputs:
 runs:
   using: "composite"
   steps:
+    # Pinned commits for third-party toolchain repos with no tagged releases. Bump these
```
Wotuu: :robot: Nit: the comment is inaccurate for one of the two repos — `cli-weakauras-parser` does have tagged releases, and the pinned `febb57faeaa805a79c9c2e0425c804e441a0e969` is exactly `refs/tags/v0.1.8` (verified via `git ls-remote`). `--rev <sha>` is still the right choice (tags are mutable, a SHA is not), so no change to the command — but the wording should be adjusted, and adding `(= v0.1.8)` beside the SHA on line 20 would tell a future bumper where they are in the version history. `Wotuu/php-lua` genuinely has no tags, so the claim only holds there.

.github/actions/php-ci-setup/action.yml
```diff
@@ -98,14 +108,14 @@ runs:
       uses: actions/cache@v5
       with:
         path: ~/.cargo/bin/cli_weakauras_parser
-        key: cli-weakauras-parser-rust-1.91.1
+        key: cli-weakauras-parser-rust-1.91.1-${{ steps.pins.outputs.cli-weakauras-parser-sha }}
```
Wotuu: :robot: The Rust toolchain version `1.91.1` is a magic string duplicated between the cache key here and `--default-toolchain 1.91.1` in the build step below — and both lines are touched by this diff. That is the same drift bug this PR exists to prevent: bump the toolchain in the install command, forget the key, and CI silently keeps reusing a `cli_weakauras_parser` binary built by the old toolchain. Non-blocking, but the `pins` step this PR just added is the natural home for it (`echo "rust-version=1.91.1" >> "$GITHUB_OUTPUT"`, then reference `steps.pins.outputs.rust-version` in both places).

Wotuu:

.github/actions/php-ci-setup/action.yml
Wotuu: :robot: Fixed in f23b12d1a.

.github/actions/php-ci-setup/action.yml
Wotuu: :robot: Fixed in f23b12d1a.

.github/actions/php-ci-setup/action.yml
Wotuu: :robot: Fixed in f23b12d1a.

Issue #2767, opened 2025-03-11
Undefined variable $activeExpansions


Wotuu:
This has been cropping up again by people requesting specific urls (malicious actors).

PR #3810, opened 2026-08-03
#3806 Route /ajax/ errors to JSON instead of crashing HTML error views

### LINKED ISSUE #3806: Error page for /ajax/* 404s crashes: undefined $currentUserGameVersion in header.blade.php
:robot: Filed from Sentry [PHP-LARAVEL-54](https://keystoneguru.sentry.io/issues/PHP-LARAVEL-54).

**3,047 occurrences, 13 users, first seen 2026-01-16, still firing** (last seen 2026-08-02). Highest-volume
unresolved issue in the project. Environment `production`, level `fatal`, unhandled.

```
Illuminate\View\ViewException: Undefined variable $currentUserGameVersion
(View: /var/www/html/resources/views/common/layout/header.blade.php)
```

Culprit and `transaction`: `/{fallbackPlaceholder}` — it crashes while rendering the **error page**, not
the originally requested page.

## Reproduce

```
GET /ajax/heatmap/data
```

with a browser-ish `Accept` header. That route is **POST-only** (`routes/web.php:592`), so a GET falls
through to the fallback, which raises a 405. Expect a rendered 405 page; get a 500 instead.

The sampled event came from `bingbot` (device family `Spider`), guest — this is mostly crawler traffic
hitting POST-only ajax endpoints with GET.

## Root cause

Two individually-sensible mechanisms that combine badly.

**1. On `/ajax/*`, no view composers are registered at all.**
`app/Providers/KeystoneGuruServiceProvider.php:310`

```php
if (!$viewService->shouldLoadViewVariables(request()->getPathInfo())) {
    return;
}
```

Every `view()->composer(...)` call sits *below* that early return, including the global one at line 326
that supplies the variable (`app/Http/View/Composers/GlobalComposer.php:38`, via
`RequestViewContext.php:45`). The blacklist is `/ajax/`, `/api/`, `/benchmark`
(`app/Service/View/ViewService.php:46`), with only `/ajax/dungeonroute/search`, `/ajax/search` and
`/ajax/view` whitelisted back in.

**2. But `/ajax/` errors are still rendered as HTML.**
`Handler::render()` only returns JSON when `$request->isJson()` **or** the path starts with `api/`
(`app/Exceptions/Handler.php:95`, `isApiRequest()` at `:143-146`). `/ajax/` is not covered, and a crawler
sends neither a JSON `Accept` nor `X-Requested-With`. So it renders
`errors/405.blade.php` → `errors/error-layout.blade.php` → `layouts.sitepage` → `common.layout.header`,
which dereferences the variable at `resources/views/common/layout/header.blade.php:35`:

```php
if ($currentUserGameVersion->key === GameVersion::GAME_VERSION_RETAIL) {
```

**The asymmetry is the bug:** `/ajax/` is blacklisted for view variables but not routed to the JSON error
branch, so any `/ajax/*` request ending in an HTML error page is guaranteed to 500.

## This is latent, not a regression

- The gate dates to `aaad686d8` (2025-03-11, #2767 — *"No longer receiving the `$activeExpansions` error
  when certain bogus URLs are called"*). Same class of bug, fixed then for a different variable by
  suppressing loading rather than making the error views resilient.
- `errors/405.blade.php` and the fallback's verb detection arrived in `4dd4327ce` (2026-07-09, #3453/#3482)
  — *after* first-seen, so 405 only added a second entry point. Before that the same crash almost
  certainly came through `errors/404.blade.php`, which extends the identical `error-layout.blade.php`.
- `GlobalComposer` was created in `9711f8d4f` (#3300/#3301) and inherited the problem.

## Fix directions

Not decided — worth weighing:

1. **Cover `/ajax/` in `Handler::isApiRequest()`** so ajax paths never render HTML error views. Narrowest
   fix and it matches the existing intent: these endpoints already return JSON in the success case. Needs
   a check that nothing relies on an HTML error body from `/ajax/` (the three whitelisted search paths
   render views deliberately).
2. **Make the header resilient**: `$currentUserGameVersion ??= GameVersion::getUserOrDefaultGameVersion();`
   — `resources/views/home/sections/featured.blade.php:10` already does exactly this. Blade-only, so
   **hotfixable** without a release. Papers over the layering problem but stops the bleeding.
3. **Register `GlobalComposer` unconditionally**, moving only the expensive per-view composers behind the
   gate. Broadest, and re-opens the performance concern #2767 was closing.

1 + 2 together is probably right: 2 now, 1 as the real fix.

## Related

- [PHP-LARAVEL-8J](https://keystoneguru.sentry.io/issues/PHP-LARAVEL-8J) (`DivisionByZeroError`) is also on
  `/ajax/heatmap/data` — worth a look while in the area, though a different defect.
- [PHP-LARAVEL-PC](https://keystoneguru.sentry.io/issues/PHP-LARAVEL-PC) /
  [PHP-LARAVEL-PD](https://keystoneguru.sentry.io/issues/PHP-LARAVEL-PD) on `/ajax/routes` have the same
  smell (crawler-shaped bad input) but different exceptions.

No user-identifying data from the Sentry events is reproduced here — the event's `user`, `server_name` and
`os.kernel_version` all carry host/IP values, deliberately omitted. See the linked Sentry issue for full
context.

Wotuu:
:robot: Closes #3806

## Summary
- `/ajax/` requests never get view composers registered (`ViewService::shouldLoadViewVariables()` blacklists them), but `Handler::render()` only routed `/api/` and explicit JSON requests to the JSON error branch — so any HTML error view rendered for an `/ajax/` request (e.g. a crawler GET to a POST-only route → 405) crashed on the resulting undefined view variables. This was the highest-volume unresolved Sentry issue in production (PHP-LARAVEL-54, 3k+ occurrences).
- Fix direction 1 from the issue: cover `/ajax/` in `Handler::isApiRequest()` so its errors always render JSON, matching how `/ajax/` already behaves in the success case.
- Fix direction 2 from the issue (defense in depth): `header.blade.php` now falls back `$currentUserGameVersion ??= GameVersion::getUserOrDefaultGameVersion();`, mirroring the existing pattern in `home/sections/featured.blade.php`.
- Updated `ErrorResponseTest` — the existing `fallback_givenGetRequestToDeleteOnlyRoute_returns405HtmlPage` test encoded the buggy behavior (a plain GET to an `/ajax/` POST-only route returning an HTML 405 page); it now asserts the JSON response instead.

Fix direction 3 (registering `GlobalComposer` unconditionally) was not taken — it re-opens the performance concern #2767 was closing, and isn't needed once `/ajax/` errors never try to render HTML at all.

## Test plan
- [x] `php artisan test --compact tests/Feature/Controller/ErrorResponseTest.php` — 5 passed
- [x] `composer run fix` — no changes
- [x] Reproduced the exact issue scenario against the running worktree stack:
  ```
  curl -A "Mozilla/5.0 (compatible; bingbot)" -H "Accept: text/html,application/xhtml+xml" http://localhost:8102/ajax/heatmap/data
  → 405 {"message":"Method Not Allowed"}
  ```

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -97,28 +97,50 @@ class Handler extends ExceptionHandler
     #[Override]
     public function render($request, Throwable $e)
     {
-        if ($request->isJson() || $this->isApiRequest($request)) {
-            if ($e instanceof ModelNotFoundException) {
+        if ($this->shouldReturnJson($request, $e)) {
+            if ($e instanceof ValidationException) {
+                // ValidationException doesn't render an HTML error view (parent::render() either
+                // returns JSON or redirects back with flashed session errors, per
+                // shouldReturnJson() below), so it was never part of the #3806 crash class - leave
+                // its normal expectsJson()-driven behavior alone rather than forcing JSON, since
+                // some /ajax/ forms rely on the redirect+flash for a non-JSON request.
+                return parent::render($request, $e);
+            } elseif ($e instanceof ModelNotFoundException) {
                 return response()->json([
                     'message' => __('exceptions.handler.api_model_not_found', [
                         'ids'   => implode(', ', $e->getIds()),
                         'model' => $e->getModel(),
                     ]),
                 ], StatusCode::NOT_FOUND);
-            } elseif ($e instanceof NotFoundHttpException) {
+            }
+
+            // Normalize the same way parent::render() would (AuthorizationException -> 403,
+            // TokenMismatchException -> 419, ...) before the instanceof checks below, so a
+            // forced-JSON /ajax/ request (which isn't necessarily an HttpException already) gets
+            // the exception's real status instead of collapsing into the generic 500 tail.
+            $normalized = $this->prepareException($this->mapException($e));
+
+            if ($normalized instanceof NotFoundHttpException) {
                 return response()->json(['message' => __('exceptions.handler.api_route_not_found')], StatusCode::NOT_FOUND);
-            } elseif ($e instanceof ThrottleRequestsException) {
+            } elseif ($normalized instanceof ThrottleRequestsException) {
                 return response()->json(['message' => __('exceptions.handler.too_many_requests')], RFC6585::TOO_MANY_REQUESTS);
-            } elseif ($e instanceof HttpExceptionInterface) {
-                // Preserve the real HTTP status (e.g. 405 Method Not Allowed) and any headers (such as
-                // the Allow header on a 405) instead of collapsing every remaining HttpException to a 500.
-                $statusCode = $e->getStatusCode();
+            } elseif ($normalized instanceof MethodNotAllowedHttpException) {
+                // The router's own message here is a verbose "The GET method is not supported for
+                // route ...", not meant for API consumers - use the plain status text instead, but
+                // still preserve the Allow header.
+                $statusCode = $normalized->getStatusCode();
 
                 return response()->json(
                     ['message' => SymfonyResponse::$statusTexts[$statusCode] ?? __('exceptions.handler.internal_server_error')],
                     $statusCode,
-                    $e->getHeaders(),
+                    $normalized->getHeaders(),
                 );
+            } elseif ($normalized instanceof HttpExceptionInterface) {
+                // Any other HttpException (e.g. a policy denial normalized above, or a deliberate
+                // abort($status, $message) elsewhere in the app) carries a message meant to be shown
+                // to the caller - delegate to parent::render() so it's preserved verbatim instead of
+                // being collapsed into a generic status text.
+                return parent::render($request, $e);
             } elseif (!config('app.debug')) {
                 return response()->json(['message' => __('exceptions.handler.internal_server_error')], StatusCode::INTERNAL_SERVER_ERROR);
             } elseif (config('app.type') !== 'local') {
@@ -138,7 +160,7 @@ class Handler extends ExceptionHandler
     #[Override]
     protected function unauthenticated($request, AuthenticationException $exception)
     {
-        if ($request->isJson() || $this->isApiRequest($request)) {
+        if ($this->shouldReturnJson($request, $exception)) {
             return response()->json(['error' => __('exceptions.handler.unauthenticated')], StatusCode::UNAUTHORIZED);
         }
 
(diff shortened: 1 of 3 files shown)


Wotuu:
:robot: Cold review: 4 findings

An independent cold-review agent reviewed the diff and found:
1. Widening `isApiRequest()` to `/ajax/` is shared with `unauthenticated()` — an unauthenticated `/ajax/` request now gets a JSON 401 instead of a login redirect. Documented in a docblock; no existing test asserted the old redirect behavior, so nothing to fix.
2. The three routes whitelisted in `ViewService::VIEW_VARIABLES_URL_WHITELIST` for deliberately rendering HTML on success (`/ajax/dungeonroute/search`, `/ajax/search`, `/ajax/view`) now also get JSON error bodies instead of HTML. Not a regression of the crash bug and very likely fine (ajax-only endpoints), documented in the same docblock.
3. `/benchmark` shares the same `ViewService` blacklist gate but isn't exposed to this crash class (doesn't render the shared header layout) — correctly left out of scope.
4. Missing regression test for a plain (non-JSON) GET to a non-existent `/ajax/` route (404) — the other common crawler shape from #3806. Added `fallback_givenPlainGetRequestToUnmatchedAjaxRoute_returns404Json`.

All addressed.

Wotuu:
:robot: Follow-up: the fix's shape changed after the cold review above.

Running the full test suite (not just the `Controller` group CI runs by filter) surfaced that widening `isApiRequest()` to `/ajax/` uniformly — as originally implemented — broke three real cases: `ValidationException` redirect+flash behavior on non-JSON ajax requests, `AuthorizationException`/`TokenMismatchException` collapsing to a generic 500 instead of their real 403/419 (since they aren't `HttpExceptionInterface` until Laravel's own exception normalization runs, which the override never invoked), and a deliberate `abort(422, $message)` policy message getting discarded in favor of a generic status text.

Reshaped the fix to centralize the /ajax/-vs-/api/ decision in a `shouldReturnJson()` override (Laravel's own hook, shared correctly by `render()`, `unauthenticated()`, and `ValidationException`'s redirect-vs-JSON choice), carve out `ValidationException` entirely (it never renders an HTML error view, so it was never part of the #3806 crash class), normalize exceptions the same way `parent::render()` would before checking status codes, and only override the message with a plain status text for the two framework exceptions whose own message isn't fit for API consumers (404, 405) — everything else delegates to `parent::render()` to preserve its real message/status/headers.

Full suite now passes: 1599 passed, 2 skipped (excluding the two CI-excluded groups). Finding 1 from the cold review above (the `unauthenticated()` side effect) still applies and is intentional/documented; findings 2-4 still stand as addressed.

Wotuu (commented):
Ok

Issue #3797, opened 2026-08-02
Split self-contained subsystems into separate Composer packages (assessment + roadmap)


Wotuu:
## Background

The combat log parsing layer was deliberately written to depend on nothing but itself, with the
intent of eventually shipping it as a standalone Composer package. This issue is the assessment of
what *else* in the codebase can be split off the same way, so each green-lit extraction can be
planned and shipped as its own issue.

Two motives:

1. **Enforce boundaries.** A package cannot accidentally reach for an Eloquent model.
2. **Publish reusable libraries.** These are intended to be independently consumable libraries that
   other people can depend on, with a release lifecycle decoupled from this repository.

A third motive — cutting test time — turned out to be weaker than expected. See the reality check
below; it is the single most important finding here and it should not be re-derived later.

## The discriminating criterion

`use App\` counts do not tell you whether something can be a package. What does is: **does it
hydrate Eloquent models?**

- Code that only transforms data (raw line → DTO) needs no database. Its tests run under plain
  PHPUnit with no Docker, no MySQL, no seed.
- Code that reads models needs the whole app. Packaging it buys a release runbook and nothing else.

That is exactly why `app/Logic/CombatLog` works as the flagship, and it is the axis everything below
is ranked on.

## Test-time reality check (measured)

CI run `30764969835`, PHPUnit **step** durations only (the rest of each ~3 min job is setup):

| shard | PHPUnit step |
|---|---|
| `unit` | 23s |
| `correct-events` | 31s |
| `combatlog-route` | 43s |
| `feature-controller` | 1m05 |
| `feature-other` | **4m42** ← critical path |

`php-tests` is already sharded 5 ways (`.github/workflows/php-tests.yml`), so wall clock is ~7 min
and is set entirely by `feature-other`. The pure-parser tests (34 files in
`tests/Unit/App/Logic/CombatLog`) live in the 23s `unit` shard.

**Extracting the parser therefore saves roughly zero CI wall-clock time.** What it does buy:

- Those 34 test files currently extend `Tests\TestCases\PublicTestCase` → `Tests\TestCase`, booting
  the full Laravel app for tests that only call `CombatLogStringParser::parseCombatLogLine()`. In a
  standalone package they run under bare PHPUnit.
- The package repo's own CI is `composer install && phpunit` — ~30s green instead of ~7 min.
- The boundary becomes enforced rather than conventional.

Also worth knowing: the ~2.5 min of per-shard setup is **mostly not the database**.
[...]

PR #3811, opened 2026-08-03
#3803 Move request DTOs out of App\Http into App\Dto, document the Stub repositories

### LINKED ISSUE #3803: Fix layering issues in app/Service/CombatLog: Http-namespaced Request and concrete repository imports
Found incidentally while assessing package split-offs (#3797). Neither of these blocks the combat
log parser extraction (#3798) — the parser itself is clean. They are in the layer that *consumes* it.

## 1. Service layer depends on an HTTP-namespaced model (28 imports)

`app/Service/CombatLog` imports `App\Http\Models\Request` in **28 files**.

This is a layering inversion: the service layer reaching upward into `App\Http`. The service layer
should not know that its caller arrived over HTTP. Practically, it means any non-HTTP caller — a
console command, a queued job, a test — has to construct an HTTP-namespaced object to talk to the
service.

`app/Http/Models` is an unusual directory in the first place (15 files): models living under the
HTTP namespace rather than `app/Models`. Worth understanding what it is for before deciding the fix,
since the answer may be that these objects belong somewhere neutral rather than that the service
layer should stop using them.

Likely shape of a fix: move the shared request/DTO types out of `App\Http` into a neutral namespace,
or have the service layer accept a plain DTO that the HTTP layer maps into.

## 2. Service layer imports concrete repository implementations instead of interfaces

`app/Service/CombatLog` imports 10 concrete repository classes rather than their interfaces:

- `App\Repositories\Swoole\Interfaces` (5)
- `App\Repositories\Stub\{KillZone,DungeonRoute}` (5)

The project's `repository-pattern` convention is to depend on `App\Repositories\Interfaces\*` and let
the container bind the implementation — which the rest of these same files already do correctly (78
interface imports across the directory). These 10 are the exception.

Depending on `Stub\*` in particular is worth a look: a stub implementation reached directly from
production service code is either a deliberate optimisation that should be documented, or a leak.
Establish which before changing it — do not assume it is a mistake.

## Scope note

Both are contained to `app/Service/CombatLog` and neither affects `app/Logic/CombatLog`, so #3798
can proceed independently. Deliberately filed separately rather than folded into the extraction, to
keep that PR's diff to the move itself.

## Key files

- `app/Service/CombatLog/**`
- `app/Http/Models/` (15 files)
- `app/Repositories/{Stub,Swoole}/`, `app/Repositories/Interfaces/`
- `.claude/skills/repository-pattern/SKILL.md` for the intended convention

## Definition of done

- The `App\Http\Models\Request` dependency is resolved, or a documented decision recorded on why it
  stays.
- The 10 concrete repository imports either become interface imports, or carry a comment explaining
  why the concrete class is required.
- `php artisan test --compact --group=CombatLog` green, plus `CombatLogRoute` and `CorrectEvents`.
- `composer run analyse` clean.

Wotuu:
:robot: Closes #3803

## 1. `App\Http\Models\Request` → `App\Dto\Request`

The whole `app/Http/Models` tree (15 DTOs + the `RequestModel` base) moves to `app/Dto`. Destination
picked by Wotuu over `App\Models\Request\**`; `app/Dto` is a new top-level folder, approved for this.

The base class had to move with the DTOs — it is extended by both the `CombatLog\Route` and `Route`
hierarchies, so leaving it under `App\Http` would have relocated the same upward dependency into the
base rather than removing it. That forces a single destination for the whole tree.

`config/l5-swagger.php` lists annotation scan paths explicitly and had `app/Http/Models` in both the
`public` and `internal_team` blocks — updated to `app/Dto`. Verified with `l5-swagger:generate --all`
that both spec files still contain the `CombatLogRoute*` / `RouteThumbnailRequest` schemas
(`api-docs.json`: 6 hits, `api-docs-internal-team.json`: 4). Without that config change the move
would have silently dropped every one of these schemas from the generated API docs while CI stayed
green.

**Correction to the issue's numbers:** it reads "28 imports in **28 files**". It is 28 import
*statements* across **5** files in `app/Service/CombatLog`. The full blast radius across the app is
29 files, which is what this diff touches.

### Scope: this removes the upward dependency, it does not make these plain DTOs

Worth stating so it doesn't read as an oversight. The issue floated two possible shapes — move the
types to a neutral namespace, *or* have the service layer accept a plain DTO. This PR does the
first. `CombatLogRouteRequestModel` still imports `DungeonRepositoryInterface`,
`DungeonRouteRepositoryInterface`, `DungeonRouteAffixGroupRepositoryInterface`,
`SeasonServiceInterface`, `SeasonAffixGroupServiceInterface` and
`App\Service\CombatLog\Exceptions\DungeonNotSupportedException`, and its `getDungeonRoute()` writes
through repositories — so `Service → Http` is gone, but `Dto → Service + Repositories` remains.

That is enough for this issue's DoD. It does matter for #3797 though: a standalone package cannot
ship a DTO that depends on the app's repositories, so flattening these into true DTOs is a
prerequisite there and belongs in that work, not here.

## 2. Concrete repository imports — half is a mischaracterisation, half is intentional

The issue lists 10 concrete repository imports. Splitting them:

- **`App\Repositories\Swoole\Interfaces\*` (5) are already interfaces.** They are bound in
  `OctaneServiceProvider` via `app()->instance(...)`. Nothing to fix; no change made.
- **`App\Repositories\Stub\*` (5) are deliberate and must stay concrete.**
  `convertCombatLogRouteToCombatLogEvents()` and `correctCombatLogRoute()` build a throwaway
  `DungeonRoute` purely as scaffolding so the builder can resolve enemies/floors, and the caller
  discards it. The builder writes through repositories, so the writing ones are passed as explicit
  `Stub\*` instances — resolving those interfaces from the container would hand back the persisting
  `Database\*` implementations and start creating rows on every API call.

Per the DoD's "or carry a comment explaining why the concrete class is required", both methods now
carry a docblock stating this, plus a note at the import block. The `$isSwoole ? ... : ...` ternary
is untouched — that is a deliberate Octane path and out of scope here.

Because a comment cannot be tested, the intent is also pinned by
`CombatLogRouteDungeonRouteServicePersistenceTest`: it runs both flows against a real fixture and
asserts `DungeonRoute`, `DungeonRouteAffixGroup`, `KillZone`, `KillZoneEnemy` and `KillZoneSpell`
row counts are unchanged. Mutation-checked — swapping two stubs for the injected repositories in
`correctCombatLogRoute()` makes it fail (`Column 'author_id' cannot be null` on a real
`insert into dungeon_routes`), so the test genuinely guards the behaviour.

## Docs

`.claude/skills/project-backend-structure` documented `app/Http/Models/Request/` as the DTO home in
three places — updated, including the `app/` directory map and a note that `config/l5-swagger.php`
must list `app/Dto` in its annotation paths.

## Verification

- `php artisan test --group=CombatLog --group=CombatLogRoute --group=CorrectEvents` — 301 passed, 1752 assertions
- `composer run analyse` — no errors
- `composer run fix` — clean
- `php artisan l5-swagger:generate --all` — both spec files regenerate with all schemas intact

No UI-visible change, so no visual verification.


Wotuu:

.claude/skills/project-backend-structure/SKILL.md
```diff
@@ -38,8 +42,8 @@ app/
 │                       (e.g. Heatmap.php, NpcCompendium.php). Pure class definitions.
 ├── Helpers             Global FUNCTION files (CustomHelper.php, ColorHelper.php), loaded via
 │                       require_once in HelperServiceProvider — NOT PSR-4 classes.
-├── Http                Controllers, FormRequests, request DTOs, middleware, resources
-│                       (see Http section).
+├── Http                Controllers, FormRequests, middleware, resources (see Http section).
+│                       Request DTOs live in app/Dto, not here.
```
Wotuu: Don’t mention this, this is already clear in the DTO section

app/Dto/Request/CombatLog/Route/CombatLogRouteChallengeModeRequestDto.php
```diff
@@ -1,8 +1,8 @@
 <?php
 
-namespace App\Http\Models\Request\CombatLog\Route;
+namespace App\Dto\Request\CombatLog\Route;
```
Wotuu, line 3: So I think all these classes that end with model can now end with DTO instead. I think that will be clearer for everyone involved just to be models or something that maps to the database one by one.

.claude/skills/project-backend-structure/SKILL.md
Wotuu: :robot: Fixed — removed the redundant "Request DTOs live in app/Dto, not here" line from the `app/` directory map under the `Http` entry (line 46). The FormRequests-vs-Request-DTOs section (`## Http layer`) already covers this in full.

app/Dto/Request/CombatLog/Route/CombatLogRouteChallengeModeRequestDto.php
Wotuu: :robot: Renamed all 15 classes this PR introduces under `app/Dto/Request/` from `*RequestModel` to `*RequestDTO`, including the base `RequestDTO.php` (was `RequestModel.php`): the base class plus `DungeonRouteThumbnailRequestDTO` and all 13 `CombatLogRoute*RequestDTO` classes. Updated every call site, `use` import, type hint, and docblock (`@method`/`@property`/`@mixin`) across controllers, services, builders, rules, resources, the console command and the test file. I left `APIFormRequest::getRequestModelClass()` (and its overrides in the other FormRequest classes) and its local `getModel()`/`$requestModelClass` alone — those are pre-existing method names on classes this PR barely touches (1-line import fixes), not part of the DTO class rename, and touching them would be a broader FormRequest-API naming change beyond this PR's scope. Happy to rename those too in a follow-up if you want full consistency there — let me know.

Wotuu: Yeah, I do want full consistency here so please rename those too

Wotuu:
Finally, can you rename all the DTO classes to Dto? Subtle difference but the casing I think is important to get right.

app/Dto/Request/CombatLog/Route/CombatLogRouteChallengeModeRequestDto.php
Wotuu: :robot: Done — renamed the FormRequest-side naming to match the `RequestDTO` convention:

- `APIFormRequest::getRequestModelClass()` (abstract) → `getRequestDTOClass()`
- `APIFormRequest::getModel()` → `getDTO()`
- Local `$requestModelClass` variable in `APIFormRequest::getDTO()` → `$requestDTOClass`
- Every override of the abstract method: `DungeonRouteRequest`, `DungeonRouteListRequest`, `DungeonRouteThumbnailRequest`, `APIOffsetPaginatedRequest`, `CombatLogRouteRequest` — all renamed `getRequestModelClass()` → `getRequestDTOClass()`
- `@method ... getModel()` docblocks on `CombatLogRouteRequest` and `DungeonRouteThumbnailRequest` → `@method ... getDTO()`
- The one call site, `APIDungeonRouteController::storeThumbnails()`, updated from `$request->getModel()` / `$model->...` to `$request->getDTO()` / `$dto->...`

Verified: `php artisan test --group=CombatLog --group=CombatLogRoute --group=CorrectEvents` (301 passed), `--filter=Thumbnail` (52 passed), `composer run fix` clean, `composer run analyse` no errors.

Wotuu:
:robot: Renamed the `DTO` casing to `Dto` across the whole `App\Dto\Request` family and every call site, per your request above.

**Base class**
- `App\Dto\Request\RequestDTO` → `App\Dto\Request\RequestDto` (file renamed too)

**Request DTOs** (all under `App\Dto\Request`, files renamed to match)
- `Route\DungeonRouteThumbnailRequestDTO` → `Route\DungeonRouteThumbnailRequestDto`
- `CombatLog\Route\CombatLogRouteRequestDTO` → `CombatLogRouteRequestDto`
- `CombatLog\Route\CombatLogRouteChallengeModeRequestDTO` → `CombatLogRouteChallengeModeRequestDto`
- `CombatLog\Route\CombatLogRouteCoordRequestDTO` → `CombatLogRouteCoordRequestDto`
- `CombatLog\Route\CombatLogRouteCorrectionRequestDTO` → `CombatLogRouteCorrectionRequestDto`
- `CombatLog\Route\CombatLogRouteMetadataRequestDTO` → `CombatLogRouteMetadataRequestDto`
- `CombatLog\Route\CombatLogRouteNpcRequestDTO` → `CombatLogRouteNpcRequestDto`
- `CombatLog\Route\CombatLogRouteNpcCorrectionRequestDTO` → `CombatLogRouteNpcCorrectionRequestDto`
- `CombatLog\Route\CombatLogRoutePlayerDeathRequestDTO` → `CombatLogRoutePlayerDeathRequestDto`
- `CombatLog\Route\CombatLogRoutePlayerDeathCorrectionRequestDTO` → `CombatLogRoutePlayerDeathCorrectionRequestDto`
- `CombatLog\Route\CombatLogRouteRosterRequestDTO` → `CombatLogRouteRosterRequestDto`
- `CombatLog\Route\CombatLogRouteSettingsRequestDTO` → `CombatLogRouteSettingsRequestDto`
- `CombatLog\Route\CombatLogRouteSpellRequestDTO` → `CombatLogRouteSpellRequestDto`
- `CombatLog\Route\CombatLogRouteSpellCorrectionRequestDTO` → `CombatLogRouteSpellCorrectionRequestDto`

**Methods / variables**
- `APIFormRequest::getRequestDTOClass()` → `getRequestDtoClass()`
- `APIFormRequest::getDTO()` → `getDto()`
- local variable `$requestDTOClass` → `$requestDtoClass` (in `APIFormRequest`)

Updated every call site referencing these: the FormRequests (`DungeonRouteRequest`, `DungeonRouteThumbnailRequest`, `DungeonRouteListRequest`, `CombatLogRouteRequest`, `APIOffsetPaginatedRequest`), controllers (`SiteController`, `APIDungeonRouteController`, `APICombatLogController`), the `CombatLogRouteNpcChronologicalRule`, the `CombatLogRouteCorrectionRequestResource` (`@property`/`@mixin` docblocks), `ChallengeModeRunDataService`, the `CombatLogRouteDungeonRouteService`/`...ServiceInterface` and their builders (`CombatLogRouteDungeonRouteBuilder`, `CombatLogRouteCorrectionBuilder`, `CombatLogRouteCombatLogEventsBuilder`), `EnemyKilled`, the `IngestCombatLogRouteJson` console command, the persistence feature test, and the `api-endpoint`/`project-backend-structure` skill docs.

Left untouched (pre-existing, unrelated generic uses of the word "DTO", not part of this class family): a comment in `CombatLogEventGridAggregationResult.php` ("used as a DTO to store...") and a comment in `DungeonRouteDiscoverController.php` ("WeeklyRoute DTOs").

**Verification**
- `php artisan test --group=CombatLog --group=CombatLogRoute --group=CorrectEvents --compact` → 301 passed
- `php artisan test --filter=Thumbnail --compact` → 52 passed
- Full suite `php artisan test --compact` → 1688 passed, 2 skipped, 1 failed (`MapTilesExistenceTest`, a pre-existing worktree-only asset-mount env issue unrelated to this change and excluded from CI)
- `composer run fix` → 0 files changed
- `composer run analyse` → no errors

Pushed as `189c4f6f0`.

Wotuu (commented):
Ok

PR #3812, opened 2026-08-03
Fix faction display control filtering only working in en_US

### LINKED ISSUE #3788: Faction display control filters on the translated faction name instead of its key, so it only works in en_US
:robot: Spun off from the cold review of #3787.

## Problem

The faction display control on the map (admin/edit mode, The Nexus) filters enemies by comparing a
**translated, lowercased faction name** against a **faction key**. It only works in `en_US`, by
coincidence.

`resources/assets/js/custom/mapcontrols/factiondisplaycontrols.js`:

```js
factionsData.push({
    name: lang.get(faction.name),
    name_lc: lang.get(faction.name).toLowerCase(),   // <- translated name, not the key
    ...
});
```

`resources/assets/js/handlebars/map_faction_display_controls_template.handlebars` binds that to the
element:

```handlebars
<a class="map_faction_display_control ..." data-faction="{{ name_lc }}" ...>
```

and the click handler compares it against a **key**:

```js
self._visibilityToggled(root.data('faction'), true);
// ...
if (mapObject.faction === faction && mapObject.faction !== 'any') { ... }
```

`mapObject.faction` holds `horde` / `alliance` / `any`. In `en_US`, `lang.get('factions.horde')`
returns `Horde`, so `name_lc` happens to equal `horde` and the comparison succeeds. In any other
locale it does not — and `lang/de_DE/factions.php`, `es_ES`, `ru_RU` etc. currently hold **empty
strings** for these keys, so `name_lc` becomes `""` and the control silently filters nothing at all.

The same mismatch means the initial state set on `map:mapobjectgroupsloaded` (which uses the literal
keys `'horde'` / `'alliance'`) does not correspond to the buttons the user then clicks.

## Fix

`name_lc` should come from the faction **key**, not the translated name:

```js
name_lc: faction.key,
```

The now-removed inline Blade copy of this template used `strtolower($faction->key)` — the correct
source — which is how this was noticed. It was dead code (nothing compiled inline
`x-handlebars-template` blocks any more), so the live JS path has always had this bug; #3787 did not
introduce it.

Consider renaming `name_lc` to `key` at the same time, since it no longer has anything to do with
the name.

## Scope / severity

Low: the control is only added in **edit mode** for **The Nexus**
(`resources/assets/js/custom/dungeonmap.js`, `resources/assets/js/custom/admin/admindungeonmap.js`,
both gated on `DUNGEON_THE_NEXUS`), so it affects route editing and admin mapping on one dungeon for
non-English users.

## Definition of done

- Clicking a faction button filters enemies correctly regardless of the active locale.
- A JS test covering the key-vs-name distinction, so an `en_US`-only run cannot pass it by
  coincidence.

Wotuu:
:robot: Closes #3788

## Summary
- `factiondisplaycontrols.js` built the faction data passed to the map control's Handlebars
  template with `name_lc` derived from `lang.get(faction.name).toLowerCase()` — the *translated*
  faction name — while the click handler compares that value against `mapObject.faction`, which
  holds the faction **key** (`horde`/`alliance`). This only worked in `en_US` by coincidence
  (`lang.get('factions.horde')` happens to equal `'horde'` lowercased); in other locales the
  faction translations are empty strings, so the filter silently matched nothing.
- Extracted the per-faction data transform into `FactionDisplayControls._buildFactionsData()` and
  changed it to use `faction.key` directly, renaming `name_lc` to `key` throughout (JS + the
  handlebars template's `data-faction` attribute), per the issue's suggested fix.

## Test plan
- [x] Added `factiondisplaycontrols.test.js`, asserting `_buildFactionsData` uses the faction key
  even when the mocked translation differs from it (simulating a non-en_US locale) — this test
  would have failed before the fix.
- [x] `npx vitest run` — all 352 JS tests pass.

### DIFF
--- a/resources/assets/js/custom/mapcontrols/factiondisplaycontrols.js
+++ b/resources/assets/js/custom/mapcontrols/factiondisplaycontrols.js
@@ -10,19 +10,8 @@ class FactionDisplayControls extends MapControl {
             onAdd: function (leafletMap) {
                 let template = Handlebars.templates['map_faction_display_controls_template'];
 
-                let factionsData = [];
                 let stateFactions = getState().getMapContext().getStaticFactions();
-                for (let index in stateFactions) {
-                    if (stateFactions.hasOwnProperty(index)) {
-                        let faction = stateFactions[index];
-                        factionsData.push({
-                            name: lang.get(faction.name),
-                            name_lc: lang.get(faction.name).toLowerCase(),
-                            icon_url: faction.icon_url,
-                            fa_class: parseInt(index) === 0 ? 'fas' : 'far'
-                        });
-                    }
-                }
+                let factionsData = self._buildFactionsData(stateFactions);
 
                 let data = $.extend({}, getHandlebarsDefaultVariables(), {factions: factionsData});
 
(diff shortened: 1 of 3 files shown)


Wotuu:
:robot: Cold review: 0 findings

Wotuu (commented):
Ok

PR #3813, opened 2026-08-03
#3789 Stop a failed segment download from being parsed as a combat log

### LINKED ISSUE #3789: Combat log parse failures: downloaded artifact is an XML document, not a combat log

**Unresolved failures in this cluster: 3** (count = unresolved `CombatLogParseFailure` rows visible at poll time — a later *decrease* means rows were resolved in the admin panel, not a regression).

All three fail on **line 1**, and the `rawLine` is the XML declaration itself — so the artifact that was handed to the parser does not appear to be a combat log at all, but an XML document. They span three different runs and two different `combatLogVersion` values, which argues for a recurring problem in the fetch/download path rather than a one-off.

### Example rows (verbatim)

```
id=745  run_id=41746652  season_id=17  combat_log_version=22012000007  line_number=1
message:   Unable to parse event <?xml version="1.0" encoding="UTF-8"?>
raw_line:  <?xml version="1.0" encoding="UTF-8"?>
```

```
id=21   run_id=40280113  season_id=17  combat_log_version=22012000005  line_number=1
message:   Unable to parse event <?xml version="1.0" encoding="UTF-8"?>
raw_line:  <?xml version="1.0" encoding="UTF-8"?>
```

(Third row: `id=744`, `run_id=41723757`, `season_id=17`, `combat_log_version=22012000007`, `line_number=1`, identical shape.)

### Hypothesis

Inferable from the message/raw line alone: the downloaded artifact is an XML document rather than a combat log — an error response from the storage/HTTP layer being written to disk and then parsed is the obvious candidate, but which one cannot be confirmed from the failure row. A follow-up session should pull the actual log segments for one of these runs (see the `combatlog-parse-failure-triage` skill) before assuming anything about the parser.

Wotuu:
:robot: Closes #3789

## The bug

`Curl::curlSaveToFile()` wrote whatever body came back from the request without ever inspecting the HTTP status. When a presigned S3 URL is expired or denied, S3 answers with an XML `Error` document — and that document was saved as the segment's `.txt` file and handed straight to the combat log parser.

It then failed on line 1 with `InvalidArgumentException: Unable to parse event <?xml version="1.0" encoding="UTF-8"?>` and was recorded as a `CombatLogParseFailure`. That is doubly wrong: it reads like a parser bug when it is really a download failure, and the run is never retried even though `ProcessCombatLogSegments` already re-throws `RuntimeException` precisely so the job can retry with fresh presigned URLs.

This matches all three failure rows in #3789 exactly — line 1, `rawLine` = the XML declaration, three different runs and two different `combatLogVersion` values.

## The fix

1. **`app/Service/Traits/Curl.php`** — `curlSaveToFile()` now returns `false` unless the request succeeded at the transport level (`curl_errno`) *and* answered 2xx. Nothing is written to disk otherwise, so the existing `RuntimeException` at `ProcessCombatLogSegments.php:88` fires and the job retries.
2. **`app/Jobs/CombatLog/ProcessCombatLogSegments.php`** — a second line of defence, `isPlausibleSegment()`: a downloaded segment is rejected if its first byte is `<`, or if it is zero bytes. Neither a combat log line (a timestamp) nor a zip archive (`PK`) can start with `<`, so an opening angle bracket means an error document — which covers the variant where a proxy or CDN in front of S3 returns an error body with a `200`, invisible to a status check. The zero-byte case is the same class of hole: an empty 200 body is a "successful" response by every measure curl reports, and would otherwise have been extracted as a silent no-op run reporting success. (A transfer truncated mid-flight is already covered — curl reports `CURLE_PARTIAL_FILE`.) Both throw `RuntimeException`, so they land on the retry path rather than recording a parse failure.

`curlGet()`'s public contract is deliberately unchanged — it has 14 other call sites (Wowhead, Cloudflare, Archon, Raider.IO, localization downloads) and widening this into a general HTTP-error overhaul is out of scope here.

### Consequences for the other `curlSaveToFile()` caller

`WowheadService::downloadSpellIcon()` (`WowheadService.php:126`) downloads spell icons through the same method. Previously a 404 HTML page from zamimg was silently saved as a `.jpg`; now nothing is written and the call returns `false`. Two knock-on effects worth stating plainly rather than calling this a pure improvement:

- `downloadMissingSpellIcons()` accumulates via `$result = $result && ...`, so `wowhead:fetchmissingspellicons` now exits non-zero if *any single* icon 404s, where before it exited 0.
- Because no garbage file is written, the `file_exists($targetFile)` skip no longer permanently short-circuits an icon that does not exist upstream — each such icon is re-requested (with its `sleep()`) on every invocation.

Both are the correct behaviour — a corrupt HTML-as-JPEG on disk being served to users is worse than either — and the command is manual, not scheduled, so the cost is bounded. Left as-is deliberately; widening this PR into a Wowhead retry/negative-cache change would be scope creep.

### Where this failure surfaces now

Worth being explicit, since the bogus `CombatLogParseFailure` row is exactly what surfaced #3789 in the first place via `combatlog-parse-failure-poll`. After this change the same failure no longer appears in the parse-failure admin panel or that poll — correctly, since it was never a parse failure. It surfaces instead as an `error`-level structured log (`handleSegmentDownloadFailed` / `handleSegmentIsNotACombatLog`), which `config/logging.php` routes to the Discord and Sentry channels at `level => error`, plus a `failed_jobs` row once the 3 retries are exhausted. So recurrence is still visible, just in error alerting rather than in the parse-failure tooling.

## Verification

- `tests/Unit/App/Service/Traits/CurlTest.php` (new) — asserts a 2xx response is written, and that a 403/404/500/302 response and a transport failure all return `false` **and leave no file on disk**.
- `tests/Feature/Jobs/CombatLog/ProcessCombatLogSegmentsTest.php` — new data-provider case asserting that an XML error document, an HTML error page, and an empty body downloaded with a "successful" status throws `RuntimeException`, never calls `extractData()`, and never records a `CombatLogParseFailure`.
- 16 tests / 79 assertions green (plus the full `--group=CombatLog` suite: 224 tests / 1095 assertions); `composer run fix` and `composer run analyse` clean.

**Why there is no real-data reproduction here**, unlike the usual `combatlog-parse-failure-triage` step 4: this failure is transient by nature. Pulling `/segments` for one of the three affected runs today issues *fresh* presigned URLs, which return 200 with real log data — the bug cannot be reproduced on demand against live segments. The tests above stub the error response directly, which is a stronger proof for this particular failure mode than a real-data run could be.

### DIFF
--- a/app/Jobs/CombatLog/ProcessCombatLogSegments.php
+++ b/app/Jobs/CombatLog/ProcessCombatLogSegments.php
@@ -94,6 +94,14 @@ class ProcessCombatLogSegments implements ShouldBeUnique, ShouldQueue
                 }
 
                 $tempFiles[] = $tempPath;
+
+                if ($this->isErrorDocument($tempPath)) {
+                    $log->handleSegmentIsErrorDocument($this->runId, $segment->id, $tempPath);
+
+                    throw new RuntimeException(
+                        sprintf('Downloaded segment %d for run %d is not a combat log', $segment->id, $this->runId),
+                    );
+                }
             }
 
             foreach ($tempFiles as $tempFile) {
@@ -142,6 +150,21 @@ class ProcessCombatLogSegments implements ShouldBeUnique, ShouldQueue
         $path = (string)parse_url($downloadUrl, PHP_URL_PATH);
 
         return str_ends_with($path, '.zip') ? 'zip' : 'txt';
+    }
+
+    /**
+     * A second line of defence against a storage error response reaching the parser: `curlSaveToFile()` already
+     * rejects a non-2xx download, but a proxy or CDN in front of S3 may hand back an error document with a 200.
+     * Neither a combat log (a timestamp) nor a zip archive (`PK`) can start with `<`, so an opening angle bracket
+     * means an XML or HTML document rather than the segment we asked for (see #3789).
+     */
+    private function isErrorDocument(string $filePath): bool
+    {
+        if (!is_file($filePath)) {
+            return false;
+        }
+
+        return file_get_contents($filePath, length: 1) === '<';
     }
 
     /**
--- a/app/Jobs/CombatLog/ProcessCombatLogSegments.php
+++ b/app/Jobs/CombatLog/ProcessCombatLogSegments.php
@@ -95,8 +95,8 @@ class ProcessCombatLogSegments implements ShouldBeUnique, ShouldQueue
 
                 $tempFiles[] = $tempPath;
 
-                if ($this->isErrorDocument($tempPath)) {
-                    $log->handleSegmentIsErrorDocument($this->runId, $segment->id, $tempPath);
+                if (!$this->isPlausibleSegment($tempPath)) {
+                    $log->handleSegmentIsNotACombatLog($this->runId, $segment->id, $tempPath);
 
                     throw new RuntimeException(
                         sprintf('Downloaded segment %d for run %d is not a combat log', $segment->id, $this->runId),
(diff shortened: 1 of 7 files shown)


Wotuu:

Dispatched to a fresh-context reviewer with no shared context with the implementation. All three were low severity; all three are addressed:

1. **Empty 200 body slipped through both guards** — real gap. An empty-but-valid response is not `false`, is 2xx, `file_put_contents` returns `0` (not `false`), and its first byte is not `<`, so a zero-byte segment reached the parser and the run reported success. Fixed in 6ea17ee83: `isErrorDocument()` → `isPlausibleSegment()`, which now also rejects a zero-byte file onto the same retry path. Covered by a new data-provider case.
2. **`WowheadService` consequences understated** — the original body called it "a strict improvement" without noting that `wowhead:fetchmissingspellicons` now exits non-zero if any single icon 404s, and that permanently-missing icons are re-requested every invocation instead of being short-circuited by a garbage file on disk. Both are correct behaviour and the command is manual, so left as-is; the PR body now states them plainly rather than glossing them.
3. **Observability trade-off unmentioned** — this failure class no longer appears in the parse-failure panel/poll that surfaced #3789. Correct (it was never a parse failure), and it still surfaces via `error`-level structured logs routed to Discord/Sentry plus a `failed_jobs` row after 3 retries. Now spelled out in the PR body.

Also confirmed by the reviewer: `$tempFiles[] = $tempPath` sits before the plausibility throw, so a rejected download is still unlinked in `finally`; and the new protected `curlGetResponse()` collides with nothing across the 10 classes that `use Curl;`.

Wotuu (commented):
Ok

Issue #745, opened 2021-04-16
Homepage should be an overview page for your account once you're logged in


Wotuu:
Relevant discussion: https://www.reddit.com/r/wow/comments/ms09xv/keystoneguru_m_routes_made_easy_redesign_many_new/gupvxo5/?utm_source=reddit&utm_medium=web2x&context=3

Wotuu:
This could also show a list of recently viewed routes. I have tjis info already.

Issue #744, opened 2021-04-16
Mobile view of Discovery page does not show subcreation ease tiers


Wotuu:
This is because the list is different in mobile and I didn't add the tiers there - my bad!

PR #3814, opened 2026-08-03
#3789 Validate HTTP status before writing curlSaveToFile() output

### LINKED ISSUE #3789: Combat log parse failures: downloaded artifact is an XML document, not a combat log

Wotuu:
:robot: Closes #3789

## Diagnosis

Pulled the real S3 segments for the three unresolved failure rows in the issue's cluster (#745, #21,
and the equivalent for #744) via the `combatlog-parse-failure-triage` skill. **All 9 segments for both
sampled runs (`41746652`, `40280113`) now download and parse cleanly with zero failures** — the
artifact that originally triggered `Unable to parse event <?xml version="1.0" ...` is no longer
reproducible from the stored S3 object, meaning the bad body was transient (an S3/HTTP error response
captured at ingestion time), not a stable, re-downloadable artifact. So this is diagnosed by
elimination from the code path rather than from a captured bad body.

Root cause: `Curl::curlSaveToFile()` (`app/Service/Traits/Curl.php`) wrote the raw curl response body
to disk **unconditionally** — no check of the HTTP status code or `curl_errno()`. `ProcessCombatLogSegments`
(the job that downloads combat log segments from their presigned S3 URLs) only treats a download as
failed when `file_put_contents()` itself fails, which it essentially never does. So an expired/invalid
presigned URL returning an S3 `<Error>` XML document (line 1) was written to the temp file as if it
were a valid segment and handed straight to the parser — producing a permanent
`InvalidArgumentException` `CombatLogParseFailure` row instead of tripping the job's existing
retry-with-fresh-URL path (`tries=3`, `backoff() = [30, 120]`, re-thrown `RuntimeException`).

## Fix

`curlSaveToFile()` now inspects the transport error (`curl_errno()`) and HTTP status
(`CURLINFO_HTTP_CODE`) before writing anything to disk, returning `false` on any non-2xx/transport
failure — which flows into the existing `RuntimeException`-retry path unchanged.
`curlGet()` keeps its prior behavior (returns whatever body it gets, even on non-2xx) since its ~8
other call sites already handle/ignore malformed bodies themselves and none of them were part of this
bug.

**Side effect worth flagging:** `WowheadService::downloadSpellIcon()` is the other `curlSaveToFile()`
caller. It previously always returned `true` (a 404 HTML page got silently written to the icon file
and reported as success); it will now correctly report `false` for a missing icon. This looks like a
pre-existing latent bug this fix also corrects, not a regression — noting it here for review.

**Consequence to be aware of:** a segment whose presigned URL still 403s after all 3 job attempts now
fails the job outright rather than ever reaching the parser/recording a `CombatLogParseFailure` row.
That's the intended behavior (nothing valid was ever parseable in that case).

## Testing

- New `tests/Feature/Service/Traits/CurlTest.php` exercises the real curl transport against a local
  PHP built-in server (started in `setUpBeforeClass`) rather than mocking curl, since the bug is
  specifically about status-code handling: 200 writes+returns true, 500/403 leave no file and return
  false, a refused connection returns false, and `curlGet()` still returns the raw body on 500 for
  backward compatibility.
- `php artisan test --group=CombatLog`: 221 passed.
- `composer run fix` / `composer run analyse`: clean.
- Real-data verification: `run_41746652` (9/9 segments) and `run_40280113` (9/9 segments) both
  re-parse with 0 failures via `CombatLogDataExtractionServiceInterface::extractData()`.

### DIFF
--- a/app/Service/Traits/Curl.php
+++ b/app/Service/Traits/Curl.php
@@ -9,6 +9,36 @@ trait Curl
      */
     public function curlGet(string $url, array $options = []): string
     {
+        return $this->curlExec($url, $options)['body'];
+    }
+
+    /**
+     * Downloads $url to $filePath, refusing to write the response body unless the request actually
+     * succeeded (transport-level success and a 2xx status). Without this, an S3/HTTP error body (e.g.
+     * an XML error document from an expired presigned URL) gets written to disk and silently treated
+     * as a valid combat log segment by the caller.
+     *
+     * @param  string $url
+     * @param  string $filePath
+     * @return bool
+     */
+    public function curlSaveToFile(string $url, string $filePath): bool
+    {
+        $result = $this->curlExec($url);
+
+        if ($result['errno'] !== 0 || $result['httpCode'] < 200 || $result['httpCode'] >= 300) {
+            return false;
+        }
+
+        return file_put_contents($filePath, $result['body']) !== false;
+    }
+
+    /**
+     * @param  array<string, mixed>                           $options
+     * @return array{body: string, httpCode: int, errno: int}
+     */
+    private function curlExec(string $url, array $options = []): array
+    {
         $ch = curl_init();
 
         curl_setopt_array($ch, $options + [
@@ -35,23 +65,17 @@ trait Curl
 
         try {
             $response = curl_exec($ch);
+            $errno    = curl_errno($ch);
+            $httpCode = (int)curl_getinfo($ch, CURLINFO_HTTP_CODE);
         } finally {
             curl_close($ch);
         }
 
-        return $response;
-    }
-
-    /**
-     * @param  string $url
-     * @param  string $filePath
-     * @return bool
-     */
-    public function curlSaveToFile(string $url, string $filePath): bool
-    {
-        $rawImage = $this->curlGet($url);
-
-        return file_put_contents($filePath, $rawImage) !== false;
+        return [
+            'body'     => $response !== false ? $response : '',
+            'httpCode' => $httpCode,
+            'errno'    => $errno,
+        ];
     }
 
     /**
--- a/tests/Feature/Service/Traits/CurlTest.php
+++ b/tests/Feature/Service/Traits/CurlTest.php
@@ -1,0 +1,164 @@
+<?php
+
+namespace Tests\Feature\Service\Traits;
+
+use App\Service\Traits\Curl;
+use PHPUnit\Framework\Attributes\Group;
+use PHPUnit\Framework\Attributes\Test;
+use Tests\TestCases\PublicTestCase;
+
+/**
+ * Exercises the real curl transport (via a local PHP built-in server) rather than mocking it, since the
+ * bug this covers (#3789) is specifically about curlSaveToFile() failing to notice a non-2xx response.
+ */
+#[Group('Service')]
+final class CurlTest extends PublicTestCase
+{
+    private static int $port;
+
+    /** @var resource|null */
+    private static $serverProcess = null;
+
+    private static string $routerPath;
+
+    public static function setUpBeforeClass(): void
+    {
+        parent::setUpBeforeClass();
+
+        self::$port       = random_int(20000, 29999);
+        self::$routerPath = sprintf('%s/curl_test_router_%s.php', sys_get_temp_dir(), uniqid());
+
+        file_put_contents(
+            self::$routerPath,
+            <<<'PHP'
+<?php
+$status = (int)($_GET['status'] ?? 200);
+http_response_code($status);
+echo $status >= 200 && $status < 300 ? 'OK-BODY' : '<?xml version="1.0" encoding="UTF-8"?><Error/>';
+PHP,
+        );
+
+        self::$serverProcess = proc_open(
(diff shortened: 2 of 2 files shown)


app/Service/Traits/Curl.php
```diff
@@ -8,6 +8,36 @@ trait Curl
...
+     *
+     * @param  string $url
+     * @param  string $filePath
+     * @return bool
+     */
+    public function curlSaveToFile(string $url, string $filePath): bool
+    {
+        $result = $this->curlExec($url);
+
+        if ($result['errno'] !== 0 || $result['httpCode'] < 200 || $result['httpCode'] >= 300) {
```
Wotuu, line 29: :robot: **Blocking (process, not code): PR #3813 is an open, in-flight fix for the same issue touching this same file.**

`#3813 Stop a failed segment download from being parsed as a combat log` (created 15:09:39Z, this PR 15:15:21Z — both still drafts) also says `Closes #3789` and also rewrites `curlSaveToFile()` here, extracting a private/protected `curlExec()`-style helper with the identical `curl_errno()` + 2xx check. It is a strict superset: same Curl change, plus a first-byte `<` sanity check in `ProcessCombatLogSegments`, a `handleSegmentIsErrorDocument()` logging method, and a job-level test.

Whichever merges second will conflict on `app/Service/Traits/Curl.php`, and #3789 would be closed twice. A human needs to reconcile the two before either merges — I'm not picking a winner here, just flagging that two parallel sessions solved the same issue.

Wotuu, line 29: :robot: The status check alone doesn't fully cover the failure this PR is meant to fix, and the PR body's own diagnosis says why: the bad body was never captured ("diagnosed by elimination"), so we don't actually know the S3 response was non-2xx. Two concrete gaps remain:

1. **2xx with an error body still slips through.** A proxy/CDN in front of S3 (or an S3 error surfaced with a 200) writes the XML document to disk exactly as before and hands it to the parser — the same `Unable to parse event <?xml version="1.0" ...` failure. PR #3813 independently reached this conclusion and added a first-byte `<` check in `ProcessCombatLogSegments`; that convergence is decent evidence a status check on its own isn't sufficient.
2. **A zero-byte 200 body returns `true`.** `file_put_contents($filePath, '')` returns `0`, and `0 !== false`, so an empty response is reported as a successful download and an empty segment file goes to the parser.

Suggest either rejecting an empty body here, or (better, since it also covers case 1) a content sanity-check at the call site in `ProcessCombatLogSegments` before `extractData()`.

app/Service/Traits/Curl.php
```diff
@@ -8,6 +8,36 @@ trait Curl
...
+        }
+
+        return file_put_contents($filePath, $result['body']) !== false;
+    }
+
+    /**
+     * @param  array<string, mixed>                           $options
+     * @return array{body: string, httpCode: int, errno: int}
+     */
+    private function curlExec(string $url, array $options = []): array
```
Wotuu, line 40: :robot: Class-member ordering: this diff inserts the new **private** `curlExec()` between the public `curlSaveToFile()` and the public `curlPost()`. Repo `CLAUDE.md` orders members public → protected → private, and the pre-diff file was correctly ordered (three public methods); this is drift introduced here, not pre-existing. `.php-cs-fixer.php` has no `ordered_class_elements` rule, so CI won't catch it. Moving `curlExec()` below `curlPost()` fixes it (that's where #3813 put its equivalent helper).

Wotuu:

Wotuu:
:robot: Closing as a duplicate of #3813, opened a few minutes earlier by another session against the same issue with a more complete fix (adds a second-layer `isPlausibleSegment()` check for a 200-status error body and zero-byte segments, plus more thorough tests and explicit handling of the `WowheadService::downloadSpellIcon()` side effect). No need to carry two competing fixes for the same root cause.

PR #3815, opened 2026-08-03
Retry combat log parse on a dropped Redis/Valkey connection instead of recording a permanent failure

### LINKED ISSUE #3791: Combat log parse aborted by Valkey read error mid-run (partially ingested run)

Wotuu:
:robot: Closes #3791

## Root cause

`CombatLogService::parseCombatLog()` (`app/Service/CombatLog/CombatLogService.php:402-405`) catches
`Exception` around the per-line parse callback and wraps *everything* into a `CombatLogParseException`
— including a `RedisException` thrown when the Redis/Valkey connection drops mid-parse (the exact
failure reported in #3791: a connection blip, not a malformed line — the `raw_line` recorded is
wherever the reader happened to be, not the actual cause).

`ProcessCombatLogSegments::handle()` (`app/Jobs/CombatLog/ProcessCombatLogSegments.php:107-126`) then
catches that wrapped exception in its generic `catch (Throwable $e)` clause, records a permanent
`CombatLogParseFailure` row, and lets the job complete **without rethrowing** — so `tries=3` /
`backoff([30,120])` never engage. A transient infra blip permanently and silently leaves the run
partially ingested.

**Answering the issue's own follow-up question directly**: run `41910839` was never retried — the
exception was caught, recorded, and the job returned normally (no rethrow), per the code path above.

## Fix

`RedisException` (which extends `RuntimeException`) is now rethrown unwrapped from
`parseCombatLog()` instead of being wrapped in `CombatLogParseException`. The job's existing
`catch (RuntimeException $e) { throw $e; }` clause — already used to retry download failures with
fresh presigned URLs — now also catches it (since it extends `RuntimeException`) and retries the job
instead of recording a permanent failure. No new catch clause was needed in the job; only its comment
was updated to document the shared path.

Scoped to `RedisException` specifically because that's the observed failure in #3791; the same
pattern would extend to other transient infra exceptions (e.g. DB connection drops) if those show up
too, but that's out of scope here.

**Retry safety**: `CombatLogDataExtractionService::extractData()` only writes the `ParsedCombatLog`
idempotency-guard row *after* a fully successful parse (`app/Service/CombatLog/CombatLogDataExtractionService.php:141-148`,
past the point where `parseCombatLog()`'s try/catch lives) — so a segment that died mid-parse leaves
no guard row, and a retry correctly re-parses just that segment. Already-successful segments in the
same run are skipped via the existing guard, and re-parsing a partially-processed segment is safe
regardless (observations are upserts; events only fire when canonical state actually changes).

**Observability trade**: after this change, a sustained Valkey outage means 3 failed attempts and
*no* `CombatLogParseFailure` row — the failures table stays a signal for genuine parser defects only,
not infra flakiness. Full outages are already visible via the job's own failure/retry-exhaustion
signals (Horizon).

The real-data reproduction step from the triage skill doesn't apply here — there is no failing input
line to reproduce (the raw_line in the issue is just parser position, not a parse defect), so this MR
is a code-path fix verified with unit/feature tests instead.

## Tests

- `CombatLogServiceTest::parseCombatLog_givenCallbackThrowsRedisException_rethrowsUnwrapped` — asserts
  a `RedisException` from the callback propagates unwrapped (not `CombatLogParseException`).
- `CombatLogServiceTest::parseCombatLog_givenCallbackThrowsOtherException_wrapsInCombatLogParseException` —
  regression guard that genuine parse errors are still wrapped so line/number info is preserved.
- `ProcessCombatLogSegmentsTest::handle_givenRedisExceptionDuringExtraction_throwsForRetryInsteadOfRecordingParseFailure` —
  asserts the job rethrows (for queue retry) and never calls `recordFailure`/`handleParseError`.

`--group=CombatLog` (225 tests) green; `composer run fix` / `composer run analyse` clean.

## Test plan

- [x] Unit test: `RedisException` rethrown unwrapped from `parseCombatLog()`
- [x] Unit test: other exceptions still wrapped in `CombatLogParseException`
- [x] Feature test: job rethrows on `RedisException` from `extractData()`, no parse failure recorded
- [x] Full `--group=CombatLog` suite green

### DIFF
--- a/app/Jobs/CombatLog/ProcessCombatLogSegments.php
+++ b/app/Jobs/CombatLog/ProcessCombatLogSegments.php
@@ -110,7 +110,11 @@ class ProcessCombatLogSegments implements ShouldBeUnique, ShouldQueue
 
             $result = true;
         } catch (RuntimeException $e) {
-            // Re-throw download failures so the job can be retried with fresh one-time URLs.
+            // Re-throw so the job is retried instead of recording a permanent CombatLogParseFailure: this
+            // covers download failures (fresh one-time URLs on retry) as well as RedisException, which
+            // extends RuntimeException and is rethrown unwrapped by CombatLogService::parseCombatLog() for
+            // the same reason - a dropped Redis/Valkey connection mid-parse is a transient infra blip, not
+            // an unparsable line (see #3791).
             throw $e;
         } catch (Throwable $e) {
             $log->handleParseError($this->runId, $this->combatLogVersion, $e->getMessage(), $e::class);
--- a/app/Service/CombatLog/CombatLogService.php
+++ b/app/Service/CombatLog/CombatLogService.php
@@ -31,6 +31,7 @@ use File;
 use Illuminate\Support\Collection;
 use Illuminate\Support\Str;
 use InvalidArgumentException;
+use RedisException;
 use Throwable;
 use ZipArchive;
 
@@ -399,6 +400,12 @@ readonly class CombatLogService implements CombatLogServiceInterface
                     $this->log->parseCombatLogParseEventsChangedCombatLogVersion($combatLogVersion, $isAdvancedLoggingEnabled);
                 }
             }
+        } catch (RedisException $exception) {
+            // A dropped Redis/Valkey connection is a transient infrastructure failure, not a malformed
+            // combat log line - the line the reader happened to be on when it hit is not the cause. Let it
+            // propagate unwrapped so the caller's retry logic (queue tries/backoff) applies instead of it
+            // being recorded as a permanent CombatLogParseFailure.
+            throw $exception;
         } catch (Exception $exception) {
             $this->log->parseCombatLogParseEventsException(sprintf('%d: %s', $lineNr, $rawEvent), $exception);
 
--- a/tests/Feature/Jobs/CombatLog/ProcessCombatLogSegmentsTest.php
+++ b/tests/Feature/Jobs/CombatLog/ProcessCombatLogSegmentsTest.php
@@ -19,6 +19,7 @@ use PHPUnit\Framework\Attributes\Group;
 use PHPUnit\Framework\Attributes\Test;
 use PHPUnit\Framework\Constraint\IsType;
 use PHPUnit\Framework\MockObject\Exception;
+use RedisException;
 use RuntimeException;
 use Tests\TestCases\PublicTestCase;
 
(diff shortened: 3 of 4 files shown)


Wotuu:

Verified independently rather than taken on trust:
- `RedisException` really does extend `RuntimeException` in this stack (checked `class_parents()` in the `app` container, phpredis 6.3.0), so the job's existing `catch (RuntimeException $e) { throw $e; }` genuinely picks it up — the fix hinges entirely on this.
- `predis` is not installed (`composer.json` requires `ext-redis`), so the `config/database.php` default of `'client' => env('REDIS_CLIENT', 'predis')` can't route around the new catch with a `Predis\Connection\ConnectionException`. The recorded exception class on #3791's row is `RedisException`, matching.
- No caller anywhere catches `CombatLogParseException` specifically (only `ProcessCombatLogSegments` type-checks it for field extraction), so letting `RedisException` escape unwrapped doesn't silently change behaviour for the splitters, `CombatLogMappingVersionService` or `extractDataAsync`. Neither the interface nor the impl docblock advertises a narrower `@throws` that would now be wrong.
- Retry safety holds beyond the `ParsedCombatLog` guard argument in the description: the only extractor that writes *during* the parse is `CreateMissingNpcDataExtractor`, and it is existence-guarded (`Npc::find()` before `Npc::create()`, with `NpcDungeon::create()` reachable only for a just-created NPC). All other extractors defer their writes to `afterExtract()`. A re-parse of a segment that died mid-flight won't duplicate rows.

Non-blocking scope note (not inline, the file isn't in the diff): `ProcessCombatLogFromS3::handle()` (dispatched by `ProcessCombatLogFanout`) has the same failure mode this MR fixes — a bare `catch (Throwable $e)` that logs `handleParseError` and returns normally, with no `tries`/`backoff` on the job at all. A Valkey blip on that ingestion path still silently drops a run. Out of scope here; may be worth a follow-up issue.

The "no `CombatLogParseFailure` row after retry exhaustion" trade-off and the deliberate scoping to `RedisException` are both argued in the MR body and I agree with both — keeping the failures table a signal for genuine parser defects is the right call given the poll/triage tooling consumes it.

Wotuu (commented):
Ok

PR #3816, opened 2026-08-03
#3804 Drop the doubled stage prefix from the SQS queue names

### LINKED ISSUE #3804: Drop the doubled stage prefix from the SQS queue names (infra step 2/3)
Step 2 of the three-deploy sequence tracked by [RaiderIO/keystoneguru-infra#16](https://github.com/RaiderIO/keystoneguru-infra/issues/16). This is the app-side half; it needs a release to reach staging/production.

## Background

Queue names are built here as `sprintf('%s-%s-<prefix>', config('app.type'), config('app.env'))`. `APP_TYPE` is the stage and `APP_ENV` is `production` in **both** AWS stages, so the deployed names carry the stage twice: `staging-production-default`, `production-production-default`.

The rename is an expand/contract across three deploys so no job is ever lost:

1. **infra — [keystoneguru-infra#47](https://github.com/RaiderIO/keystoneguru-infra/pull/47)** (open, draft): creates `<stage>-<prefix>` queues alongside the legacy ones; every worker runs `--queue=<new>,<legacy>` and drains both. IAM is granted on both ARNs for the workers *and* the queue writers (web/swoole/cron).
2. **this issue — app**: emit the new names.
3. **infra**: drop `<legacy>` from the `--queue` list and delete the old queues once drained.

**Step 1 must be deployed — not merely merged — before this ships.** Until those queues exist, dispatching to the new names fails.

Once step 1 is deployed this change is safe on its own: the workers keep draining the legacy queues, so anything already in flight is still picked up.

## Remaining work

Drop `config('app.env')` from the queue name in each of these. `sprintf('%s-<prefix>', config('app.type'))` — keep using `sprintf`, per project convention.

| File | Line | Queue |
|---|---|---|
| `app/Jobs/RefreshDiscoverCache.php` | 25 | `long-running` |
| `app/Jobs/RegenerateCombatLogRoute.php` | 27 | `long-running` |
| `app/Jobs/ProcessRouteFloorThumbnail.php` | 33 | `thumbnail` |
| `app/Jobs/ProcessRouteFloorThumbnailCustom.php` | 33 | `thumbnail-api` |
| `app/Jobs/CombatLog/ProcessCombatLogSegments.php` | 50 | `cl-process` |
| `app/Jobs/CombatLog/ProcessCombatLogFromS3.php` | 32 | `cl-process` |
| `app/Jobs/CombatLog/ProcessCombatLogFanout.php` | 27 | `cl-fanout` |
| `app/Http/Resources/DungeonRouteThumbnailJob/DungeonRouteThumbnailJobResource.php` | 45 | `thumbnail-api` (`Queue::size()`, not a dispatch) |
| `config/horizon.php` | 145-207 | all supervisors, both the `production` and `local` blocks |
| `.env.example` | 181 | `SQS_QUEUE` |

Two notes on the last three:

- `DungeonRouteThumbnailJobResource` **reads** the queue depth rather than dispatching, and it is the one place where a partial rename is user-visible: between this deploy and the legacy queues draining, it reports the new queue's depth while stragglers still sit on the old one. Rename it with the rest and accept the short window.
- `config/horizon.php` is redis/local only (AWS uses SQS), but its supervisor queue names must keep matching what the jobs emit or local queue processing silently stops. Note the existing `supervisor-combat-log-process` entry watches `%s-local-combat-log-process`, which no job has ever dispatched to — pre-existing, worth fixing or deleting while in there.

## Decided: `SQS_QUEUE` stays a manual edit

Jobs that never set `$this->queue` (broadcasts, notifications, anything queued but unrouted) fall back to the SQS connection's default queue, which is `env('SQS_QUEUE')`. That value lives in each stage's `.env` **in the S3 config bucket** — not in this repo and not in keystoneguru-infra — so it rides neither deploy.

We considered deriving it in `config/queue.php` instead (`sprintf('%s-default', env('APP_TYPE'))`) so it would ship with the release. **Rejected**: `config/queue.php` currently defines only `failed` and leans on Laravel's base config, so adding a `connections.sqs` block means reproducing the framework's SQS defaults in-repo and keeping them in sync across Laravel upgrades. Not worth it for a one-time edit.

So: **after this ships, update `SQS_QUEUE` from `<stage>-production-default` to `<stage>-default` in the staging and production `.env` in the config bucket, and roll the services.** This is a precondition for infra step 3 — deleting the old queues before it is done silently drops every unrouted job. It is recorded as the first item of step 3's checklist on keystoneguru-infra#16.

## Verify

- `php artisan tinker` in the dev stack: dispatch a thumbnail job and confirm the queue name is `local-thumbnail`, not `local-local-thumbnail`.
- After the release reaches staging: dispatch a thumbnail and confirm it lands on `staging-thumbnail` (`aws sqs get-queue-attributes`), and that the worker logs picking it up.
- Confirm the legacy queues drain to zero — that is the gate for infra step 3.

## Definition of done

- All ten sites above renamed; no remaining `config('app.env')` in a queue name (`grep -rn "app.env" app/ config/`).
- Tests green, `composer run fix` and `composer run analyse` clean.
- Shipped in a release, and the `SQS_QUEUE` edit above done in both stages afterwards.

Wotuu:
:robot: Closes #3804

Step 2/3 of the queue rename tracked by RaiderIO/keystoneguru-infra#16.

## ✅ Infra step 1 is deployed — this is unblocked

RaiderIO/keystoneguru-infra#47 merged at 20:41 and its "Deploy infra on merge" run succeeded, creating all 12 `<stage>-<prefix>` queues (staging + production × default/long-running/thumbnail/thumbnail-api/cl-fanout/cl-process). The workers now run `--queue=<new>,<legacy>` and keep draining anything already in flight, so this change is safe on its own.

*(The original body warned this must not ship first; that warning is resolved and kept here only for the record.)*

## What changed

`sprintf('%s-%s-<prefix>', config('app.type'), config('app.env'))` → `sprintf('%s-<prefix>', config('app.type'))`. `APP_TYPE` is the stage and `APP_ENV` is `production` in both AWS stages, so the deployed names carried the stage twice (`staging-production-default`).

All ten sites from the issue: the 7 jobs, `DungeonRouteThumbnailJobResource` (`Queue::size()`), `config/horizon.php` (both the `production` and `local` blocks, including the commented-out thumbnail supervisors), and `.env.example`'s `SQS_QUEUE`.

## Deliberate extra: the dead Horizon combat-log supervisor

The issue flagged `supervisor-combat-log-process` as watching `<type>-local-combat-log-process`, a queue **no job has ever dispatched to** — meaning combat log jobs were never processed locally at all — and said it was "worth fixing or deleting while in there".

Fixed rather than deleted, and this goes one step past the issue text: renaming it to `cl-process` alone would still leave local combat log processing broken, because `ProcessCombatLogFanout` dispatches to `cl-fanout` and nothing consumed that either. So the entry is renamed to `supervisor-cl-process` (queue `<type>-cl-process`) **and** a `supervisor-cl-fanout` entry was added, matching the two workers infra already runs in AWS. Local-only (`config/horizon.php` is redis; AWS uses SQS); the effect is one extra local Horizon process. Say the word if you'd rather have the plain rename and a separate issue for the fanout supervisor.

## Two short user-visible windows after the release

Both resolve as soon as the legacy queues drain:

- `DungeonRouteThumbnailJobResource` reports the *new* queue's depth while stragglers still sit on the old one (called out in the issue).
- `QueueSize.php` telemetry iterates `horizon.environments.<app.env>` and calls `Queue::size()` on each name, so the graphs for `production-default` / `production-long-running` will read near-zero briefly — a metrics dip, not an outage. Worth not chasing.

## Still manual after this ships

`SQS_QUEUE` (the fallback queue for jobs that never set `$this->queue` — broadcasts, notifications) lives in each stage's `.env` in the S3 config bucket, so it rides neither deploy. After the release: change it from `<stage>-production-default` to `<stage>-default` in staging and production and roll the services. **This is a precondition for infra step 3** — deleting the old queues before it is done silently drops every unrouted job. Recorded on keystoneguru-infra#16.

## Verification

- `grep -rn -- "-default\|-long-running\|-thumbnail\|-cl-process\|-cl-fanout" app/ config/ tests/ sh/ .env.example` is clean of doubled names. (Note the issue's `grep app.env` would *not* have caught `config/horizon.php`, which spelled the stage out literally.) The sibling `keystone.guru.combatlogstreamer` repo constructs no queue names.
- Dev stack tinker: `(new ProcessRouteFloorThumbnail(...))->queue` → `local-thumbnail`, not `local-local-thumbnail`.
- New `tests/Feature/Jobs/JobQueueNameTest.php`: asserts the queue name of all 7 jobs under a fixed `app.type`/`app.env`, and asserts no Horizon supervisor (either environment) watches a queue no job dispatches to — the drift that produced the dead combat-log entry. Negative-checked: reverting the supervisor to its old name makes that test fail.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -178,7 +178,7 @@ AWS_S3_BUCKET_COMBAT_LOGS=
 AWS_USE_PATH_STYLE_ENDPOINT=false
 
 SQS_PREFIX=https://sqs.us-east-1.amazonaws.com/your-account-id
-SQS_QUEUE=production-production-default
+SQS_QUEUE=production-default
 
 SESSION_ENCRYPT=false
 SESSION_PATH=/
--- a/tests/Feature/Jobs/JobQueueNameTest.php
+++ b/tests/Feature/Jobs/JobQueueNameTest.php
@@ -1,0 +1,126 @@
+<?php
+
+namespace Tests\Feature\Jobs;
+
+use App\Jobs\CombatLog\ProcessCombatLogFanout;
+use App\Jobs\CombatLog\ProcessCombatLogFromS3;
+use App\Jobs\CombatLog\ProcessCombatLogSegments;
+use App\Jobs\ProcessRouteFloorThumbnail;
+use App\Jobs\ProcessRouteFloorThumbnailCustom;
+use App\Jobs\RefreshDiscoverCache;
+use App\Jobs\RegenerateCombatLogRoute;
+use App\Models\DungeonRoute\DungeonRoute;
+use App\Models\DungeonRoute\DungeonRouteThumbnailJob;
+use App\Models\Season;
+use Illuminate\Support\Facades\Config;
+use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\Attributes\Group;
+use PHPUnit\Framework\Attributes\Test;
+use Tests\TestCases\PublicTestCase;
+
+#[Group('Jobs')]
+final class JobQueueNameTest extends PublicTestCase
+{
+    private const string APP_TYPE = 'teststage';
+
+    /**
+     * The queue name must carry the stage (APP_TYPE) exactly once - APP_ENV is 'production' in both AWS stages,
+     * so including it produced doubled names such as 'staging-production-thumbnail'. See #3804.
+     *
+     * @param callable(): object $jobFactory
+     */
+    #[Test]
+    #[DataProvider('queueNameProvider')]
+    public function construct_givenAppType_setsQueueWithoutAppEnv(callable $jobFactory, string $expectedQueue): void
+    {
+        // Arrange
+        Config::set('app.type', self::APP_TYPE);
+        Config::set('app.env', 'production');
+
+        // Act
+        $job = $jobFactory();
+
+        // Assert
+        $this->assertSame($expectedQueue, $job->queue);
+    }
+
+    /**
+     * @return array<string, array{0: callable(): object, 1: string}>
+     */
+    public static function queueNameProvider(): array
+    {
+        return [
+            'RefreshDiscoverCache' => [
+                static fn() => new RefreshDiscoverCache(),
+                sprintf('%s-long-running', self::APP_TYPE),
+            ],
+            'RegenerateCombatLogRoute' => [
+                static fn() => new RegenerateCombatLogRoute(1),
+                sprintf('%s-long-running', self::APP_TYPE),
+            ],
+            'ProcessRouteFloorThumbnail' => [
+                static fn() => new ProcessRouteFloorThumbnail(new DungeonRoute(), 1),
+                sprintf('%s-thumbnail', self::APP_TYPE),
+            ],
+            'ProcessRouteFloorThumbnailCustom' => [
+                static fn() => new ProcessRouteFloorThumbnailCustom(new DungeonRouteThumbnailJob(), new DungeonRoute(), 1),
+                sprintf('%s-thumbnail-api', self::APP_TYPE),
+            ],
+            'ProcessCombatLogSegments' => [
+                static fn() => new ProcessCombatLogSegments(new Season(), 1, 1),
+                sprintf('%s-cl-process', self::APP_TYPE),
+            ],
+            'ProcessCombatLogFromS3' => [
+                static fn() => new ProcessCombatLogFromS3('bucket', 'path.log.zip', 1),
+                sprintf('%s-cl-process', self::APP_TYPE),
+            ],
+            'ProcessCombatLogFanout' => [
+                static fn() => new ProcessCombatLogFanout('bucket', 'path/', 1),
+                sprintf('%s-cl-fanout', self::APP_TYPE),
+            ],
+        ];
+    }
+
+    /**
+     * Horizon drives local queue processing; if a supervisor watches a queue no job dispatches to, that job type
+     * silently stops being processed locally (which is what happened to 'local-combat-log-process' before #3804).
+     */
+    #[Test]
+    #[DataProvider('horizonEnvironmentProvider')]
+    public function horizonConfig_givenEnvironment_onlyWatchesQueuesJobsDispatchTo(string $environment): void
+    {
+        // Arrange
+        // 'default' is the connection's fallback queue for jobs that never set $this->queue
+        $knownSuffixes = ['default'];
+        foreach (self::queueNameProvider() as [1 => $expectedQueue]) {
+            $knownSuffixes[] = substr($expectedQueue, strlen(self::APP_TYPE) + 1);
+        }
+
+        // Act
+        $supervisors = config(sprintf('horizon.environments.%s', $environment));
+
+        // Assert
+        $this->assertNotEmpty($supervisors);
+        foreach ($supervisors as $supervisorName => $supervisorConfig) {
+            foreach ($supervisorConfig['queue'] as $queueName) {
+                // Everything past the APP_TYPE prefix must be a queue some job actually dispatches to
+                $this->assertContains(
+                    substr($queueName, strpos($queueName, '-') + 1),
+                    $knownSuffixes,
+                    sprintf('Horizon supervisor "%s" watches a queue no job dispatches to', $supervisorName),
+                );
+            }
+        }
+    }
+
+    /**
+     * @return array<string, array{0: string}>
+     */
+    public static function horizonEnvironmentProvider(): array
+    {
+        return [
+            'production' => ['production'],
+            'local'      => ['local'],
+        ];
+    }
+}
--- a/tests/Feature/Jobs/JobQueueNameTest.php
+++ b/tests/Feature/Jobs/JobQueueNameTest.php
@@ -90,11 +90,7 @@ final class JobQueueNameTest extends PublicTestCase
     public function horizonConfig_givenEnvironment_onlyWatchesQueuesJobsDispatchTo(string $environment): void
     {
         // Arrange
-        // 'default' is the connection's fallback queue for jobs that never set $this->queue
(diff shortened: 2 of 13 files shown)


tests/Feature/Jobs/JobQueueNameTest.php
```diff
@@ -0,0 +1,126 @@
...
+        $knownSuffixes = ['default'];
+        foreach (self::queueNameProvider() as [1 => $expectedQueue]) {
+            $knownSuffixes[] = substr($expectedQueue, strlen(self::APP_TYPE) + 1);
+        }
+
+        // Act
+        $supervisors = config(sprintf('horizon.environments.%s', $environment));
+
+        // Assert
+        $this->assertNotEmpty($supervisors);
```
Wotuu, line 99: :robot: The new drift test only asserts one direction — supervisor → job ("no supervisor watches a queue no job dispatches to"). But the bug this PR actually fixed in the `local` block was the *opposite* direction: `cl-fanout` had **no supervisor at all**, so those jobs were never processed locally. That regression class is still uncovered — delete `supervisor-cl-fanout` from `config/horizon.php` again and this test stays green.

Suggest adding a reverse assertion, **scoped to `local` only**: every suffix produced by `queueNameProvider()` must be watched by some `local` supervisor. It has to stay local-only because `production` deliberately omits the thumbnail supervisors (commented out) and has no `cl-*` entries — AWS runs SQS + ECS workers rather than Horizon — so a symmetric check across both environments would fail on `production`.

Negative-checked against the config: the six `local` supervisors after this PR cover all five job suffixes (`long-running`, `thumbnail`, `thumbnail-api`, `cl-process`, `cl-fanout`) plus `default`, so the assertion passes today; before this PR neither `cl-fanout` nor `cl-process` was watched, so it would have failed — which is exactly the guard the test is meant to provide.

.env.example
```diff
@@ -178,7 +178,7 @@ AWS_S3_BUCKET_COMBAT_LOGS=
 AWS_USE_PATH_STYLE_ENDPOINT=false
 
 SQS_PREFIX=https://sqs.us-east-1.amazonaws.com/your-account-id
-SQS_QUEUE=production-production-default
+SQS_QUEUE=production-default
```
Wotuu, line 181: :robot: Low severity, but the rename missed a sibling file: `.env.ci.example:179` still reads `SQS_QUEUE=production-production-default`. Functionally inert (CI runs `QUEUE_CONNECTION=redis`, so the SQS default queue is never resolved), but it's now the only place in the repo carrying a doubled name, and it will read as authoritative to the next person who copies it.

Worth noting *why* it slipped: the verification grep in the PR body enumerates paths (`app/ config/ tests/ sh/ .env.example`) rather than grepping the repo, so the sibling env file was outside the search by construction. A repo-wide `grep -rIn "production-production-\|-local-\|local-local-"` catches it.

Wotuu:

1. `tests/Feature/Jobs/JobQueueNameTest.php` — the drift test only guards supervisor → job, not job → supervisor, so the missing-`cl-fanout`-supervisor half of the fix is uncovered.
2. `.env.ci.example:181` — missed rename site (`SQS_QUEUE=production-production-default`), functionally inert but the last doubled name in the repo. Commented inline on `.env.example` since the CI file is not in the diff.

Nit, not posted inline (file not in diff): `.claude/skills/combatlog-data-pipeline/SKILL.md` lines 55 and 84 document the queue as `*-combat-log-process`. That string now exists nowhere in the repo (the jobs already dispatched to `cl-process` before this PR; this PR removes the last matching supervisor name), so the skill is fully stale on that point.

Checked and deliberately not raised: the `Queue::size()` round-trip in `DungeonRouteThumbnailJobResource::toArray()` (pre-existing, unchanged by the rename), `QueueSize.php` telemetry throwing if the new queues do not exist yet (covered by the documented infra-step-1-deployed precondition), and `env('APP_TYPE')` lacking the `'local'` default that `config/app.php` has (pre-existing, and both `.env.example` and `.env.ci.example` ship `APP_TYPE=local`). Ran `JobQueueNameTest` in the worktree: 9 passed.

tests/Feature/Jobs/JobQueueNameTest.php
Wotuu: :robot: Fixed in 0d025f9. Added `horizonConfig_givenLocalEnvironment_watchesEveryQueueJobsDispatchTo` — the mirror assertion, scoped to `local` for exactly the reason you gave (the `production` block is deliberately partial since AWS runs its workers as ECS services against SQS). Negative-checked: removing the `supervisor-cl-fanout` entry now fails it.

.env.example
Wotuu: :robot: Fixed in 0d025f9 — `.env.ci.example` is now `production-default`. Also fixed the skill nit from your summary: `combatlog-data-pipeline/SKILL.md` was doubly wrong there, since it attributed `ProcessCombatLogFanout` to the process queue rather than `cl-fanout`.

Wotuu (commented):
Ok

Wotuu:
:robot: All 3 cold-review findings fixed in `0d025f9` (reverse Horizon guard, `.env.ci.example`, stale skill doc); both threads replied to and resolved. All 9 checks green.

Confirmed the infra precondition is now satisfied — RaiderIO/keystoneguru-infra#47 merged at 20:41 and its deploy run created all 12 `<stage>-<prefix>` queues, so the "do not ship before step 1" warning in the original body is resolved and the body has been updated accordingly. Marking ready for review.

No UI surface in this diff, so no visual verification. **Reminder for after the release:** `SQS_QUEUE` still needs the manual `<stage>-production-default` → `<stage>-default` edit in both stages' S3 `.env` — that is the precondition for infra step 3.

Issue #3824, opened 2026-08-03
Sentry migration 5/5: Sentry project settings needed to back the error-triage loop


Wotuu:
:robot: Phase 4 of moving combat log parse failures onto Sentry (#3820, #3821, #3822, #3823). These are Sentry org/project settings, not code — **no agent can do them**, but each one decides whether the code phases actually hold up.

## Checklist

- [ ] **Enable spike protection** on the `php-laravel` project. The app-side throttle in #3820 is the first line of defence; this is the backstop. On a 30-day/Developer plan a WoW patch that breaks parsing could otherwise burn the monthly quota in hours and drop all other errors with it.
- [ ] **Consider a per-key rate limit** on the DSN as a second backstop.
- [ ] **Confirm the GitHub integration with commit tracking is installed.** This is what makes `Fixes <SHORT-ID>` auto-resolve an issue when the fix deploys. If it is not installed, the explicit `resolvedInNextRelease` call from #3823 carries the whole load.
- [ ] **Confirm production has a release stream.** All 204 production events over the last 7 days carry `release: null` — the `version`-file derivation added by #3794 has only been observed on `local` so far. `resolvedInNextRelease` anchors to the next release Sentry *sees*; with no release stream it degrades to a plain `resolved`, which quietly removes the regression-detection safety net the new resolve policy leans on.
- [ ] **Consider an inbound filter on `environment:local`** as a backstop to the code fix in #3820. Over the last 7 days `local` accounted for 108 events and `testing` 41, against 204 from production — 42% of volume was non-production noise.
- [ ] **Clear the DSN from any local `.env`** that still has one, on machines that generate `environment: local` events. #3820 ships the example files empty, but an existing `.env` is not touched by that.

## Related

Whether structured errors reach Sentry in production at all depends on `LOG_CHANNEL`/`LOG_STACK`, tracked separately in #3805.

---

### Sentry migration stages

Moving combat log parse failures off their bespoke DB pipeline and onto Sentry, so there is one error feed and one triage skill instead of two of each.

| Stage | Issue | Ships |
|---|---|---|
| 1/5 | #3820 | Fingerprint, tag and throttle structured errors; stop dev/test noise reaching the project |
| 2/5 | #3821 | Run-keyed segments endpoint; remove the parse-failure API, admin page and writes |
| 3/5 | #3822 | Drop the table (contract half — one release after 2/5) |
| 4/5 | #3823 | Collapse the three triage skills into one; new resolve policy |
[...]

Issue #3823, opened 2026-08-03
Sentry migration 4/5: unify error triage into one skill and let the agent resolve what it fixed


Wotuu:
:robot: Phase 3 of moving combat log parse failures onto Sentry. Docs-only; can be worked in parallel with #3820, but should land after it so the runbook describes shipped behaviour.

There are three overlapping skills today: `sentry-error-triage`, `combatlog-parse-failure-poll` and `combatlog-parse-failure-triage`. Wotuu wants **one skill that reads Sentry and fixes whatever comes its way**, combat log or otherwise. The right reading is a single *entry point*, not a single file — the repo already works this way, since `sentry-error-triage` delegates per-issue debugging to `sentry:sentry-debug-issue` rather than restating it.

| Skill | Fate |
|---|---|
| `sentry-error-triage` | **The one loop.** Already has the better bones: stable `<!-- sentry-issue: SHORT-ID -->` dedupe key, security allowlist, quiet hours, bounded fix phase, boundaries section. |
| `combatlog-parse-failure-poll` | **Deleted.** Its bespoke sha1 signature and digit-normalising clustering become redundant once the fingerprint is computed server-side by #3820 — the Sentry short ID *is* the stable key. |
| `combatlog-parse-failure-triage` | **Kept as a referenced runbook**, renamed `combatlog-parse-failure-reproduce` so "triage" means one thing. Its presigned-URL / gzip / `docker cp` / tinker recipe is domain knowledge; inlining it would bloat every non-combat-log pass. |

## Changes to `sentry-error-triage`

1. **Invert its own description** — it currently says "not for combat log parse failures (combatlog-parse-failure-poll)", which will point at a deleted skill and state the opposite of the new truth.
2. **Rewrite the resolve policy.** Hard rule #2 is currently "Never resolve or ignore a Sentry issue by hand". Wotuu's instruction overrides it: *"you manage it, you fix it, you resolve it. Leave me out of it ideally."* The safe mechanism is not blanket resolution:
   - After a fix PR **merges**, set `status='resolvedInNextRelease'` with a `reason` comment. If it recurs after that release Sentry marks it a **regression** and reopens it, so the "is it still happening" signal the old rule protected is preserved by Sentry itself rather than by refusing to act.
   - For genuine noise (scanner traffic, third-party outages), `status='ignored'` with `ignoreMode='untilEscalating'` — self-managing, resurfaces if volume changes.
[...]

Issue #3822, opened 2026-08-03
Sentry migration 3/5: drop the combat_log_parse_failures table and its model


Wotuu:
:robot: Contract half of the expand/contract started in #3821. **Must not ship in the same release as #3821** — per `.claude/CLAUDE.md`, destructive schema changes wait a full release after the last code consumer is removed, or still-running old containers 500 against the missing schema (this is what broke staging in #3497).

Once #3821 has shipped and deployed:

- Add the drop migration for `combat_log_parse_failures` under `database/migrations_combatlog/` (the table lives on the `combatlog` connection).
- Delete `app/Models/CombatLog/CombatLogParseFailure.php`.
- Delete `database/factories/CombatLog/CombatLogParseFailureFactory.php`.
- Delete `tests/Feature/App/Repository/CombatLog/CombatLogParseFailureRepositoryTest.php` and any remaining fixtures referencing the model.

Rows are discarded deliberately — Sentry has carried the same information since #3820, and the data was never recoverable from seeders anyway.

## Verification

`php artisan test --compact --group=CombatLog`, and confirm a fresh `migrate` on an isolated worktree DB applies cleanly.

---

### Sentry migration stages

Moving combat log parse failures off their bespoke DB pipeline and onto Sentry, so there is one error feed and one triage skill instead of two of each.

| Stage | Issue | Ships |
|---|---|---|
| 1/5 | #3820 | Fingerprint, tag and throttle structured errors; stop dev/test noise reaching the project |
| 2/5 | #3821 | Run-keyed segments endpoint; remove the parse-failure API, admin page and writes |
| 3/5 | #3822 | Drop the table (contract half — one release after 2/5) |
| 4/5 | #3823 | Collapse the three triage skills into one; new resolve policy |
| 5/5 | #3824 | Sentry project settings (spike protection, release stream, integrations) |

Order matters: **1/5 must deploy before 2/5**, and **3/5 must not ship in the same release as 2/5**. 4/5 is docs-only and can run in parallel, but should land after 1/5. 5/5 is Wotuu's.

Issue #3821, opened 2026-08-03
Sentry migration 2/5: replace the parse-failure API with a run-keyed segments endpoint and remove the parse-failure surface


Wotuu:
:robot: Phase 2 of moving combat log parse failures onto Sentry. **Blocked on #3820 having merged _and deployed_** — that is what puts the equivalent data into Sentry, so removing the DB surface before it is live would lose failures outright.

Expand half of an expand/contract: this removes every consumer of `CombatLogParseFailure` but **leaves the table, model, factory and migration in place**. Deploys are not atomic (a cron runs `migrate` independently of the ECS rollout), so old containers must not 500 against missing schema — see the backward-compatibility rules in `.claude/CLAUDE.md`. The drop is #3822.

## A run-keyed segments endpoint

The reproduction loop currently fetches Raider.IO log segments by *parse-failure row id*. With the table gone it must key on what Sentry carries as tags: `seasonId` + `runId`.

- New route in `routes/api.php`, e.g. `GET /api/v1/combatlog/seasons/{season}/runs/{runId}/segments`, under the same `api_role:admin` middleware as the endpoints it replaces.
- Retarget `app/Http/Controllers/Traits/ResolvesCombatLogParseFailureSegments.php` — rename to `ResolvesCombatLogRunSegments` and take `(Season $season, int $runId)` (significant-first param order). It keeps delegating to `RaiderIOApiServiceInterface::getCombatLogSegmentsForRun()`.
- Add a **Form Request** — there is none for these endpoints today. Validate the season id with an `exists` rule and expose a cached `season(): Season` getter, per `.claude/CLAUDE.md`.

## Remove

- The three API endpoints (`routes/api.php:30-34`) and `APICombatLogParseFailureController`.
- `AdminToolsCombatLogParseFailureController`, its routes (`routes/web.php:510-512`), `resources/views/admin/tools/combatlog/parsefailures.blade.php`, and its card in `admin/tools/list.blade.php`.
- The `numCombatLogParseFailures` badge in `GlobalComposer.php:40` and its rendering in `common/layout/nav/user.blade.php`.
- `CombatLogParseFailureResource` and `CombatLogParseFailureEnvelopeResource`, plus their `@OA\Schema` annotations. Check `config/l5-swagger.php` scan paths still resolve — it hardcodes them and drops schemas silently with CI green.
- The repository, its interface, and the binding in `RepositoryServiceProvider.php:259`.
[...]

PR #3825, opened 2026-08-03
Make structured errors usable in Sentry: fingerprint, tag and throttle them

### LINKED ISSUE #3820: Sentry migration 1/5: make structured errors usable — fingerprint, tag, throttle, and stop dev noise
:robot: #3794 wired `StructuredLogging` errors into Sentry via a `sentry` Monolog channel. The feed works, but it is not yet usable for triage, and it is currently being polluted by non-production traffic.

This is phase 1 of moving combat log parse failures off their bespoke DB pipeline and onto Sentry entirely (see the follow-up issues linked below). **It must ship and deploy before anything is removed.**

## Problems

**1. Every structured error collapses into one Sentry issue.**
Sentry groups message events by their message, and the message is the bare `ClassLogging::method`. No fingerprint is set, so every combat log parse failure ever — regardless of root cause — lands in a single issue titled `ProcessCombatLogSegmentsLogging::handleParseError`.

**2. The data needed to reproduce a failure never reaches Sentry.**
`ProcessCombatLogSegmentsLogging::handleParseError` is passed only `(runId, combatLogVersion, message, class)`. `rawLine`, `lineNumber` and `seasonId` exist only in the `combat_log_parse_failures` row.

**3. Local dev and the test suite ship events to the production Sentry project.**
Last 7 days: `production` 204, `local` 108, `testing` 41 — **42% of volume is noise**. `PHP-LARAVEL-R9` is a PHPUnit assertion string; `PHP-LARAVEL-R8` is a worktree seeder failure carrying a multi-KB SQL statement. On a 30-day/Developer plan that competes with real errors for quota.

**4. Nothing throttles the Sentry mirror.**
A WoW patch that breaks parsing could produce thousands of failures and burn the monthly quota in hours, dropping all other errors with it. The DB table's `updateOrCreate(run_id, line_number)` provides natural dedup today; Sentry grouping does **not** reduce quota consumption.

## Work

**Tag/fingerprint tap.** Add a `FingerprintsStructuredErrors` tap to the `sentry` channel in `config/logging.php`, modelled on the existing `SkipsExceptionMirrors` pair. It should set `context['fingerprint'] = [$exceptionClass, $normalizedMessage]` (digit runs replaced with `#`) and promote `runId`, `seasonId`, `combatLogVersion`, `exceptionClass`, `lineNumber` into `context['tags']`. `SentryHandler::consumeContextAndApplyToScope()` already reads all three keys — no new SDK plumbing.

Do this as a **tap, not by editing call sites**: adding `tags`/`fingerprint` keys to the `error()` context would leak them into stderr/daily/discord and would need repeating at every future log site.

Known limitation, inherited from the clustering `combatlog-parse-failure-poll` does by hand today: two different root causes with the same normalised shape merge into one issue. Document it; don't claim it's solved.

**Widen the log site.** `ProcessCombatLogSegmentsLogging::handleParseError` and its interface gain `?int $seasonId`, `?int $lineNumber`, `?string $rawLine` — values `ProcessCombatLogSegments::handle()` already pulls off `CombatLogParseException` one line below. Keep `rawLine` out of tags (too long, and it is the PII-carrying field); it belongs in context, which becomes the `log_context` extra.

**Cover the second log site.** `ProcessCombatLogPartLogging::handleParseError` (called from `ProcessCombatLogFromS3.php:56`) is a different signature with no `runId`. The tap must cover it too — that is why the tap approach was chosen.

**Throttle per fingerprint.** Suggest max 10 events per fingerprint per hour, via the cache. **Do not reuse `DeduplicateHandlers`** — #3794 removed it from this channel because `DeduplicationHandler` forwards via `handleBatch()`, and `SentryHandler::handleBatch()` compares a Monolog 3 `Level` enum against an integer, so the batch empties and every record is silently discarded. The new tap must not buffer: filter in `handle()` and return `false` so records keep bubbling (the `NullHandler`/`NoopHandler` trap from the same PR). Note in code and docs that a throttled issue's event count becomes a floor, not the true count.

**Stop the noise.**
- Tests: diagnose before suppressing. `phpunit.xml` already sets `LOG_CHANNEL=daily` (no `sentry`), so the leak is not the channel stack — something resolves the SDK client directly. Find the emitting test, fix that path, and add `SENTRY_LARAVEL_DSN=` to `phpunit.xml` as a backstop.
- Local: ship `SENTRY_LARAVEL_DSN=` empty in the `.env*.example` files and let the existing `empty(env(...))` → `NoopHandler` guard do the rest. Do **not** add an `APP_TYPE === 'local'` refusal to `config/sentry.php` — that would also kill the scratch-DSN path needed to verify this work end to end.
- CI: check `.env.ci.example` for a real DSN.

## Verification

Unit tests in the style #3794 established (`tests/Unit/App/Logging/`), resolving the real DSN-shaped channel against `CapturingTransport`. **Do not test by calling the handler directly** — that is the one path production never takes, and it is exactly how #3794's first round shipped a channel that sent nothing.

- Two records of different exception classes produce different fingerprints.
- Two of the same class differing only in digits produce the same fingerprint.
- `runId`/`seasonId`/`combatLogVersion`/`lineNumber` arrive as tags.
- N+1 identical records produce N events.
- An event still arrives through a stack with a no-op channel in front.

End to end, in a worktree with a **scratch** Sentry DSN (never the production project): run a known-bad segment through `ProcessCombatLogSegments` and confirm via the Sentry MCP that the issue is fingerprinted separately, carries the tags, and that `rawLine` is readable back out of `extra`.

## Note

Whether structured errors reach Sentry in production at all depends on `LOG_CHANNEL`/`LOG_STACK`, which is set outside this repo and is tracked separately in #3805.

---

### Sentry migration stages

Moving combat log parse failures off their bespoke DB pipeline and onto Sentry, so there is one error feed and one triage skill instead of two of each.

| Stage | Issue | Ships |
|---|---|---|
| 1/5 | #3820 | Fingerprint, tag and throttle structured errors; stop dev/test noise reaching the project |
| 2/5 | #3821 | Run-keyed segments endpoint; remove the parse-failure API, admin page and writes |
| 3/5 | #3822 | Drop the table (contract half — one release after 2/5) |
| 4/5 | #3823 | Collapse the three triage skills into one; new resolve policy |
| 5/5 | #3824 | Sentry project settings (spike protection, release stream, integrations) |

Order matters: **1/5 must deploy before 2/5**, and **3/5 must not ship in the same release as 2/5**. 4/5 is docs-only and can run in parallel, but should land after 1/5. 5/5 is Wotuu's.

Wotuu:
:robot: Closes #3820

Stage 1/5 of moving combat log parse failures onto Sentry (#3820 → #3821 → #3822 → #3823 → #3824). **This must deploy before #3821 removes the DB fallback.**

#3794 wired `StructuredLogging` errors into Sentry, but the resulting feed cannot be triaged:

- Sentry groups message events by their message, and `StructuredLogging` logs under a constant `ClassLogging::method`. **Every combat log parse failure ever recorded collapses into one issue**, no matter how many distinct root causes it represents.
- `rawLine`, `lineNumber` and `seasonId` never left the database, so an issue carried nothing you could reproduce from.
- Nothing bounded the volume. Sentry meters every event even though it groups them, so a bad WoW patch failing every line of every run would burn the account's event budget and start dropping genuinely novel errors.

## What changed

Three taps on the `sentry` channel. Their order in `config/logging.php` **reads backwards** — each tap wraps the handlers the previous one produced, so the last entry is outermost and runs first. Execution is `SkipsExceptionMirrors` → `FingerprintsStructuredErrors` → `ThrottlesRepeatedEvents` → `SentryHandler`.

| Handler | Does |
|---|---|
| `FingerprintsStructuredErrorsHandler` | Fingerprints as `[logSite, exceptionClass, normalizedMessage]` (digit runs → `#`), and promotes `runId`/`seasonId`/`combatLogVersion`/`lineNumber`/`exceptionClass` to tags |
| `ThrottlesRepeatedEventsHandler` | Caps events per fingerprint per hour |

Done as **taps rather than at the call sites**, so the `tags`/`fingerprint` keys never leak into the file, stderr and Discord output — and so every future `{Service}Logging` error benefits without further work. That also covers the second parse-failure log site (`ProcessCombatLogPartLogging`, S3-sourced), which has no `runId` and would otherwise have stayed unfingerprinted.

`rawLine` is **deliberately not a tag**: tags are capped at 200 characters and indexed, and a raw combat log line is long and carries player character names and GUIDs. It stays in the context, which the SDK files as extra data.

The normalisation is a direct port of the clustering `combatlog-parse-failure-poll` does by hand today, and **inherits its known limitation**: two genuinely different root causes sharing a normalised shape merge into one issue. Documented, not claimed solved.

## Two inconsistencies found while wiring this up

Both are pre-existing and fixed here, because the tap keys on exactly these names:

1. `ProcessCombatLogSegments` logged `$e::class` — the `CombatLogParseException` **wrapper** — while `recordFailure()` one line below stored the unwrapped cause. The log line and the DB row could not be correlated. Now resolved once and shared.
2. `ProcessCombatLogPartLogging::handleParseError` declared `($combatLogVersion, $offendingLine, $reason, $filePath)`, but its only caller passes `($combatLogVersion, $e->getMessage(), $e::class, $filePath)`. **Every logged context had the message under `offendingLine` and the exception class under `reason`.** Renamed to match reality.

Because `StructuredLogging` builds context from `get_defined_vars()`, renaming a Logging method's parameter silently changes Sentry grouping. Both methods now carry a docblock saying so.

## Test-suite leak

`sentryChannel_givenEmptyDsn_logsWithoutEmittingPhpErrors` logs for real rather than against a stub transport and — unlike its sibling four lines above — had **no guard for a configured DSN**. On any machine whose `.env` holds a real DSN it delivered straight to that project. That is `PHP-LARAVEL-R9`, a PHPUnit assertion string sitting in the production issue list.

Diagnosed before suppressing, per the issue: `phpunit.xml` already pins `LOG_CHANNEL=daily` (which contains no `sentry`), so the channel stack was never the leak. Fixed the test, and pinned both DSN keys empty in `phpunit.xml` as a suite-wide backstop so no future test can reach the real transport either.

Over the last 7 days `local` accounted for 108 events and `testing` 41, against 204 from production — 42% of volume was non-production noise. The remaining `local` share comes from a developer `.env` holding a real DSN, which is not this repo's to change; that is tracked in #3824.

## Verification

- `php artisan test --compact tests/Unit/App/Logging/ tests/Feature/Jobs/CombatLog/` → **65 passed**.
- `composer run fix` → 0 of 3804 files changed. `composer run analyse` → no errors.
- **Each new test was verified to fail with its tap removed** — asserting on configuration shape or calling the handler directly is exactly how #3792 shipped three silent failures past a green suite, so the tests resolve the real DSN-shaped channel against `CapturingTransport` instead.

Backend-only; no visual verification applies.

### DIFF
--- a/app/Jobs/CombatLog/ProcessCombatLogSegments.php
+++ b/app/Jobs/CombatLog/ProcessCombatLogSegments.php
@@ -117,16 +117,31 @@ class ProcessCombatLogSegments implements ShouldBeUnique, ShouldQueue
             // an unparsable line (see #3791).
             throw $e;
         } catch (Throwable $e) {
-            $log->handleParseError($this->runId, $this->combatLogVersion, $e->getMessage(), $e::class);
+            // CombatLogParseException only wraps the real failure, so the class worth grouping and reporting on is
+            // the one it wrapped. Resolved once here because the log line used to report the wrapper while the
+            // recorded failure reported the cause, which made the two impossible to correlate.
+            $exceptionClass = $e instanceof CombatLogParseException ? $e->getOriginalExceptionClass() : $e::class;
+            $lineNumber     = $e instanceof CombatLogParseException ? $e->lineNumber : null;
+            $rawLine        = $e instanceof CombatLogParseException ? $e->rawLine : null;
+
+            $log->handleParseError(
+                $this->runId,
+                $this->season->id,
+                $this->combatLogVersion,
+                $lineNumber,
+                $exceptionClass,
+                $e->getMessage(),
+                $rawLine,
+            );
 
             $parseFailureRepository->recordFailure(
                 $this->runId,
                 $this->season->id,
                 $this->combatLogVersion,
-                $e instanceof CombatLogParseException ? $e->lineNumber : null,
-                $e instanceof CombatLogParseException ? $e->rawLine : null,
+                $lineNumber,
+                $rawLine,
                 $e->getMessage(),
-                $e instanceof CombatLogParseException ? $e->getOriginalExceptionClass() : $e::class,
+                $exceptionClass,
             );
         } finally {
             foreach ($tempFiles as $tempFile) {
--- a/app/Logging/Handlers/FingerprintsStructuredErrorsHandler.php
+++ b/app/Logging/Handlers/FingerprintsStructuredErrorsHandler.php
@@ -1,0 +1,155 @@
+<?php
+
+namespace App\Logging\Handlers;
+
+use Monolog\Handler\HandlerWrapper;
+use Monolog\LogRecord;
+use Override;
+
+/**
+ * Makes error-level {@see \App\Logging\StructuredLogging} records triageable once they reach the error tracker.
+ *
+ * These records are messages, not exceptions, and the error tracker groups message events by their message. Since
+ * StructuredLogging logs under a constant `ClassLogging::method`, every failure of a given log site collapses into a
+ * single issue no matter how many distinct root causes it represents - a hundred different combat log parse failures
+ * look like one problem. All the distinguishing detail sits in the context, where nothing can search or group on it.
+ *
+ * This handler fixes both halves:
+ *
+ * 1. It derives an explicit fingerprint from the log site plus the failure's exception class and the shape of its
+ *    message, so one root cause becomes one issue.
+ * 2. It promotes the short domain identifiers to tags, which are the only searchable and aggregatable surface the
+ *    error tracker offers. That is what lets a triage pass answer "which runs does this affect" without opening
+ *    events one by one.
+ *
+ * Everything not promoted stays in the context and is filed as extra data by the SDK's handler.
+ *
+ * Deliberately done as a channel tap rather than at the call sites: adding `tags`/`fingerprint` keys to a
+ * StructuredLogging context would leak them into the file, stderr and Discord output, which are the human-facing
+ * narrative, and would have to be repeated at every log site that ever wants them.
+ */
+class FingerprintsStructuredErrorsHandler extends HandlerWrapper
+{
+    /**
+     * Context keys promoted to tags, in the order they should appear. Restricted to short, bounded identifiers -
+     * tags are capped at 200 characters and a high-cardinality free-text tag is worse than useless.
+     *
+     * Notably absent: `rawLine`. It is long, and it is the field carrying player names and GUIDs, so it stays in the
+     * context where it is readable during triage but never indexed.
+     */
+    private const array TAG_CONTEXT_KEYS = [
+        'runId',
+        'seasonId',
+        'combatLogVersion',
+        'lineNumber',
+        'exceptionClass',
+    ];
+
+    /** Context key holding the class of the exception the record describes. */
+    private const string EXCEPTION_CLASS_KEY = 'exceptionClass';
+
+    /** Context key holding the failure's message, whose digit-stripped shape is the second fingerprint component. */
+    private const string MESSAGE_KEY = 'message';
+
+    /** Hard cap the error tracker applies to tag values; a longer value is dropped rather than truncated. */
+    private const int MAX_TAG_LENGTH = 200;
+
+    #[Override]
+    public function handle(LogRecord $record): bool
+    {
+        return parent::handle($this->fingerprint($this->promoteTags($record)));
+    }
+
+    /**
+     * Records are forwarded in batches when a buffering handler sits in front of this one, which would otherwise
+     * bypass the enrichment entirely.
+     *
+     * @param array<int, LogRecord> $records
+     */
+    #[Override]
+    public function handleBatch(array $records): void
+    {
+        foreach ($records as $record) {
+            $this->handle($record);
+        }
+    }
+
+    /**
+     * Copies the recognised identifiers into the `tags` context key, which the SDK's handler reads out and applies to
+     * the scope. Values that cannot be expressed as a short string are skipped rather than truncated - a silently
+     * cut-off identifier is a wrong identifier.
+     */
+    private function promoteTags(LogRecord $record): LogRecord
+    {
+        $tags = [];
+        foreach (self::TAG_CONTEXT_KEYS as $key) {
+            $value = $record->context[$key] ?? null;
+
+            if ($value === null || is_array($value) || is_object($value)) {
+                continue;
+            }
+
+            $value = (string)$value;
+            if ($value === '' || strlen($value) > self::MAX_TAG_LENGTH) {
+                continue;
+            }
+
+            $tags[$key] = $value;
+        }
+
+        if ($tags === []) {
+            return $record;
+        }
+
+        // Merge rather than replace - a caller that set its own tags keeps them, and wins on a key collision
+        return $record->with(context: array_merge($record->context, [
+            'tags' => array_merge($tags, is_array($record->context['tags'] ?? null) ? $record->context['tags'] : []),
+        ]));
+    }
+
+    /**
+     * Splits a log site's single issue into one issue per root cause.
+     *
+     * The log site itself stays the first component, so two different log sites that happen to hit the same exception
+     * class never merge. A record carrying neither an exception class nor a message is left alone: there is nothing to
+     * distinguish causes by, and the error tracker's default message grouping is already the right answer.
+     *
+     * A fingerprint set by another handler is never overwritten - {@see SkipsExceptionMirrorsHandler} sets its own for
+     * uncaught-exception mirrors, and it has better information about those than this handler does.
+     */
+    private function fingerprint(LogRecord $record): LogRecord
+    {
+        if (isset($record->context['fingerprint'])) {
+            return $record;
+        }
+
+        $exceptionClass = $record->context[self::EXCEPTION_CLASS_KEY] ?? null;
+        $message        = $record->context[self::MESSAGE_KEY] ?? null;
+
+        $components = array_values(array_filter([
+            is_string($exceptionClass) ? $exceptionClass : null,
+            is_string($message) ? self::normalizeMessage($message) : null,
+        ]));
+
+        if ($components === []) {
+            return $record;
+        }
+
+        return $record->with(context: array_merge($record->context, [
+            'fingerprint' => [$record->message, ...$components],
+        ]));
+    }
+
+    /**
+     * Reduces a failure message to its stable shape by collapsing every digit run to a `#`, so
+     * "Unable to find combat log version 21" and "... version 22" group together.
+     *
+     * Known limitation, inherited from the manual clustering this replaces: two genuinely different root causes that
+     * happen to share a normalised shape merge into one issue. That is acceptable for a triage feed - re-cluster by
+     * hand once the dominant cause is fixed.
+     */
+    private static function normalizeMessage(string $message): string
+    {
+        return trim((string)preg_replace('/\d+/', '#', $message));
+    }
+}
--- a/app/Logging/Handlers/FingerprintsStructuredErrorsHandler.php
+++ b/app/Logging/Handlers/FingerprintsStructuredErrorsHandler.php
@@ -54,6 +54,12 @@ class FingerprintsStructuredErrorsHandler extends HandlerWrapper
     /** Hard cap the error tracker applies to tag values; a longer value is dropped rather than truncated. */
     private const int MAX_TAG_LENGTH = 200;
 
+    /**
+     * How much of the normalised message contributes to the fingerprint. Long enough to tell failure shapes apart,
+     * short enough that a message with an entire combat log line appended cannot make every event unique.
+     */
+    private const int MAX_FINGERPRINT_MESSAGE_LENGTH = 120;
+
     #[Override]
     public function handle(LogRecord $record): bool
     {
@@ -141,8 +147,15 @@ class FingerprintsStructuredErrorsHandler extends HandlerWrapper
     }
 
     /**
-     * Reduces a failure message to its stable shape by collapsing every digit run to a `#`, so
-     * "Unable to find combat log version 21" and "... version 22" group together.
+     * Reduces a failure message to its stable shape, so "Unable to find combat log version 21" and "... version 22"
+     * group together.
+     *
+     * Collapsing digit runs alone is not enough here. Parsing exceptions routinely interpolate the offending combat
+     * log line into their message - `CombatLogStringParser` throws "Unbalanced quotes in string <line>", and
+     * `CombatLogService` wraps that as the message - and such a line carries character names and hex GUIDs. Left
+     * alone those make every single failing line its own fingerprint, which would defeat both the grouping and the
+     * throttle that keys on it. Quoted segments and GUID-shaped tokens are therefore flattened first, and the result
+     * is capped so a very long line cannot reintroduce uniqueness past the cap.
      *
      * Known limitation, inherited from the manual clustering this replaces: two genuinely different root causes that
      * happen to share a normalised shape merge into one issue. That is acceptable for a triage feed - re-cluster by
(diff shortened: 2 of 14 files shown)


app/Logging/Handlers/ThrottlesRepeatedEventsHandler.php
```diff
@@ -0,0 +1,98 @@
...
+     * Counts this record against its signature's window and reports whether it is still within the allowance.
+     *
+     * A cache failure must never cost us an error report, so anything unexpected forwards the record.
+     */
+    private function shouldForward(LogRecord $record): bool
+    {
+        $key = sprintf('%s:%s', self::CACHE_KEY_PREFIX, sha1($this->signature($record)));
+
+        // add() only succeeds when the key is absent, which both starts the window and counts the first event
+        if (Cache::add($key, 1, self::WINDOW_SECONDS)) {
```
Wotuu: :robot: **A cache outage now turns every error-level log call into a thrown exception.**

`shouldForward()` calls `Cache::add()` / `Cache::increment()` with no `try`/`catch`, but the docblock two lines above it promises the opposite: *"A cache failure must never cost us an error report, so anything unexpected forwards the record."* Monolog rethrows anything a handler throws (`Logger::handleException()` — `if (null === $this->exceptionHandler) { throw $e; }`), and nothing in Laravel installs an exception handler on the channel.

Verified end to end in this worktree with the real channel and a cache store that throws:

```
$ch->error('ProcessCombatLogSegmentsLogging::handleParseError', [...]);
THREW: RedisException Connection refused
```

Blast radius: `sentry` is a member of `stack_docker`, `stack_docker_local` and `scheduler`, so with Redis unavailable *every* `Log::error()` on the app's default stack throws — including the one inside `Handler::report()`, i.e. an exception thrown while reporting an exception, and inside queue workers. Redis being down is precisely the moment you want the error reports. This failure mode does not exist on master; it is introduced here.

Suggested fix:

```php
try {
    if (Cache::add($key, 1, self::WINDOW_SECONDS)) {
        return true;
    }

    $count = Cache::increment($key);
} catch (Throwable) {
    // A cache failure must never cost us an error report
    return true;
}
```

app/Logging/Handlers/FingerprintsStructuredErrorsHandler.php
```diff
@@ -0,0 +1,155 @@
...
+     * Reduces a failure message to its stable shape by collapsing every digit run to a `#`, so
+     * "Unable to find combat log version 21" and "... version 22" group together.
+     *
+     * Known limitation, inherited from the manual clustering this replaces: two genuinely different root causes that
+     * happen to share a normalised shape merge into one issue. That is acceptable for a triage feed - re-cluster by
+     * hand once the dominant cause is fixed.
+     */
+    private static function normalizeMessage(string $message): string
+    {
+        return trim((string)preg_replace('/\d+/', '#', $message));
```
Wotuu: :robot: **Collapsing only digit runs is not enough for the messages this actually sees — the fingerprint stays near-unique and the throttle never engages.**

`CombatLogStringParser.php:65` throws `sprintf('Unbalanced quotes in string %s', $line)`, and `CombatLogService.php:412` wraps that with `$exception->getMessage()` — so the `message` context key for a whole family of parse failures *contains the raw combat log line*. After `preg_replace('/\d+/', '#')` a line like

```
SPELL_DAMAGE,Player-1084-0B4087DE,"Somebody-Realm",...
```

becomes `SPELL_DAMAGE,Player-#-#B#DE,"Somebody-Realm",...` — the hex letters of the GUID and the character name survive, so the third fingerprint component differs per player, per pull, per event. Consequences, both of which are the exact things this PR sets out to fix:

1. **Grouping**: one Sentry issue per failing line instead of one per root cause.
2. **Throttling**: `ThrottlesRepeatedEventsHandler` keys on the fingerprint, so a near-unique fingerprint means the allowance is never reached and the volume is unbounded — the bad-WoW-patch scenario the throttle exists for.

The other message family (`ValidatesParameterCount`: class name + digits) normalises fine, so this is specifically about messages that interpolate the offending line.

Worth noting the new tests never exercise this shape: `'Invalid parameter count, wanted 30-32, got 29'` and `'Unable to find combat log version 21'` are both digit-only variance, so the suite is green on the shape that works and silent on the shape that doesn't. A test using a real `Unbalanced quotes in string ...` message would fail today.

Also a consistency point: `rawLine` is deliberately kept out of the tags because it carries player names and GUIDs — but via the message it lands in the fingerprint, which is indexed grouping metadata, so the same content reaches Sentry anyway.

Options: truncate the normalised message to a bounded prefix, and/or normalise the known variable shapes (`Player-\d+-[0-9A-F]+`, quoted `"Name-Realm"` segments) alongside digit runs.

app/Jobs/CombatLog/ProcessCombatLogSegments.php
```diff
@@ -117,16 +117,31 @@ public function handle(
...
         } catch (Throwable $e) {
-            $log->handleParseError($this->runId, $this->combatLogVersion, $e->getMessage(), $e::class);
+            // CombatLogParseException only wraps the real failure, so the class worth grouping and reporting on is
+            // the one it wrapped. Resolved once here because the log line used to report the wrapper while the
+            // recorded failure reported the cause, which made the two impossible to correlate.
+            $exceptionClass = $e instanceof CombatLogParseException ? $e->getOriginalExceptionClass() : $e::class;
+            $lineNumber     = $e instanceof CombatLogParseException ? $e->lineNumber : null;
+            $rawLine        = $e instanceof CombatLogParseException ? $e->rawLine : null;
+
+            $log->handleParseError(
```
Wotuu, line 127: :robot: The headline fix here — the log line reported the `CombatLogParseException` **wrapper** while `recordFailure()` reported the **cause** — has no test asserting it.

`tests/Feature/Jobs/CombatLog/ProcessCombatLogSegmentsTest.php:183` is still just:

```php
$log->expects($this->once())->method('handleParseError');
```

while the sibling `recordFailure` expectation ten lines above it (line ~170) asserts all seven arguments. So this call could pass the wrapper class, the wrong argument order, or a null `lineNumber`/`rawLine` and the suite stays green — on the one behaviour the PR body calls out as previously broken, and on a signature that just changed from 4 positional args to 7.

Please mirror the `recordFailure` assertion: `->with(self::RUN_ID, $this->isNull(), self::COMBAT_LOG_VERSION, 257080, InvalidArgumentException::class, 'Unbalanced quotes in string ...', 'SPELL_DAMAGE,Player-1084-0B4087DE,"Pa')`.

phpunit.xml
```diff
@@ -30,6 +30,12 @@
         <env name="TELESCOPE_ENABLED" value="false"/>
         <env name="PENNANT_STORE" value="array"/>
         <env name="LOG_CHANNEL" value="daily"/>
+        <!-- No test may ever reach the real error tracker. LOG_CHANNEL=daily is not enough on its own: any test that
+             resolves the 'sentry' channel by name, or triggers the SDK directly, sends to whichever project the
+             developer's .env points at. Both keys are pinned because config/sentry.php falls back from the first to
+             the second. Tests that need the DSN branch build it themselves against a stub transport. -->
+        <env name="SENTRY_LARAVEL_DSN" value=""/>
```
Wotuu: :robot: Nit: without `force="true"` this backstop is conditional, not absolute.

PHPUnit only applies an `<env>` entry when the variable is not already present in the real environment — `vendor/phpunit/phpunit/src/TextUI/Configuration/PhpHandler.php:112`:

```php
if ($force || getenv($name) === false) {
```

So a developer or CI runner that *exports* `SENTRY_LARAVEL_DSN` (rather than only setting it in `.env`) keeps their real DSN and the suite can still reach the live transport — which is the scenario the PR body describes as closed off ("no future test can reach the real transport either"). Nothing exports it in the local `app` container today, so the practical severity is low, but `force="true"` on both lines makes the claim actually hold.

Wotuu:

Wotuu:
:robot: All four cold-review findings addressed in 2028d7e73 — thank you, the two blockers were both real and neither would have shown up in CI.

**1. Cache outage breaking error logging** — fixed, and it was worse than the finding described. `shouldForward()` now guards its whole body and fails open. Confirmed the direction of the fix: reverting the `try`/`catch` makes the new `sentryChannel_givenAnUnreachableCache_stillLogsWithoutThrowing` test fail.

While in there I also removed a latent way the throttle could silence a signature *permanently*: `Cache::increment()` creates a key without a lifetime (Redis `INCRBY` sets none), so if the previous key expired between the `add()` and the `increment()`, the counter never reset. The key is now bucketed by wall-clock window and given an explicit lifetime on first use.

**2. Fingerprint near-unique** — fixed, and the finding was sharper than I first credited. The fix needed one more pass than quoted-segment + GUID collapse: an unbalanced quote *by definition* leaves a dangling `"Somebody-Realm` at the end of the message, which `/"[^"]*"/` cannot match because there is no closing quote. My first attempt still produced two different fingerprints for the two real messages — that is exactly the case you pointed at. There is now a dangling-quote pass as well, plus a length cap.

You were also right that the old tests were green on the shape that worked and silent on the shape that didn't. Added `sentryChannel_givenParseErrorsCarryingDifferentCombatLogLines_fingerprintsThemTogether` using two real unbalanced-quote messages from different players.

**3. Untested headline fix** — `handleParseError` now asserts all seven arguments, mirroring the sibling `recordFailure` expectation, so the wrapper-vs-cause correction is actually pinned.

**4. `force="true"`** — added to both DSN pins, with a comment naming the `PhpHandler` behaviour.

Verification after the changes: 67 tests pass (`tests/Unit/App/Logging/` + `tests/Feature/Jobs/CombatLog/`), `composer run fix` changed 0 of 3804 files, `composer run analyse` reports no errors. Both new tests were verified to fail with their respective fix reverted.

Also noting for the record the three things you explicitly checked and cleared — tap ordering verified by reflection rather than by trusting the config comment, no new silent-drop path, and the `ProcessCombatLogPartLogging` rename being complete. That is the kind of verification that makes the review worth having.

phpunit.xml
Wotuu: :robot: Fixed in 2028d7e73 — added force="true" to both DSN pins with a comment naming the PhpHandler behaviour. See the consolidated summary comment for full detail.

app/Logging/Handlers/FingerprintsStructuredErrorsHandler.php
Wotuu: :robot: Fixed in 2028d7e73 — added a dangling-quote normalization pass (an unbalanced quote leaves no closing quote for the quoted-segment regex to match) plus a length cap, and added sentryChannel_givenParseErrorsCarryingDifferentCombatLogLines_fingerprintsThemTogether using two real unbalanced-quote messages from different players to pin it. See the consolidated summary comment for full detail.

app/Jobs/CombatLog/ProcessCombatLogSegments.php
Wotuu: :robot: Fixed in 2028d7e73 — handleParseError's test expectation now asserts all seven arguments, mirroring the sibling recordFailure assertion, so the wrapper-vs-cause correction is actually pinned. See the consolidated summary comment for full detail.

app/Logging/Handlers/ThrottlesRepeatedEventsHandler.php
Wotuu: :robot: Fixed in 2028d7e73 — shouldForward() now wraps its whole body in try/catch and fails open on a cache outage (verified: reverting the catch fails the new sentryChannel_givenAnUnreachableCache_stillLogsWithoutThrowing test). Also fixed a related latent bug found while in there: Cache::increment() created a key with no lifetime, so the throttle counter could never reset if the add() key expired first — now bucketed by wall-clock window with an explicit lifetime. See the consolidated summary comment for full detail.

phpunit.xml
```diff
@@ -30,6 +30,14 @@
         <env name="TELESCOPE_ENABLED" value="false"/>
         <env name="PENNANT_STORE" value="array"/>
         <env name="LOG_CHANNEL" value="daily"/>
+        <!-- No test may ever reach the real error tracker. LOG_CHANNEL=daily is not enough on its own: any test that
```
Wotuu: Now this is way too many comments, so just say no test may ever reach the real error tracker, full stop.

Wotuu (commented):
One tiny thing, that's it.

phpunit.xml
Wotuu: :robot: Fixed in 38c631dc5 — condensed the two comment blocks above `SENTRY_LARAVEL_DSN` into a single line: "No test may ever reach the real error tracker, full stop."

PR #3827, opened 2026-08-04
Drop the vestigial icon_file_id columns from affixes, factions, character_classes and character_class_specializations

### LINKED ISSUE #3786: Drop the vestigial icon_file_id columns from affixes, factions, character_classes and character_class_specializations

Wotuu:
:robot: Closes #3786

## Summary
Follow-up to #3775 (released in v15.9.1) — drops the vestigial `icon_file_id` column from
`affixes`, `factions`, `character_classes` and `character_class_specializations`, mirroring the
`Expansion` migration from #3350.

- One migration per table: deletes orphaned `files` rows for that model class, then drops the
  index and column. `down()` restores the correct type/position per table (`int` after a specific
  column for `factions`/`affixes`, `varchar(255)` appended last for the two character-class
  tables) — verified with an actual `migrate:rollback` + `migrate` round trip, not just read from
  the issue table.
- Removed `icon_file_id` from `$fillable`/`$hidden`/PHPDoc on all four models.
- Removed the `icon_file_id => -1` placeholder from `AffixSeeder`, `FactionsSeeder`,
  `CharacterClassesSeeder`, `CharacterClassSpecializationsSeeder` (the latter's `->map()` closure
  collapses into a plain `insert()`).
- Removed `AffixController::store()`'s `$validated['icon_file_id'] = -1`.
- Updated `AffixTest` and `AdminAffixControllerTest` (renamed off the `-1` placeholder), plus
  `FactionTest`/`CharacterClassTest`, which also asserted on the now-nonexistent column.

## Test plan
- [x] `php artisan migrate` — all four migrations apply cleanly
- [x] `php artisan migrate:rollback --step=4` then `php artisan migrate` — verified restored
      column type/position for all four tables against the pre-migration schema
- [x] `php artisan db:seed` (full) — succeeds, no `icon_file_id` references left in any seeder
- [x] Verified `files.model_class` is stored as `$model::class` (FQCN), matching what each
      migration's delete-orphans query targets
- [x] `php artisan test --compact --filter="CharacterClass|Faction|Affix"` — 106 passed
- [x] `composer run fix` — clean (only PHPDoc realignment on files already in the diff)

Backend-only, no rendered-output change — visual verification skipped per the MR checklist.


Wotuu:
:robot: Cold review: 0 findings.

Verified as part of this data-destructive/schema review:
- **`down()` type/position fidelity**: checked all four restores against `database/schema/migrate-schema.dump` — `factions` `int(11) NOT NULL` after `id`, `affixes` `int(11) NOT NULL` after `key`, `character_classes`/`character_class_specializations` `varchar(255) NOT NULL` appended last. All four match exactly, including the restored index names (all < 64 chars).
- **`files` orphan delete**: `files.model_class` is written as `$model::class` (FQCN, `File.php:178`), matching each migration's `where('model_class', X::class)` filter; the query-builder delete deliberately bypassing the `deleting` hook (which would unlink shared static assets) mirrors the reviewed #3350/#3362 expansions precedent verbatim.
- **No missed readers**: grepped the PR head tree — zero remaining `icon_file_id`/`iconfile` references for these four models outside the schema dumps (stale by design, same as `expansions`) and a `migrations_legacy` file that runs before these migrations anyway. `AffixFormRequest` never validated the field.
- **Expand/contract safety**: confirmed #3787 (the last code reader/writer removal) is an ancestor of the `v15.9.1` tag, so the running code tolerates the drop. The one same-release writer removal (`AffixController::store`'s `icon_file_id = -1`) creates only an admin-panel-affix-creation failure window during the deploy — identical to the accepted expansions precedent, so not raised as a finding.

PR #3830, opened 2026-08-04
#3826 Detect and display player spell counters (Vanish/Shadowmeld) in the NPC Compendium

### LINKED ISSUE #3826: NPC Compendium: detect and display player counter tricks (Vanish/Shadowmeld) for NPC abilities
:robot: Implementation plan below was designed and approved in a planning session with Wotuu on 2026-08-04. It is written to be executed cold by an agent.

**Before starting:** the reference combat log `tmp/WoWCombatLog-061026_122601.zip` lives in the main checkout on the dev machine (`tmp/` is not in git) — copy it into your worktree for the e2e verification step. Replace `<issue>` in the worktree command with this issue's number.

---

## Context

Rogues (Vanish) and Night Elves (Shadowmeld) can cancel certain incoming NPC abilities by
dropping themselves as the target while the ability is cast / in flight / channeled. Which NPC
abilities can be "vanished" is valuable routing knowledge. This adds:

1. An extensible "spell counter" concept (many more trick types will follow).
2. Detection inside the existing combat-log extraction pipeline.
3. Display in the NPC Compendium (spells table column + activity feed events).

Reference log: `tmp/WoWCombatLog-061026_122601.zip` — Skyreach (zone 1209), key run starts
12:33:24, rogue **Jaxeek-TarrenMill-EU** (`Player-1084-0A5F8492`). Vanish = cast **1856**
(self-aura 11327), Shadowmeld = **58984**.

Work in an isolated worktree (`sh/worktree.sh create <issue>-spell-counters`), per default flow.

## Verified log signatures (all confirmed in the reference log)

ε must be **two-sided** (~500 ms): same-instant events can log in either order (the Lens Flare
removal logs 1 ms *before* the Vanish cast-success line).

**A. Targeting-debuff dropped (High Sage Viryx — Lens Flare).** 12:52:50.374 NPC
`SPELL_CAST_START` of cast spell 1253840 (dest nil); same timestamp, a targeting debuff with
**nil source guid** and a *different* spell id (1253531, same name) is applied to the player.
12:52:52.174 that debuff is removed 1 ms before the Vanish cast line (normal duration ~12 s on
the two later victims: 12:53:29→12:53:41, 12:54:08→12:54:20). Cast completes with no effect.
Attach the fact to the **cast** spell id (that's what lands in `npc_spells`), linked via
same-timestamp CAST_START correlation (prefer same spell name).

**B. Channel broken (Solar Construct — Solar Flame 1253446).** Channel = `SPELL_CAST_SUCCESS`
+ debuff + periodic ticks on the player (12:40:13.576). Debuff removed at the exact Vanish
timestamp (12:40:15.2302), ~1.7 s into a normally ~6 s channel; no retarget.

**C. Cast aborted (Dread Raven — Dread Wind 1258174, ~3 s cast).** `SPELL_CAST_START` (dest
nil) 12:35:06.752; Vanish 12:35:08.544; the cast **never reaches SPELL_CAST_SUCCESS** — same
guid restarts the cast 12:35:09.186. Heuristic attribution; exclusions: cast ended by
`SPELL_CAST_SUCCESS`/`SPELL_CAST_FAILED`/`SPELL_INTERRUPT`, caster `UNIT_DIED`, or any debuff
applied to the caster mid-cast ("disturbed"). **User decision: C facts are treated identically
to A/B** — first sighting sets the bit; the 3-day staleness window self-corrects false
positives.

**Negative control (verified):** offensive Vanish at 12:37:33.849 — an NPC debuff on the rogue
expired naturally 2.4 s later (outside ε); Ranjit's in-flight cast succeeded 0.55 s after the
Vanish (C must not flag a completed cast).

## Design: reuse the spell-property machinery (no new models)

A counter is semantically identical to existing spell properties (aura/debuff/miss types): a
per-spell, combat-log-derived boolean kept alive by the rolling observation window. So:

- `spells.counters_mask` int bitmask + `Spell::ALL_COUNTERS` const (bit ⇒ key), mirroring
  `miss_types_mask` / `ALL_MISS_TYPES`.
- New `SpellProperty` cases `counter_vanish` / `counter_shadowmeld` — both
  `combat_log_spell_property_observations.property` and `combat_log_spell_events.property` are
  plain strings, so **no combatlog-DB migration** and **no new event-type enum cases** (the
  three exhaustive `match` blocks in `event_list.blade.php` keyed on `event_type` are safe).
- Extensibility lives at the detection layer: one small definition class per counter.
  A future counter = definition class + `ALL_COUNTERS` bit + `SpellProperty` case + lang keys.

### 1. Vocabulary
- `app/Models/Spell/SpellConstants.php`: `COUNTER_VANISH = 1`, `COUNTER_SHADOWMELD = 2`,
  `ALL_COUNTERS = [1 => 'vanish', 2 => 'shadowmeld']` (same style as miss types).
- `app/Models/CombatLog/SpellProperty.php`: cases `CounterVanish = 'counter_vanish'`,
  `CounterShadowmeld = 'counter_shadowmeld'`; add `fromCounterBit(int $bit): self` (mirrors
  `fromMissTypeBit`) and `isCounter(): bool` (`str_starts_with($this->value, 'counter_')`).
- `app/Models/Spell/Spell.php`: `counters_mask` in `@property` docblock, `$fillable`, `casts()`.

### 2. Migration (main DB only, additive/backward-compatible, no FK)
- `database/migrations/2026_08_04_000000_add_counters_mask_to_spells_table.php`:
  `unsignedInteger('counters_mask')->default(0)->after('miss_types_mask')`.

### 3. Counter definitions (the extensible layer)
`app/Service/CombatLog/DataExtractors/SpellCounters/` — one class per counter, carrying BOTH
detection and display metadata so "add a counter = one class + one bit + one enum case + lang
keys" holds:
- `SpellCounterDefinitionInterface`:
  - `getProperty(): SpellProperty` / `getCounterBit(): int`
  - `getTriggerCastSpellIds(): array` (player `SPELL_CAST_SUCCESS` ids marking a counter use)
  - `getIconName(): string` (e.g. `ability_vanish` / `ability_ambush`, rendered via
    `ksgAssetImage('spells/<icon>.jpg')` like characteristics — verify the jpgs exist in the
    assets at implementation time, else fetch/add them)
  - `getCharacterClassKey(): ?string` — `'rogue'` for Vanish (`CharacterClass` key)
  - `getCharacterRaceKey(): ?string` — `CharacterRace::CHARACTER_RACE_NIGHT_ELF` for
    Shadowmeld. Exactly one of class/race key is non-null per definition (class ability vs
    racial).
- `SpellCounterDefinitions::all(): Collection<SpellCounterDefinitionInterface>` — small static
  registry consumed by both the extractor and `ClassCompendiumController` (the seeded player
  Spell rows are useless for this: Vanish 227736 / Shadowmeld 58984 carry
  `category = spellcategory.unknown`, so the class page's category query can't find them).
- `VanishSpellCounterDefinition` (trigger 1856; aura 11327 documented in a const, unused for
  v1), `ShadowmeldSpellCounterDefinition` (trigger 58984).

### 4. Detector: `SpellCounterDataExtractor`
`app/Service/CombatLog/DataExtractors/SpellCounterDataExtractor.php`, implements
`DataExtractorInterface`; register in `CombatLogDataExtractionService::__construct` **after**
`SpellDataExtractor` (afterExtract flush order keeps `SpellCreated` before the counter
`PropertyChanged` in the feed, and Spell rows created in the same file run exist at flush).
Logging companion `DataExtractors/Logging/SpellCounterDataExtractorLogging{,Interface}.php`
(auto-bound by `LoggingServiceProvider`'s glob convention): info per detection, debug per
rejected candidate.

Consts: `EPSILON_MS = 500`, `CAST_START_LINK_WINDOW_MS = 500`, `C_SUPERSEDE_WINDOW_MS = 3000`,
`PENDING_CAST_MAX_AGE_MS = 15000`, plus an activeDebuffs age bound for memory.

State (Collections, pruned opportunistically on insert; reset in `afterExtract` and on
`ChallengeModeStart/End` + `ZoneChange`):
- `activeDebuffs` keyed `"{playerGuid}-{spellId}"` → `{appliedAt, sourceGuid?, linkedNpcCastSpellId?}`
  — nil-source or NPC-creature-source DEBUFFs currently on players (A/B).
- `recentNpcCastStarts` list `{ts, casterGuid, spellId, spellName}` — links nil-source
  targeting debuffs to the concurrent NPC cast (A).
- `pendingNpcCasts` keyed by caster guid → `{spellId, startedAt, disturbed}` (C).
- `recentCounterCasts` list `{ts, playerGuid, definition}`.
- `recentDebuffRemovals` list — the backward half of the two-sided ε (removal logged before
  the counter cast).
- `pendingCounterObservations` keyed `"{spellId}-{property}"` — dedup'd flush queue.

Event handling (`extractData`): SpecialEvents first — `UnitDied`/`UnitDestroyed` forget the
pending cast; run-boundary events reset state. Then `CombatLogEvent` with `Spell` prefix; NPC
test = `Guid\Creature` + `CREATURE_UNIT_TYPE_CREATURE`; timestamps via Carbon
`diffInMilliseconds` (absolute):
1. **CastStart (NPC)**: if the guid already has an unresolved pending cast → C decision:
   match when `!disturbed`, a counter cast happened in `[startedAt - ε, now]`, within
   `C_SUPERSEDE_WINDOW_MS` of the counter, and the pending cast isn't stale → queue
   `{pendingSpellId, property}`. Then insert new pending cast + append `recentNpcCastStarts`.
2. **CastSuccess**: player + trigger spell id → record counter cast, then scan
   `recentDebuffRemovals` (same player, within ε) → queue matches. NPC → forget pending cast
   (C negative control).
3. **CastFailed (NPC)** / **Interrupt (dest NPC)** → forget pending cast, never attributed.
4. **AuraApplied DEBUFF**: dest player + (nil source | NPC source) → insert `activeDebuffs`;
   nil source → link to `recentNpcCastStarts` within window, prefer same spell name, store
   `linkedNpcCastSpellId`. Dest NPC with pending cast → `disturbed = true`.
5. **AuraRemoved DEBUFF, dest player, known in `activeDebuffs`**:
   `attributedSpellId = linkedNpcCastSpellId ?? spellId`. Natural-expiry guard: if the spell's
   `duration > 0` and lifetime ≥ duration − ε → skip. Scan `recentCounterCasts` (same player,
   ±ε) → queue match; else append to `recentDebuffRemovals`. Remove from `activeDebuffs`.

Flush (`afterExtract`), mirroring `SpellPropertyObservationCollector::afterCollect`:
1. Batch `CombatLogSpellPropertyObservation::upsert` (`spell_id`, `property`,
   `observed_on = today`, `combat_log_path`), conflict `['spell_id','property','observed_on']`.
2. Fresh `Spell::whereIn('id', ...)` (not a constructor-time cache — must see spells created
   earlier in the same run); for each spell missing the bit: set bit, `save()`,
   `$result->addedSpellCounter()`, create `CombatLogSpellEvent`
   (`PropertyChanged`, property, log path). Missing Spell row → skip mask/event, keep the
   observation.
3. Reset all state. (For the expiry guard: lazy memoized per-spell duration lookup.)

### 5. Result DTO
`app/Service/CombatLog/Dtos/DataExtraction/ExtractedDataResult.php`: private
`$addedSpellCounters`, method `addedSpellCounter()`, `toArray()` key — assertable in e2e.

### 6. Staleness (MANDATORY same release as the enum cases, else hourly sweep throws)
`app/Console/Commands/CombatLog/DetectStaleCombatLogDataCommand.php`:
- `removeStaleSpellProperty()` (~line 159): the `match` default currently routes everything
  non-aura/debuff to `getMissTypeBit` (throws `LogicException` on counter cases). Add an
  explicit `isCounter()` branch → `whereRaw('counters_mask & <bit> != 0')`.
- `clearSpellProperty()` (~line 197): counter branch → `counters_mask &= ~bit`.
- New `getCounterBit(SpellProperty): int` mirroring `getMissTypeBit` (~line 210).
- `removeStaleSpellProperties()` iterates `SpellProperty::cases()` — picks up new cases
  automatically; `pruneOldObservations()` unchanged; `PropertyRemoved` events with
  `combat_log_path = null` come for free.

### 7. Display
- New `lang/en_US/spellcounters.php`: `['vanish' => 'Vanish', 'shadowmeld' => 'Shadowmeld']`.
- `resources/views/compendium/npc/sections/spells.blade.php`: "Counters" column after "Miss
  types" — `Spell::maskToReadableString(Spell::ALL_COUNTERS, $spell->counters_mask,
  'spellcounters')`, header + tooltip lang keys.
- `resources/views/compendium/spell/sections/details.blade.php`: same row next to miss types.
- `resources/views/compendium/sections/event_list.blade.php`: in `$spellPropertyName`
  (lines 64–71) add before the miss-type fallback:
  `if ($property->isCounter()) { return __('spellcounters.' . substr($property->value, 8)); }`.
  In `$eventDescription`, branch on `$event->property?->isCounter()` for
  PropertyChanged/PropertyRemoved to use dedicated keys `view_compendium.event.counter_added`
  ("`:spell` can now be countered by `:property`") / `counter_removed` — "gained property
  Vanish" reads poorly. The exhaustive `match` blocks on `event_type` are untouched.
- `lang/en_US/view_compendium.php`: `header_counters` + tooltip (npc + spell sections),
  `event.counter_added`, `event.counter_removed`, plus the class-page section keys below.
  (en_US only.)
- `NpcCompendiumService` unchanged — counter events are ordinary `CombatLogSpellEvent` rows.

### 7b. Class compendium page (`/compendium/class/{class}`) — user-required
The rogue class page must answer "which NPC abilities can I Vanish away?".
- `app/Http/Controllers/Compendium/ClassCompendiumController.php::show()`: from
  `SpellCounterDefinitions::all()`, take definitions where either
  `getCharacterClassKey() === $characterClass->key`, or (racial) `getCharacterRaceKey()` names
  a race this class can actually be — checked against the `character_race_class_couplings`
  pivot via the existing `CharacterRace::classes()` / `CharacterClass::races()` relations
  (eager-load `races` on the class; match on `CharacterRace->key`). So Shadowmeld shows for a
  Rogue (Night Elves can be Rogues) but never for a class the seeded couplings say can't be a
  Night Elf, e.g. Paladin. Racial entries render with a "Racial (<race name>)" badge. For each
  selected counter, query the counterable NPC spells in the context dungeon, mirroring the
  existing characteristics query shape:
  `Spell::query()->whereRaw(sprintf('counters_mask & %d != 0', $bit))` joined through
  `npc_spells` → `npcs` → `enemies` → `mapping_versions` filtered to `$contextDungeon` (and
  current mapping version), selecting the spells with their casting NPCs grouped per counter.
  Pass e.g. `counterSections: Collection<int(bit), {definition, Collection<Spell + npcs>}>`.
- `resources/views/compendium/class/show.blade.php`: new section below the existing
  Spell → Characteristic → NPCs table — per counter: icon + `__('spellcounters.<key>')`
  (+ "Racial (<race name>)" badge for race-based counters), then a table of counterable NPC spells
  (`common.spell.link`) with the NPCs that cast them (`common.npc.link`), empty-state string
  when the context dungeon has none. Layout need not match the existing table exactly.
- Lang keys under `view_compendium.class.show.counters.*` (`title`, `racial`, `no_spells`,
  table headers).

### 8. Seeder round-trip & admin
- `app/Service/Mapping/MappingExportService.php::serializeSpells()`: add `'counters_mask'` to
  the `makeHidden` list (verified: `aura`/`debuff`/`miss_types_mask` already hidden there as
  combat-log-derived; loader side needs nothing — absent key → default 0).
- Admin spell edit: intentionally NOT editable (like `miss_types_mask` — the sweep would
  revert manual edits). State this in the MR body.

## Tests

`tests/Feature/App/Service/CombatLog/DataExtractors/SpellCounterDataExtractorTest.php`,
following `NpcCharacteristicDataExtractorTest.php` exactly: `final ... extends
PublicTestCase`, `#[Group('SpellCounterDataExtractor')]`, mocked logging interface binding,
raw-line fixtures parsed via `new CombatLogEntry($raw)->parseEvent([],
CombatLogVersion::RETAIL_11_0_5)`, an sprintf line factory taking ms offsets (timestamp format
`8/2/2024 16:24:18.477-4`), lifecycle helper before→extract*→after, try/finally teardown
(high test ids, e.g. spells 999xxxx, creature guid with test npc id; clean spells,
observations, spell events; reset `counters_mask`).

Cases (AAA, `[fn]_given[Cond]_returns[Result]`):
- A forward (removal 1 ms before Vanish; fact lands on the **cast** id, not the debuff id).
- A reversed (removal 300 ms after Vanish — two-sided ε).
- B channel (NPC-source debuff, removal at Vanish ts) → counter on channel id.
- Shadowmeld variant → `counter_shadowmeld` bit.
- C happy path (CAST_START → Vanish +1 s → same-guid re-CAST_START +1.6 s, no success).
- C negatives: cast succeeds 550 ms after Vanish; interrupted; caster died; caster disturbed
  (debuff applied mid-cast); re-cast 4 s after Vanish (outside window). Each asserts mask 0 +
  no observation.
- ε negative: removal 2400 ms after Vanish (mirrors real-log control).
- Natural-expiry guard: duration == lifetime → no counter.
- Dedup: same match twice in one file → one observation, one event, `addedSpellCounters == 1`.

Extend `DetectStaleCombatLogDataCommandTest`: stale counter bit cleared +
`PropertyRemoved(counter_vanish)` event; fresh observation → bit retained.

## End-to-end verification (real log, in the worktree Docker stack)

1. Copy the zip into the worktree, run
   `docker compose exec -T app php artisan combatlog:extractdata <path>.zip --force`.
2. Output: `addedSpellCounters >= 3`.
3. DB: spells **1253840**, **1253446**, **1258174** have `counters_mask & 1`;
   `counter_vanish` observation rows for those ids (`observed_on` today); three
   `PropertyChanged`/`counter_vanish` spell events with the log path.
4. Review the FULL list of `counter_vanish` rows — any spell beyond the three expected must be
   vetted against the log by hand (signature C is the likely source of extras).
5. Compendium pages for High Sage Viryx / Solar Construct / Dread Raven: "Counters" column
   shows Vanish; feed renders the counter events without `UnhandledMatchError`. And
   `/compendium/class/rogue` (context dungeon set to Skyreach) lists Vanish with the three
   counterable NPC spells, plus Shadowmeld as "Racial (Night Elf)" (Night Elves can be
   Rogues); a class the couplings say cannot be Night Elf (e.g. `/compendium/class/paladin`)
   must NOT show Shadowmeld. (Headless-browser screenshots per the visual-verification rule since rendered output
   changes.)
6. `composer run fix` + `composer run analyse`; then the standard MR checklist (cold review,
   green CI) before ready-for-review.

Wotuu:
:robot: Closes #3826

Implements the approved plan from the issue: detection and display of player "spell counters" — abilities like Rogue **Vanish** and Night Elf **Shadowmeld** that make an incoming NPC ability fizzle, retarget, or drop its channel.

## What's in here

**Vocabulary & storage** — `spells.counters_mask` bitmask (additive migration, default 0) + `Spell::ALL_COUNTERS`, new `SpellProperty` cases `counter_vanish` / `counter_shadowmeld`. Reuses the entire existing spell-property machinery: observation upserts, `PropertyChanged`/`PropertyRemoved` audit events, and the hourly staleness sweep (which gained counter-bit branches — mandatory, since `getMissTypeBit` would throw on the new enum cases otherwise). `counters_mask` is hidden from the mapping export like `aura`/`debuff`/`miss_types_mask`, and deliberately **not** admin-editable (the sweep would revert manual edits within the 3-day window).

**Extensible definitions** — one class per counter in `app/Service/CombatLog/DataExtractors/SpellCounters/` carrying detection triggers + display metadata (icon, class/race association). Adding a future counter = one definition class + one bit + one enum case + lang entries.

**Detection** — `SpellCounterDataExtractor`, a cross-line state machine recognizing three signatures (all verified against real logs):
- **A** — a nil-source *targeting debuff* (e.g. High Sage Viryx's Lens Flare) dropped at the counter cast; the fact attaches to the NPC's cast spell via same-instant CAST_START linking.
- **B** — an NPC channel debuff (e.g. Solar Construct's Solar Flame) removed at the counter cast, well before natural duration.
- **C** — an in-flight NPC cast (e.g. Dread Raven's Dread Wind) abandoned right after a counter, with no other explanation (interrupt, death, loss-of-control debuff). Only debuffs whose `mechanic` is a loss of control veto this — plain damage debuffs land on trash constantly and would otherwise suppress it entirely.

The detector also assigns the countered spell to its casting NPC (`NpcSpell` + `SpellDungeon` + `SpellAssigned` event): cast-only spells with a nil destination never pass `SpellDataExtractor`'s existing assignment gate, so without this the countered spell would be invisible on its NPC's compendium page.

**Display** — "Counters" column on the NPC compendium spells table and the spell detail page; counter-aware activity-feed wording ("*Lens Flare can now be countered by Vanish*"); a **Counterable abilities** section on the class compendium pages listing, per counter, the counterable spells and their casting NPCs in the context dungeon. Racial counters (Shadowmeld) only appear for classes that can actually be that race, via the `character_race_class_couplings` pivot — Rogues see it, Paladins don't.

**CLI for validation** (per Wotuu's request) — `combatlog:searchruns` and `combatlog:downloadruns` wrap the Raider.IO API (dungeon/class/spec/key-level filters, presigned segment download) so real logs can be fetched for detector validation, now and for future counter types.

## Validation evidence

- **Reference log** (Skyreach, rogue-supplied ground truth): all three reported tricks detected — Lens Flare 1253840 (A), Solar Flame 1253446 (B), Dread Wind 1258174 (C) — and the rogue's fourth, purely offensive mid-fight Vanish correctly produced nothing.
- **Bulk validation on 11 real Raider.IO runs** (93 segments, ~722 MB; 6 runs with rogues, 5 without as negative controls): 6 additional detections, 5 of them verified true against the raw logs (including a Shadowmeld-drops-Fixate in a no-rogue party, from a German-locale client — id-based trigger matching). The one false positive (a ~1 s churning AoE debuff whose natural removal landed 392 ms from an unrelated Vanish) drove tightening the correlation window from 500 ms to 100 ms: every verified true positive correlates within 0–1 ms (same server tick). After tightening: 8/8 true facts detected, 0 false positives.
- Detected counters across the validation set: Lens Flare, Solar Flame, Dread Wind, Repel (Skyreach), Ethereal Shackles, Terror Wave (Magisters' Terrace), Gunk Splatter ~~(Windrunner Spire)~~ *rejected as false positive*, Devouring Rift (Xal'atath — no NPC link since summoned NPCs have no `npcs` row, by design), Fixate → Shadowmeld (Nexus-Point Xenas).

## Tests

- `SpellCounterDataExtractorTest`: 16 cases / 86 assertions — all three signatures forward+reversed ordering, Shadowmeld, every C exclusion (success, interrupt, death, stun, window), ε boundary, natural-expiry guard, dedup, NPC assignment.
- Staleness command counter cases, class-page rogue/paladin rendering tests, CLI command tests (mocked Raider.IO service). 124 tests / 360 assertions green across the affected suites; `composer run fix` + `composer run analyse` clean.

Screenshots follow in a comment. The trivial-change rule does not apply (migration + UI); cold review to follow before this leaves draft.


Wotuu:
:robot: Visual verification (worktree stack, admin session, Skyreach dungeon context; all pages 200, no page errors):

**High Sage Viryx NPC page** — Lens Flare now listed with Counters: Vanish, feed shows the counter event:

viryx

**Rogue class page** — Counterable abilities section with Vanish + Shadowmeld (Racial, Night Elf):

rogue

**Paladin class page** — negative control: no Shadowmeld (Paladins cannot be Night Elves per the seeded race/class couplings):

paladin

Note: these pages did not exist in this form on master (purely additive sections), so no A/B baseline pair is included.

app/Service/CombatLog/DataExtractors/SpellCounterDataExtractor.php
```diff
@@ -0,0 +1,775 @@
...
+        if ($npc === null) {
+            return;
+        }
+
+        $spellId = $observation['spell_id'];
+        if ($npc->npcSpells->filter(fn(NpcSpell $npcSpell) => $npcSpell->spell_id === $spellId)->isNotEmpty()) {
+            return;
+        }
+
+        NpcSpell::create([
```
Wotuu, line 408: :robot: `assignSpellToNpc()` skips the gate `NpcSpellAssignmentCollector::collect()` applies before creating an `NpcSpell`: there, a spell is only ever assigned when `$spell->category === spellcategory.unknown`. Without that gate, a misdetection here can attach an already-categorized (player/CC) spell to an NPC. That matters because — unlike the `counters_mask` bit, which the observation-staleness sweep self-corrects — `NpcSpell`/`SpellDungeon` rows created here are permanent: `combatlog:detectstaledata` never removes NpcSpell rows, and these tables are combat-log-derived (not recoverable from seeders). The docblock on `isCastDisturbingDebuff()` claims "the observation staleness window self-corrects rare misattributions", but that is only true for the mask bit, not for the NpcSpell/SpellDungeon side effects of the same false positive. Consider adding the same category-unknown gate here (all three signatures attribute NPC-cast spells, so a categorized spell id showing up is itself a misfire signal).

app/Service/CombatLog/DataExtractors/SpellCounterDataExtractor.php
```diff
@@ -0,0 +1,775 @@
...
+            }
+
+            $fallback ??= ['spellId' => $npcCastStart['spellId'], 'npcId' => $npcCastStart['npcId']];
+        }
+
+        return $fallback;
+    }
+
+    /**
+     * @return int|null The spell's duration in milliseconds, or null when the spell is unknown or has no duration
```
Wotuu: :robot: Stranded docblock: this `@return int|null The spell's duration...` block belongs to `getSpellDurationMs()` below, but it sits directly above the docblock for `isCastDisturbingDebuff()`, leaving that method double-documented and `getSpellDurationMs()` without its intended doc. Move it down (or drop it).

app/Models/CombatLog/SpellProperty.php
```diff
@@ -6,22 +6,34 @@
...
+    case MissInterrupt     = 'miss_interrupt';
+    case CounterVanish     = 'counter_vanish';
+    case CounterShadowmeld = 'counter_shadowmeld';
 
     public static function fromMissTypeBit(int $bit): self
     {
         return self::from(sprintf('miss_%s', Spell::ALL_MISS_TYPES[$bit]));
     }
+
+    public static function fromCounterBit(int $bit): self
```
Wotuu: :robot: `fromCounterBit()` has no callers anywhere in app/tests (the miss-type analogue `fromMissTypeBit()` is used by `SpellPropertyObservationCollector`, but the counter path queues its `SpellProperty` directly from the definition). Either drop it, or wire it in where it was intended — dead public API on an enum tends to linger.

app/Http/Controllers/Compendium/ClassCompendiumController.php
```diff
@@ -56,6 +59,65 @@ public function show(CharacterClass $characterClass): View
...
+                || ($definition->getCharacterRaceKey() !== null && $classRaceKeys->contains($definition->getCharacterRaceKey())),
+        );
+
+        $result = [];
+
+        foreach ($definitions as $definition) {
+            $bit = $definition->getCounterBit();
+
+            $spells = Spell::query()
+                ->whereRaw(sprintf('counters_mask & %d != 0', $bit))
```
Wotuu, line 90: :robot: Scope mismatch with the copy: this spells query is game-version-wide (any dungeon), while the section empty-state string says "No counterable NPC spells found for this dungeon." (`view_compendium.class.show.counters.no_spells`). Once any spell in the game version carries the bit, that string is unreachable and spells countered only in *other* dungeons render as rows with "-" NPCs on every dungeon context. Either scope `$spells` to the context dungeon via `spell_dungeons` (matching the "in the context dungeon" framing in the PR body), or reword the lang string to match the game-version-wide behavior.

app/Service/CombatLog/DataExtractors/SpellCounterDataExtractor.php
```diff
@@ -0,0 +1,775 @@
...
+
+        // Signature A attributes the fact to the NPC's cast, signature B to the debuff itself
+        $attributedSpellId = $activeDebuff['linkedNpcCastSpellId'] ?? $spellId;
+        $signature         = $activeDebuff['linkedNpcCastSpellId'] === null ? 'B' : 'A';
+        $observedLifetime  = $this->millisecondsBetween($activeDebuff['appliedAt'], $timestamp);
+
+        // The duration is that of the debuff that was actually removed, not of the cast it is attributed to
+        $duration = $this->getSpellDurationMs($spellId);
+        if ($duration !== null && $duration > 0 && $observedLifetime >= $duration - self::EPSILON_MS) {
+            $this->log->extractDataDebuffExpiredNaturally($spellId, $observedLifetime, $duration);
```
Wotuu, line 607: :robot: SPELL_AURA_REFRESH is not handled (`AuraRefresh extends AuraBase` directly, so it matches neither the applied nor removed branch), which means `appliedAt` stays at the *initial* application when a debuff is refreshed. A refreshed debuff that is then genuinely countered will show `observedLifetime >= duration` and be rejected here as natural expiry — a false negative only (no false-positive risk), so this may be acceptable, but worth either handling refresh by bumping `appliedAt`, or noting the limitation in the natural-expiry docblock.

Wotuu:

app/Service/CombatLog/DataExtractors/SpellCounterDataExtractor.php
Wotuu: :robot: Fixed in 419a04a2e — the gate now rejects categorized player spells. Deliberately unlike the collector, a **null** category stays assignable: spells first created from a combat log have no category until their Wowhead data is fetched (the collector's own gate silently skips all such new spells — pre-existing upstream limitation), and the countered spell is typically exactly such a spell. Pinned by extractData_givenCounteredSpellWithPlayerCategory_doesNotAssignSpellToNpc.

app/Service/CombatLog/DataExtractors/SpellCounterDataExtractor.php
Wotuu: :robot: Fixed in 419a04a2e — docblock moved back onto getSpellDurationMs().

app/Models/CombatLog/SpellProperty.php
Wotuu: :robot: Fixed in 419a04a2e — fromCounterBit() removed.

app/Http/Controllers/Compendium/ClassCompendiumController.php
Wotuu: :robot: Fixed in 419a04a2e — the spells query is now scoped to the context dungeon via spell_dungeons, matching the section copy. (The detector also creates SpellDungeon rows for countered spells now, so in-dungeon spells reliably qualify.)

app/Service/CombatLog/DataExtractors/SpellCounterDataExtractor.php
Wotuu: :robot: Fixed in 419a04a2e — SPELL_AURA_REFRESH now restarts the tracked appliedAt, with a regression test (refresh at 5s of a 6s debuff, counter at 8s → detected).

lang/en_US/view_compendium.php
```diff
@@ -92,6 +94,8 @@
                 'header_schools_tooltip'     => 'What type of damage does this spell do?',
                 'header_miss_types'          => 'Miss types',
                 'header_miss_types_tooltip'  => 'What can you do to avoid this spell?',
+                'header_counters'            => 'Counters',
+                'header_counters_tooltip'    => 'Player abilities that can make this spell fizzle or retarget (observed from combat logs).',
```
Wotuu: Remove the observed from combat logs.

app/Console/Commands/CombatLog/DownloadCombatLogRunsCommand.php
```diff
@@ -0,0 +1,202 @@
...
+        $response = $this->raiderIOApiService->searchAdvancedRuns($filter);
+
+        return array_map(fn(SearchAdvancedRun $run): int => $run->id, $response->runs);
+    }
+
+    /**
+     * Derive the file extension from the (presigned) download URL, mirroring
+     * {@see \App\Jobs\CombatLog\ProcessCombatLogSegments::resolveSegmentExtension()}.
+     */
+    private function resolveSegmentExtension(string $downloadUrl): string
```
Wotuu: If these two functions mirror other functions, can we put them in a generic place where they can be shared, or is that too difficult?

lang/en_US/view_compendium.php
Wotuu: :robot: Fixed in b9c67eec8 — dropped "(observed from combat logs)" from the counters tooltip on both the NPC and class spell compendium sections (`view_compendium.php` lines 98 and 135, which carried the identical string).

app/Console/Commands/CombatLog/DownloadCombatLogRunsCommand.php
Wotuu: :robot: Fixed in b9c67eec8 — `resolveSegmentExtension()` and `isPlausibleSegment()` were exact duplicates of the same-named private methods on `App\Jobs\CombatLog\ProcessCombatLogSegments`. Extracted both into a new `App\Service\Traits\CombatLogSegmentFile` trait (following the existing `Service/Traits/Curl` convention) and used it from both classes.

PR #3838, opened 2026-08-04
Exempt heatmap embed data endpoint from CSRF verification

### LINKED ISSUE #3836: Heatmap embed: CSRF token mismatch loading data inside iframe
:robot:

## Problem

Loading a heatmap in embed mode (`/heatmap/{gameVersion}/{dungeon}/embed/{floorIndex}`) reliably fails to load its data. The `POST /ajax/heatmap/data` request (`AjaxHeatmapController::getData`, routed in `routes/web.php:592`) comes back with:

```json
{
    "message": "CSRF token mismatch."
}
```

This happens consistently for embedded views, not just occasionally.

## Suspected root cause

The embed page is, by definition, meant to be loaded inside an `<iframe>` on a third-party site (that's the whole point of "embed"). Both the session cookie and the `XSRF-TOKEN` cookie default to `SameSite=Lax` (`config/session.php:207`, `env('SESSION_SAME_SITE', 'lax')`), and there's no override for the embed routes.

`SameSite=Lax` cookies are sent on a top-level cross-site navigation (so the initial iframe `GET` that renders the page works fine and gets a valid session + CSRF token), but they are **not** sent for subsequent same-origin `fetch`/`XHR` requests issued from *inside* a cross-site iframe — the browser evaluates "same-site" relative to the top-level browsing context, not the iframe's own origin. So the follow-up `POST /ajax/heatmap/data` call is sent without the session cookie, the session (and the CSRF token tied to it) is effectively anonymous/new server-side, and `VerifyCsrfToken` middleware rejects the request.

This would explain why the failure is consistent rather than intermittent, and why it's specific to the embed flow (the non-embed heatmap page is loaded top-level, not iframed, so it doesn't hit this).

## Where to look

- `config/session.php:207` — `same_site` config, currently `lax` app-wide.
- `app/Http/Controllers/Dungeon/DungeonHeatmapController.php` — `embed()` method, and how the CSRF token is bootstrapped into the embed view.
- `resources/views/dungeon/heatmap/gameversion/embed.blade.php` — embed page markup/meta tags.
- `routes/web.php:592` — `ajax.heatmap.data` route, currently under the default `web` middleware group (session + CSRF enforced).
- `app/Http/Controllers/Ajax/**/AjaxHeatmapController.php` — the endpoint returning the error.

- Set `SameSite=None; Secure` specifically for the embed routes/cookies (would need `Secure` since `SameSite=None` requires HTTPS), rather than globally weakening the site's CSRF/session cookie posture.
- Move the embed data endpoint off session-based CSRF entirely (e.g. token-based auth using the existing `token` query param already passed through the embed URL, bypassing `VerifyCsrfToken` for this specific route) since embed views are already designed to carry a `token` for data scoping.
- Confirm in a real cross-site iframe (not same-origin dev testing) that this is actually what's happening — check `Set-Cookie` headers' `SameSite` attribute and whether the cookie is present on the `POST /ajax/heatmap/data` request in the Network tab.

## Definition of done

- Embedding a heatmap on a real third-party origin (e.g. a local static HTML page with an `<iframe src="https://keystone.guru/heatmap/...">`) loads data successfully without a CSRF error.
- Existing CSRF protection for non-embed routes is untouched.
- Test coverage added for whatever fix is chosen.

Wotuu:
:robot:

Closes #3836

## Problem

`POST /ajax/heatmap/data` returns `419 CSRF token mismatch` when called from a heatmap embed page, because the embed is designed to be loaded in a cross-site `<iframe>`. `SESSION_SAME_SITE=lax` (the default) means the session/`XSRF-TOKEN` cookies are sent on the initial top-level iframe navigation, but not on the subsequent same-origin XHR issued from *inside* that cross-site iframe — so the token the page renders never reaches the server.

## Fix

Add `ajax/heatmap/data` to the CSRF exemption list in `bootstrap/app.php`, alongside the existing `webhook/*` exemption. `AjaxHeatmapController::getData()` is a read-only, unauthenticated data fetch (no state mutation, no auth), so exempting it from CSRF carries no meaningful risk — same reasoning as the `pr cold reviewed` exceptions applied elsewhere for genuinely side-effect-free endpoints.

## Note: this has already been hotfixed to production (v15.9.1)

Given the severity (heatmap embeds are broken for all viewers) this was pushed out via the `hotfix` skill ahead of this PR landing, with explicit go-ahead in-conversation. This PR lands the same change permanently on `master` so the next release includes it and the hotfix key doesn't need to be re-applied.

- RED baseline confirmed against production before the hotfix: `POST /ajax/heatmap/data` → `419 {"message":"CSRF token mismatch."}`
- Uploaded surgically via `--file=bootstrap/app.php` (verified via `Storage::disk('s3_hotfixes')->allFiles('v15.9.1')`)
- Production container restart being done manually by @Wotuu; GREEN re-verification against production pending that restart.

## Testing

- Added `tests/Unit/App/Http/Middleware/CsrfExemptRoutesTest.php` asserting `ajax/heatmap/data` (and the pre-existing `webhook/*`) are present in `ValidateCsrfToken::getExcludedPaths()` after boot.
- Existing `tests/Feature/Controller/Ajax/AjaxHeatmapControllerTest.php` still green (CSRF is bypassed for feature tests regardless, so it only confirms no regression to the endpoint itself).

Cold review not yet done — @Wotuu asked for Fable to review this one given it's a security-relevant middleware change. Left as draft until that happens.

### DIFF
--- a/bootstrap/app.php
+++ b/bootstrap/app.php
@@ -48,10 +48,17 @@ return Application::configure(basePath: dirname(__DIR__))
         $middleware->redirectGuestsTo(fn() => route('login'));
         $middleware->redirectUsersTo('/home');
 
-        // Only the external webhook endpoints (called by GitHub / Wowhead without a session)
-        // are exempt from CSRF verification; every other route sends a token.
+        // Only the external webhook endpoints (called by GitHub / Wowhead without a session), and
+        // the heatmap embed data endpoint, are exempt from CSRF verification; every other route
+        // sends a token. The heatmap embed page is designed to be loaded in an <iframe> on
+        // third-party sites, where SameSite=Lax session/XSRF cookies aren't sent on the
+        // same-origin XHR issued from inside that cross-site iframe, so the token never arrives.
+        // ajax.heatmap.data is a read-only, unauthenticated data fetch (see
+        // AjaxHeatmapController::getData()), so exempting it from CSRF carries no state-changing
+        // risk (see #3836).
         $middleware->validateCsrfTokens(except: [
             'webhook/*',
+            'ajax/heatmap/data',
         ]);
 
         // Prepend so every log line of the request - including those of other global middleware - carries the trace_id
--- a/tests/Unit/App/Http/Middleware/CsrfExemptRoutesTest.php
+++ b/tests/Unit/App/Http/Middleware/CsrfExemptRoutesTest.php
@@ -1,0 +1,41 @@
+<?php
+
+namespace Tests\Unit\App\Http\Middleware;
+
+use Illuminate\Foundation\Http\Middleware\ValidateCsrfToken;
+use PHPUnit\Framework\Attributes\Group;
+use PHPUnit\Framework\Attributes\Test;
+use Tests\TestCases\PublicTestCase;
+
+#[Group('Middleware')]
+#[Group('CsrfExemptRoutes')]
+class CsrfExemptRoutesTest extends PublicTestCase
+{
+    #[Test]
+    public function getExcludedPaths_givenBootedApplication_includesHeatmapEmbedDataRoute(): void
+    {
+        // Arrange
+        /** @var ValidateCsrfToken $middleware */
+        $middleware = $this->app->make(ValidateCsrfToken::class);
+
+        // Act
+        $excludedPaths = $middleware->getExcludedPaths();
+
+        // Assert
+        self::assertContains('ajax/heatmap/data', $excludedPaths);
+    }
+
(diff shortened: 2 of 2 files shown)


Wotuu:
:robot: Superseded by the proper fix in #3840, which makes `ajax.heatmap.data` a GET route instead of adding a CSRF exemption — the exemption in this PR is then never needed on `master`. Production stays covered by the v15.9.1 hotfix overlay until the next release ships #3840.

PR #3839, opened 2026-08-05
#3837 Keep non-filter query params in the URL when the heatmap sidebar syncs its search

### LINKED ISSUE #3837: Heatmap embed URL loses non-filter params (defaultZoom, showHeader, ...) after client-side search sync
:robot:

## Problem

When viewing a heatmap embed, the server-side redirect (e.g. `/heatmap/retail/skyreach/embed/1?...` → `/embed/3?...` when the requested floor index doesn't match the resolved floor) correctly preserves every query parameter. But shortly after the page loads, client-side JS silently rewrites the address bar to a URL missing several of the original params. Pressing F5 at that point reloads from the truncated URL and the page comes up without those settings applied (wrong zoom, header visible again, etc).

Example - what the server sends the browser to:

```
.../heatmap/retail/skyreach/embed/3?type=player_death&dataType=enemy_position&minMythicLevel=2&maxMythicLevel=99&minPeriod=1056&maxPeriod=1075&maxTimerFraction=999&token=...&season=season-mn-1&mapFacadeStyle=facade&showSidebar=0&showHeader=0&showDataSourceSnackbar=0&defaultZoom=1.4
```

What ends up in the address bar shortly after (and what F5 then applies):

```
.../heatmap/retail/skyreach/embed/3?type=player_death&dataType=enemy_position&maxMythicLevel=99&minPeriod=1056&maxPeriod=1075&maxTimerFraction=999&showSidebar=0&token=...&season=season-mn-1
```

Missing: `defaultZoom`, `showHeader`, `minMythicLevel`, `showDataSourceSnackbar`, `mapFacadeStyle`.

## Root cause

`SearchInlineBase._updateUrl()` (`resources/assets/js/custom/inline/base/searchinlinebase.js:115-137`), invoked from `CommonMapsHeatmapsearchsidebar._search()` (`resources/assets/js/custom/inline/common/maps/heatmapsearchsidebar.js:303-317`) on the initial search, throws away the *entire* existing query string and rebuilds it from scratch via `history.pushState`, using only `SearchParams.params` (`resources/assets/js/custom/searchparams.js`).

`SearchParams` only knows about registered *search filters* (type, dataType, mythic level range, periods, etc.) - it has no concept of embed/display-only options, so:

- `defaultZoom`, `showHeader`, `showDataSourceSnackbar`, `mapFacadeStyle` aren't filters at all, so they're never added back in and vanish the moment `_updateUrl` runs.
- `minMythicLevel` *is* a filter, but `SearchParams` explicitly excludes any filter value equal to `filter.getDefaultValueOverride(key)`:
  ```js
  if (paramsOverride.hasOwnProperty(key) && paramsOverride[key] !== filter.getDefaultValueOverride(key)) {
      this.params[key] = paramsOverride[key];
  }
  ```
  Since the season's default min mythic level happens to also be `2`, the explicit `minMythicLevel=2` from the URL is indistinguishable from "not set" and gets dropped, while `maxMythicLevel=99` survives because it differs from that filter's default.

So the in-memory page state is correct on first load (rendered with the right values), but the URL has already been silently truncated - the bug only becomes visible on a subsequent reload (F5, or sharing/reusing the copied embed URL).

## Where to look

- `resources/assets/js/custom/inline/base/searchinlinebase.js` - `_updateUrl()` (rebuilds URL from scratch instead of merging).
- `resources/assets/js/custom/searchparams.js` - `SearchParams` constructor (drops filter values equal to their default).
- `resources/assets/js/custom/inline/common/maps/heatmapsearchsidebar.js` - `_search()`, the call site for the heatmap embed/explore sidebar.

1. Make `_updateUrl` preserve any existing query params that aren't owned by a known filter, instead of wholesale replacing `window.location.search`.
2. And/or register the embed display options (`defaultZoom`, `showHeader`, `showSidebar`, `showDataSourceSnackbar`, `mapFacadeStyle`, `style`, `headerBackgroundColor`, `mapBackgroundColor`, `showEnemyInfo`, `showTitle`) as pass-through params in `SearchParams`/the sidebar's filter set so they round-trip.
3. For the `minMythicLevel`-style default-value suppression: only treat a value as "unset" if it was never explicitly present in the URL/initial state, not just because it currently equals the default.

## Definition of done

- Loading a heatmap embed URL with explicit `defaultZoom`/`showHeader`/`showDataSourceSnackbar`/`mapFacadeStyle`/`minMythicLevel` (even when the latter equals the season default) keeps all of them in the address bar after the sidebar's initial search runs.
- Refreshing (F5) the page reproduces the same rendered state as the original URL.
- Test coverage added (there's already a `resources/assets/js/custom/inline/common/maps/heatmapsearchsidebar.urlrestore.test.js` covering related URL-restore behavior - extend it).

Wotuu:
:robot:

Closes #3837

## What was wrong

`SearchInlineBase._updateUrl()` threw away the entire existing query string and rebuilt it from the
search filters via `history.pushState`. A heatmap embed URL carries display-only options next to the
filters, and those aren't filters at all — so `defaultZoom`, `showHeader`, `showDataSourceSnackbar`
and `mapFacadeStyle` vanished from the address bar the moment the sidebar's initial search ran.

Separately, `SearchParams` omits any params-override value equal to `filter.getDefaultValueOverride()`.
An explicit `?minMythicLevel=2` on a season whose own minimum is also 2 is then indistinguishable from
"not set", so it was dropped too (while `maxMythicLevel=99` survived, because it differs from its default).

The page rendered correctly on first load — the bug only surfaced on F5, or when the (already truncated)
URL was copied and reused.

## What changed

- `_updateUrl()` carries over every query parameter that none of the filters own, instead of replacing
  the whole query string. Ownership is computed from each filter's `getParamsOverride()` keys, or its
  name (plus `name[]`) when it has none. **Disabled** filters count as owners, so their params are still
  dropped from the URL rather than preserved — the existing `dataType`-on-player-deaths behaviour is
  unchanged.
- Restoring a params-override value from the URL now marks it explicit (`SearchFilter.markValueOverrideExplicit()`),
  and `SearchParams` only applies the equals-default suppression to values that were *not* explicitly asked for.
- "Ours" also covers every param name `_updateUrl()` has written itself, not just the filters' — `_search()` merges
  caller-supplied `queryParameters` into the URL too, and those would otherwise be preserved forever once written
  (`npc_id` on the enemy failures search). Preserved keys are re-encoded, since `getQueryParams()` hands them back
  decoded.

The display-only options deliberately stay out of the filter set, so they round-trip through the URL
without being posted to the heatmap endpoint (asserted by a test).

## Testing

Extended `resources/assets/js/custom/inline/common/maps/heatmapsearchsidebar.urlrestore.test.js` with a
second suite (7 tests). The real `SearchFilterMythicLevel`/`SearchFilterKeyLevel` are now wired into the
harness instead of being stubbed — it is the one filter there that claims its URL params through
`getParamsOverride()`, which is where the `minMythicLevel` truncation lived.

New `resources/assets/js/custom/inline/base/searchinlinebase.urlsync.test.js` (5 tests) pins the ownership
boundary on the base class the four search inlines share: preserved / filter-cleared / injected-and-cleared /
blacklisted / reserved-character keys.

- Reverting only the source files makes 4 of the heatmap tests fail; dropping the written-params ownership
  makes the injected-param test fail. With the fix all 23 in the two files pass.
- Full JS suite green: 364 tests / 40 files.

## Verification

The heatmap page itself can't be opened in this dev stack (`App\Features\Heatmap` requires
`opensearch-laravel.host` to be configured when `APP_ENV=local`, so every heatmap URL redirects to the
dungeon-select page). Instead the built bundle was driven directly in real headless Chrome on
`/routes/retail/pit-of-saron?title=abc&someUnknownParam=keepme&minMythicLevel=2&maxMythicLevel=99`,
constructing a `SearchInlineBase` with a title filter and a `SearchFilterMythicLevel(2, 99)`, restoring
from the URL and calling `_updateUrl()`:

| bundle | resulting `window.location.search` |
| --- | --- |
| before | `?title=abc` |
| after | `?someUnknownParam=keepme&title=abc&minMythicLevel=2&maxMythicLevel=99` |

No rendered output changed, so no screenshots.

### DIFF
--- a/resources/assets/js/custom/inline/base/searchinlinebase.js
+++ b/resources/assets/js/custom/inline/base/searchinlinebase.js
@@ -7,6 +7,8 @@ class SearchInlineBase extends InlineCode {
         this.searchHandler = searchHandler;
         // Previous search params are used to prevent searching for the same thing multiple times for no reason
         this._previousSearchParams = null;
+        // Every query param name we've written into the URL ourselves - see _updateUrl()
+        this._writtenUrlParamNames = [];
         this.filters = {};
     }
 
@@ -142,9 +144,12 @@ class SearchInlineBase extends InlineCode {
     /**
      * Updates the URL according to the passed searchParams (so users can press F5 and be where they left off, ish)
      *
-     * Query parameters that none of the filters own (embed display options such as defaultZoom or showHeader) are
-     * carried over untouched - rebuilding the query string from the filters alone silently truncated shared embed
-     * URLs (#3837).
+     * Query parameters that we don't own (embed display options such as defaultZoom or showHeader) are carried over
+     * untouched - rebuilding the query string from the filters alone silently truncated shared embed URLs (#3837).
+     *
+     * "Ours" is everything the filters claim plus everything a previous call wrote: a param the caller injected
+     * through `queryParameters` (npc_id on the enemy failures search) would otherwise be preserved forever once it
+     * had been written, even after the search stopped sending it.
      *
      * @param searchParams {SearchParams}
      * @param blacklist {Array}
@@ -159,18 +164,23 @@ class SearchInlineBase extends InlineCode {
         blacklist.push('limit');
 
         // Everything we're not responsible for stays exactly as it was
-        let ownedParams = this._getFilterParamNames().concat(blacklist);
+        let ownedParams = this._getFilterParamNames().concat(blacklist).concat(this._writtenUrlParamNames);
         let existingParams = getQueryParams();
         for (let index in existingParams) {
             if (existingParams.hasOwnProperty(index) && !ownedParams.includes(index) &&
                 !searchParams.params.hasOwnProperty(index)) {
-                urlParams.push(`${index}=${encodeURIComponent(existingParams[index])}`);
+                // The key came out of getQueryParams() decoded, so it needs re-encoding just like the value
+                urlParams.push(`${encodeURIComponent(index)}=${encodeURIComponent(existingParams[index])}`);
             }
         }
 
         for (let index in searchParams.params) {
             if (searchParams.params.hasOwnProperty(index) && !blacklist.includes(index)) {
                 urlParams.push(`${index}=${encodeURIComponent(searchParams.params[index])}`);
+
+                if (!this._writtenUrlParamNames.includes(index)) {
+                    this._writtenUrlParamNames.push(index);
+                }
             }
         }
 
--- a/resources/assets/js/custom/inline/base/searchinlinebase.js
+++ b/resources/assets/js/custom/inline/base/searchinlinebase.js
@@ -50,6 +50,9 @@ class SearchInlineBase extends InlineCode {
                     if (paramsOverride !== null && filter.getParamsOverride().hasOwnProperty(filtersKey)) {
                         // It does! Set the value
                         filter.setValueOverride(filtersKey, queryParams[key]);
+                        // The URL asked for this value explicitly, so it must survive a round trip even when it
+                        // happens to equal the filter's default (#3837)
+                        filter.markValueOverrideExplicit(filtersKey);
                         valueAssigned = true;
                         break;
                     }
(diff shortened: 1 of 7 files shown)


Wotuu:
:robot: Cold review: 2 findings

Both fixed in 4fa17f9.

1. **Injected `queryParameters` became sticky in the URL (real regression, non-heatmap consumer).** `_getFilterParamNames()` derived ownership from `this.filters` only, but `_search()` also merges caller-supplied `queryParameters` into what gets written. `CommonMapsCombatlogrouteenemyfailures` injects `npc_id` only while an npc is selected and blacklists nothing — so selecting then clearing an npc left `npc_id=123` in the address bar permanently, with nothing to restore it on F5. `_updateUrl()` now also treats every param name it has written itself as ours. Pinned by `search_givenAnInjectedParamIsNoLongerSent_dropsItFromTheUrl`, which fails without the fix.
2. **Preserved param *keys* weren't URL-encoded.** They come out of `getQueryParams()` decoded but the new loop wrote them back raw, so a key containing `&`/`=`/`#` split into a different param set than it came in as. The pre-existing loop got away with this because its keys are known filter names. Pinned by `search_givenAPreservedParamWithReservedCharacters_writesItBackEncoded`.

Two observations the reviewer raised and I agree don't need action:

- `explicitValueOverrides` is never cleared, so a restored `minMythicLevel=2` keeps being sent even after the user moves the slider back to the default. `HeatmapDataFilter::fromRequest()` maps that to `setKeyLevelMin(2)` where it previously got `null` — semantically identical, since 2 *is* the season minimum; only the cache key differs.
- `minSamplesRequired` is now preserved in the URL for non-admins (the filter is only constructed for admin/internal-team). Cosmetic, and arguably better for sharing.

Also noted while fixing: clearing the *last* remaining filter still leaves the old query string in place, because `_updateUrl()` skips `pushState` when the rebuilt query string is empty. That is pre-existing and deliberate, so it's left alone.

PR #3840, opened 2026-08-05
#3836 Make the heatmap data endpoint a GET so cross-site embeds work without a CSRF exemption

### LINKED ISSUE #3836: Heatmap embed: CSRF token mismatch loading data inside iframe

Wotuu:
:robot:

Closes #3836

## Problem

The heatmap embed page runs in a cross-site `<iframe>`, where `SameSite=Lax` session/`XSRF-TOKEN` cookies are not sent on the XHR it issues — so `POST /ajax/heatmap/data` always failed CSRF verification with a 419. This was emergency-hotfixed to production (v15.9.1, PR #3838) by exempting the path from CSRF verification; this PR is the proper fix and supersedes #3838.

## Fix: make `ajax.heatmap.data` a GET route

`AjaxHeatmapController::getData()` is a stateless, read-only, unauthenticated data fetch (no `Auth`, no session reads, no writes — it forwards validated filters to the Raider.IO API). Laravel only verifies CSRF on POST/PUT/PATCH/DELETE, so as a GET the endpoint is structurally outside CSRF verification — **no exemption list entry needed at all**, and the site's CSRF posture stays exactly as it was (`webhook/*` remains the only exemption). This also matches the codebase's own convention: every other read-only `/ajax/*` search endpoint is already GET (`/ajax/routes`, `/ajax/search`, `/ajax/tag`, mapcontext); this one was the outlier.

Changes:
- `routes/web.php` — `Route::post` → `Route::get` for `ajax.heatmap.data`, with a comment explaining why GET is load-bearing here.
- `searchhandlerheatmap.js` — `type: 'POST'` → `type: 'GET'`. jQuery serializes the same filter object into the query string (`includeAffixIds[]=…`), so `AjaxGetDataFormRequest`'s array rules validate unchanged.
- `ReadOnlyMode` middleware — dropped the `ajax/heatmap/data` whitelist entry; that middleware already short-circuits on GET, so it became dead code.
- No `bootstrap/app.php` change: the hotfix exemption never landed on master (#3838 stayed draft), so there is nothing to revert there.

## Verification

Cross-site iframe end-to-end in headless Chrome: an opaque-origin host page embedding `/heatmap/retail/operation-mechagon-workshop/embed/5` in an `<iframe>` — the exact SameSite condition that broke production. The embed's XHR is now `GET /ajax/heatmap/data` → **200**, heatmap renders with data, no page errors:

cross-site embed loads data

(The same harness run against the old POST bundle reproduced the failure mode, and the RED baseline against production — POST → 419 — was captured in #3838.)

## Tests

- `AjaxHeatmapControllerTest` — both data tests switched to GET with query-string params, proving the FormRequest still validates them.
- New `HeatmapDataCsrfRegressionTest` — asserts `ajax/heatmap/data` is **not** in `ValidateCsrfToken::getExcludedPaths()` (only `webhook/*` is), guarding against the hotfix exemption quietly returning, and that POST to the route now 405s.
- `composer run fix` clean, `composer run analyse` clean.

## Rollout note

The JS change ships in release assets, so the fix takes effect with the next release; the v15.9.1 hotfix overlay keeps production working until then (hotfix overlays are keyed per version tag, so the next release sheds it automatically).

## Follow-up (out of scope)

`POST /ajax/dungeonroute/search/{gameVersion}/{dungeon}` is the same read-only-search-as-POST pattern and would break identically if ever iframed — worth its own issue.

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -601,9 +601,6 @@ Route::middleware(['viewcachebuster', 'language', 'debugbarmessagelogger', 'read
             Route::delete('/{tag}/all', new AjaxTagController()->deleteAll(...))->name('ajax.tag.deleteall');
         });
         Route::prefix('heatmap')->group(static function () {
-            // GET on purpose: the heatmap embed page calls this from inside a cross-site <iframe>,
-            // where SameSite=Lax session/XSRF cookies aren't sent on the XHR, so a POST would
-            // always fail CSRF verification (#3836). The endpoint is a stateless read-only fetch.
             Route::get('/data', new AjaxHeatmapController()->getData(...))->name('ajax.heatmap.data');
         });
 
--- a/routes/web.php
+++ b/routes/web.php
@@ -601,7 +601,10 @@ Route::middleware(['viewcachebuster', 'language', 'debugbarmessagelogger', 'read
             Route::delete('/{tag}/all', new AjaxTagController()->deleteAll(...))->name('ajax.tag.deleteall');
         });
         Route::prefix('heatmap')->group(static function () {
-            Route::post('/data', new AjaxHeatmapController()->getData(...))->name('ajax.heatmap.data');
+            // GET on purpose: the heatmap embed page calls this from inside a cross-site <iframe>,
+            // where SameSite=Lax session/XSRF cookies aren't sent on the XHR, so a POST would
+            // always fail CSRF verification (#3836). The endpoint is a stateless read-only fetch.
+            Route::get('/data', new AjaxHeatmapController()->getData(...))->name('ajax.heatmap.data');
         });
 
         Route::middleware('throttle:create-reports')->group(static function () {
--- a/tests/Unit/App/Http/Middleware/HeatmapDataCsrfRegressionTest.php
+++ b/tests/Unit/App/Http/Middleware/HeatmapDataCsrfRegressionTest.php
@@ -1,0 +1,45 @@
+<?php
+
+namespace Tests\Unit\App\Http\Middleware;
+
+use Illuminate\Foundation\Http\Middleware\ValidateCsrfToken;
+use PHPUnit\Framework\Attributes\Group;
+use PHPUnit\Framework\Attributes\Test;
+use Tests\TestCases\PublicTestCase;
+
+/**
+ * Guards the proper fix for #3836: ajax.heatmap.data is a GET route (structurally outside CSRF
+ * verification, safe to call from the cross-site heatmap embed <iframe>), and the temporary
+ * hotfix CSRF exemption for it must not quietly come back.
+ */
+#[Group('Middleware')]
+#[Group('HeatmapDataCsrfRegression')]
+class HeatmapDataCsrfRegressionTest extends PublicTestCase
+{
+    #[Test]
+    public function getExcludedPaths_givenBootedApplication_doesNotIncludeHeatmapDataRoute(): void
+    {
+        // Arrange
+        /** @var ValidateCsrfToken $middleware */
+        $middleware = $this->app->make(ValidateCsrfToken::class);
+
+        // Act
+        $excludedPaths = $middleware->getExcludedPaths();
+
+        // Assert
+        self::assertNotContains('ajax/heatmap/data', $excludedPaths);
+        self::assertEquals(['webhook/*'], $excludedPaths);
+    }
+
+    #[Test]
+    public function getData_givenPostRequest_returnsMethodNotAllowed(): void
+    {
+        // Arrange
+
+        // Act
+        $response = $this->post(route('ajax.heatmap.data'));
+
+        // Assert
+        $response->assertMethodNotAllowed();
+    }
+}
--- a/tests/Unit/App/Http/Middleware/HeatmapDataCsrfRegressionTest.php
+++ b/tests/Unit/App/Http/Middleware/HeatmapDataCsrfRegressionTest.php
@@ -28,7 +28,7 @@ class HeatmapDataCsrfRegressionTest extends PublicTestCase
 
         // Assert
         self::assertNotContains('ajax/heatmap/data', $excludedPaths);
-        self::assertEquals(['webhook/*'], $excludedPaths);
+        self::assertContains('webhook/*', $excludedPaths);
     }
 
     #[Test]
--- a/app/Http/Middleware/ReadOnlyMode.php
+++ b/app/Http/Middleware/ReadOnlyMode.php
@@ -14,7 +14,6 @@ class ReadOnlyMode
     private const array ROUTE_WHITELIST = [
         'login',
         'logout',
-        'ajax/heatmap/data',
     ];
 
     public function __construct(private readonly ReadOnlyModeServiceInterface $readOnlyModeService)
--- a/resources/assets/js/custom/inline/common/search/searchhandlerheatmap.js
+++ b/resources/assets/js/custom/inline/common/search/searchhandlerheatmap.js
@@ -44,7 +44,7 @@ class SearchHandlerHeatmap extends SearchHandler {
 
     getAjaxOptions() {
         return {
-            type: 'POST',
+            type: 'GET',
             dataType: 'json'
         };
     }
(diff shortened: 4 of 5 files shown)


routes/web.php
```diff
@@ -601,7 +601,10 @@
             Route::delete('/{tag}/all', new AjaxTagController()->deleteAll(...))->name('ajax.tag.deleteall');
         });
         Route::prefix('heatmap')->group(static function () {
-            Route::post('/data', new AjaxHeatmapController()->getData(...))->name('ajax.heatmap.data');
+            // GET on purpose: the heatmap embed page calls this from inside a cross-site <iframe>,
+            // where SameSite=Lax session/XSRF cookies aren't sent on the XHR, so a POST would
+            // always fail CSRF verification (#3836). The endpoint is a stateless read-only fetch.
+            Route::get('/data', new AjaxHeatmapController()->getData(...))->name('ajax.heatmap.data');
```
Wotuu, line 604: :robot: Low severity — mixed-deploy window: this verb flip has the same non-atomic-deploy property as a schema change. During the ECS rollout (and for any embed/heatmap page already open in a browser, or served by an old container with old release-keyed JS assets), the old bundle still issues `POST /ajax/heatmap/data`, which new containers now answer with **405** — so the heatmap silently fails to load for those users until they refresh onto the new assets. A read-only feature and a short window, so arguably acceptable, but the zero-breakage variant is the expand/contract analog: `Route::match(["GET", "POST"], ...)` in this release, drop `POST` in the next.

tests/Unit/App/Http/Middleware/HeatmapDataCsrfRegressionTest.php
```diff
@@ -0,0 +1,45 @@
...
+    {
+        // Arrange
+        /** @var ValidateCsrfToken $middleware */
+        $middleware = $this->app->make(ValidateCsrfToken::class);
+
+        // Act
+        $excludedPaths = $middleware->getExcludedPaths();
+
+        // Assert
+        self::assertNotContains('ajax/heatmap/data', $excludedPaths);
```
Wotuu, line 30: :robot: Nit — `assertEquals(["webhook/*"], $excludedPaths)` makes this test fail whenever any *legitimate* future CSRF exemption is added, and the failure will surface under the misleading name "HeatmapDataCsrfRegression" in a completely unrelated PR. The `assertNotContains` on the line above already fully guards the actual regression (the heatmap exemption returning). If the exact-list assertion is an intentional tripwire forcing conscious review of every new exemption, keep it but say so in the docblock; otherwise drop it and keep only the `assertNotContains`.

Wotuu:

Wotuu:
:robot: Cold review findings addressed:

**Finding 2 (brittle `assertEquals`)**: fixed in 5e20b1730 — the test now asserts `webhook/*` present and `ajax/heatmap/data` absent, instead of pinning the whole exclusion list.

**Finding 1 (mixed-deploy window, suggested `Route::match(['GET','POST'])`)**: intentionally not adopted. Keeping POST routable without the CSRF exemption doesn't help the page this PR is about — an old-bundle *embed* page POSTing cross-site still fails (419 instead of 405) because its CSRF token never reaches the server either way. The only beneficiary would be a *non-embed* heatmap tab left open across the deploy, whose next search would 405 until refresh; code and JS assets flip together in a single ECS rollout (no schema/cron skew as in #3497), so that window is minutes plus stale tabs, self-healing on refresh, and shows as the existing 'error loading data' snackbar. Against that, `Route::match` would re-add a POST surface, muddy the 405 regression test, and require a follow-up contract release to drop POST again. Not worth it for this endpoint.

tests/Unit/App/Http/Middleware/HeatmapDataCsrfRegressionTest.php
Wotuu: :robot: Fixed in 5e20b1730 — the test now asserts `webhook/*` is present and `ajax/heatmap/data` is absent from the exclusion list, instead of pinning the whole list with `assertEquals`.

routes/web.php
Wotuu: :robot: Intentionally not adopted — full reasoning: https://github.com/RaiderIO/keystone.guru/pull/3840#issuecomment-5189442195. Short version: keeping POST routable without the CSRF exemption does not help the embed page this PR targets, since its cross-site POST still fails on the missing CSRF token either way (419 instead of 405). The only beneficiary would be a non-embed heatmap tab left open across the deploy, and that window is minutes plus stale tabs, self-healing on refresh. Not worth re-adding a POST surface and a follow-up contract release to drop it again for that.

routes/web.php
```diff
@@ -601,7 +601,10 @@
             Route::delete('/{tag}/all', new AjaxTagController()->deleteAll(...))->name('ajax.tag.deleteall');
         });
         Route::prefix('heatmap')->group(static function () {
-            Route::post('/data', new AjaxHeatmapController()->getData(...))->name('ajax.heatmap.data');
+            // GET on purpose: the heatmap embed page calls this from inside a cross-site <iframe>,
```
Wotuu: Remove these comments, not necessary.

Wotuu: :robot: Removed the unnecessary comment in d7dcd5125.

PR #3841, opened 2026-08-05
#3834 Show reflectable spells on the by-class Compendium page

### LINKED ISSUE #3834: Show reflectable spells on the by-class Compendium page (e.g. Warrior)
:robot: Immune and Reflect outcomes are already captured identically today:
`SpellPropertyObservationCollector::collectFromEvent()`
(`app/Service/CombatLog/DataExtractors/SpellDataCollectors/SpellPropertyObservationCollector.php`)
maps any `MissedInterface` suffix via `Spell::GUID_MISS_TYPE_MAPPING`
(`app/Models/Spell/SpellConstants.php`) into a bit on `Spell::miss_types_mask` — Immune and
Reflect are just two bits among dodge/parry/absorb/etc, with no asymmetry in what's collected.

That data is already rendered via `Spell::maskToReadableString(...)` on the per-Spell page
(`resources/views/compendium/spell/sections/details.blade.php`) and per-NPC spell list
(`resources/views/compendium/npc/sections/spells.blade.php`) — but **not** on the by-class
Compendium page (`app/Http/Controllers/Compendium/ClassCompendiumController.php`,
`resources/views/compendium/class/show.blade.php`), which today only lists player-ability
"counter" spells (the Vanish/Shadowmeld/Feign-Death-style data from #3826 / #3833) and never
queries `miss_types_mask` at all.

## Why this is useful

Immunity (Divine Shield, Ice Block, Aspect of the Turtle, ...) is a full no-sell — once you know
a cooldown is an immunity, you already know it stops everything, so listing "what can this
immunity stop" per-class isn't useful information.

Spell Reflect is different: it only reflects certain spells, and which ones is genuinely useful
information a Warrior (or other reflect-capable class) player doesn't already know. Right now
that data exists in the DB but has no page that surfaces it *by class*.

## Remaining work

- Add a query in `ClassCompendiumController::show()` (or a class-specific branch) selecting
  spells where `miss_types_mask` has the Reflect bit set (see `SpellConstants` for the exact
  mask constant), scoped the same way the existing counter-spell query is scoped
  (dungeon/mapping-version).
- Add a new section to `resources/views/compendium/class/show.blade.php` (or a partial) listing
  those spells, following the existing conventions on that page.
- This is pure aggregation/display — no new combat-log detection logic is required, the data is
  already collected.

## Definition of done

- Warrior's Compendium class page shows a "Reflectable spells" list.
- Feature test covering the new query/view.

Wotuu:
:robot: Closes #3834

`Spell::miss_types_mask` already carries a `MISS_TYPE_REFLECT` bit, collected from combat logs by
`SpellPropertyObservationCollector`. It was rendered on the per-Spell page and the per-NPC spell
table, but the by-class Compendium page never queried it.

Immunity is a full no-sell — once you know a cooldown is an immunity you already know it stops
everything — so listing "what this immunity stops" is not useful and is deliberately not included.
Spell Reflect *is* selective, so which dungeon abilities actually get reflected is real information
a Warrior does not already have.

## What changed

- `ClassCompendiumController::getReflectSection()` — returns `null` for classes without a reflect
  ability, otherwise the spells in the context dungeon whose `miss_types_mask` has the Reflect bit,
  scoped exactly like the existing counter query (game version + `spell_dungeons` for the context
  dungeon), plus `->visible()` so hidden spells stay hidden.
- Reflect-capable classes live in a small `SPELL_REFLECT_CLASS_ICONS` const. Retail Mythic+ has
  exactly one entry (Warrior / Spell Reflection); the Warlock's Nether Ward is PvP-only. The const
  is the extension point if that ever changes.
- The duplicated "NPCs in this dungeon that cast these spells" query is extracted out of
  `getCounterSections()` into a shared `getNpcsBySpellId()` helper used by both sections.
- New "Reflectable spells" section in `resources/views/compendium/class/show.blade.php`, following
  the counter sections' markup, plus `en_US` translations.
- Three feature tests in `ClassCompendiumControllerTest`: Warrior shows the section, Paladin does
  not, and a spell carrying the Reflect bit renders together with the NPC that casts it.

Pure aggregation/display — no combat-log detection logic was touched.

## Verification

- `php artisan test tests/Feature/Controller/Compendium/` — 51 passed.
- `composer run fix` + `composer run analyse` clean.
- Visually verified in headless Chrome against a seeded reflect row; before/after below.


Wotuu:
:robot: Visual verification — Warrior class Compendium page (Pit of Saron context), with three spells seeded as reflected in the dev DB.

**Before (master)**
before

**After**
after

Wotuu:

Reviewed by a fresh-context agent with no shared context with the implementing session. All findings resolved in e39a3bfa7 except where noted.

| # | Finding | Resolution |
|---|---|---|
| 1 | The new test could pass off the counter sections — Warrior renders the Shadowmeld racial section, scoped by the very `spell_dungeons` row the test creates, so a spell with a `counters_mask` bit would satisfy both assertions with the reflect section broken | Fixed — spell pick now requires `counters_mask = 0`, and the NPC pick excludes NPCs that have any characteristic or any `npc_spells` row, so neither the CC table nor a counter section can render them. Verified the test goes red against the pre-change controller/view |
| 2 | ~30 lines of table markup copy-pasted between the counter and reflect sections | Fixed — extracted into `compendium.class.sections.spell_npc_table`, used by both. Both call sites re-verified in headless Chrome with populated data (screenshots below) |
| 3 | `visible()` asymmetry: the reflect query filtered `hidden_on_map`, the counter query 10 lines above did not | Fixed — `->visible()` applied to the counter query too. Hidden spells are hidden everywhere else on the site, so the asymmetry was the bug, not the filter |
| 4 | `SPELL_REFLECT_CLASS_ICONS` is keyed on class alone, so the section also renders on a game version where the class has no reflect | Not gated — the section lists nothing there, which reads identically to any dungeon without reflect data. The const's docblock no longer claims an invariant the code doesn't enforce |
| 5 | Translation array alignment | Not applied — `composer run fix` actively produces the unaligned form (the preceding `],` breaks php-cs-fixer's alignment group), so the tool's output stands |
| 6 | `alt=` duplicates the adjacent visible text | Not applied — the counter markup directly above does the same; changing only the new section would make the file less consistent, not more |

Explicitly out of scope per repo convention: model-cache invalidation findings.

resources/views/compendium/class/sections/spell-npc-table.blade.php
```diff
@@ -0,0 +1,44 @@
+<?php
```
Wotuu, line 1: I noticed you put underscores in this file's name, and I don't think it's wrong. It's just not the convention that I do. If you think we should be adding underscores to the blade files, if you think it's a genuine improvement, can you then create a follow-up issue to apply this convention to the entire codebase? If you think this is not something we want to do, rename this file.

Wotuu: :robot: Renamed to `spell-npc-table.blade.php` (kebab-case) to match the existing convention in b2b938ff5 — no case for changing the project-wide convention here (the repo already uses kebab-case for blade views, e.g. `error-layout.blade.php`, `sitepage.blade.php`), so no follow-up issue filed.

PR #3842, opened 2026-08-05
#3833 Track Feign Death, Mage Invisibility and Cloak of Shadows as counter spells

### LINKED ISSUE #3833: Track Feign Death, Mage Invisibility and Cloak of Shadows as counter spells (follow-up to #3826)
:robot: Follow-up to #3826, which added combat-log detection of Rogue Vanish / Night Elf
Shadowmeld as "counters" to specific boss abilities via `SpellCounterDataExtractor`.

## Background

While researching what else could be tracked in the same way, we identified three more
class abilities worth adding, split into two tiers of effort:

1. **Hunter Feign Death** and **Mage Invisibility** — mechanically identical to Vanish/Shadowmeld.
   Both drop a targeted cast/pull/channel aimed at the player. The existing 3-signature detector
   in `SpellCounterDataExtractor` (A: targeting debuff dropped, B: channel broken, C: cast
   aborted, vetoed by loss-of-control-mechanic debuffs) should apply with no structural changes —
   just register the new player buff spell IDs alongside Vanish/Shadowmeld.
2. **Rogue Cloak of Shadows** — a different shape of "counter": it doesn't pre-empt a cast, it
   *strips* a magic debuff that already landed. Detectable via `SPELL_AURA_REMOVED` on a magic
   debuff coinciding with Cloak's `SPELL_AURA_APPLIED`. This needs a new signature (debuff-strip)
   in the extractor, not just a new spell ID.

## Remaining work

- Add Feign Death + Invisibility spell IDs to wherever Vanish/Shadowmeld are currently
  registered (see `SpellCounterDataExtractor` and its NpcSpell-assignment code, per #3826).
- Validate against real logs using `combatlog:searchruns` / `combatlog:downloadruns`
  (class/spec/dungeon/key-level filters) — same validation approach used for #3826.
- Add a new debuff-strip signature for Cloak of Shadows: watch for `SPELL_AURA_REMOVED` on a
  magic-type debuff at (approximately) the same timestamp as Cloak's `SPELL_AURA_APPLIED` on the
  same player. Reuse the existing ε tolerance approach from #3826 (100ms, evidence-tuned) as a
  starting point, but re-validate against real Cloak logs since the signature shape differs from
  A/B/C.

- Cloak of Shadows only strips *magic* debuffs — need to confirm which debuff types actually get
  removed (poison/disease/curse are separate categories in WoW and may need excluding) to avoid
  false positives.
- Same NULL-category gotcha as #3826: combat-log-created spells have NULL category until a
  Wowhead fetch; the counter extractor's own NpcSpell assignment must allow null but reject
  player categories (see `[[project_3826_spell_counters]]` internal note from #3826's
  implementation).

## Key files

- `app/Service/CombatLog/...SpellCounterDataExtractor` (see #3826's MR #3830 for exact path/shape)
- Whatever registers the Vanish/Shadowmeld spell ID list today

## Definition of done

- Feign Death and Invisibility detected on real validation logs with no new false positives.
- Cloak of Shadows debuff-strip signature added, validated against real logs.
- Tests added/updated, `composer run analyse` + `composer run fix` clean.

Wotuu:
:robot: Closes #3833

Follow-up to #3826, adding three more player abilities as tracked counters: **Hunter Feign
Death**, **Mage Invisibility** and **Rogue Cloak of Shadows**.

## What changed

`SpellCounterDefinitionInterface` grew three knobs, all defaulted on a new abstract
`SpellCounterDefinition` base so the existing Vanish/Shadowmeld definitions only shrank:

| knob | why |
| --- | --- |
| `getTriggerAuraSpellIds()` | Invisibility's cast only starts a 3 second fade — the threat drop lands with the buff aura, far outside the ±100 ms correlation window of its cast |
| `dropsThreat()` | Cloak leaves the rogue on the threat table, so it cannot explain an abandoned cast; signature C is off for it |
| `getUnstrippableDebuffDispelTypes()` | Cloak only strips magic, so a poison/disease/curse/enrage falling off in its window is provably not its doing |

Feign Death and Greater Invisibility ride the existing A/B/C signatures unchanged — only their
spell ids are new. Cloak needed the two gates above; the debuff-strip itself is what signature B
already recognises (an NPC-sourced debuff on the player removed well before its natural duration,
within ε of the counter), so no new signature was necessary.

An unresolved dispel type (`n_a`, `unknown`, empty) never rejects — combat-log-created spells
carry none until their Wowhead data is fetched, and those are exactly the spells this extractor
exists to discover. Only positively contradicting types disqualify.

## Validation against real logs

15 real M+ runs / 129 segments pulled with `combatlog:downloadruns`, run through
`combatlog:extractdata`, and cross-checked against a raw-log correlation script.

**Two spell ids were wrong until the logs corrected them** — worth knowing for future counters:

- Greater Invisibility applies its buff under **110960**, not the **110959** it is cast with. Real
  logs also show it *cast* under 110960, so both ids are registered as both kinds of trigger.
- Vanish's aura is 11327, not 1856 (already documented in #3826) — same class of trap.

**Cloak of Shadows — 14 uses, 8 debuff removals inside ε, 5 recorded, 3 correctly rejected:**

| removed debuff | dispel type | outcome |
| --- | --- | --- |
| Permeating Cold (1258437) | magic | recorded |
| Blight Splatter (1259116), Chains of Subjugation (1262509), Shade Shift (1264246), Suppression Field (1269283) | unresolved (`n_a`) | recorded |
| Creeping Void (1281636) | **curse** | rejected by the dispel filter |
| Shadowbind (1264186) | **curse** | rejected by the dispel filter |
| Luciferin Flare (1264354) | `n_a` | rejected — it had run its full 5 s duration |

Two of the eight candidates were curses, i.e. the dispel filter removed two false positives that
would otherwise have shipped. Every recorded strip correlated at **0–1 ms**, matching #3826's
finding that a true positive lands in the same server tick.

**Invisibility — 16 triggers, 0 detections.** No debuff removal fell inside any trigger's window,
so nothing was recorded. No false positives; no true positive available in this corpus either.

**Feign Death — not exercised by the corpus.** All 10 occurrences across the 15 runs are
`SPELL_CAST_FAILED` (the ability on cooldown); no hunter successfully used it. Its registration is
mechanically identical to Vanish/Shadowmeld and is covered by unit tests (signatures B and C), but
it has **not** been confirmed against a real successful use. It produced no false positives, since
it never fired.

## Notes

- `combatlog:searchruns --class=` / `--spec=` do not filter — the returned run set is identical
  with and without them. Pre-existing, not touched here; runs for this validation were picked by
  hand from the unfiltered listing by their member spec ids. Worth its own issue.
- No migration: `counters_mask` is an unsigned int with 3 of 32 bits used, and the `property`
  columns are `varchar`, not enums.
- Backend only — the by-class Compendium page renders counter sections straight off
  `SpellCounterDefinitions::all()`, so the three new ones appear with no view change.

## Test plan

- `SpellCounterDataExtractorTest`: 10 new tests (28 total) — Feign Death via signatures B and C,
  Greater Invisibility via its cast, Invisibility via its fade aura *and* a negative proving its
  cast alone does not trigger, Cloak strips (magic and unresolved dispel types), Cloak rejecting a
  poison in both window directions, Cloak not claiming an abandoned cast, and Cloak stripping
  several debuffs at once.
- `DetectStaleCombatLogDataCommandTest` re-run — the new `SpellProperty` cases flow through its
  `counter_<name>` ↔ `ALL_COUNTERS` bit lookup.

### DIFF
--- a/app/Models/Spell/SpellConstants.php
+++ b/app/Models/Spell/SpellConstants.php
@@ -72,12 +72,18 @@ trait SpellConstants
         Resist::class  => self::MISS_TYPE_RESIST,
     ];
 
-    public const int COUNTER_VANISH     = 1;
-    public const int COUNTER_SHADOWMELD = 2;
+    public const int COUNTER_VANISH           = 1;
+    public const int COUNTER_SHADOWMELD       = 2;
+    public const int COUNTER_FEIGN_DEATH      = 4;
+    public const int COUNTER_INVISIBILITY     = 8;
+    public const int COUNTER_CLOAK_OF_SHADOWS = 16;
 
     public const array ALL_COUNTERS = [
-        self::COUNTER_VANISH     => 'vanish',
-        self::COUNTER_SHADOWMELD => 'shadowmeld',
+        self::COUNTER_VANISH           => 'vanish',
+        self::COUNTER_SHADOWMELD       => 'shadowmeld',
+        self::COUNTER_FEIGN_DEATH      => 'feign_death',
+        self::COUNTER_INVISIBILITY     => 'invisibility',
+        self::COUNTER_CLOAK_OF_SHADOWS => 'cloak_of_shadows',
     ];
 
     public const string DISPEL_TYPE_MAGIC         = 'magic';
--- a/app/Models/Spell/SpellConstants.php
+++ b/app/Models/Spell/SpellConstants.php
@@ -86,6 +86,9 @@ trait SpellConstants
         self::COUNTER_CLOAK_OF_SHADOWS => 'cloak_of_shadows',
     ];
 
+    /** The prefix `spells`.`dispel_type` carries since its values became translation keys. */
+    public const string DISPEL_TYPE_TRANSLATION_KEY_PREFIX = 'spelldispeltype.';
+
     public const string DISPEL_TYPE_MAGIC         = 'magic';
     public const string DISPEL_TYPE_DISEASE       = 'disease';
     public const string DISPEL_TYPE_POISON        = 'poison';
--- a/app/Service/CombatLog/DataExtractors/SpellCounterDataExtractor.php
+++ b/app/Service/CombatLog/DataExtractors/SpellCounterDataExtractor.php
@@ -66,9 +66,6 @@ class SpellCounterDataExtractor implements DataExtractorInterface
      */
     public const int EPSILON_MS = 100;
 
-    /** @var string Prefix `spells`.`dispel_type` carries since the values became translation keys. */
-    public const string DISPEL_TYPE_TRANSLATION_KEY_PREFIX = 'spelldispeltype.';
-
     /** @var int Window in which a nil-source debuff is linked back to a temporally adjacent NPC SPELL_CAST_START. */
     public const int CAST_START_LINK_WINDOW_MS = 500;
 
@@ -790,7 +787,7 @@ class SpellCounterDataExtractor implements DataExtractorInterface
         }
 
         // Rows predating the translation-key migration still hold the bare value, so compare on that
-        $dispelType = Str::after($dispelType, self::DISPEL_TYPE_TRANSLATION_KEY_PREFIX);
+        $dispelType = Str::after($dispelType, SpellModel::DISPEL_TYPE_TRANSLATION_KEY_PREFIX);
 
         if (!in_array($dispelType, $unstrippableDispelTypes, true)) {
             return true;
--- a/app/Service/CombatLog/DataExtractors/SpellCounterDataExtractor.php
+++ b/app/Service/CombatLog/DataExtractors/SpellCounterDataExtractor.php
@@ -38,19 +38,23 @@ use App\Service\CombatLog\Dtos\DataExtraction\DataExtractionCurrentDungeon;
 use App\Service\CombatLog\Dtos\DataExtraction\ExtractedDataResult;
 use Illuminate\Support\Carbon;
 use Illuminate\Support\Collection;
+use Str;
 
 /**
- * Detects "spell counter" tricks - a player using a counter ability (Rogue Vanish, Night Elf Shadowmeld) to make an
- * incoming NPC ability fizzle - and records them on the Spell that was countered.
+ * Detects "spell counter" tricks - a player using a counter ability (Rogue Vanish, Night Elf Shadowmeld, Hunter Feign
+ * Death, Mage Invisibility, Rogue Cloak of Shadows) to make an incoming NPC ability fizzle - and records them on the
+ * Spell that was countered.
  *
- * Three signatures are recognized:
+ * A counter is triggered by its own SPELL_CAST_SUCCESS, by a player buff SPELL_AURA_APPLIED, or by both - see
+ * SpellCounterDefinitionInterface::getTriggerAuraSpellIds(). Three signatures are recognized from there:
  * - A: an NPC targeting debuff (applied with a nil source guid, linked back to the NPC's SPELL_CAST_START) is dropped
- *   within EPSILON_MS of a counter cast. The fact attaches to the NPC's *cast* spell id.
- * - B: an NPC-sourced debuff (a channel) on the counter-caster is removed within EPSILON_MS of the counter cast, well
- *   before its natural duration. The fact attaches to the debuff's own spell id.
+ *   within EPSILON_MS of a counter trigger. The fact attaches to the NPC's *cast* spell id.
+ * - B: an NPC-sourced debuff (a channel, or - for Cloak of Shadows - a stripped magic debuff) on the counter-caster is
+ *   removed within EPSILON_MS of the counter trigger, well before its natural duration. The fact attaches to the
+ *   debuff's own spell id.
  * - C: an NPC SPELL_CAST_START never reaches SPELL_CAST_SUCCESS and the same caster starts a fresh cast shortly after
- *   a counter cast, with no other explanation (no fail, no interrupt, no death, no loss-of-control debuff on the
- *   caster).
+ *   a counter trigger, with no other explanation (no fail, no interrupt, no death, no loss-of-control debuff on the
+ *   caster). Only counters that drop threat can produce this signature.
  */
 class SpellCounterDataExtractor implements DataExtractorInterface
 {
@@ -61,6 +65,9 @@ class SpellCounterDataExtractor implements DataExtractorInterface
      * without admitting coincidences.
      */
     public const int EPSILON_MS = 100;
+
+    /** @var string Prefix `spells`.`dispel_type` carries since the values became translation keys. */
+    public const string DISPEL_TYPE_TRANSLATION_KEY_PREFIX = 'spelldispeltype.';
 
     /** @var int Window in which a nil-source debuff is linked back to a temporally adjacent NPC SPELL_CAST_START. */
     public const int CAST_START_LINK_WINDOW_MS = 500;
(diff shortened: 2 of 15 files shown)


app/Service/CombatLog/DataExtractors/SpellCounters/CloakOfShadowsSpellCounterDefinition.php
```diff
@@ -0,0 +1,74 @@
...
+     * @return array<string>
+     */
+    public function getUnstrippableDebuffDispelTypes(): array
+    {
+        return [
+            Spell::DISPEL_TYPE_POISON,
+            Spell::DISPEL_TYPE_DISEASE,
+            Spell::DISPEL_TYPE_CURSE,
+            Spell::DISPEL_TYPE_ENRAGE,
+        ];
```
Wotuu, line 78: :robot: `Spell::DISPEL_TYPE_NONE` is missing from this list, and its absence contradicts the rule the interface docblock states ("Only positively contradicting types belong here. A dispel type that is merely unresolved - `n_a`, `unknown`, or the empty string ... - must not be listed").

`none` is not unresolved — it is a *resolved* "this effect is not dispellable at all", which contradicts Cloak of Shadows just as hard as `poison` does: Cloak removes harmful **magic** effects, so a non-dispellable debuff falling off within ε of a Cloak is provably not its doing. As written, every `spelldispeltype.none` debuff removal near a Cloak is recorded as a `counter_cloak_of_shadows` fact.

There is a real counter-argument — `database/migrations/2026_04_20_224500_update_spell_dispel_type_to_translation_keys.php` maps the old `''` to `spelldispeltype.none`, so for rows that predate that migration `none` genuinely conflates "not dispellable" with "never fetched". If that is why `none` was deliberately left out, please say so in the docblock here; otherwise the list should include it. Right now a reader applying the stated rule would conclude it was simply overlooked.

tests/Feature/App/Service/CombatLog/DataExtractors/SpellCounterDataExtractorTest.php
```diff
@@ -492,6 +495,228 @@ public function extractData_givenDebuffOnAnotherPlayerThanTheCounterCaster_doesN
...
+            $this->debuffRemoved(1700, $channelSpellId, 'Solar Flame', self::CREATURE_GUID),
+        ]);
+
+        // Assert
+        $this->assertSame(0, $this->result->toArray()['addedSpellCounters']);
+        $this->assertNoCounterRecorded($channelSpellId);
+    }
+
+    #[Test]
+    public function extractData_givenMagicDebuffStrippedAtCloakOfShadows_setsCloakOfShadowsCounter(): void
```
Wotuu, line 605: :robot: Coverage gap: none of the Cloak of Shadows tests exercise `getTriggerCastSpellIds()` — every one of them drives the detection through `playerBuffApplied()` only. Same for `SPELL_ID_GREATER_INVISIBILITY_AURA` (110960), which is registered as *both* a cast and an aura trigger but is never used in a test.

That matters because the shape real logs actually have is the one that is untested: per the definitions' own docblocks, Cloak's `SPELL_CAST_SUCCESS` and its `SPELL_AURA_APPLIED` carry the same id 31224 and land in the same instant, so in production `handleCounterTrigger()` runs **twice** for one use, pushing two entries into `recentCounterCasts` and running the backward `recentDebuffRemovals` scan twice. Nothing here proves the `queueCounterObservation()` dedup absorbs that and `addedSpellCounters` stays at 1 rather than 2.

Suggest one extra test emitting cast-success + aura-applied at the same offset with a stripped magic debuff, asserting `addedSpellCounters === 1`.

app/Service/CombatLog/DataExtractors/SpellCounterDataExtractor.php
```diff
@@ -62,6 +66,9 @@ class SpellCounterDataExtractor implements DataExtractorInterface
      */
     public const int EPSILON_MS = 100;
 
+    /** @var string Prefix `spells`.`dispel_type` carries since the values became translation keys. */
+    public const string DISPEL_TYPE_TRANSLATION_KEY_PREFIX = 'spelldispeltype.';
```
Wotuu: :robot: Minor: `'spelldispeltype.'` is now hardcoded in a fifth place. It already exists as an inline `sprintf('spelldispeltype.%s', ...)` in `WowheadService` (the writer), `SpellController`, `ExportCsv` and the `2026_04_20_224500_update_spell_dispel_type_to_translation_keys` migration.

Since it is a property of `spells`.`dispel_type` rather than of this extractor, it would be better placed in `SpellConstants` next to `ALL_DISPEL_TYPES` / `DISPEL_TYPE_*` — that is where the next person changing the storage format will look, and where the other consumers could be pointed at over time.

Wotuu:

app/Service/CombatLog/DataExtractors/SpellCounters/CloakOfShadowsSpellCounterDefinition.php
Wotuu: :robot: Kept off the list, rationale now documented in the method docblock. `2026_04_20_224500_update_spell_dispel_type_to_translation_keys` maps both `'None'` **and** the legacy empty string to `spelldispeltype.none`, so post-migration the value cannot distinguish "known to be undispellable" from "never filled in" — rejecting on it would rule out spells that only look undispellable because nobody populated the field, which is the same failure mode the unresolved-types carve-out exists to prevent. (In the validation corpus this is moot either way: exactly one spell in the whole DB carries `spelldispeltype.none`, and it was not near a Cloak.)

tests/Feature/App/Service/CombatLog/DataExtractors/SpellCounterDataExtractorTest.php
Wotuu: :robot: Fixed — added `extractData_givenCounterLoggedAsBothCastAndAura_recordsOneCounter`, which drives Cloak of Shadows and Greater Invisibility (110960) through cast success *and* buff aura in the same millisecond and asserts `addedSpellCounters` stays at 2, not 4. `assertCounterRecorded` also asserts exactly one `CombatLogSpellEvent` per spell, so the dedup is pinned on both the counter and the audit row.

app/Service/CombatLog/DataExtractors/SpellCounterDataExtractor.php
Wotuu: :robot: Fixed — moved to `SpellConstants::DISPEL_TYPE_TRANSLATION_KEY_PREFIX`, right above `DISPEL_TYPE_MAGIC`, and the extractor now uses it. Left the four pre-existing call sites alone to keep the diff on topic.

Wotuu:
:robot: All 3 cold-review findings addressed in fecead3ce; threads replied to and resolved. CI green (9/9). Backend-only change — no visual verification needed. Marking ready for review.

app/Models/Spell/SpellConstants.php
```diff
@@ -72,14 +72,23 @@ trait SpellConstants
         Resist::class  => self::MISS_TYPE_RESIST,
     ];
 
-    public const int COUNTER_VANISH     = 1;
-    public const int COUNTER_SHADOWMELD = 2;
+    public const int COUNTER_VANISH           = 1;
```
Wotuu, line 75: Please add a follow-up ticket to split up this class into a bunch of different files, each with their separate enum. I think this Constance class is now housing multiple concerns, and it should just not do that.

Wotuu: :robot: Filed #3852 to track splitting SpellConstants into separate files per concern.

PR #3843, opened 2026-08-05
#3835 Detect NPC spells that land despite an active player immunity

### LINKED ISSUE #3835: Detect spells that land despite an active player immunity (Divine Shield / Ice Block / Aspect of the Turtle)
:robot: Related to #3826/#3833 (counter detection) and #3834 (surfacing reflect data), but a
genuinely new detection pattern rather than a display change.

## Background

Full immunities (Divine Shield, Ice Block, Aspect of the Turtle, Anti-Magic Shell, etc.) are
common knowledge to stop "everything" — that's already assumed and not worth calling out per
mechanic (see #3834's reasoning for why we're *not* building an "immune to X" list). The genuinely
useful, much rarer signal is the inverse: **which spells still land on a player despite an active
immunity window** — i.e. mechanics that are documented/expected to be un-immunable. Today nothing
in the combat log pipeline detects this at all.

## Why this is non-trivial

Unlike #3826/#3833's counters, this isn't a small addition to the existing 3-signature detector —
it's a different signal shape entirely (presence of an unexpected event, not absence of an
expected one), so treat it as comparable in scope to the original #3826 detector build, not a
quick add.

## Recommended approach

1. Define a hardcoded list of known full-immunity player buff spell IDs (Divine Shield, Ice
   Block, Aspect of the Turtle, Anti-Magic Shell, and any others worth including) — follow the
   existing convention for this kind of static spell-ID list, e.g. `BLOODLUSTY_SPELLS` /
   `CHARM_SPELLS` in `app/Models/Spell/SpellConstants.php`.
2. Track each such buff's active window per player (`SPELL_AURA_APPLIED` → `SPELL_AURA_REMOVED`/
   expiry) via a new extractor/collector, following the combatlog-data-pipeline conventions.
3. For any harmful `SPELL_DAMAGE`/effect event targeting that same player GUID within the active
   window, and where the event is **not already a miss/immune outcome** (i.e. it actually landed),
   flag it as bypassing the immunity.
4. Aggregate flagged spells up to NpcSpell/Spell level (mirroring how #3826's counters and #3834's
   reflect data get surfaced), and add a Compendium display for it — likely fits best as a
   per-spell or per-NPC flag rather than a by-class list, since it's about specific boss abilities.

## Risks / false-positive sources

- **Partial immunities**: some "immunity" cooldowns don't block 100% of effects (DoTs already
  ticking before the immunity started, physical vs magic distinctions, etc.) — need to confirm
  per-immunity-spell what it's actually supposed to block before treating a landed hit as a real
  bypass rather than expected behavior.
- **Timing edge cases**: a hit landing in the same event batch as the immunity's expiry needs the
  same kind of ε-tolerance tuning #3826 needed for its signatures — don't assume a fixed window
  without validating against real logs.
- Validate against real logs using `combatlog:searchruns` / `combatlog:downloadruns`, same
  approach as #3826/#3833.

## Definition of done

- New extractor/collector validated against real logs with a defensible false-positive rate
  (document the validation, same as #3826's memory notes).
- Compendium display added for flagged "bypasses immunity" spells.
- Tests added, `composer run analyse` + `composer run fix` clean.

Wotuu:
:robot: Closes #3835

Detects NPC abilities that **land on a player despite an active immunity** — the rare inverse of
"this cooldown stops everything", which is the signal a route author actually needs.

## How it works

`ImmunityBypassDataExtractor` opens a window per player on the `SPELL_AURA_APPLIED` of a known
immunity buff, closes it on `SPELL_AURA_REMOVED`, and flags any NPC-sourced harmful event that
lands inside it. `SPELL_MISSED` is the immunity working and is ignored by construction — only
events that actually landed (`SPELL_DAMAGE`, a harmful `SPELL_AURA_APPLIED`) are considered.

The detection is aggregated onto `spells`.`bypasses_immunities_mask` and surfaced as a
**"Bypasses immunity"** column on the NPC and spell compendium pages, with the same
observation → staleness-sweep lifecycle as the miss types and the #3826 counters.

## The false-positive problem, and what solves it

The issue flags partial immunities as the main risk, and the real logs confirm it is *the* risk.
Every immunity therefore declares its own coverage in
`DataExtractors/ImmunityBypasses/ImmunityDefinitions.php`:

| Immunity | Protects against | Damage | Harmful auras |
| --- | --- | --- | --- |
| Divine Shield, Ice Block | all schools | ✅ | ✅ |
| Aspect of the Turtle | all schools | ✅ | ❌ (never stopped debuffs) |
| Blessing of Protection | physical only | ✅ | ✅ |
| Blessing of Spellwarding | magic only | ✅ | ✅ |
| Anti-Magic Shell | magic only | ❌ (absorbs, not immune) | ✅ |

Deliberately excluded: Cloak of Shadows and Nether Ward (version-dependent coverage), and
Dispersion / Evasion / Die by the Sword (reduction or avoidance chance, never immunity).

Pre-existing damage-over-time ticks — the other partial case named in the issue — are dropped by
skipping periodic events; a DoT *applied* inside the window is still caught by the harmful-aura
branch.

## Real-log validation

**28 real M+ runs (234 segments, ~1.7 GB)** downloaded with `combatlog:downloadruns` across
paladin, mage, hunter and death knight, replayed through the detector's logic:

| | |
| --- | --- |
| Immunity windows | **310** (AMS 226, Divine Shield 56, Aspect of the Turtle 14, BoP 10, BoS 4, Ice Block 0) |
| Runs | 28 (`--class` paladin/mage/hunter/death knight, one class per invocation — passing several at once returns nothing) |
| NPC-sourced landed hits inside a window | **132** |
| Rejected — immunity does not cover this kind | 68 |
| Rejected — immunity does not cover this school | 57 |
| Rejected — ability resolved before the window (in flight) | 1 |
| **Detected as bypasses** | **5** (4 distinct spells) |

(126 rejected + 6 surviving = 132; the sixth sat in a window that expired without a removal line
and was dropped there.)

All 5 detections are harmful magic *debuffs* applying through Anti-Magic Shell — Light Ray,
Scorching Ray, Void Binding and one more. Light Ray was verified by hand in the raw log:
`SPELL_ABSORBED` lines show AMS absorbing its damage for the whole window while its debuff keeps
landing, which is exactly the "still lands despite the immunity" fact the issue asks for.

**No hit landed inside a full-immunity window at all** across 56 Divine Shield and 14 Aspect of
the Turtle windows — full immunities were airtight in this corpus, which is what should happen.
Ice Block was never used, so it has no real-log coverage yet (unit tests cover it).

Worth stressing what the 126 rejections mean: a detector without the coverage model and the
provenance gate would have reported **126 false "bypasses immunity" claims** on this corpus,
most of them shadow/frost/arcane damage during physical-only Blessing of Protection.

⚠️ **These figures were measured against the pre-cold-review revision (`72ceb42`).** Three of the
five review fixes would move them, and I did not re-run: the corpus is deleted and Raider.IO's
segment URLs are presigned, so a fresh download is a *different* set of runs and would produce
non-comparable numbers rather than a confirmation. The directions are understood:

- **AMS 5s → 7s** can only *add* detections — 25 windows force-expired under the old duration, and
  AMS was 226 of the 310, so any hit between 5s and 7s of an AMS window was never even considered.
- **Trailing ε on the expiry path** can only *remove* detections, from those same expired windows.
- **"any cast before the window" instead of "newest cast"** can only *remove* detections. Note the
  5 surviving detections are all auras, and I confirmed only Light Ray has no `SPELL_CAST_SUCCESS`
  lines at all — Scorching Ray, 星宇飞升 and Void Binding were not checked for that, so they are
  the ones that could drop out.

The qualitative conclusion is unaffected: the coverage model is what stands between this detector
and a flood of false claims, and every fix above tightens or corrects it.

### Reproducing the corpus

The segment files are gone (hundreds of MB, presigned URLs long expired), but the run ids are
permanent — this re-downloads the exact same 28 runs:

```sh
php artisan combatlog:downloadruns --output-dir=combatlogs/immunity \
  --run=40286260 --run=40286263 --run=40288638 --run=40291459 --run=40317435 \
  --run=40276237 --run=40276891 --run=40285649 --run=40314755 --run=40368441 \
  --run=40445481 --run=41592922 --run=40292528 --run=41100727 --run=41216069 --run=42304841 \
  --run=40286525 --run=40367812 --run=40395335 --run=40393865 --run=40091598 --run=41429856 \
  --run=40352147 --run=40420944 --run=40415388 --run=40512754 --run=40547525 --run=40615396
```

(Originally selected as `--min-level=8 --from-days=21` with `--class=paladin --limit=10`, then
`--limit=6` each for mage, hunter and death knight — but *that* is not reproducible, since
`--from-days` is relative to now. The ids are.) The two logs called out above live in
`run_40415388_segment_7.txt` (Arcane Bolt / Aspect of the Turtle) and `run_40420944_segment_4.txt`
(Light Ray / Anti-Magic Shell). The `combatlog-download-runs` skill now carries this as a standing
convention so the next detector's corpus is recorded up front.

## Cold review

Reviewed by a fresh-context agent: **5 findings, all fixed** in 1f46794 with a regression test
each. Three were real correctness bugs the real-log run could not have surfaced — cast provenance
leaking across combat logs (the extractor instance is reused per file), a `ZoneChange` attributing
a spell to the dungeon the party had just left for, and a re-cast inside the window laundering an
earlier still-in-flight projectile. The other two tightened the trailing ε on the unlogged-expiry
path and corrected Anti-Magic Shell's duration to 7 s (Anti-Magic Barrier).

## Timing

Both ends of the window are shrunk by **ε = 100 ms** (the value #3826's bulk validation settled
on): the log orders same-tick lines arbitrarily, so a hit sharing a tick with the application or
the expiry cannot be told apart from one landing just outside the window. Detections are held as
candidates until the window closes so the trailing edge can be applied retroactively.

## Notes for review

- The immunity buff spell ids live in `SpellConstants` as the issue asked (`IMMUNITY_SPELLS`,
  next to `BLOODLUSTY_SPELLS`); the per-immunity *behaviour* lives in the definition objects,
  following the `SpellCounterDefinitions` convention from #3826.
- Periodic events are recognized by event name, not by the `SpellPeriodic` prefix class:
  `Prefix::PREFIX_CLASS_MAPPING` matches `SPELL` before `SPELL_PERIODIC`, so a periodic event
  never actually carries that prefix class. Reordering the map would change which events
  `SpellDataExtractor` and `NpcCharacteristicDataExtractor` process, so it is left alone here.
- Migration is additive (`bypasses_immunities_mask`, default 0) and backward-compatible with the
  currently-running code.
- #3842 (#3833, still draft) appends to `SpellProperty.php` and `SpellConstants.php` as well.
  Both changes are pure appends, so whichever lands first leaves the other a trivial conflict.
- Known narrow gap, left deliberately: `closeAllImmunityWindows()` (end of log, run boundary)
  still flushes with an unknown end, so a candidate sitting within ε of an *unlogged* expiry is
  confirmed there. Closing it properly needs a last-seen-event timestamp to tell "the window
  expired" from "the log stopped", which felt like more machinery than the case is worth — happy
  to add it if you disagree.
- Not fixed here, but noticed while working on the school gate: `SpellCreationCollector` does
  `(int)$prefix->getSpellSchool()` on a hex string like `"0x20"`, which is `0` in PHP — so
  combat-log-created spells get `schools_mask = 0` (605 of 7916 on a fresh seed). Filed as
  **#3845**; it needs a backfill of existing rows, which is why it is not a drive-by fix here.

## Verification

NPC compendium page, before and after (demo data seeded locally — the seeded dev DB has no
`npc_spells` rows). The table stays within its `.table-responsive` container; no page errors.

**Before**
npc before

**After**
npc after

**Spell details page**
spell after


app/Service/CombatLog/DataExtractors/ImmunityBypassDataExtractor.php
```diff
@@ -0,0 +1,657 @@
...
+
+            foreach ($this->pendingBypassObservations as $observation) {
+                $this->applyBypassToSpell($result, $spells, $observation['spell_id'], $observation['property']);
+                $this->assignSpellToNpc($result, $spells, $observation);
+            }
+        }
+
+        $this->pendingBypassObservations = collect();
+        $this->currentCombatLogFilePath  = null;
+        $this->currentDungeonId          = null;
```
Wotuu, line 286: :robot: `recentNpcCastSuccesses` is never reset between combat logs. `afterExtract()` clears `pendingBypassObservations`, `currentCombatLogFilePath`, `currentDungeonId` and (via `closeAllImmunityWindows()`) the windows, but leaves the cast-provenance map populated. The precedent this mirrors does reset everything: `SpellCounterDataExtractor::afterExtract()` ends with `$this->resetCorrelationState()`.

This matters because the extractor instance is shared across files — `CombatLogDataExtractionService` builds the extractors once in its constructor and `combatlog:extractdata` calls `extractData($filePath)` in a loop over many paths. Carried-over entries are keyed by `casterGuid-spellId` (creature GUIDs repeat across logs of the same dungeon), and `lastNpcCastSuccessAt()` only rejects entries with `millisecondsBetween($castAt, $timestamp) <= 15000` — for a log whose timestamps are *earlier* than the previous log's (batch runs are not guaranteed chronological) that difference is negative and therefore passes, handing a bogus `$resolvedAt` to the in-flight gate. It also keeps up to `CAST_PROVENANCE_PRUNE_THRESHOLD` entries alive for the whole process lifetime.

Suggest clearing it here (and ideally at the run boundary in `extractData()` alongside `closeAllImmunityWindows()`).

app/Service/CombatLog/DataExtractors/ImmunityBypassDataExtractor.php
```diff
@@ -0,0 +1,657 @@
...
+            $this->log->extractDataBypassAlreadyQueued($spellId, $property->value);
+
+            return;
+        }
+
+        $this->pendingBypassObservations->put($dedupKey, [
+            'spell_id'   => $spellId,
+            'property'   => $property,
+            'npc_id'     => $npcId,
+            'dungeon_id' => $this->currentDungeonId,
```
Wotuu: :robot: `dungeon_id` is read at **flush** time, not at candidate time, so a candidate can be attributed to the wrong dungeon.

`queueBypassObservation()` is only reached from `flushWindowCandidates()`, which runs when the window closes. Two of those close paths happen after `currentDungeonId` has already moved on:

- `extractData()` assigns `$this->currentDungeonId = $currentDungeon->dungeon->id` as its *first* statement, and the service updates `$currentDungeon` before dispatching the event to the extractors. So on a `ZoneChange` into a different dungeon, `closeAllImmunityWindows()` flushes the previous dungeon's candidates with the **new** dungeon's id.
- `afterExtract()` flushes any still-open window with whatever dungeon the log ended in.

The consequence is a wrong `SpellDungeon` row in `assignSpellToNpc()`, which then feeds the season filter in `DetectStaleCombatLogDataCommand::removeStaleSpellProperty()` and the compendium "Dungeons" section. `SpellCounterDataExtractor` does not have this problem because it queues the observation at detection time.

Cheap fix: capture `dungeon_id` (and the timestamp-relevant context) on the candidate when it is recorded in `recordBypassCandidates()`, or on the window when it is opened, and pass it through to `queueBypassObservation()`.

app/Service/CombatLog/DataExtractors/ImmunityBypasses/ImmunityDefinitions.php
```diff
@@ -0,0 +1,90 @@
...
+            ),
+            // Absorb-based: magic damage still lands (and is absorbed), only harmful magic *effects* are immune
+            new ImmunityDefinition(
+                SpellProperty::BypassAntiMagicShell,
+                Spell::IMMUNITY_ANTI_MAGIC_SHELL,
+                [Spell::SPELL_ANTI_MAGIC_SHELL],
+                Spell::SCHOOLS_MASK_MAGIC,
+                false,
+                true,
+                5000,
```
Wotuu: :robot: Anti-Magic Shell's duration is not fixed at 5 s: the DK talent **Anti-Magic Barrier** increases it by 2 s (to 7 s), and it is a very commonly taken talent. Because `getMaxDurationMs()` is used as a hard cutoff in `recordBypassCandidates()` (`$windowAgeMs > $definition->getMaxDurationMs()` → drop the window), a talented AMS window is force-expired at 5 s even though its `SPELL_AURA_REMOVED` arrives at ~7 s, so:

- every harmful aura landing in the 5–7 s slice is silently discarded (false negatives), and
- the candidates gathered in 0–5 s are flushed through the `$removedAt === null` path, losing the trailing-edge ε protection.

That is significant here specifically: per the PR body, AMS accounted for 226 of the 310 observed windows and produced *all 5* detections, so this is the one duration where being short actually costs data. The other five durations look right (DS 8 s, Ice Block 10 s, AotT 8 s, BoP/BoS 10 s).

Since the interface doc says the removal line is the authority and the max duration only exists for logs where it never arrives, bumping AMS to the talented 7000 ms costs nothing in false positives — the removal still closes the window earlier in the normal case.

app/Service/CombatLog/DataExtractors/ImmunityBypassDataExtractor.php
```diff
@@ -0,0 +1,657 @@
...
+     *
+     * @param array{definition: ImmunityDefinitionInterface, buffSpellId: int, appliedAt: Carbon, candidates: array<int, array{spellId: int, npcId: int, ts: Carbon, kind: string}>} $window
+     */
+    private function flushWindowCandidates(string $playerGuid, array $window, ?Carbon $removedAt): void
+    {
+        $definition = $window['definition'];
+        $property   = $definition->getProperty();
+
+        foreach ($window['candidates'] as $candidate) {
+            if ($removedAt !== null) {
```
Wotuu, line 559: :robot: The trailing-edge ε is only applied when an explicit `SPELL_AURA_REMOVED` closed the window; on the expiry path (`$windowAgeMs > getMaxDurationMs()` in `recordBypassCandidates()`) `$removedAt` is `null` and every candidate is accepted unconditionally — including one recorded at, say, window age 7 999 ms of an 8 000 ms Divine Shield, which is exactly the ambiguous case the ε exists for. The window end is known in that path (`appliedAt + getMaxDurationMs()`), so the same ε could be applied. End-of-log / run-boundary flushes genuinely have no known end and are fine as-is.

app/Service/CombatLog/DataExtractors/ImmunityBypassDataExtractor.php
```diff
@@ -0,0 +1,657 @@
...
+        Creature $sourceGuid,
+        int      $schoolMask,
+        string   $kind,
+        Carbon   $timestamp,
+    ): void {
+        $windows = $this->activeImmunityWindows->get($playerGuid);
+        if ($windows === null) {
+            return;
+        }
+
```
Wotuu, line 480: :robot: The in-flight gate only ever sees the *most recent* cast for a `(casterGuid, spellId)` pair, because `rememberNpcCastSuccess()` overwrites the map entry. For an NPC that repeats the same ability quickly — the exact profile of the validated Arcane Bolt case, cast every ~2 s — a re-cast *inside* the window overwrites the earlier cast, so a hit that was actually in flight from the pre-window cast is judged against the newer `$resolvedAt` and is no longer rejected. The bias is toward false positives, i.e. the direction this detector cares about most.

Not necessarily blocking (a projectile's travel time is short and the ε already absorbs some of this), but worth noting in the class docblock alongside the other provenance caveats, or fixing by keeping a small ring of recent casts per key and picking the newest one at or before the hit that is at least the projectile travel time old.

Wotuu:

app/Service/CombatLog/DataExtractors/ImmunityBypassDataExtractor.php
Wotuu: :robot: Fixed in 1f46794. `afterExtract()` now clears `recentNpcCastSuccesses`, and the age check rejects a negative age outright rather than treating a cast from the future as recent. Covered by `extractData_givenACastRememberedFromAPreviousCombatLog_doesNotAffectTheNextOne` — the second log deliberately starts later than the leftover cast, so a surviving entry would veto its hit as in-flight and fail the test.

app/Service/CombatLog/DataExtractors/ImmunityBypassDataExtractor.php
Wotuu: :robot: Fixed in 1f46794. The candidate now carries the dungeon it was seen in, captured at record time, and the flush uses that instead of `currentDungeonId`. Covered by `extractData_givenAZoneChangeAfterTheHit_attributesTheSpellToTheDungeonItWasSeenIn`, which asserts the `SpellDungeon` row names the original dungeon and not the one the ZoneChange moved to.

app/Service/CombatLog/DataExtractors/ImmunityBypasses/ImmunityDefinitions.php
Wotuu: :robot: Fixed in 1f46794 — raised to 7000 ms with a comment saying why. Worth noting the blast radius was real: AMS was 226 of the 310 windows in the validation corpus, so this was the one duration where being wrong actually costs detections.

app/Service/CombatLog/DataExtractors/ImmunityBypassDataExtractor.php
Wotuu: :robot: Fixed in 1f46794. The expiry path now passes `appliedAt + maxDurationMs` as the window end, so the trailing ε applies there too; only a truncated window (end of log, run boundary) still flushes with an unknown end, where there genuinely is no end to measure against. Covered by `extractData_givenDamageWithinEpsilonOfAnUnloggedExpiry_doesNotSetImmunityBypass`.

app/Service/CombatLog/DataExtractors/ImmunityBypassDataExtractor.php
Wotuu: :robot: Fixed in 1f46794. Provenance keeps up to 8 recent resolutions per (caster, spell) and the gate rejects the hit if **any** of them resolved before the window — the conservative direction, since a re-cast inside the window cannot be told apart from the earlier projectile finally landing. Covered by `extractData_givenARecastInsideTheWindowAfterAnEarlierCast_doesNotSetImmunityBypass`.

Wotuu:
I have just a question after reading what you just wrote: what exactly is it going to show? If I have what I want to display on the side, it is this: if a boss casts a certain spell, namely Haha, I win. I want it to show that it will pierce an Ice Block or a Divine Shield, because that is useful information. If he is casting that ability, I cannot immune it. Maybe I should avoid it by standing behind the rock, which is the mechanic that the boss wants me to use, but I cannot avoid it by just immuting it and attempting to do it anyway. That's what I want to display.

What I don't want to display is "Oh, I have an anti-magic shell up, but I'm still getting hit by physical attacks," because that's just how anti-magic shell works. You should know that it is anti-magic and not anti-getting hit in the face by an axe. If there's something the other way around, like Warriors' Retaliation, that blocks all physical attacks, if I'm not mistaken, and it does nothing against magic, then if you put up Retaliation and a Frostbolt hits you, I don't want it to be displayed anywhere because that doesn't make sense.

Just to fully align on what's being displayed here, I think maybe your screenshots put some doubt in me, since the screenshots were generated with seeded data, but I just want to confirm with you that that's how it works. I don't want to display stuff that shouldn't be displayed. 

One addition to this: if I have anti-magic shell up and there is a magic spell that does still hit me, then of course I still want to be able to see it. If I have retaliation on and some sort of melee attack still hits me, then yes, that's something that I want to display. Effectively, things that I should be immune to, but I am not immune to them because of some reason, maybe some gameplay reason or whatever. I want it displayed. Otherwise, I don't care.

app/Service/CombatLog/DataExtractors/ImmunityBypasses/ImmunityDefinition.php
```diff
@@ -0,0 +1,64 @@
...
+     * @param array<int> $buffSpellIds
+     */
+    public function __construct(
+        private readonly SpellProperty $property,
+        private readonly int           $immunityBit,
+        private readonly array         $buffSpellIds,
+        private readonly int           $protectedSchoolsMask,
+        private readonly bool          $blocksDamage,
+        private readonly bool          $blocksHarmfulAuras,
+        private readonly int           $maxDurationMs,
```
Wotuu, line 23: One thing I do want you to be careful about is these max durations. They may be affected by talents, so I don't know which talent specifically, but I can imagine that there is a talent that says, "Oh, your divine shield doesn't have a 6-second duration, but it now has a 10-second duration." In that case, the divine shield should have a max duration of 10.

In theory, this information is findable in the combat log. The combatant info will have all the talents that a player has selected, but I think it's a bit overkill to start parsing that and to start tracking individual talents to see what they do and see if they have a talent, and then adjust the duration based on that. It is something I do want to mention. I don't know if you've taken any of that into account or not?

Wotuu:
:robot: Confirmed against the code, not just the description - what gets displayed is exactly "this boss ability lands despite an immunity that should have stopped it," never "this immunity works as everyone already knows it works."

**Mechanism.** Each of the six tracked immunities (`ImmunityDefinitions.php`) declares two things:
- `blocksDamage()` / `blocksHarmfulAuras()` - which *kind* of landed event that immunity actually claims to stop
- `protectedSchoolsMask` - which damage schools it covers

`recordBypassCandidates()` in `ImmunityBypassDataExtractor.php` gates every candidate hit through both, in order:
1. Kind gate (line 515): a hit that dealt damage is only even considered if `blocksDamage()` is true for that immunity; a debuff application only if `blocksHarmfulAuras()` is true.
2. School gate (line 520): `($schoolMask & $definition->getProtectedSchoolsMask()) === 0` discards it.

So for your two examples:
- **Retaliation** isn't tracked at all - it has no `ImmunityDefinition`, so it can never appear on either side of this feature. You used it as a "shouldn't show up" example; it structurally can't.
- **Anti-Magic Shell**: its `protectedSchoolsMask` is magic-only, so a physical hit dies at the kind gate before school is even checked (`extractData_givenPhysicalDamageInsideBlessingOfSpellwarding_doesNotSetImmunityBypass` is the Spellwarding version of that same case; AMS gets its own test too). Blessing of Protection is physical-only and blocks the inverse for the same reason.

**One asymmetry worth flagging explicitly**, since it's the one place the code says "no" to something you described as "yes": AMS's `blocksDamage()` is `false`. That's not an omission - AMS absorbs magic damage into runic power rather than preventing it, so damage landing while AMS is up is how the ability works, not a failure of it (`extractData_givenMagicDamageInsideAntiMagicShell_doesNotSetImmunityBypass`). What AMS is genuinely immune to is harmful magic *auras/effects* (CC, debuffs) - a magic debuff landing through AMS **is** flagged (`extractData_givenHarmfulAuraInsideAntiMagicShell_setsImmunityBypass`). So: magic debuff through AMS -> shown. Magic damage through AMS -> not shown, by design. If you'd rather see the damage case too, it's a single boolean flip (`blocksDamage: true` on that one definition) - say the word and I'll change it, but I believe the current behaviour matches actual AMS mechanics.

**The list is deliberately closed at six abilities** whose coverage can be stated exactly (Divine Shield, Ice Block, Aspect of the Turtle, Blessing of Protection, Blessing of Spellwarding, Anti-Magic Shell). Cloak of Shadows, Nether Ward, Dispersion, Evasion and Die by the Sword are named exclusions in the code (`ImmunityDefinitions.php` docblock) - their coverage is either version-dependent or they're damage reduction/avoidance rather than true immunity, so they're left out rather than guessed at.

**Self-correcting if something wrong does get in**: the hourly `combatlog:detectstaledata` sweep re-checks every recorded bypass against recent observations and clears the bit (`immunity_bypass_removed`) if it stops being seen, the same mechanism already used for counters and miss types.

The screenshots in the PR description used seeded/synthetic data purely to have something to screenshot with - the gating logic above is what actually decides what ends up on a spell's row in production, and it's covered by 28 tests in `ImmunityBypassDataExtractorTest.php` exercising exactly these boundary cases (physical during a magic-only immunity, magic during a physical-only immunity, damage vs. aura per immunity, etc).

app/Service/CombatLog/DataExtractors/ImmunityBypasses/ImmunityDefinition.php
Wotuu: :robot: Good catch, and yes I'd considered it for one of the six but not systematically - here's the actual mechanics.

In the normal case (a `SPELL_AURA_REMOVED` line is present) the window boundary comes from that real timestamp, not from `maxDurationMs` - so a talent that extends the buff's actual duration is already handled for free, no talent parsing needed.

`maxDurationMs` only matters in two situations, and it behaves differently in each:
1. **The removal line never arrives at all** (log ends, run ends, zone change) - here the constant *is* the boundary, since there's nothing else to go on.
2. **A hit lands on the player after `maxDurationMs` has elapsed, but before the real removal line has been processed** - I checked this while answering, and it force-closes the window right there (`ImmunityBypassDataExtractor.php`, `recordBypassCandidates`), even though the actual `SPELL_AURA_REMOVED` may arrive later and prove the immunity was still up. I've fixed the docblock on that branch, which incorrectly claimed this only happens when no removal was ever logged.

So a talent-extended Divine Shield (say, hypothetically, 10s instead of 8s) with a genuine bypass at 9s would currently be dropped, not shown - a missed detection, never a false "bypasses immunity" claim. That failure direction matches the rest of the detector's deliberately conservative design (see the file's top-level docblock and `ImmunityDefinitions.php`'s "not included, and why" section).

Anti-Magic Shell already got this treatment: its constant is 7000ms against a 5s base specifically because Anti-Magic Barrier is a commonly-taken talent that adds 2s, and that's called out in the code comment. I did not blanket-pad the other five, because padding only ever matters in the "no removal line" case above - and there, too little padding costs a missed detection (safe), but too much costs a false positive (the one thing this feature must never produce). Padding without a known talent and duration would be a guess in the wrong direction. I'm not aware of a specific duration-extending talent for Divine Shield, Ice Block, Aspect of the Turtle, Blessing of Protection or Blessing of Spellwarding, but I'd rather ask than assert that from memory - if you know of one (talent + duration), point me at it and I'll pad that immunity's constant the same way AMS's was.

I did not implement talent parsing from combatant info, per your comment that it's overkill for this.

PR #3844, opened 2026-08-05
#3821 Replace the parse-failure API with a run-keyed segments endpoint

### LINKED ISSUE #3821: Sentry migration 2/5: replace the parse-failure API with a run-keyed segments endpoint and remove the parse-failure surface

Wotuu:
:robot: Closes #3821

Phase 2/5 of moving combat log parse failures onto Sentry. #3820 shipped in **v15.9.2** and deployed
to production at 07:45 today, so the same failures are already reaching Sentry as fingerprinted,
tagged errors — the bespoke DB pipeline that used to carry them has no reader left.

## The one thing that had to survive: the reproduction loop

Fetching a failing run's Raider.IO log segments is the step that makes a parse failure actionable,
and it was keyed on a parse-failure row id. It is now keyed on `seasonId` + `runId` — exactly the
two tags `FingerprintsStructuredErrorsHandler` promotes — so a Sentry event alone is enough to reach
the logs:

```
GET /api/v1/combatlog/seasons/{season}/runs/{runId}/segments
```

Both identifiers arrive in the path, which validation does not read from, so the new form request
merges them into `validationData()`. That is what makes the `exists` rule the thing that rejects an
unknown season — a 422 instead of a `firstOrFail()` 500 further down. The `..._no_season` error
message is gone with it: the season is a validated precondition now, not a nullable column.

## Expand half of an expand/contract

The table, model, factory and migration **stay in place**. Deploys are not atomic — the `migrate`
cron runs independently of the ECS rollout — so old containers must never meet a missing schema.
The drop is #3822, one release later. Unresolved rows are inaccessible-but-present until then, which
is intentional: Sentry has carried the same information since v15.9.2, so there is nothing to
migrate.

## Removed

The three parse-failure API endpoints and their controller, `AdminToolsCombatLogParseFailureController`
+ its routes + view + tools card, the `numCombatLogParseFailures` nav badge, both resources (and their
`@OA\Schema` annotations), the repository + interface + binding, the `recordFailure()` call in
`ProcessCombatLogSegments`, the dead lang keys, and the feature tests for the removed controllers.

`ResolvesCombatLogParseFailureSegments` went too. It only existed because the API and admin
controllers shared it, and with the admin controller gone it was one caller wide, so the logic is
inlined into `APICombatLogRunController` rather than renamed (the issue text asked for a rename;
Wotuu confirmed folding it instead).

## One judgement call worth a look

- **The parse-failure skills are only partially updated.** #3823 owns the rewrite around the Sentry
  feed. Leaving them pointing at endpoints that now 404 would be a defect this PR creates, so the
  dead sections carry a banner and the one step that survives (segment download) is retargeted at the
  new endpoint. `admin-tools-page` referenced the deleted page as its reference implementation and now
  points at the run data page.

## Verification

- `--group=CombatLog`: 253 passed. New `APICombatLogRunControllerTest`: 5 passed (happy path, no
  segments → 404, unknown season → 422 without calling Raider.IO, non-admin → 403, guest → 403).
- `route:list` no longer lists any parse-failure route; the new one is registered under `api_role:admin`.
- `l5-swagger:generate --all` regenerates both specs with no dangling `CombatLogParseFailure*` refs,
  and the new endpoint present in the internal-team spec.
- `composer run fix` (one alignment reflow in `view_common.php` from the removed key) and
  `composer run analyse` — no errors.
- Visual: before/after below.

### Admin user nav dropdown

| Before (master) | After |
|---|---|
| before | after |

The "Combat log parse failures" entry and its unresolved-count badge are gone; nothing else moved.

### Admin tools landing page

| Before (master) | After |
|---|---|
| before | after |

The Combat Log card drops from five entries to four. Baseline captured in this same worktree by
checking the tree out at the merge-base, so the two shots differ only by this branch's changes.


app/Repositories/Interfaces/CombatLog/CombatLogParseFailureRepositoryInterface.php
```diff
@@ -1,35 +0,0 @@
...
- * @method bool                                   delete(CombatLogParseFailure $model)
- * @method Collection<int, CombatLogParseFailure> all()
- * @method bool                                   exists(array<int, string> $columns)
- */
-interface CombatLogParseFailureRepositoryInterface extends BaseRepositoryInterface
-{
-    /**
-     * Records (or refreshes) a parse failure. Deduplicated on `(run_id, line_number)` so repeated failures
-     * for the same run and line update the existing row rather than creating a new one.
-     */
```
Wotuu, line 25: :robot: This PR deletes `CombatLogParseFailureRepositoryInterface` / `CombatLogParseFailureRepository`, but `.claude/skills/combatlog-parsing-internals/SKILL.md` still documents them as live: the whole `## How parse failures get recorded` section (lines ~166-174) describes `ProcessCombatLogSegments::handle()` calling `CombatLogParseFailureRepositoryInterface::recordFailure()` and `updateOrCreate`-ing a `CombatLogParseFailure` row deduplicated on `(run_id, line_number)`, and the `## Key files` table has a `Failure record model/repo` row pointing at `app/Repositories/{Interfaces,Database}/CombatLog/CombatLogParseFailureRepository*.php`.

That skill was not touched by this PR and is not in #3823 scope (which owns the two `combatlog-parse-failure-*` skills), so it will keep telling future sessions to look for classes that no longer exist — the exact defect the PR body says it wanted to avoid creating. It needs either the same banner treatment or a rewrite pointing at `ProcessCombatLogSegmentsLogging::handleParseError()` + `FingerprintsStructuredErrorsHandler`.

.claude/skills/admin-tools-page/SKILL.md
```diff
@@ -9,9 +9,9 @@ description: >
 
 # Admin tools page conventions
 
-How admin pages are wired in this project. Reference implementation: the "combat log parse failures"
-page (controller `AdminToolsCombatLogParseFailureController`, view
-`resources/views/admin/tools/combatlog/parsefailures.blade.php`).
+How admin pages are wired in this project. Reference implementation: the "combat log run data"
+page (controller `AdminToolsCombatLogRunDataController`, view
+`resources/views/admin/tools/combatlog/rundata.blade.php`).
```
Wotuu, line 14: :robot: Two problems with retargeting this skill at the run-data page:

1. **Stale reference left behind in the same file.** Line 73 still says the nav count badge is fed by `GlobalComposer` — "see `numUserReports` / `numCombatLogParseFailures`" — but this PR deletes `numCombatLogParseFailures` from `GlobalComposer::compose()`. Since the point of editing this file was to stop it pointing at removed code, that reference should go too (`numUserReports` alone still works as the example).

2. **The new reference implementation is a questionable thing to copy.** `AdminToolsCombatLogRunDataController::index()` is built entirely on `DB::connection('combatlog')->table(...)` with `DB::raw()` / `selectRaw()`. That is a deliberate choice there (covering-index scan avoiding the mediumtext column), but it directly contradicts the repo rule "Avoid `DB::`; prefer `Model::query()`". Naming it as *the* reference implementation for new admin tools pages risks propagating raw query-builder use into pages that have no reason for it — worth either picking a different reference controller or adding one sentence noting the query style there is a deliberate exception, not part of the pattern.

.claude/skills/combatlog-parse-failure-triage/SKILL.md
```diff
@@ -65,20 +73,20 @@ typically 90%+ of the rows. `exceptionClass` + the stable part of `message` is t
...
 don't cache the URLs. Response:
 
 ```json
 {"segments": [{"id": 1, "type": "trash", "downloadUrl": "https://s3..."}, ...]}
 ```
 
-If it 404s with `combatlog_parse_failure_no_segments`, or 422s with `..._no_season`, that run's
-segments aren't available (Raider.IO expired them, or the failure predates season resolution) — pick
-a different row from the cluster.
+A 422 means the season id isn't a known season. A 404 with `apicombatlogrun.error.no_segments` means
```
Wotuu: :robot: This guidance won't match what the endpoint actually returns. `ResolvesCombatLogRunSegments` returns the *translated* string:

```php
'error' => __('controller.apicombatlogrun.error.no_segments'),
```

so the 404 body is `{"error":"No combat log segments are available for this run."}` — the literal token `apicombatlogrun.error.no_segments` never appears in the response (and the full key is `controller.apicombatlogrun.error.no_segments` anyway, so it's also written wrong here). An operator or agent grepping the response for that string per this runbook will never match it.

Same applies to the 422 sentence just above: the validation failure comes back through `APIFormRequest::failedValidation()` as `{"success":false,"message":"Validation errors","data":{"season":[...]}}`, which is worth stating since it's a differently-shaped body from the 404 (`error` key) and this is the one step of the skill the banner says is still trustworthy.

Wotuu:

All three are documentation-accuracy issues in `.claude/skills/`; the PHP side came out clean.

Verified and found no issue with:
- **`CombatLogRunSegmentsRequest`** — the `validationData()` merge is correct and complete. Route params are merged *last*, so a `?season=` query param cannot override the path value; `integer` rejects non-numeric and out-of-`int`-range path segments before `exists` runs; there is no `Route::model`/`Route::bind` for `season` anywhere and the first-class-callable signature has no `Season` type hint, so implicit route-model binding never engages and `$this->route('season')` really is the raw path string. There is no path from a bad `{season}`/`{runId}` to `firstOrFail()` or to `RaiderIOApiServiceInterface`. Member ordering (protected `getRequestDtoClass()` first) matches every sibling `APIFormRequest`.
- **`CombatLogParseFailure` consumers** — a repo-wide grep (all file types, excluding vendor/node_modules) finds zero remaining code, view, route, lang or test references outside the intentionally-retained model, factory and migration. No dangling `$ref` to the removed `@OA\Schema`s either.
- **`ProcessCombatLogSegments`** — retry semantics are unchanged (the `RuntimeException` rethrow branch is untouched), and `handleParseError()` still carries every field the deleted row did: `runId`, `seasonId`, `combatLogVersion`, `lineNumber`, `exceptionClass`, `message`, `rawLine`. `FingerprintsStructuredErrorsHandler` promotes the first five to tags and deliberately leaves `rawLine` in context, so nothing that was persisted is lost.
- **Tests** — the new controller test covers happy path, no-segments 404, unknown-season 422 with an explicit `never()` on the Raider.IO mock, non-admin 403 and guest 403; the reworked job test correctly makes `handleParseError` the sole assertion.

app/Repositories/Interfaces/CombatLog/CombatLogParseFailureRepositoryInterface.php
Wotuu: :robot: Fixed in 24365f0f7. Rewrote `## How parse failures get recorded` → `## How parse failures get reported` around `handleParseError()` and the Sentry tap that consumes it, and replaced the key-files row with rows for the fingerprint handler and the new segments controller. Good catch that this file was in neither this PR nor #3823 — it would have outlived both.

.claude/skills/admin-tools-page/SKILL.md
Wotuu: :robot: Fixed in 24365f0f7. Dropped `numCombatLogParseFailures` from the badge example (`numUserReports` still stands), and the rundata pointer now says explicitly to take the *view* from it and not the data access, since its controller goes through `DB::connection(...)`. I kept it as the reference rather than hunting for a `Model::query()`-based tools page — the blade is what the section is actually about.

.claude/skills/combatlog-parse-failure-triage/SKILL.md
Wotuu: :robot: Fixed in 24365f0f7. You are right that the key never reaches the wire. The section now gives both bodies verbatim — the 404 as `{"error": "No combat log segments are available for this run."}` with a note to match on the status code rather than the key, and the 422 in the `APIFormRequest::failedValidation()` envelope. Worth getting right precisely because the banner points operators at this one section.

Wotuu:
This is the second time I've noticed that there is no issue number in the title, so please ensure you do that.

Wotuu:
:robot: Fixed — title now prefixed with `#3821`.

PR #3846, opened 2026-08-05
#3845 Parse the hex prefix spell school so combat-log-created spells keep their school

### LINKED ISSUE #3845: Combat-log-created spells get schools_mask 0 because the hex prefix school is cast with (int)
:robot: Noticed while building the immunity-bypass detector in #3835 / #3843, which needed the spell
school of a combat-log event. Not fixed there because it is a different feature and the fix rewrites
existing spell data.

## The bug

`SpellCreationCollector::ensureSpellExists()` casts the *prefix* spell school straight to `int`:

```php
// app/Service/CombatLog/DataExtractors/SpellDataCollectors/SpellCreationCollector.php:47
$createdSpell = $this->createSpellModel($result, $spellId, $prefix->getSpellName(), (int)$prefix->getSpellSchool());
```

`Range::getSpellSchool()` returns the raw combat-log field, which is a **hex string** — `"0x20"`,
`"0x1"`, `"0x40"`. In PHP `(int)"0x20"` is `0`, not `32`:

```
> var_dump((int)"0x20", intval("0x20", 16));
int(0)
int(32)
```

So **every spell first created from a combat log gets `schools_mask = 0`** — no school at all. The
Compendium's "Schools" column renders empty for them, and anything filtering or reasoning on
`schools_mask` silently sees nothing.

## Evidence

**Correction to the original filing:** I first wrote that all 605 spells with `schools_mask = 0` on
a fresh seed were the combat-log-created ones. That was wrong. Broken down:

| | |
| --- | --- |
| Spells with `schools_mask = 0` | 605 of 7916 |
| …of which have a `category` (class spells from `SpellService::importFromCsv`, which hardcodes `'schools_mask' => 0`) | 569 |
| …of which have a null `category` — the combat-log-created signature | **36** |
| Spells with a null `category` in total | 38 |

So the real blast radius is **36 spells, not 605** — but **36 of the 38** combat-log-created spells
have no school, which is about as clean a confirmation of the bug as the data can give. The reason
it is not worse is that `wowhead:fetchspelldata` overwrites `schools_mask` correctly whenever it
runs; only spells it has never covered stay broken. That command is not scheduled, so "eventually"
is doing a lot of work there.

The 569 CSV-imported class spells are a separate matter and out of scope here — that importer sets
0 deliberately and those spells get their school from Wowhead.

Confirmed against a real log (run `40415388`, Raider.IO):

```
SPELL_DAMAGE,...,275779,"Judgment",0x2,...     ← prefix school, hex
SPELL_INTERRUPT,...,396640,"Healing Touch",8   ← suffix extraSchool, decimal
```

Only the **prefix** is affected. `SpellBase::getExtraSchool()` and `DamageInterface::getSchool()`
are typed `int` and those fields are decimal, so the interrupt and damage paths are fine as-is —
which is why nothing else has tripped over this.

## Suggested fix

Parse hex-or-decimal at the single call site (or, better, have `Range::getSpellSchool()` expose a
parsed `int` alongside the raw string, since a hex string is not a useful public type):

```php
str_starts_with(strtolower($school), '0x') ? (int)hexdec(substr($school, 2)) : (int)$school
```

A migration cannot backfill the broken rows: the school is not derivable from anything in the
database, only from a log that mentions the spell again (or from Wowhead). So the repair belongs in
the collector — when it sees an event for a spell whose `schools_mask` is 0, fill it in from the
event.

## Definition of done

- [ ] `SpellCreationCollector` writes the correct `schools_mask` for combat-log-created spells
- [ ] Test covering a hex prefix school (a `"0x20"` event produces `schools_mask = 32`)
- [ ] Repair path for existing `schools_mask = 0` spells (no migration can recompute them)
- [ ] `composer run analyse` + `composer run fix` clean

Wotuu:
:robot: Closes #3845

The combat log writes the **prefix** spell school as a hex string (`0x20`), and `(int)"0x20"` is
`0` in PHP — not `32`. `SpellCreationCollector` did exactly that cast, so every spell it created
from a combat log was stored with no school at all.

## Scale

I filed #3845 claiming all 605 `schools_mask = 0` spells were combat-log-created. That was wrong,
and the issue body is now corrected — the real breakdown on a fresh seed:

| | |
| --- | --- |
| Spells with `schools_mask = 0` | 605 of 7916 |
| …with a `category` — class spells from `SpellService::importFromCsv`, which hardcodes `'schools_mask' => 0` | 569 |
| …with a null `category` — the combat-log-created signature | **36** |
| Combat-log-created spells in total | 38 |

So this bug owns **36 spells, not 605** — but it owns 36 of 38, which is about as clean a
confirmation as the data gives. It is not worse only because `wowhead:fetchspelldata` rewrites
`schools_mask` correctly whenever it runs; that command is not scheduled, so spells it has never
covered stay school-less indefinitely. The 569 CSV-imported class spells are a separate, deliberate
`0` and are out of scope.

## The fix

`Range::getSpellSchoolMask()` parses hex or decimal and is used at the one call site. Decimal is
accepted too, since that is how the *suffix* schools are logged and it costs nothing to be
tolerant.

Verified against a real log (run `40415388`, from #3843's recorded corpus) that only the prefix is
affected:

```
SPELL_DAMAGE,...,275779,"Judgment",0x2,...     ← prefix school, hex
SPELL_INTERRUPT,...,396640,"Healing Touch",8   ← suffix extraSchool, decimal
```

`SpellBase::getExtraSchool()` and `DamageInterface::getSchool()` are typed `int` and read decimal
fields, so the interrupt and damage paths were never wrong — which is why nothing else tripped over
this.

## Repairing the existing rows

**No migration can backfill these.** The school is not derivable from anything in the database; it
exists only in the logs (or on Wowhead). So the repair lives in the collector: when it sees an
event for a spell it already knows whose `schools_mask` is still `0`, it fills it in from the
event. Guarded both ways — it never overwrites a school that is already set, and an event that
itself carries no school (`0x0`) writes nothing.

That is not only a one-off cleanup: `SpellService::importFromCsv` and `MDTMappingImportService`
both create spells with `schools_mask = 0` on an ongoing basis, and this repairs any of them the
next time an NPC is seen casting the spell.

Spells that never appear in a log again still need `wowhead:fetchspelldata`, which is the only
other source of truth for a school.

## Cold review

Reviewed by a fresh-context agent: **2 findings, both fixed** in 06794fe.

1. The repair mutated a `spells` row without writing a `CombatLogSpellEvent`, unlike every other
   spell mutation in this pipeline — so it was invisible in the Compendium activity feed and not
   attributable to a combat log. It now writes one, under a new `SchoolRecorded` event type.
   **Knock-on worth knowing:** the feed blade has *three* exhaustive `match` statements over
   `CombatLogSpellEventType`, so adding an enum case without arms throws `UnhandledMatchError` on
   every Compendium page showing spell events. All three are handled.
2. `ensureInterruptSpellExists()` kept its plain early return, so a spell that only reappears as
   the interrupted spell of a `SPELL_INTERRUPT` was never repaired — even though its `extraSchool`
   is already trusted to *create* a spell. It repairs now too.

## Verification

The activity feed gained a new line type, so this was checked in headless Chrome rather than on
the tests alone — status 200, no page errors, all three match arms handled:

activity feed

## Tests

- `RangeTest` — 11 cases over the parse: hex lower/upper case, multi-school masks, `0x0`, decimal
  and `nil`. Every hex case reads as `0` under the old cast.
- `SpellDataExtractorTest` — an unknown spell is created with the hex school; a known school-less
  spell is repaired (and writes its audit event); an interrupted school-less spell is repaired; a
  known spell with a school is left alone; a school-less spell hit by a school-less event is left
  alone (asserted through the repair log call, which must never fire).

## Note for whoever merges

`resources/views/compendium/sections/event_list.blade.php` and `lang/en_US/view_compendium.php` are
also touched by #3843 and #3842. All three changes are additive in different places, so conflicts
should be trivial.

### DIFF
--- a/app/Service/CombatLog/DataExtractors/SpellDataCollectors/SpellCreationCollector.php
+++ b/app/Service/CombatLog/DataExtractors/SpellDataCollectors/SpellCreationCollector.php
@@ -60,7 +60,14 @@ class SpellCreationCollector implements SpellDataCollectorInterface
     public function ensureInterruptSpellExists(ExtractedDataResult $result, Interrupt $interrupt): void
     {
         $spellId = $interrupt->getExtraSpellId();
-        if ($this->allSpells->has($spellId)) {
+
+        /** @var SpellModel|null $existingSpell */
+        $existingSpell = $this->allSpells->get($spellId);
+        if ($existingSpell !== null) {
+            // The interrupted spell's school is trusted to create a spell on the line below, so it is good enough
+            // to repair one too
+            $this->repairMissingSchoolsMask($result, $existingSpell, $interrupt->getExtraSchool());
+
             return;
         }
 
@@ -100,6 +107,15 @@ class SpellCreationCollector implements SpellDataCollectorInterface
         $spell->schools_mask = $schoolsMask;
 
         if ($spell->save()) {
+            // Every other spell mutation in this pipeline is auditable from the Compendium's activity feed and
+            // attributable to a combat log; a repaired school is no different
+            CombatLogSpellEvent::create([
+                'spell_id'        => $spell->id,
+                'event_type'      => CombatLogSpellEventType::SchoolRecorded,
+                'property'        => null,
+                'combat_log_path' => $this->currentCombatLogFilePath,
+            ]);
+
             $result->updatedSpell();
             $this->log->ensureSpellExistsRepairedSchoolsMask($spell->id, $schoolsMask);
         }
--- a/app/Service/CombatLog/DataExtractors/SpellDataCollectors/SpellCreationCollector.php
+++ b/app/Service/CombatLog/DataExtractors/SpellDataCollectors/SpellCreationCollector.php
@@ -39,12 +39,18 @@ class SpellCreationCollector implements SpellDataCollectorInterface
 
     public function ensureSpellExists(ExtractedDataResult $result, Spell $prefix): void
     {
-        $spellId = $prefix->getSpellId();
-        if ($this->allSpells->has($spellId)) {
+        $spellId     = $prefix->getSpellId();
+        $schoolsMask = $prefix->getSpellSchoolMask();
+
+        /** @var SpellModel|null $existingSpell */
+        $existingSpell = $this->allSpells->get($spellId);
+        if ($existingSpell !== null) {
+            $this->repairMissingSchoolsMask($result, $existingSpell, $schoolsMask);
+
             return;
         }
 
-        $createdSpell = $this->createSpellModel($result, $spellId, $prefix->getSpellName(), (int)$prefix->getSpellSchool());
+        $createdSpell = $this->createSpellModel($result, $spellId, $prefix->getSpellName(), $schoolsMask);
         $this->allSpells->put($spellId, $createdSpell);
         $this->pendingNewSpells->put($spellId, $createdSpell);
 
@@ -78,6 +84,25 @@ class SpellCreationCollector implements SpellDataCollectorInterface
 
         $this->pendingNewSpells         = collect();
         $this->currentCombatLogFilePath = null;
+    }
+
+    /**
+     * Backfills the school of a spell that has none. Spells created from a combat log before #3845 all got
+     * schools_mask 0, and their real school cannot be recovered from the database - only from a log that mentions
+     * them again, which is exactly here. Spells that legitimately have no school log 0x0 and stay untouched.
+     */
+    private function repairMissingSchoolsMask(ExtractedDataResult $result, SpellModel $spell, int $schoolsMask): void
+    {
+        if ($schoolsMask === 0 || $spell->schools_mask !== 0) {
+            return;
+        }
+
+        $spell->schools_mask = $schoolsMask;
+
+        if ($spell->save()) {
+            $result->updatedSpell();
+            $this->log->ensureSpellExistsRepairedSchoolsMask($spell->id, $schoolsMask);
+        }
     }
 
     private function createSpellModel(ExtractedDataResult $result, int $spellId, string $name, int $schoolsMask): SpellModel
--- a/app/Logic/CombatLog/CombatEvents/Prefixes/Range.php
+++ b/app/Logic/CombatLog/CombatEvents/Prefixes/Range.php
@@ -28,6 +28,20 @@ class Range extends Prefix
         return $this->spellSchool;
     }
 
+    /**
+     * The spell school as a bitmask, ready to be stored in `spells`.`schools_mask`.
+     *
+     * The raw field is a hex string in every retail log ("0x20"), and a plain (int) cast turns that into 0 rather
+     * than 32 - which is how combat-log-created spells ended up school-less (#3845). Decimal is still accepted
+     * because the *suffix* schools are logged that way, and a future format change is cheaper to absorb here.
+     */
+    public function getSpellSchoolMask(): int
+    {
+        return str_starts_with(strtolower($this->spellSchool), '0x')
+            ? (int)hexdec(substr($this->spellSchool, 2))
+            : (int)$this->spellSchool;
+    }
+
     #[Override]
     public function setParameters(array $parameters): HasParameters
     {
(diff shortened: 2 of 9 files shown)


app/Service/CombatLog/DataExtractors/SpellDataCollectors/SpellCreationCollector.php
```diff
@@ -80,6 +86,25 @@ public function afterCollect(ExtractedDataResult $result, string $combatLogFileP
...
+     */
+    private function repairMissingSchoolsMask(ExtractedDataResult $result, SpellModel $spell, int $schoolsMask): void
+    {
+        if ($schoolsMask === 0 || $spell->schools_mask !== 0) {
+            return;
+        }
+
+        $spell->schools_mask = $schoolsMask;
+
+        if ($spell->save()) {
```
Wotuu, line 109: :robot: The repair mutates a `spells` row without writing a `CombatLogSpellEvent`, which every other spell mutation in this extractor does: `ensureSpellExists` writes a `SpellCreated` event for a created spell, and `SpellPropertyObservationCollector::afterCollect` writes a `PropertyChanged` event next to its `$result->updatedSpell()`. Since those events feed the public Compendium activity/event feed (`resources/views/compendium/spell/sections/event_feed.blade.php`, `compendium/activity/*`), a school backfilled here changes a spell silently — `updatedSpells` goes up in the extraction result but nothing shows up in the feed, and there is no record of which combat log the school came from. Was that deliberate? If not, this probably wants a `PropertyChanged` event (which would need a new `SpellProperty` case for the school, since the existing enum only covers `aura`/`debuff`/miss/counter properties).

app/Service/CombatLog/DataExtractors/SpellDataCollectors/SpellCreationCollector.php
```diff
@@ -39,12 +39,18 @@ public function beforeCollect(string $combatLogFilePath): void
...
     {
-        $spellId = $prefix->getSpellId();
-        if ($this->allSpells->has($spellId)) {
+        $spellId     = $prefix->getSpellId();
+        $schoolsMask = $prefix->getSpellSchoolMask();
+
+        /** @var SpellModel|null $existingSpell */
+        $existingSpell = $this->allSpells->get($spellId);
+        if ($existingSpell !== null) {
+            $this->repairMissingSchoolsMask($result, $existingSpell, $schoolsMask);
```
Wotuu, line 48: :robot: The interrupt path is left out of the repair. `ensureInterruptSpellExists()` (line 63) still does the plain `if ($this->allSpells->has($spellId)) { return; }` early return, so a known spell with `schools_mask = 0` that only ever shows up as the *interrupted* spell of a `SPELL_INTERRUPT` never gets repaired — even though `Interrupt::getExtraSchool()` carries exactly the school needed and is already trusted enough to create new spells with it on the line below. Given the PR states the repair exists precisely because a log mention is the only recoverable source, calling `repairMissingSchoolsMask()` in that branch too looks like a one-line win (the interrupt suffix is a plausible source for some of the 38 combat-log-created spells, since that path creates spells as well).

Wotuu:

app/Service/CombatLog/DataExtractors/SpellDataCollectors/SpellCreationCollector.php
Wotuu: :robot: Fixed in 06794fe. The repair now writes a `CombatLogSpellEvent` with a new `SchoolRecorded` type, so it lands in the activity feed and carries the combat log path. Worth noting the knock-on you implied: the feed blade has **three exhaustive `match` statements** over `CombatLogSpellEventType`, so adding the case without arms would have thrown `UnhandledMatchError` on every Compendium page showing spell events — all three are handled, and I verified the rendered feed in headless Chrome rather than trusting the tests (screenshot on the PR).

app/Service/CombatLog/DataExtractors/SpellDataCollectors/SpellCreationCollector.php
Wotuu: :robot: Fixed in 06794fe — `ensureInterruptSpellExists()` now repairs a known school-less spell from `getExtraSchool()`, with a comment making your point explicit: that value is already trusted to *create* a spell on the next line, so it is good enough to repair one. Covered by `extractData_givenAnInterruptedSpellWithoutASchool_repairsItsSchoolsMask`, which uses the decimal `extraSchool` form real logs use.

Issue #1389, opened 2022-09-23
Faction specific enemies are not shown in the admin editor


Wotuu:

Wotuu:
Shadowfang Keep has this problem now, first floor.

Wotuu:
## Implementation plan — faction enemies hidden in admin editor

**Root cause (confirmed).** The backend applies **no** faction filter — `ListsEnemies::listEnemies()` (`app/Http/Controllers/Traits/ListsEnemies.php:32-89`) returns all enemies for the floor. The hiding happens purely front-end in `MapObject.shouldBeVisible()` (`resources/assets/js/custom/models/mapobject.js:749-801`).

The teeming/seasonal filters are already wrapped in `if (!state.isMapAdmin()) { … }` (lines 764-789) so admins bypass them — but the **faction** filter (lines 791-798) sits **outside** that guard. In the admin editor the map context faction is the non-`'any'` "Unspecified" name (`app/Logic/MapContext/Map/MapContextMappingVersion.php:41`), and `AdminDungeonMap.isSandboxModeEnabled()` always returns false (`resources/assets/js/custom/admin/admindungeonmap.js:45-47`), so a horde/alliance enemy (e.g. Shadowfang Keep floor 1) satisfies `this.faction !== 'any' && faction !== 'any' && this.faction !== faction` → hidden. `isMapAdmin()` is true in the edit context (`resources/assets/js/custom/statemanager.js:711-715`).

**Fix.** Move the faction block (`mapobject.js:791-798`) **inside** the existing `if (!state.isMapAdmin())` guard (i.e. append it after line 788, before the closing brace at 789), matching how teeming/seasonal are already exempted for admins. No backend change needed.

**Tests.** vitest (jsdom) is set up. Add a `shouldBeVisible` test with a mocked state:
- `shouldBeVisible_givenAdminEditorAndFactionSpecificEnemy_returnsTrue` (isMapAdmin → true);
- `shouldBeVisible_givenRouteViewAndOppositeFactionEnemy_returnsFalse` (isMapAdmin → false, preserves existing route-view behavior).

**Verify.** `npm run test`; then `npm run dev`/`npm run build`, open Shadowfang Keep floor 1 in the admin mapping editor, and confirm the horde/alliance enemies now appear.

---
<sub>🤖 Investigation + plan generated with [Claude Code](https://claude.com/claude-code)</sub>

Wotuu:
[...]

PR #3847, opened 2026-08-05
#3674 Cover the MapObject visibility/filter chain with Vitest tests

### LINKED ISSUE #3674: JS tests: cover the map object layer (Phase 2 continuation of #3292)
:robot: Phase 2 continuation child story for #3292. The three original Phase 2 stories (#3293 shared setup, #3294 pure utilities, #3295 `KillZone` model) are all closed, so Phase 2 has no open work despite coverage still being thin.

## Where coverage actually stands

| | count |
|---|---|
| JS source files in `resources/assets/js/` | 335 |
| files with a co-located `*.test.js` | 19 |
| files carrying a UMD export footer (importable by Vitest today) | 19 |

The framework question is settled and the pattern is proven. What is missing is reach: the **map object layer** — the most complex and highest-churn code in the project — is almost entirely untested, and it is exactly where a regression is most expensive and least likely to be caught by the PHP suite.

## Highest-value untested targets

Ordered by size, excluding vendored `handlebars.js` (3553 lines, not ours):

| file | lines | why it matters |
|---|---|---|
| `custom/models/enemy.js` | 1358 | the single most-touched model; drives selection, visuals, forces |
| `custom/models/mapobject.js` | 1095 | base class for every map object — `shouldBeVisible()` lives here, and #1389 is an open bug in exactly that method |
| `custom/statemanager.js` | 812 | global state; a regression here is felt everywhere |
| `custom/mapcontext/mapcontext.js` | 686 | server→client contract boundary |
| `custom/mapobjectgroups/mapobjectgroup.js` | 680 | collection lifecycle, add/remove/sync |

`mapobject.js` is the natural first target: #3295 already proved the mocked-globals pattern works for a *subclass* (`KillZone`), so covering the base class is a direct extension of an established pattern rather than new ground. It also has a known open bug (#1389 — the faction filter in `shouldBeVisible()` sits outside the `!isMapAdmin()` guard), which makes it a case where a test would have caught a real defect.

## Suggested split

- **A** — `mapobject.js`: `shouldBeVisible()` and the visibility/filter chain, including a regression test for #1389.
- **B** — `enemy.js`: forces calculation, selection state, visual state transitions.
- **C** — `statemanager.js`: state transitions and subscriber notification.

Each is independently shippable, test-only, and reviewable in isolation. A is the one to do first.

## Cost of the approach

Each target needs a guarded UMD export footer added to the source file, per the bridge strategy already agreed in #3292. That footer is a no-op in the browser bundle. This remains a stepping stone toward real ESM exports if the TypeScript migration in Phase 3 ever starts — it is not additional debt.

Wotuu:
:robot: Part of #3674 (target **A** of the suggested split: `mapobject.js`)

**Deliberately does not close #3674**: the issue enumerates three independently shippable targets (A `mapobject.js`, B `enemy.js`, C `statemanager.js`); this is A only, so the issue stays open for B and C.

## What this does

Adds `resources/assets/js/custom/models/mapobject.test.js` — 33 tests covering the
`MapObject` visibility/filter chain that every map object on every map inherits:

- **`shouldBeVisible()`**, filter by filter: floor, teeming, seasonal index, faction.
  Includes the map-admin bypass (teeming/seasonal are skipped for admins), the sandbox-mode
  escape hatch on the faction filter, the seasonal types the filter does *not* apply to,
  and filter precedence (floor wins over everything, admin or not).
- **the surrounding chain**: `setVisible`/`isVisible` (including the `_visible === null`
  first-transition case), `isDefaultVisible`/`setDefaultVisible`, and `isVisibleOnScreen`
  (no layer / hidden / in bounds / out of bounds / bounds fallback).

## #1389

`shouldBeVisible()` applies the faction filter **outside** the `!isMapAdmin()` guard the
teeming and seasonal checks live in, which is exactly the bug #1389 reports (faction-specific
enemies not shown in the admin editor). Since #3674 is scoped test-only and #1389 is a separate open bug with its own UI-visible behaviour change, this MR does **not** fix it — it adds clearly-labelled characterisation tests that pin the current behaviour.

Cold review turned up the other half of the bug: the admin map context never sends the faction *key* `any` in the first place. `MapContextMappingVersion::toArray()` emits `__(strtolower(Faction 'unspecified'->name))`, and `unspecified` is not a resolvable translation key, so the string `unspecified` reaches the client verbatim — it can never equal `any`, so the escape hatch that spares every other map never fires. The tests now model that real value, and the comment documents both halves so a fix on either side flips the expectation.

## Source change

`mapobject.js` gains only the guarded UMD export footer already used by `killzone.js`
(8 lines). It is a no-op in the browser, where `module` is undefined, and the `custom-*.js`
bundle is plain concatenation — verified by running `npm run production`.

## Verification

- `npm test` — 40 files / 386 tests green (34 new).
- Mutation-checked the new tests: inverting the `!isMapAdmin()` guard fails 2 tests, removing the sandbox-mode condition fails 3, widening the teeming condition fails 1. The cold reviewer ran 13 further mutations against a sandboxed copy; the one survivor it found is covered by the follow-up commit.
- `npm run production` builds clean.
- Backend-only/test-only: no rendered output changed, so no visual verification.

### DIFF
--- a/resources/assets/js/custom/models/mapobject.test.js
+++ b/resources/assets/js/custom/models/mapobject.test.js
@@ -9,8 +9,8 @@
 //
 // `shouldBeVisible()` is the interesting one: it is a chain of four independent
 // filters (floor, teeming, seasonal index, faction) of which only the middle two
-// sit behind the `!isMapAdmin()` guard. That asymmetry is the subject of the open
-// bug #1389 - see the last describe block.
+// sit behind the `!isMapAdmin()` guard. That asymmetry is half of the open bug
+// #1389 - see the `(#1389)` describe block for the other half.
 // ---------------------------------------------------------------------------
 
 // 1a. Seasonal type constants referenced as bare globals by shouldBeVisible().
@@ -29,37 +29,44 @@ global.Attribute = class Attribute {
 
 // 1c. A working - if small - stand-in for Signalable. MapObject registers on its own
 // signals from the constructor and emits 'shown'/'hidden' from setVisible(), so a
-// stub that swallows signals would not exercise those paths. Emitted signals are
-// also recorded on `_emittedSignals` so tests can assert on them.
+// stub that swallows signals would not exercise those paths. It mirrors the real
+// signalable.js on the parts a test can observe: listeners receive the
+// `{name, context, data}` wrapper (NOT the raw payload), and unregister() takes an
+// optional `fn` to narrow the removal. Emitted signals are additionally recorded on
+// `_emittedSignals` so tests can assert on them.
 global.Signalable = class Signalable {
     constructor() {
-        this._signals = {};
+        this._signals = [];
         this._emittedSignals = [];
         this._cleanedUp = false;
     }
 
     register(names, listener, fn) {
         for (const name of Array.isArray(names) ? names : [names]) {
-            (this._signals[name] = this._signals[name] ?? []).push({listener, fn});
+            this._signals.push({name, listener, callback: fn});
         }
     }
 
-    unregister(names, listener) {
+    unregister(names, listener, fn = null) {
         for (const name of Array.isArray(names) ? names : [names]) {
-            this._signals[name] = (this._signals[name] ?? []).filter((entry) => entry.listener !== listener);
+            this._signals = this._signals.filter((caller) => !(
+                caller.name === name && caller.listener === listener && (fn === null || caller.callback === fn)
+            ));
         }
     }
 
-    signal(name, data) {
+    signal(name, data = {}) {
         this._emittedSignals.push({event: name, data});
 
-        for (const entry of this._signals[name] ?? []) {
-            entry.fn(data);
+        for (const caller of this._signals) {
+            if (caller.name === name) {
+                caller.callback({name: name, context: this, data: data});
+            }
         }
     }
 
     _cleanupSignals() {
-        this._signals = {};
+        this._signals = [];
     }
 
     cleanup() {
@@ -194,6 +201,12 @@ describe('MapObject.shouldBeVisible - teeming filter', () => {
         expect(makeMapObject({teeming: 'visible'}).shouldBeVisible()).toBe(true);
     });
 
+    it('shows a teeming-invisible object on a non-teeming map', () => {
+        setFakeState({teeming: false});
+
+        expect(makeMapObject({teeming: 'invisible'}).shouldBeVisible()).toBe(true);
+    });
+
     it('shows an object without a teeming preference on either map', () => {
         setFakeState({teeming: true});
         expect(makeMapObject().shouldBeVisible()).toBe(true);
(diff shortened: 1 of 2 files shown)


Wotuu:

Dispatched to a fresh-context reviewer (no shared context with the implementing session), which also ran 13 mutations against a sandboxed copy of `mapobject.js`. All three findings are fixed in b299cc18b:

1. **The #1389 characterisation test did not reproduce #1389** — it modelled the admin editor as `faction: 'horde'`, a state that cannot occur. `isMapAdmin()` is only true for `MapContextMappingVersionEdit`, whose faction comes from `MapContextMappingVersion::toArray()` as `__(strtolower(Faction 'unspecified'->name))` → the literal string `unspecified`. That is the *other half* of the bug: it never equals `any`, so the escape hatch never fires. Test now uses the real value and documents both halves; the "one-word fix" claim in the MR body is corrected.
2. **A surviving mutant** — widening the teeming condition to `this.teeming !== null` kept all tests green. Added an assertion that a teeming-`invisible` object stays visible on a non-teeming map; that mutation now fails.
3. **The `Signalable` stub lied about the callback payload** — the real class hands listeners `{name, context, data}`, not the raw payload, and `unregister()` takes an optional `fn`. Since this file is the recipe for the remaining ~25 model tests, the stub now mirrors both.

Verified after the fixes: `npm test` 40 files / 386 tests green; the finding-2 mutation is killed.

resources/assets/js/custom/models/mapobject.test.js
```diff
@@ -0,0 +1,471 @@
...
+
+    it('shows an object without a teeming preference on either map', () => {
+        setFakeState({teeming: true});
+        expect(makeMapObject().shouldBeVisible()).toBe(true);
+
+        setFakeState({teeming: false});
+        expect(makeMapObject().shouldBeVisible()).toBe(true);
+    });
+
+    it('ignores the teeming filter for map admins, who must see every object', () => {
```
Wotuu: This is not for map admins. This is for when the map is in admin mode, as in when editing the mapping version.

resources/assets/js/custom/models/mapobject.test.js
```diff
@@ -0,0 +1,471 @@
...
+        expect(makeMapObject({faction: 'any'}).shouldBeVisible()).toBe(true);
+    });
+
+    it('shows a faction-specific object on a faction-agnostic map', () => {
+        setFakeState({faction: 'any'});
+
+        expect(makeMapObject({faction: 'alliance'}).shouldBeVisible()).toBe(true);
+    });
+
+    it('does not apply the faction filter in sandbox mode', () => {
```
Wotuu: Why is this, though? In sandbox mode, it's just a temporary route. It's still a regular route, so faction enemies should still work the same way as normal.

resources/assets/js/custom/models/mapobject.test.js
Wotuu: :robot: Reworded in ca67c5308 — the test now reads "ignores the teeming filter when the map is in admin (mapping version edit) mode". You're right the old title conflated "map admin" (a user permission) with `isMapAdmin()`, which actually reflects whether the map itself is being rendered in the mapping-version-edit context (`MapContextMappingVersionEdit`), not who the current user is — same distinction the file's top comment and the `#1389` describe block already draw for the faction filter.

resources/assets/js/custom/models/mapobject.test.js
Wotuu: :robot: You're right — fixed in ca67c5308. The `mapobject.js` code had `if (!this.map.isSandboxModeEnabled() && ...)` guarding the faction filter, with the comment `// Only when not in sandbox mode! (no idea why, it was like this)` — literally undocumented legacy behavior. Removed the sandbox special-case entirely, so a sandbox route now filters by faction exactly like a normal route. Updated the test (`does not apply the faction filter in sandbox mode` → `applies the faction filter the same way in sandbox mode`, now expecting `false` instead of `true`) and reran the full JS suite (41 files / 398 tests green).

PR #3848, opened 2026-08-05
#3674 Cover Enemy's forces, ignore filter and selection state with Vitest tests

### LINKED ISSUE #3674: JS tests: cover the map object layer (Phase 2 continuation of #3292)

Wotuu:
:robot: Part of #3674 (target **B** of the suggested split: `enemy.js`), alongside #3847 (target A).

Does **not** close #3674 — target C (`statemanager.js`) is still open.

## What this does

Adds `resources/assets/js/custom/models/enemy.state.test.js` — 75 tests over the parts of
`Enemy` the issue names: forces calculation, selection state, and the visual/popup state
transitions. The existing `enemy.test.js` covers the #3670 tooltip and #3703 popup regressions
only; this file deliberately does not repeat them.

- **`getEnemyForces()`** — the whole override precedence chain: shrouded and Zul'Gamux beat
  teeming, a teeming override beats the npc's teeming value, a plain override applies only off
  teeming, and a shrouded enemy without the Shrouded affix falls back to its npc value. Plus
  `setNpc()`'s forces bookkeeping and `_getPercentageString()` rounding.
- **`shouldBeIgnored()` / `shouldBeVisible()`** — awakened enemies linked to the last boss,
  each seasonal type against its affix, MDT placeholders, the `no_shrouded` swap, the Nathrezim
  Infiltrator / Zul'Gamux special case (and that a *regular* enemy marked shrouded stays visible
  without the affix — the MDT-mapping fix), dungeon difficulty, critters, and MDT enemies.
  Includes which of these the mapping editor bails out of, and that the MDT check deliberately
  does not.
- **selection state** — `isSelectable()`'s dependence on the visual, kill zone attach/detach
  signalling (including that re-setting the same id does not signal a detach), overpull and
  obsolete signalling, and the popup bind/unbind transitions driven by `map:mapstatechanged`.
- **NPC classification** — including that a rare (`classification_id` 5) also satisfies
  `isBossNpc()` (`>= 3`) and therefore counts as `isImportant()`. Pinned deliberately: it decides
  what stays visible.

## Notes for the reviewer

- **No source change.** `enemy.js` already carries the guarded UMD export footer from #3670.
- The fake state exposes a single `mappingEditor` knob that drives `isMapAdmin()`, the map context
  type, `hasAffix()` and `getTeeming()` together, because in production they move together:
  `StateManager#isMapAdmin()` is `this._mapContext instanceof MapContextMappingVersionEdit`, and
  `MapContextMappingVersion` hard-codes `hasAffix()`/`getTeeming()` to `true` ("when mapping a
  dungeon assume we have all affixes"). Independent knobs would let a test assert an editor state
  the app can never be in. The stub `extends` chain mirrors the real one for the same reason.
- **Zul'Gamux**: `shouldBeIgnored()`'s shrouded special case names `NPC_ID_ZUL_GAMUX`, but it is
  gated on `seasonal_type === 'shrouded'` — and every seeded npc 190128 row carries
  `shrouded_zul_gamux`, so that arm is unreachable for it. The test pins what actually happens
  (a Zul'Gamux stays visible without the affix, and `getEnemyForces()` falls back to its npc
  value) rather than the unreachable intent. Whether the mapping or the check is wrong is a
  question for @Wotuu — deliberately not changed here.
- `makeEnemy()` seeds `enemy_forces_override`/`_teeming` to `null` because in production those
  come from `MapObject#_setDefaults()`, which the stubbed base class does not run — `undefined`
  would take the override branch and make the fake *stricter* than the real class.

## Verification

- `npm test` — 40 files / 444 tests green (92 new).
- Mutation-checked (7): neutering `isShrouded()` in the forces chain, removing the mapping-editor
  bail-out, always signalling `killzone:detached`, re-adding the `isPopupEnabled` guard on disable,
  dropping the critter check, turning the dungeon-difficulty early return into a fall-through, and
  disabling the plain-override branch — each fails at least one test.
- Test-only: no rendered output changed, so no visual verification.

### DIFF
--- a/resources/assets/js/custom/models/enemy.state.test.js
+++ b/resources/assets/js/custom/models/enemy.state.test.js
@@ -38,9 +38,16 @@ global.KillZone = class KillZone {
         this.id = id;
     }
 };
-global.MapContextDungeonRoute = class MapContextDungeonRoute {
+// The real chain is MapContextMappingVersionEdit -> MapContextMappingVersion -> MapContext, with
+// MapContextDungeonRoute a SIBLING branch off MapContext - the mapping editor's context is not a
+// dungeon route context. Enemy#getVisualData() branches on exactly that distinction.
+global.MapContext = class MapContext {
 };
-global.MapContextMappingVersionEdit = class MapContextMappingVersionEdit extends global.MapContextDungeonRoute {
+global.MapContextDungeonRoute = class MapContextDungeonRoute extends global.MapContext {
+};
+global.MapContextMappingVersion = class MapContextMappingVersion extends global.MapContext {
+};
+global.MapContextMappingVersionEdit = class MapContextMappingVersionEdit extends global.MapContextMappingVersion {
 };
 
 // 1d. Constants referenced as bare globals, with their real values from constants.js.
@@ -67,7 +74,8 @@ global.NPC_TYPE_CRITTER = 3;
 global.NPC_ID_NATHREZIM_INFILTRATOR = 189878;
 global.NPC_ID_ZUL_GAMUX = 190128;
 // Awakened/Prideful NPC ids are hardcoded in enemy.js itself, mirrored here for readability.
-const NPC_ID_AWAKENED_OBELISK = 161124;
+const NPC_IDS_AWAKENED = [161124, 161241, 161244, 161243];
+const NPC_ID_AWAKENED_OBELISK = NPC_IDS_AWAKENED[0];
 const NPC_ID_PRIDEFUL = 173729;
 
 // 1e. Lightweight base class standing in for VersionableMapObject -> MapObject, providing only
@@ -106,12 +114,28 @@ global.VersionableMapObject = class VersionableMapObject {
         return true;
     }
 
+    // Mirrors MapObject#_assignPopup, guard included: it only binds on an editable map object of an
+    // editable map. Enemy#isEditable() returns false, so a plain enemy never gets a popup bound -
+    // dropping that guard here would let a test assert a bind production never performs.
     _assignPopup(layer = null) {
         const targetLayer = layer === null ? this.layer : layer;
 
         if (targetLayer !== null) {
-            targetLayer.bindPopup('<div>popup</div>', {});
+            targetLayer.off('popupopen');
+            targetLayer.unbindPopup();
+
+            if (this.map.options.edit && this.isEditable() && this.isEditableByPopup()) {
+                targetLayer.bindPopup('<div>popup</div>', {});
+            }
         }
+    }
+
+    isEditable() {
+        return true;
+    }
+
+    isEditableByPopup() {
+        return true;
     }
 
     unbindTooltip() {
@@ -170,8 +194,12 @@ function setFakeState({
         unregister: () => {
         },
         getAuras: () => [],
-        getTeeming: () => teeming,
-        hasAffix: (affix) => affixes.includes(affix),
+        // MapContextMappingVersion hard-codes both of these ("when mapping a dungeon assume we have
+        // all affixes so things show up properly"), so the editor cannot be short an affix or
+        // non-teeming. Honouring that here keeps the caller's affixes/teeming from describing a
+        // state the editor can never be in.
+        getTeeming: () => mappingEditor || teeming,
+        hasAffix: (affix) => mappingEditor || affixes.includes(affix),
         getEnemyForcesShrouded: () => enemyForcesShrouded,
         getEnemyForcesShroudedZulGamux: () => enemyForcesShroudedZulGamux,
         getDungeonDifficulty: () => dungeonDifficulty,
@@ -232,18 +260,19 @@ function makeFakeLayer() {
  * @param {Object} options.killZonesById Kill zones resolvable by id.
  * @returns {Object}
  */
-function makeFakeMap({mapState = null, enemyForcesRequired = 100, killZonesById = {}} = {}) {
+function makeFakeMap({mapState = null, edit = false, enemyForcesRequired = 100, killZonesById = {}} = {}) {
     const listeners = {};
+    let currentMapState = mapState;
 
     return {
-        options: {edit: false},
+        options: {edit},
         listeners,
         register(name, listener, fn) {
             (listeners[name] = listeners[name] ?? []).push(fn);
         },
         unregister() {
         },
-        getMapState: () => mapState,
+        getMapState: () => currentMapState,
         enemyForcesManager: {
             getEnemyForcesRequired: () => enemyForcesRequired,
         },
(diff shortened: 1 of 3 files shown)


Wotuu:

Dispatched to a fresh-context reviewer (no shared context with the implementing session). All six are addressed in cc330b4c6:

1. **The popup fakes bound a popup the real code never binds** — `MapObject#_assignPopup()` only binds when the map is editable *and* the object says it is, and `Enemy#isEditable()` returns `false`. The stub dropped that guard, so two tests asserted a bind production never performs. Guard mirrored; transitions now go through a `setMapState()` that swaps the state *before* notifying (as `DungeonMap` does), so a listener calling back into `getMapState()` sees a consistent world; the duplicated-with-`enemy.test.js` cases are gone, replaced by one proving a plain enemy never gets a popup.
2. **Zul'Gamux was driven with a `seasonal_type` it never carries** — every seeded npc 190128 row is `shrouded_zul_gamux`, and the special case is gated on `'shrouded'`, so that arm is unreachable for it. Now pins the real behaviour and flags the discrepancy in the MR body rather than asserting the intent.
3. **Map context fake fidelity** — `MapContextMappingVersion` hard-codes `hasAffix()`/`getTeeming()` to `true`, and `MapContextMappingVersionEdit` descends from it, not from `MapContextDungeonRoute`. The editor could therefore be modelled as short an affix, which it can never be. Both fixed, hierarchy included.
4. **The override-seeding rationale was wrong** — neither override declares an Attribute `default:`, so `_setDefaults()` skips them; they are `null` because the server sends them and `loadRemoteMapObject()` assigns them. Comment corrected (it is the template future model tests will copy).
5. **Coverage gaps in the precedence chains** — added: a plain override on a teeming map, Zul'Gamux forces without the affix, a critter shadowed by a matching dungeon difficulty (the early `return` hides the critter check), and all four awakened npc ids.
6. **Convention** — split the tests that ran two condition→assert cycles under a one-outcome name, and added the Arrange/Act/Assert comments the sibling file uses.

Verified after the fixes: `npm test` 40 files / 444 tests green (75 → 92 in this file); 7 mutations re-run, all killed.

resources/assets/js/custom/models/enemy.state.test.js
```diff
@@ -0,0 +1,1069 @@
...
+ */
+function makeNpc({id = 1, enemyForces = 10, enemyForcesTeeming = null, classificationId = 1, npcTypeId = 1} = {}) {
+    return {
+        id,
+        name: 'Murkbrine Shorerunner',
+        classification_id: classificationId,
+        npc_type_id: npcTypeId,
+        enemy_forces: {enemy_forces: enemyForces, enemy_forces_teeming: enemyForcesTeeming},
+    };
+}
```
Wotuu, line 341: There's about 300 lines of setup here, and some of the functions I've already seen before in other tests. It's up to you, since I think you manage these tests more than I do. I only really read the test names and see if the test makes sense and if it adds to the coverage. I don't read the test body per se, but is it an idea to move some of this test setup into shared files? Maybe it's not. It's up to you, but that's just a thought I had.

resources/assets/js/custom/models/enemy.state.test.js
```diff
@@ -0,0 +1,1069 @@
...
+        expect(makeEnemy({npc: null}).getEnemyForces()).toBe(0);
+    });
+
+    test('getEnemyForces_givenAnNpcWithoutEnemyForces_returnsZero', () => {
+        setFakeState();
+
+        expect(makeEnemy({npc: {id: 1, enemy_forces: null}}).getEnemyForces()).toBe(0);
+    });
+
+    test('getEnemyForces_givenAPlainEnemy_returnsTheNpcForces', () => {
```
Wotuu: What does a plain enemy mean? Do you just mean a regular enemy that has an NPC assigned? A plain enemy is not a term that's used anywhere.

resources/assets/js/custom/models/enemy.state.test.js
Wotuu: :robot: Looked into it. The ~300 lines are the "global-script model recipe" (documented at the top of killzone.test.js): stub the globals `enemy.js` touches at load time, then hand the class fake collaborators.

I checked `enemy.test.js`, `killzone.test.js`, `mapobject.test.js` and `enemyforcescheckpoint.test.js` for real overlap rather than just similarly-named helpers:
- `makeNpc()` — unique to this file, no duplicate anywhere.
- `makeFakeLayer()` / `makeFakeMap()` / `setFakeState()` — same names reused across files, but each is shaped for that suite's specific regression (e.g. `enemy.test.js`'s layer models the exact `getTooltip()` undefined-vs-null asymmetry that #3670 is about; this file's tracks popup bind/unbind counts instead). Unifying them would mean one union-of-everything fake, which is worse than several small, purpose-built ones.
- The genuinely near-identical bit is the `global.L` Leaflet stub (`divIcon`/`Marker.extend`/`Draw.Marker.extend`/`Draw.Feature.prototype.initialize`), duplicated ~8-10 lines across three files, plus a couple of 1-2 line globals (`MapState`/`EditMapState`, `$.extend`).

Decided not to extract: it is only ~15-25 total duplicated lines, `resources/assets/js/test/setup.js` already exists as "the single place to add a new global stub" with an explicit "keep minimal, override locally" charter, and the files that would need touching (enemy.test.js, killzone.test.js, mapobject.test.js) are all already merged from earlier PRs, so pulling them into this one is a bigger footprint than the line count justifies. Happy to revisit if it starts adding up across more test files.

resources/assets/js/custom/models/enemy.state.test.js
Wotuu: :robot: Yes — an enemy backed by an npc with enemy forces set, no `enemy_forces_override`/`enemy_forces_override_teeming`, and no seasonal type: the baseline case for `getEnemyForces()`. Renamed to `getEnemyForces_givenAnNpcWithEnemyForces_returnsTheNpcForces`, which pairs with the neighbouring `givenAnNpcWithoutEnemyForces_returnsZero` test right above it.

"Plain enemy" turned out to be doing double duty in this file for a second, unrelated concept - a base `Enemy` as opposed to an `AdminEnemy` (`Enemy#isEditable()` always returns `false`; `AdminEnemy#isEditable()` returns `!this.is_mdt`). Reworded those two comments (lines 118 and 925) to spell that out instead.

Also found the same term in two already-merged sibling files with yet another meaning (an enemy with no special classification - not prideful/awakened/linked to the last boss), so swept those too: `enemy.test.js` and `killzone.test.js`.

PR #3849, opened 2026-08-05
#3674 Cover StateManager's state transitions and subscriber notification with Vitest tests

Wotuu:
:robot: Part of #3674 (target **C** of the suggested split: `statemanager.js`), alongside #3847 (A) and #3848 (B).

With this one every target in #3674 has an MR, so **merging all three closes the issue** — but none of them claims `Closes` on its own.

## What this does

Adds `resources/assets/js/custom/statemanager.test.js` — coverage over the global state every map
page reads — and, following review, **fixes the bugs that coverage turned up** instead of pinning
them.

### Tests

- **map context** — `setMapContext()` builds the right class per `MAP_CONTEXT_TYPE_*` (all five)
  and leaves the previous context in place on an unknown type; `isMapAdmin()` is true only for the
  mapping editor.
- **floor state** — `setFloorId()` notification payloads, `getCurrentFloor()` resolving through
  `Number()` on both sides, and the new visible-floor guard.
- **zoom** — `setMapZoomLevel()` reports the previous level and stays silent when unchanged
  (`setFloorId()` deliberately has no such guard; that asymmetry is pinned).
- **cookie-backed display settings** — enemy display type, both opacities, both borders, both
  heatmap toggles, facade style and number style, including the unset-cookie defaults.
- the rest of the surface: `isCurrentDungeonFacadeEnabled()`, focused enemy, MDT mapping mode,
  user roles / patreon benefits, snackbar id bookkeeping, `updateKillZones()` flattening, pull
  gradient parsing, and echo state.

### Fixes

1. **`getPullGradientHandlers()` threw on a malformed gradient.** It dereferenced `values[1]`
   without checking it existed, so a `pull_gradient` with a trailing comma — or any handler with no
   space in it — threw a `TypeError`. `pull_gradient` is persisted with **no format validation at
   all** (`storePullGradient()` takes a bare `Request`; the other path is only
   `nullable|string|max:2000`), so those strings really can be stored. Reproduced against a live
   route holding `0% #ff0000,`. Unusable handlers are now skipped, and the defaults are used only
   when *nothing* parses — `pickHexFromHandlers()` indexes `handlers[0]` unconditionally, so an
   empty list is its own `TypeError`.

   Blast radius is the route's **own editor**, not its viewers: the three callers are the pull
   settings tab's init, `createNewPull()`, and the sidebar drag-reorder — all edit-mode only.

2. **Both opacity getters returned `NaN` on a missing cookie.** `Number(undefined)` is `NaN`, and
   `opacity: NaN%` is dropped by the browser, rendering unkilled enemies fully opaque. Given the
   same `isNaN()` fallback `getKillZonePathWeight()` already has (50 / 80, matching what `map.js`
   writes the cookies with). Same class as the `getMapZoomSpeed()` `NaN` fixed earlier in the branch.

3. **`setFloorId()` accepted any floor id** (@Wotuu's review). `getCurrentFloor()` answers `false`
   for a floor it cannot resolve, and callers read `.id`/`.index` straight off that — so `false.id`
   is `undefined` rather than a throw, and that `undefined` reached the tile URL's `{index}` segment
   and `floor_id` on ajax payloads. It now refuses a floor not visible in the current map context
   and logs why, sharing `getCurrentFloor()`'s lookup via a new `isVisibleFloorId()` so the string
   ids the floor `<select>` produces are still accepted.

### Also fixed here: an unrelated CI flake

`AjaxKillZoneControllerTest::store_givenAFacadeLocationWhileTheMapFacadeStyleSaysSplitFloors_persistsTheLocationOnItsRealFloor`
(added in master's `1788bef1b`) took the first non-facade enemy with a `lat` and *asserted* it lived
inside a floor union area. Union areas cover only part of a floor, so that is an assumption, and
`findDungeon()` shuffles — the test failed for whichever facade dungeon it landed on.
`lowerkarazhan` is the one of 53 where it bites: 406 of its 408 candidate enemies convert onto the
facade floor, but the first two do not.

The conversion is now a requirement on the fixture (a `resolve()` closure) rather than an assertion
after the fact. Verified across all 53 facade dungeons: every one resolves a usable enemy, none is
rejected. Not related to this branch — fixed here rather than deferred, since it was red on this PR.

## Notes for the reviewer

- **Signalable is not stubbed.** Subscriber notification *is* the subject of this file, so the
  tests register real listeners on the real event bus and assert on the delivered
  `{name, context, data}` payloads. That is why `signalable.js` gets the guarded UMD export footer
  too — the same no-op already shipping in `killzone.js` / `enemy.js`.
- The cookie jar is a real read-back store rather than the shared setup's write-less stub, because
  several getters read straight back out of `Cookies`.
- **`updateKillZones()` appears to be dead code** — `grep -rn "updateKillZones" resources/ app/`
  finds only its definition, and the `this.killZones` it writes is never read. I've left it (and
  its tests) alone rather than deleting a public method unilaterally, but it's a candidate for
  removal if you agree. Its tests can only ever pin the shape it invents, since nothing calls it.

## Verification

- `npm test` — 46 files / 627 tests green. `composer run fix` (0 files) and `composer run analyse`
  (no errors) both clean.
- **Mutation-checked**: dropping the `values.length` guard fails 4 tests; removing the empty-result
  fallback fails 4; reinstating the over-eager `< 2` threshold the cold reviewer caught fails 2;
  dropping both opacity `isNaN` fallbacks fails 2; dropping the `setFloorId` guard fails 2. Earlier
  set still killed: both `Number()` floor coercions, `isMapAdmin()` widening, the `isNaN` guard in
  `getKillZonePathWeight()`, the zoom changed-check, and the snackbar existence check.
- **Headless Chrome against a real route page** (`/route/halls-of-atonement/...`): map renders, no
  page errors, initial floor accepted on load, string `'78'` from the DOM accepted (resolves to
  floor index 2), `999` refused with the console error and the floor left unchanged. With fix 1
  reverted, a route holding `pull_gradient = '0% #ff0000,'` made `getPullGradientHandlers()` throw a
  real `TypeError`; with it, parsing succeeds.
- No rendered output changed, so no before/after screenshots. The map screenshot posted as a comment
  is evidence the floor guard does not break page load, not a visual diff.

## Follow-ups filed

- #3946 — move focused-enemy state off `StateManager` onto `EnemyMapObjectGroup` (as requested).
- #3947 — use `enemyDisplayType` consistently instead of `enemyVisualType`/`enemy_visual_type`.

### DIFF
--- a/resources/assets/js/custom/statemanager.js
+++ b/resources/assets/js/custom/statemanager.js
@@ -814,3 +814,11 @@ class StateManager extends Signalable {
         });
     }
 }
+
+// Guarded export for the test runner (Vitest). This is a no-op in the browser,
+// where `module` is undefined, so it does not affect the concatenated bundle.
+if (typeof module !== 'undefined' && module.exports) {
+    module.exports = {
+        StateManager,
+    };
+}
--- a/resources/assets/js/custom/statemanager.js
+++ b/resources/assets/js/custom/statemanager.js
@@ -316,7 +316,12 @@ class StateManager extends Signalable {
      */
     getUnkilledEnemyOpacity() {
         console.assert(this instanceof StateManager, 'this is not a StateManager', this);
-        return Number(Cookies.get('map_unkilled_enemy_opacity'));
+        let opacity = Number(Cookies.get('map_unkilled_enemy_opacity'));
+
+        // Same hazard as getKillZonePathWeight(): the cookie may be missing on a cookie-less
+        // headless render, and `opacity: NaN%` is simply dropped by the browser - rendering
+        // unkilled enemies fully opaque. Fall back to the default map.js first sets it to.
+        return isNaN(opacity) ? 50 : opacity;
     }
 
     /**
@@ -337,7 +342,10 @@ class StateManager extends Signalable {
      */
     getUnkilledImportantEnemyOpacity() {
         console.assert(this instanceof StateManager, 'this is not a StateManager', this);
-        return Number(Cookies.get('map_unkilled_important_enemy_opacity'));
+        let opacity = Number(Cookies.get('map_unkilled_important_enemy_opacity'));
+
+        // See getUnkilledEnemyOpacity() - same missing-cookie fallback, different default.
+        return isNaN(opacity) ? 80 : opacity;
     }
 
     /**
@@ -401,16 +409,18 @@ class StateManager extends Signalable {
     getCurrentFloor() {
         console.assert(this instanceof StateManager, 'this is not a StateManager', this);
 
-        if (this._visibleFloorsByFloorId === null) {
-            this._visibleFloorsByFloorId = new Map();
+        return this._getVisibleFloorsByFloorId().get(Number(this._floorId)) ?? false;
+    }
 
-            let floors = this._mapContext.getVisibleFloors();
-            for (let i = 0; i < floors.length; i++) {
-                this._visibleFloorsByFloorId.set(Number(floors[i].id), floors[i]);
-            }
-        }
+    /**
+     * Checks if the given floor is one of the floors visible in the current map context.
+     * @param {Number|String} floorId
+     * @returns {boolean}
+     */
+    isVisibleFloorId(floorId) {
+        console.assert(this instanceof StateManager, 'this is not a StateManager', this);
 
-        return this._visibleFloorsByFloorId?.get(Number(this._floorId)) ?? false;
+        return this._getVisibleFloorsByFloorId().has(Number(floorId));
     }
 
     /**
@@ -421,10 +431,41 @@ class StateManager extends Signalable {
      */
     setFloorId(floorId, center = null, zoom = null) {
         console.assert(this instanceof StateManager, 'this is not a StateManager', this);
+
+        // Refuse a floor that isn't visible in this map context. getCurrentFloor() answers `false`
+        // for an unknown floor, and every consumer reads `.id`/`.index` straight off that result -
+        // which is `undefined` rather than a throw, so a bad floor id would otherwise propagate
+        // silently into tile URLs and ajax payloads. Keeping the floor we're on is the safer of the
+        // two bad options, and the error says why. Skipped while the map context is still unset,
+        // so this cannot fire before statemanager.blade.php has installed one.
+        if (this._mapContext !== null && !this.isVisibleFloorId(floorId)) {
+            console.error(`Refusing to set floor id '${floorId}'; it is not visible in the current map context`);
+
+            return;
+        }
+
         this._floorId = floorId;
 
         // Let everyone know it's changed
         this.signal('floorid:changed', {floorId: this._floorId, center: center, zoom: zoom});
+    }
+
+    /**
+     * Builds - once - a lookup of the map context's visible floors, keyed by their numeric id.
+     * @returns {Map<Number, Object>}
+     * @private
+     */
+    _getVisibleFloorsByFloorId() {
+        if (this._visibleFloorsByFloorId === null) {
+            this._visibleFloorsByFloorId = new Map();
+
+            let floors = this._mapContext.getVisibleFloors();
+            for (let i = 0; i < floors.length; i++) {
+                this._visibleFloorsByFloorId.set(Number(floors[i].id), floors[i]);
+            }
+        }
+
+        return this._visibleFloorsByFloorId;
     }
 
     /**
@@ -668,15 +709,23 @@ class StateManager extends Signalable {
                 if (handlers.hasOwnProperty(index)) {
                     let handler = handlers[index];
                     let values = handler.trim().split(' ');
-                    // Only RGB values
-                    if (values[1].indexOf('#') === 0) {
+                    // Only RGB values. The color may be missing entirely - a trailing comma, or a
+                    // handler with no space in it, leaves values[1] undefined - so check for it
+                    // before dereferencing rather than throwing on the way to drawing the map.
+                    if (values.length > 1 && values[1].indexOf('#') === 0) {
                         result.push([parseInt(('' + values[0]).replace('%', '')), values[1]]);
                     } else {
                         console.warn('Invalid handler found:', handler);
                     }
                 }
             }
-        } else {
+        }
+
+        // pull_gradient is persisted without any format validation, so a route may well carry a
+        // gradient that yields fewer than the two handlers pickHexFromHandlers() requires (it
+        // indexes handlers[0] and handlers[length - 1] unconditionally). Fall back to the defaults
+        // rather than handing it a list it cannot work with.
+        if (result.length < 2) {
             result = c.map.editsidebar.pullGradient.defaultHandlers;
         }
 
--- a/resources/assets/js/custom/statemanager.js
+++ b/resources/assets/js/custom/statemanager.js
@@ -434,7 +434,11 @@ class StateManager extends Signalable {
     getMapZoomSpeed() {
         console.assert(this instanceof StateManager, 'this is not a StateManager', this);
 
-        return parseInt(Cookies.get('map_zoom_speed'));
+        let zoomSpeed = parseInt(Cookies.get('map_zoom_speed'));
+
+        // The cookie may be missing; fall back to the same default used when it's first set
+        // (see map.js/mapsettings.blade.php) rather than returning NaN.
+        return isNaN(zoomSpeed) ? 50 : zoomSpeed;
     }
 
     /**
--- a/resources/assets/js/custom/statemanager.js
+++ b/resources/assets/js/custom/statemanager.js
@@ -722,10 +722,13 @@ class StateManager extends Signalable {
         }
 
         // pull_gradient is persisted without any format validation, so a route may well carry a
-        // gradient that yields fewer than the two handlers pickHexFromHandlers() requires (it
-        // indexes handlers[0] and handlers[length - 1] unconditionally). Fall back to the defaults
-        // rather than handing it a list it cannot work with.
-        if (result.length < 2) {
+        // gradient from which nothing usable survives. pickHexFromHandlers() indexes handlers[0]
+        // unconditionally, so an empty list is its own TypeError - fall back to the defaults.
+        // A single handler is deliberately kept: it is reachable (grapick lets you remove stops
+        // down to one) and works, since handlers[0] === handlers[length - 1] makes every weight
+        // resolve to that one color. Replacing it would also feed the defaults back into grapick
+        // on SettingsTabPull.activate(), overwriting the user's saved gradient on the next edit.
+        if (result.length < 1) {
             result = c.map.editsidebar.pullGradient.defaultHandlers;
         }
 
--- a/resources/assets/js/custom/statemanager.js
+++ b/resources/assets/js/custom/statemanager.js
@@ -312,11 +312,11 @@ class StateManager extends Signalable {
 
     /**
      * Get the opacity at which unkilled enemies should be rendered at.
-     * @returns {string}
+     * @returns {Number}
      */
     getUnkilledEnemyOpacity() {
         console.assert(this instanceof StateManager, 'this is not a StateManager', this);
-        return Cookies.get('map_unkilled_enemy_opacity');
+        return Number(Cookies.get('map_unkilled_enemy_opacity'));
     }
 
     /**
@@ -333,11 +333,11 @@ class StateManager extends Signalable {
 
     /**
      * Get the opacity at which unkilled important enemies should be rendered at.
-     * @returns {string}
+     * @returns {Number}
      */
     getUnkilledImportantEnemyOpacity() {
         console.assert(this instanceof StateManager, 'this is not a StateManager', this);
-        return Cookies.get('map_unkilled_important_enemy_opacity');
+        return Number(Cookies.get('map_unkilled_important_enemy_opacity'));
     }
 
     /**
--- a/resources/assets/js/custom/statemanager.test.js
+++ b/resources/assets/js/custom/statemanager.test.js
@@ -29,18 +29,33 @@ global.cookieDefaultAttributes = undefined;
 // 1c. Map context classes. Only their identity matters to StateManager: setMapContext() picks one
 // by `type` and isMapAdmin() answers `instanceof MapContextMappingVersionEdit`. Each records the
 // raw context it was constructed with so a test can prove the payload was handed through.
-global.MapContextDungeonRoute = class MapContextDungeonRoute {
+// The `extends` chain mirrors the real one (custom/mapcontext/*.js) exactly:
+//
+//   MapContext
+//   |- MapContextDungeonRoute -> MapContextLiveSession
+//   `- MapContextMappingVersion
+//      |- MapContextDungeonExplore -> MapContextDungeonRouteSearch
+//      `- MapContextMappingVersionEdit
+//
+// isMapAdmin() would answer identically under a flatter stub, but only this shape can tell a
+// widening to `instanceof MapContextMappingVersion` (which would wrongly admit explore/search)
+// apart from a widening to `instanceof MapContextDungeonRoute` (which would not).
+global.MapContext = class MapContext {
     constructor(mapContext) {
         this.mapContext = mapContext;
     }
 };
+global.MapContextDungeonRoute = class MapContextDungeonRoute extends global.MapContext {
+};
 global.MapContextLiveSession = class MapContextLiveSession extends global.MapContextDungeonRoute {
 };
-global.MapContextMappingVersionEdit = class MapContextMappingVersionEdit extends global.MapContextDungeonRoute {
+global.MapContextMappingVersion = class MapContextMappingVersion extends global.MapContext {
 };
-global.MapContextDungeonExplore = class MapContextDungeonExplore extends global.MapContextDungeonRoute {
+global.MapContextMappingVersionEdit = class MapContextMappingVersionEdit extends global.MapContextMappingVersion {
 };
-global.MapContextDungeonRouteSearch = class MapContextDungeonRouteSearch extends global.MapContextDungeonRoute {
+global.MapContextDungeonExplore = class MapContextDungeonExplore extends global.MapContextMappingVersion {
+};
+global.MapContextDungeonRouteSearch = class MapContextDungeonRouteSearch extends global.MapContextDungeonExplore {
 };
 
 // 1d. A cookie jar with real read-back, unlike the shared setup's write-less stub. The settings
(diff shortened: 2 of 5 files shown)


Wotuu:
:robot: Cold review: 4 findings

Dispatched to a fresh-context reviewer (no shared context with the implementing session). All four are addressed in 046f2f529:

1. **The `Number()` floor-id coercion was not actually pinned** — the test used a string id on *both* sides, so dropping both coercions kept all 55 tests green (I reproduced this before fixing). A string id in `visibleFloors` also cannot occur: `MapContextBase::toArray()` sends integers. The test now models the real asymmetry — integer ids from the server, a string id from the floor `<select>` — which kills the both-sided mutation as well as either one-sided one.
2. **The stale-lookup comment cited a mechanism that cannot happen** — it named a facade style change as the in-page context swap to worry about, but that path always ends in `window.location.reload()`. Reworded: nothing swaps `_mapContext` in-page today (`setMapContext()` has one caller, once per page load), so the pin is aimed at the first change that does.
3. **The stubbed MapContext hierarchy was inverted** — all five descended from `MapContextDungeonRoute`. `isMapAdmin()` answered identically, but the stub could not express the genuinely dangerous widening (`instanceof MapContextMappingVersion`, which would wrongly admit the explore/search contexts). Chain now mirrors the real one, and that mutation is part of the re-run set. The old body's "fails 3" claim was an artifact of the flat stub and has been corrected.
4. **Coverage gap** — `getKillZonePathWeight()` is the one cookie getter with a real `isNaN` guard (and a documented cookie-less-headless-render incident behind it) and was untested. Added, with its setter and `getMapZoomSpeed()` — which has the same `parseInt` shape *without* the guard, pinned as current behaviour rather than endorsed.

Verified after the fixes: `npm test` 40 files / 411 tests green (55 → 59 in this file); 5 mutations re-run, all killed.

Wotuu:

resources/assets/js/custom/statemanager.test.js
```diff
@@ -0,0 +1,653 @@
...
+        expect(stateManager.getCurrentFloor()).toBe(floor);
+    });
+
+    test('getCurrentFloor_givenAFloorThatIsNotVisible_returnsFalse', () => {
+        const stateManager = makeStateManager(makeFakeMapContext({visibleFloors: [{id: 1}]}));
+
+        stateManager.setFloorId(99);
+
+        expect(stateManager.getCurrentFloor()).toBe(false);
+    });
```
Wotuu: If an invalid floor is set, should it just.. not do that? Right now it seems like this will break the state?

resources/assets/js/custom/statemanager.test.js
```diff
@@ -0,0 +1,653 @@
...
+        const stateManager = makeStateManager(makeFakeMapContext({visibleFloors: [{id: 1}]}));
+        stateManager.setFloorId(1);
+        expect(stateManager.getCurrentFloor()).toEqual({id: 1});
+
+        stateManager._mapContext = makeFakeMapContext({visibleFloors: [{id: 1, name: 'Renamed'}, {id: 5}]});
+        stateManager.setFloorId(5);
+
+        expect(stateManager.getCurrentFloor()).toBe(false);
+    });
+});
```
Wotuu: FYI a map context is set once, and then NEVER changed again during the course of the lifetime of the page. It decides what "Version" of the map is loaded. That version never changes, only when you load a different page entirely.

resources/assets/js/custom/statemanager.test.js
```diff
@@ -0,0 +1,653 @@
...
+describe('StateManager cookie-backed display settings', () => {
+    test('setEnemyDisplayType_givenAType_storesItAndNotifies', () => {
+        const stateManager = makeStateManager();
+        const received = listenFor(stateManager, 'enemydisplaytype:changed');
+
+        stateManager.setEnemyDisplayType('aggressiveness');
+
+        expect(stateManager.getEnemyDisplayType()).toBe('aggressiveness');
+        expect(Cookies.get('enemy_display_type')).toBe('aggressiveness');
+        expect(received[0].data).toEqual({enemyDisplayType: 'aggressiveness'});
```
Wotuu: I heavily dislike these magic strings. Please extract them into `constants.js` and label them `DISPLAY_TYPE_<something>`. I also see it mixes up `enemyDisplayType` with `enemyVisualType` (in map.blade.php for example).. I think `enemyDisplayType` should be consistently used.

resources/assets/js/custom/statemanager.test.js
```diff
@@ -0,0 +1,653 @@
...
+    });
+
+    test('setUnkilledImportantEnemyOpacity_givenAnOpacity_storesItAndNotifies', () => {
+        const stateManager = makeStateManager();
+        const received = listenFor(stateManager, 'unkilledimportantenemyopacity:changed');
+
+        stateManager.setUnkilledImportantEnemyOpacity(0.8);
+
+        expect(stateManager.getUnkilledImportantEnemyOpacity()).toBe('0.8');
+        expect(received[0].data).toEqual({opacity: 0.8});
```
Wotuu: Should these values not be a Number instead of a string? `getUnkilledEnemyOpacity()` returns the raw cookie value but it should be cast to a Number I think.

resources/assets/js/custom/statemanager.test.js
```diff
@@ -0,0 +1,653 @@
...
+        stateManager.setMapZoomSpeed(3);
+
+        expect(stateManager.getMapZoomSpeed()).toBe(3);
+    });
+
+    test('getMapZoomSpeed_givenNoStoredValue_returnsNaN', () => {
+        // Unlike getKillZonePathWeight() this one has no isNaN() guard, so a missing cookie yields
+        // NaN rather than a usable default. Pinned as the current behaviour, not endorsed.
+        expect(makeStateManager().getMapZoomSpeed()).toBeNaN();
+    });
```
Wotuu: Ok so this sounds like a bug you adjusted the test to, instead of fixing the bug.

resources/assets/js/custom/statemanager.test.js
```diff
@@ -0,0 +1,653 @@
...
+    test('setFocusedEnemy_givenNull_notifiesWithNull', () => {
+        const stateManager = makeStateManager();
+        stateManager.setFocusedEnemy({id: 5});
+        const received = listenFor(stateManager, 'focusedenemy:changed');
+
+        stateManager.setFocusedEnemy(null);
+
+        expect(received[0].data).toEqual({focusedenemy: null});
+    });
+});
```
Wotuu, line 493: I think this function `setFocusedEnemy` should be moved to the EnemyMapObjectGroup. It's too global for this state. All listeners should listen to the new area instead.

Wotuu: Create a follow-up issue for this work.

resources/assets/js/custom/statemanager.test.js
Wotuu: :robot: This is existing production behaviour, not something introduced by this PR — `setFloorId()` never validated its input, and `getCurrentFloor()` already gracefully falls back to `false` when the given floor ID is not in the visible-floors map, rather than throwing or corrupting state. The rest of the codebase (e.g. `dungeonmap.js`) already handles a falsy `getCurrentFloor()` result. Since nothing calls `setFloorId()` with an untrusted/invalid value today (it comes from the floor `<select>` or the map facade switch, both of which only offer valid floor IDs), I left this as a pinning test documenting the current graceful-degradation behaviour rather than adding new validation for a scenario that cannot currently occur — happy to add a guard if you want it hardened regardless.

resources/assets/js/custom/statemanager.test.js
Wotuu: :robot: Thanks for confirming — that matches exactly what this test's own inline comment already documents (line 251-256: "Nothing swaps `_mapContext` on a live StateManager today... So this is latent, not live"). The test is deliberately pinning a scenario that cannot currently happen through any real code path (no in-place context swap exists), so that if someone *does* introduce one in the future, this test fails and points them here instead of the stale floor list silently causing a confusing bug. No change needed unless you'd rather this test be removed since the scenario is unreachable — happy to drop it if you prefer not to carry a test for dead code paths.

resources/assets/js/custom/statemanager.test.js
Wotuu: :robot: Fixed — added `DISPLAY_TYPE_ENEMY_PORTRAIT`/`DISPLAY_TYPE_NPC_CLASS`/`DISPLAY_TYPE_NPC_TYPE`/`DISPLAY_TYPE_ENEMY_FORCES`/`DISPLAY_TYPE_ENEMY_GROUP`/`DISPLAY_TYPE_ENEMY_SKIPPABLE` to `constants.js` (matching the real option set in `enemyvisualtype.blade.php`) and switched the test to use `DISPLAY_TYPE_NPC_CLASS`. Turned out the test's original `'aggressiveness'` value was not just a magic string but an outright wrong one — that name belongs to the separate aggressiveness-border feature (`setEnemyAggressivenessBorder`), not to any real display type, so it was accidentally testing a value that could never occur in practice.

On the `enemyDisplayType`/`enemyVisualType` naming mismatch: I left that as-is for now. Unifying it means renaming the PHP `$enemyVisualType` variable (and the `enemyvisualtype.blade.php` file itself) across `map.blade.php`/`enemyvisualtype.blade.php` plus any other call sites, which is a larger cross-cutting rename outside a JS-test-coverage PR and I didn't want to do blind without visual verification. Happy to open a follow-up issue for it, or do it here if you'd rather bundle it in — let me know.

resources/assets/js/custom/statemanager.test.js
Wotuu: :robot: Fixed — `getUnkilledEnemyOpacity()` and `getUnkilledImportantEnemyOpacity()` now cast the cookie value through `Number()` instead of returning the raw string, with the tests updated to assert on `0.5`/`0.8` rather than `'0.5'`/`'0.8'`. Checked the two real call sites (`enemyvisual.js`) — both only use the value inside `${...}%` template literals, which behave identically for a string or a Number, so this is a safe type-only fix with no behavior change there.

resources/assets/js/custom/statemanager.test.js
Wotuu: :robot: Agreed, and fixed — `getMapZoomSpeed()` now mirrors `getKillZonePathWeight()`'s `isNaN()` guard: a missing cookie falls back to `50` (the same default `map.js`/`mapsettings.blade.php` already use when first setting the cookie) instead of returning `NaN`. That `NaN` was feeding straight into `dungeonmap.js`'s `wheelPxPerZoomLevel` calculation, so this closes a real latent bug, not just a test-pinning exercise. Test updated to `getMapZoomSpeed_givenNoStoredValue_returnsTheDefault` asserting `50`.

resources/assets/js/custom/statemanager.test.js
Wotuu: :robot: That's a reasonable architectural direction, but I don't think it's something to do blind inside a JS-test-coverage PR — moving `setFocusedEnemy`/`getFocusedEnemy` off `StateManager` onto `EnemyMapObjectGroup` means finding every current listener of `focusedenemy:changed` and every caller of `getState().setFocusedEnemy(...)`/`getState().getFocusedEnemy()`, re-wiring them to the map object group instance instead of the global state singleton, and verifying nothing relies on it being reachable globally. That's a real behavioral/architectural change I'd want to verify visually rather than land as a side effect of adding tests. I'll leave this test coverage in place for the current (global) design and would suggest filing it as its own follow-up issue if you'd like it tackled — let me know if you'd rather I scope and do it here instead.

Wotuu: So create a follow-up as I asked please.

resources/assets/js/custom/statemanager.test.js
Wotuu: :robot: You were right, and I've hardened it in cc918088a rather than leaving the pin.

`setFloorId()` now refuses a floor that isn't visible in the current map context, logs why, and
leaves the current floor alone. My earlier reply undersold the risk: `getCurrentFloor()` returning
`false` doesn't throw, so `getCurrentFloor().id` silently evaluates to `undefined` — which then
flows into the tile URL's `{index}` segment and into `floor_id` on ajax payloads. That's a wrong
value written to the database, not a loud failure, which is worse than the crash I implied it
wasn't.

The lookup `getCurrentFloor()` already built is now shared with a new `isVisibleFloorId()`, so the
guard uses exactly the same `Number()` coercion and can't reject the string ids the floor `<select>`
produces.

Verified in headless Chrome against a real route page: initial floor accepted on load, a string
`'78'` from the DOM accepted (resolves to floor index 2), `999` refused with the error and the floor
left unchanged. Also confirmed the two server-side floor resolvers (`indexOrFacade` /
`defaultOrFacade`) key off the same `User::getCurrentUserMapFacadeStyle()` + `facade_enabled` pair
that `getVisibleFloors()` does, so facade mode agrees with the guard.

resources/assets/js/custom/statemanager.test.js
Wotuu: :robot: Follow-up filed for the naming half of this: #3947 (use `enemyDisplayType` consistently
instead of `enemyVisualType` / `enemy_visual_type`). It lists the four spellings currently in play,
the call sites, and flags that the translation-key rename under
`view_common.maps.controls.elements.enemyvisualtype.*` needs visual verification — which is why I'd
rather not bundle it in here.

The `DISPLAY_TYPE_*` constants added on this PR already use the target naming, so that issue closes
the gap rather than widening it.

resources/assets/js/custom/statemanager.test.js
Wotuu: :robot: Follow-up on this one — the `Number()` cast I added was half a fix.

`Number(undefined)` is `NaN`, so with the cookie missing these getters went from returning
`undefined` to returning `NaN`. Both interpolate into `` `opacity: ${x}%` ``, which the browser
drops either way, rendering unkilled enemies fully opaque. cc918088a gives both getters the same
`isNaN()` fallback `getKillZonePathWeight()` has, defaulting to 50 / 80 — the same values `map.js`
writes the cookies with when it first sets them.

That's the same class as the `getMapZoomSpeed()` NaN you caught below; these two were the remaining
unguarded ones.

resources/assets/js/custom/statemanager.test.js
Wotuu: :robot: Follow-up issue created as asked: #3946.

It scopes the whole surface — `_focusedEnemy` plus `getFocusedEnemy()`/`setFocusedEnemy()` on
`StateManager`, the four call sites (`enemyvisual.js` ×2, `enemyvisualmanager.js` ×2) and the single
`focusedenemy:changed` listener in `models/enemypatrol.js` — notes that the `EnemyPatrol` constructor
registers the listener so `MapObjectGroupManager` ordering needs checking, and flags that the move
needs visual verification of patrol highlighting on hover.

Wotuu:
:robot: Evidence for fix 3 (the `setFloorId()` visible-floor guard), which is the one change that runs on every map page load: a real route page still renders normally with it in place — enemies, pulls, kill zone paths and enemy forces all intact, no page errors.

route map with the floor guard in place

Nothing here is *supposed* to look different from master; the screenshot is proof the guard does not refuse the initial floor set, not a visual diff.

resources/assets/js/custom/statemanager.js
```diff
@@ -664,15 +709,23 @@ class StateManager extends Signalable {
...
                 }
             }
-        } else {
+        }
+
+        // pull_gradient is persisted without any format validation, so a route may well carry a
+        // gradient that yields fewer than the two handlers pickHexFromHandlers() requires (it
+        // indexes handlers[0] and handlers[length - 1] unconditionally). Fall back to the defaults
+        // rather than handing it a list it cannot work with.
+        if (result.length < 2) {
```
Wotuu: :robot: The `< 2` half of this fallback throws away a configuration that works today.

The stated justification — "a single-handler gradient, which `pickHexFromHandlers()` asserts on" (commit message, repeated in the test comment "handing it a single handler is its own crash") — doesn't hold:

- `console.assert(handlers.length > 1, ...)` in `colorutil.js:126` **logs**, it does not throw.
- With one handler, `handlers[0]` and `handlers[handlers.length - 1]` are the same element, so both early-return branches hit and `pickHexFromHandlers()` returns that colour for every weight. A single-handler gradient is simply a solid colour, and it renders correctly.

And it is user-reachable: grapick's per-handler `data-toggle="handler-close"` × calls `remove()` with no minimum, and `settingstabpull.js` wires `handler:remove` → `onChange` → `setPullGradient()` → `_savePullGradientSettings()`. So `pull_gradient = '50 #ff0000'` really does get persisted by a user who deletes one of the two default stops.

Consequence of the change, beyond a wrong colour:

1. `applyPullGradient()` now colours the pulls red→green instead of that user's solid colour.
2. Worse, `SettingsTabPull.activate()` restores grapick from `getPullGradientHandlers()`. The user opens the pull tab, sees the two *default* handlers, touches anything, and `onChange` serialises grapick's current state back over their saved gradient — their configuration is gone, silently.

`result.length < 1` (i.e. `=== 0`) keeps the crash fix intact — that is the genuine hazard, since the old `else` only fired for an empty string, so an all-invalid gradient returned `[]` and `pickHexFromHandlers([], w)` did throw on `handlers[0][0]` — while leaving the single-handler case working.

Also worth a `console.warn` on the fallback path: a route that silently reverts to the defaults is currently indistinguishable from one that never had a gradient.

Wotuu:

resources/assets/js/custom/statemanager.js
Wotuu: :robot: Fixed in 2bb62b763 — you were right on both counts, and I'd checked neither.

`console.assert()` logs and returns; it does not throw. And with a single handler
`handlers[0] === handlers[length - 1]`, so the first branch of `pickHexFromHandlers()` returns that
color for every weight. Confirmed both by running the function standalone: one handler resolves to
its color at weights 0/50/100, and only the empty list throws (`handlers[0][0]` on `undefined`).

The `SettingsTabPull.activate()` consequence is the one that makes this worth catching rather than
merely wrong — the user would have been shown the two default stops in grapick and the next
interaction would have serialised those back over their own gradient. Silent data loss on saved
user data is strictly worse than the crash I was fixing.

Threshold is now `< 1`, which still covers the original defect: an all-invalid gradient parses to
`[]` and falls back. Tests updated — `'0% #ff0000,'` (trailing comma) now asserts the one handler
that parsed is *kept*, and `'50% #ff0000'` asserts a single-stop gradient survives untouched.
Mutation-checked both directions: `< 0` (never fall back) fails 4 tests, and reinstating `< 2`
fails 2 — so the regression you found is now pinned.

I did not add the extra `console.warn` on the fallback path: each unusable handler already warns
individually inside the loop, so an all-invalid gradient warns once per segment before falling back,
and an empty `pull_gradient` is the ordinary "no gradient set" case that shouldn't warn at all.

PR #3854, opened 2026-08-05
#3853 Anchor the Playwire side rails to the viewport edges

### LINKED ISSUE #3853: Left/right rail ads are positioned too far inward and overlap page content
## Problem

The Playwire left and right rail ad units (`#pw-oop-left_rail` / `#pw-oop-right_rail`) are not
anchored to the edges of the viewport. They sit too far towards the middle of the screen and end
up overlapping the page content underneath them, making the content unreadable/unclickable in
that band.

## Suggested fix (from Wotuu's contact at Playwire)

A CSS override to force the rails back to the viewport edges:

```css
#pw-oop-right_rail {
  right: 0 !important;
}

#pw-oop-left_rail {
  left: 0 !important;
}
```

This is being relayed to Playwire as well, but we want the override in our own CSS in the
meantime so the site is usable now.

## Where it goes

Custom CSS lives under `resources/assets/css`, concatenated into `public/css/custom-{version}.css`
in the order `base/` -> `lib/` -> `sections/` (see `scripts/build/custom-styles-order.mjs`). Each
subfolder is a glob, so a new file is picked up automatically without touching the build config.

Suggested: a new `resources/assets/css/sections/ads.css` holding the two rules. `sections/` is
last in the cascade order, so the override lands after the base/library rules. The rails are
injected by Playwire's ramp script (`resources/views/common/thirdparty/playwire/`) and are not
markup we control, hence the `!important`.

## Notes / gotchas

- Both rails are ad-network markup, so nothing in the repo references those ids today; they only
  exist at runtime once ramp injects them.
- Playwire only loads when the ad config is present, so this is **not verifiable on a local dev
  stack** - it needs staging/production (or a page where the ramp script actually loads) to
  confirm the rails snap to the edges.
- **This is a compiled asset change (CSS), so it cannot be hotfixed** - the fix ships in a new
  release. See the `hotfix` skill for why.
- If Playwire fixes it upstream, the override should be harmless (it just re-asserts the intended
  position), but worth revisiting later so we do not accumulate dead `!important` rules.

## Definition of done

- New CSS rules present and included in the built `public/css/custom-{version}.css`.
- `composer run fix` / `composer run analyse` clean (no PHP touched, but run them anyway).
- Verified on staging with the rails rendering flush against the viewport edges and no overlap
  with page content, at both a wide and a narrow-desktop viewport width. Before/after screenshots
  on the PR.

Wotuu:
:robot: Closes #3853

## What

The Playwire left/right rail ad units (`#pw-oop-left_rail` / `#pw-oop-right_rail`) sat too far
towards the middle of the viewport and overlapped the page content underneath them. This forces
them back to the viewport edges using the override Playwire suggested:

```css
#pw-oop-left_rail  { left: 0 !important; }
#pw-oop-right_rail { right: 0 !important; }
```

The rails are injected at runtime by the ramp script and are not markup we control, hence
`!important`. This has been relayed to Playwire as well - if they fix it upstream these rules only
re-assert the intended position.

## Where

Added to the existing `resources/assets/css/lib/thirdparty/playwire.css`, which already holds the
bottom-rail and sticky in-content overrides, rather than starting a second home for Playwire CSS.

## Verification

The rails only exist once the ramp script runs, which it does not on a dev stack - so there is
**nothing to screenshot locally** and no way to visually confirm the ads themselves. The
checklist's visual-verification item does not apply here; positioning has to be confirmed on
staging/production after release.

What was verified mechanically instead, against the real built bundle
(`public/css/custom-<sha>.css`) served by the worktree stack, in headless Chrome: stub elements
carrying the ids and an inline `left:250px` / `right:250px` compute to `left: 0px` / `right: 0px`.
That proves the selectors match and the `!important` beats the kind of inline positioning ramp
applies - it does not prove the ads look right.

- `npm run development` - rules present in `public/css/custom-<sha>.css` (the `lib/**/*.css` glob
  in `scripts/build/custom-styles-order.mjs` picks them up).
- `npm test` - 40 files / 364 tests pass, including the css-concat ordering tests.
- `composer run fix` - 0 files changed. `composer run analyse` - no errors.

**This is a compiled-asset change, so it cannot be hotfixed** - it ships in a release.

Cold review skipped under the trivial-change rule (13-line CSS-only diff), by Wotuu's explicit
call in the implementing session. The `pr cold reviewed` label is set so `babysit-prs` does not
dispatch its own review.

### DIFF
--- a/resources/assets/css/lib/thirdparty/playwire.css
+++ b/resources/assets/css/lib/thirdparty/playwire.css
@@ -12,6 +12,19 @@
     pointer-events: auto !important;
 }
 
+/**
+ * The side rails were positioned too far towards the middle of the viewport and overlapped the
+ * page content underneath them; anchor them back to the viewport edges. Reported to Playwire as
+ * well - once they fix it upstream these two rules only re-assert the intended position.
+ */
+#pw-oop-left_rail {
+    left: 0 !important;
+}
+
+#pw-oop-right_rail {
+    right: 0 !important;
+}
+
 /** Otherwise the video will be above the blank space for the bottom rail */
 .pw-corner-ad-video {
     margin-bottom: 0 !important;


Wotuu:

PR #3855, opened 2026-08-05
#3851 Rework the site header: single in-flow stack, CSS-driven shrink, mobile route bar

### LINKED ISSUE #3851: Rework the site header: single in-flow stack, CSS-driven shrink, mobile route bar
:robot: 

## Background

The header has grown from a plain Bootstrap navbar into three separately-`position: fixed` bars — the game-version bar (+ dungeon context strip), the main navbar, and on the map view a route-specific third bar — each placed by hardcoded `top:` offsets (126/94/151/30/86px) spread across `header.css`, `map-header.css`, inline blade styles and theme variables. Symptoms:

- **1px lines / imperfect stacking**: the offsets must exactly equal the rendered height of the bar above and don't (3px shrink-delta mismatch, lux's 1px `--theme-header-border`, em-based shrink padding → fractional heights). Three disagreeing "header height" definitions exist: inline spacer 190px, `--theme-header-height: 114px` (dead), `top: 126px`.
- **Jumpy scroll-shrink**: `navbarshrink.js` binds a handler per `[data-toggle="navbar-shrink"]` (two elements); both write the same `.navbar-top-fixed-spacer` height (last-writer-wins) while unshrink restores the sum → page height collapses too far, content jumps, scroll re-fires (feedback loop, previously patched with a one-sided debounce in #3118).
- **Mobile**: the game-version bar and the map's route bar are `d-none d-lg-block` — rendered but invisible; the map view shows no route title/options at all on mobile.

## Remaining work

Rework into a single in-flow stack, keeping the current look and feel:

1. `common/layout/header.blade.php` gets a single root wrapper `<header id="site_header" class="ksg-header">`; both bars become static children stacking in normal flow. `position: sticky; top: 0` on the wrapper; delete the spacer div, all `top:` offsets, `data-toggle="navbar-shrink"` and per-bar shrink classes. (`layout.css` needs `height: 100%` → `min-height: 100%` on `#app` for sticky to work; fallback plan B is fixed + `body { padding-top: var(--ksg-header-height) }`.)
2. Map view: `#map_header` becomes the one positioned element (`position: fixed; display: flex; flex-direction: column`), the included `.ksg-header` degrades to a flow child; the `:has(> .has_dungeon_context_header)` repositioning rules die naturally.
3. Shrink rebuilt CSS-driven: one `.ksg-header--shrink` class toggled by a small vanilla script (rAF-throttled scroll, hysteresis: shrink > 140px, unshrink < 40px — separated by more than the height delta so the feedback loop is impossible by construction; never shrink on short pages). Replaces `navbarshrink.js`. A `ResizeObserver` writes `--ksg-header-height` for dependents (`.route_sidebar` tops, later the `map.css` 100vh hacks).
4. **Click-through gutters (hard requirement)**: on the map view the header floats over the map with transparent sides; every full-width layer (`#map_header`, wrapper, bars) gets `pointer-events: none`, only the centered `max-width: 1400px` `.container`s restore `pointer-events: auto` — generalizing the existing `map-header.css:115-127` trick to the new wrappers.
5. Mobile route bar: drop `d-none d-lg-block` from `.navbar-third`; below `lg` show favourite star + truncated route title + a kebab dropdown whose items target the same existing modals as the desktop buttons.

Full plan with selector audit, file-by-file steps and verification: agreed in-session (plan file `so-i-want-to-noble-creek.md`).

## Key files

- `resources/views/common/layout/header.blade.php`, `resources/views/common/maps/controls/header.blade.php`, `resources/views/common/maps/map.blade.php`, `resources/views/layouts/sitepage.blade.php`
- `resources/assets/css/sections/header.css` (rewrite), `resources/assets/css/sections/map-header.css` (trim), `resources/assets/css/base/layout.css`
- `resources/assets/sass/theme/_variables.scss`, `theme.scss`, `resources/assets/sass/custom/map.scss`
- `resources/assets/js/custom/inline/common/general/navbarshrink.js` → `siteheader.js` (+ JS test)

## Definition of done

- No hardcoded `top:` offsets or JS-written spacer heights remain; `git grep` clean for `navbar-top-fixed-spacer`, `theme-header-height`, `has_dungeon_context_header`.
- Headless-Chrome verification: zero-gap bar stacking on all three themes (lux especially), natural scroll with exactly one shrink/unshrink flip, map header click-through in both gutters (`elementFromPoint` resolves to the map pane), route title + kebab usable at 375px.
- JS hysteresis test green (`npm run test`), `composer run fix` + `composer run analyse` clean.

Wotuu:
Closes #3851

:robot:

## What changed

The header's three bars (game-version bar + dungeon context, main navbar, and the map view's route bar) were independent `position: fixed` elements placed by hardcoded `top:` offsets spread across five files — the source of the 1px stacking gaps, the jumpy scroll-shrink, and the missing mobile route bar.

- **One in-flow stack**: `common/layout/header.blade.php` now renders a single `<header id="site_header" class="ksg-header">` wrapper, `position: sticky; top: 0`; both bars are static children. All `top:` offsets, the `.navbar-top-fixed-spacer` div, the dead `--theme-header-height` theme var and the `:has()` map repositioning rules are deleted.
- **Shrink rebuilt**: `siteheader.js` replaces `navbarshrink.js`. One `.ksg-header--shrink` class toggle with scroll hysteresis (shrink > 140px, unshrink < 40px — separated by more than the height delta, so the old spacer feedback loop is impossible by construction) plus a short-page guard. The decision is a pure, vitest-covered function. A `ResizeObserver` publishes `--ksg-header-height`, which now positions the route sidebar instead of hardcoded 150px/70px.
- **Map header**: `#map_header` is one fixed flex column; the included header degrades to a flow child. Click-through gutters are a hard requirement and generalized: every full-width layer is `pointer-events: none`, only the centered `max-width: 1400px` containers catch events.
- **Mobile route bar**: `.navbar-third` is no longer desktop-only — below `lg` it shows favourite star + truncated route title + a kebab overflow menu whose items retarget the same modals the desktop buttons use (modals render once, unchanged).
- `layout.css`: `height: 100%` → `min-height: 100%` on `body`/`#app` so the sticky header's travel isn't capped at one viewport.

## Verification (headless Chrome, this branch's stack)

- Home page desktop: **0px gap** between bars (was misaligned by design), sticky pinned at `top: 0` while scrolled, exactly **one** shrink flip down and one up across a 0→600→0px scroll sweep — no oscillation. Same 0px gap on the lux theme (the 1px-border suspect).
- Map view desktop: 0px gap between navbar and route bar, `elementFromPoint` in both gutters resolves to non-header elements while the center still hits the header, route sidebar sits exactly at the measured header height.
- Mobile (375px): route title + kebab render, the Share modal opens from the kebab; home page content starts at ~86px instead of the old desktop-tuned 190px spacer.
- `npx vitest` (6 hysteresis tests green), `composer run fix` + `composer run analyse` clean, `git grep` clean for `navbar-top-fixed-spacer` / `theme-header-height` / `has_dungeon_context_header` / `navbar-shrink`.

The navbar items wrapping on narrow-ish desktop map viewports is pre-existing (confirmed against the master baseline, which additionally shows the navbar half-overlapped under the game-version bar — the stacking bug this PR fixes).

## Follow-up (deliberately not in this diff)

- Migrating the `map.css` `calc(100vh - 106px/158px/...)` constants to `--ksg-header-height` — kept out to keep this reviewable.

Screenshots incoming in a comment (before/after desktop map, mobile).


Wotuu:
:robot: Verification screenshots (headless Chrome):

**Map view desktop — master (before)**: note the navbar overlapping up under the game-version bar (the hardcoded-offset stacking bug).
master map desktop

**Map view desktop — this branch (after)**: bars stack in flow, zero gap, gutters click through to the map. (The nav-item text wrapping also appears on the master shot — pre-existing.)
branch map desktop

**Home page — this branch** (darkly): full header at top of page, sticky + shrink verified programmatically (one class flip per direction, 0px inter-bar gap; same on lux).
branch home

**Map view mobile (375px) — this branch**: route bar now exists on mobile — title + kebab overflow menu (Share modal verified opening from it). Previously this bar was not rendered at all on mobile.
branch map mobile

resources/assets/css/base/layout.css
```diff
@@ -4,7 +4,14 @@ body {
...
-html, body, #app {
+html {
     height: 100%;
     width: 100%;
 }
+
+/* min-height, not height: a fixed height would cap the sticky site header's
+   travel at one viewport, so it would stop sticking further down the page. */
+body, #app {
+    min-height: 100%;
```
Wotuu, line 15: :robot: **Blocker — this collapses every non-embed map page to a 0px-high map.**

The chain `html{height:100%}` → `body{min-height:100%}` → `#app{min-height:100%}` → `.wrapper{height:100%}` (map.css:2) → `#map{height:100%}` no longer resolves: a percentage height resolves against the containing block's *height*, and `min-height` alone leaves it `auto` (`#app` is a flex item in a **column** flex `body`, so `align-items:stretch` sizes it horizontally, not vertically). Measured on this branch's stack:

```
/explore/retail/mechagon-workshop      -> {app:0, wrapper:0, map:0}
                                          pageErrors: ["IndexSizeError: Failed to execute 'getImageData' ... source height is 0"]
/route/scarlet-monastery-cathedral/... -> {wrapper:0, map:0}   (desktop AND --mobile)
```

Causation proven in-page: re-adding `document.body.style.height='100%'; #app.style.height='100%'` on the very same load gives `{app:967, wrapper:967, map:967}`. The embed pages survive only because `embed.scss` sizes `.embed_wrapper` with `calc(100vh - …)` (measured 946px, fine) — every non-embed map view (dungeonroute view/edit, live session, dungeon explore, admin floor mapping, the search-page map) renders a blank map.

Note that the PR body's map verification cannot catch this: a 0-height map still satisfies "`elementFromPoint` in the gutters resolves to a non-header element" (it hits `#app`/`body`), and the header itself lays out fine.

The sticky-travel problem the comment describes doesn't actually require dropping `height:100%` on `#app` — sticky travel is bounded by the *scrolling ancestor*, and the home page scrolls/sticks fine with the old rule. If a change here is genuinely needed, it has to come with a replacement height source for `.wrapper`/`#map`.

resources/views/common/layout/header.blade.php
```diff
@@ -24,13 +24,17 @@
...
  */
 
 $navs                   = [];
 $showMore               ??= false;
 $showDungeonContext     ??= true;
 $forceShrink            ??= false;
 $dungeonContextLinks    ??= null;
+// The map view passes null - its own #map_header wraps this include, and duplicate ids would
+// confuse the header height measurement in siteheader.js.
+$headerId               ??= 'site_header';
```
Wotuu: :robot: **Major — `??=` does not honour the explicit `null` that `map.blade.php` passes, so the map header still renders `id="site_header"`.**

`??=` assigns when the variable is `null`, and `@include('common.layout.header', ['headerId' => null])` defines `$headerId` as exactly that — so `$headerId` becomes `'site_header'` again and the `@if($headerId !== null)` guard always passes. Verified on the served page (`/route/…`):

```
document.getElementById('site_header')  -> <header id="site_header" class="ksg-header ksg-header--shrink ksg-header--shrink-forced">   (inside #map_header)
```

Consequence: `siteheader.js` line 34 picks `#site_header` first, so `--ksg-header-height` is published as the **inner** bar stack instead of the whole floating header:

| viewport | `#map_header` real height | published `--ksg-header-height` |
|---|---|---|
| 1600x1000 | 152.5px | **105px** (`.route_sidebar.right` computed `top: 105px`) |
| iPhone (`--mobile`) | 125.6px | **54px** |

So the route sidebar / `.route_manipulation_tools.mobile` are positioned *underneath* the route bar — on mobile the first tool buttons are visibly covered by the header. That's a regression against the hardcoded 150px/70px this PR replaced, and it also means the "duplicate ids would confuse the measurement" rationale in the comment isn't actually in effect.

A sentinel that isn't `null` (`false`, or `array_key_exists`) is needed — and `siteheader.js` should prefer `#map_header` over `#site_header` so the fallback ordering is right even if a stray id ever reappears.

resources/views/common/maps/controls/header.blade.php
```diff
@@ -232,6 +239,103 @@ class="btn-sm btn-success w-100">
...
+                            aria-label="{{ __('view_common.maps.controls.header.route_actions') }}">
+                        <i class="fas fa-ellipsis-v"></i>
+                    </button>
+                    <ul class="dropdown-menu dropdown-menu-end">
+                        @isset($dungeonroute)
+                            @if($mapContext instanceof MapContextLiveSession)
+                                @if($liveSession->expires_at === null)
+                                    <li>
+                                        <a class="dropdown-item" href="#"
+                                           data-bs-toggle="modal" data-bs-target="#stop_live_session_modal">
```
Wotuu, line 264: :robot: **Major — the mobile "Stop" item never actually stops the live session.**

The desktop control is `<button id="stop_live_session" …>` and `resources/assets/js/custom/inline/dungeonroute/livesession.js:24` binds it (`stopLiveSessionSelector => '#stop_live_session'`, wired in `dungeonroute/livesession/view.blade.php:41`) to `_stopLiveSession()`, which fires the `DELETE /ajax/{key}/live/{sessionKey}` that ends the session; the `data-bs-target` only opens the rating modal alongside it.

This kebab item has no `id` and no other binding, so on mobile it opens the "live session concluded" modal (rating + Edit route link) while the live session keeps running — the user is told it ended and it hasn't. Two knock-on gaps from the same asymmetry:

- `livesession.js:36` does `$('#stop_live_session').hide()` after a stop; the kebab item has no id, so it stays offered after the session has already been stopped.
- The desktop `#stopped_live_session_container` (expiry countdown + "Edit route"/"View route" link, gated on `$mayUserEdit`) has no mobile equivalent, so a mobile user in a concluded live session has no way back to the route.

The other kebab entries do match their desktop conditions 1:1 (save-to-profile `@auth`+`$showCreateRouteBtn`, admin settings `$isUserAdmin`, simulate, route settings `!isSandbox() && $edit`, mapping-version, share) — this is the only pair that diverges.

Wotuu:

Wotuu:
:robot: Good catch on the mobile map controls rendering behind the header — confirmed and fixed in 564654561.

Root cause: the map view passed `'headerId' => null` to suppress the `site_header` id, but PHP's `??=` treats null as unset, so the inner header still rendered `id="site_header"` — and `siteheader.js` measured that element (navbar only, 54px) instead of the full `#map_header` stack (126px on mobile). Every `--ksg-header-height` consumer, including the left map controls, then sat 72px too high. The sentinel is now `false`, and the JS prefers `#map_header` when present as a belt-and-braces.

Re-verified: mobile controls now start exactly at the header's bottom edge (126px), and on desktop the right sidebar now sits below the full three-bar stack (153px) — previously it would have tucked under the route bar.

mobile map controls fixed

resources/assets/css/base/layout.css
Wotuu: :robot: Correction to my last paragraph — I retract it. Sticky travel is bounded by the element's **containing block** (`#app`'s box), not by the scrolling ancestor, so `min-height` rather than `height` on `#app` is very likely *required* for the header to keep sticking past one viewport, exactly as the code comment says. I never tested a `height: 100%` + long-scroll combination, so I shouldn't have implied the comment was mistaken.

That makes the two requirements genuinely conflicting: sticky travel wants `#app` unbounded, while `.wrapper`/`#map` need a resolvable parent height. So the fix is a replacement height source for the map chain rather than a revert — e.g. `flex: 1` on `#app` with a flex chain down to `#map`, `min-height: 100dvh` on `.wrapper`, or scoping `height: 100%` to the map layout path only (map pages don't need a sticky header at all, `#map_header` is `position: fixed` there). Your call which. The blocker itself stands unchanged.

Wotuu: :robot: Fixed in f99e040f8. Correct on both counts — the min-height on #app is required for the sticky header's travel, so the map chain got its own definite height instead: `.wrapper { height: 100dvh }` (with a 100vh fallback) in map.css. All `.wrapper` consumers extend `layouts.map` and are full-viewport map pages, so this is safe. Re-verified in headless Chrome: route view and /explore/retail/ataldazar both render wrapper/#map at full viewport height with leaflet tiles loading, and the IndexSizeError is gone.

resources/views/common/layout/header.blade.php
Wotuu: :robot: Fixed in 564654561 (pushed before this review landed — Wotuu spotted the symptom on the mobile screenshot). The sentinel is now `false` (`??=` treats null as unset), and siteheader.js prefers `#map_header` over `#site_header` as a second line of defense. Re-verified: no stray #site_header on map pages, --ksg-header-height reports the full stack (126px mobile / 153px desktop), controls and sidebar sit exactly at the header's bottom edge.

resources/views/common/maps/controls/header.blade.php
Wotuu: :robot: Fixed in f99e040f8. Both triggers now share a `.stop_live_session_trigger` class and `livesession/view.blade.php` binds the stop handler to that class instead of the id — the mobile kebab Stop performs the DELETE and both triggers hide once stopped (the hide also runs through the same selector). The desktop-only countdown/edit-link container stays desktop-only for now: after stopping, mobile users get the same "session concluded" modal (with rating + navigation), which covers the flow; happy to add a mobile countdown row if wanted.

Wotuu:
:robot: Corrected desktop map screenshot after the `.wrapper` height fix — the earlier "after" shot inadvertently documented the 0px-map blocker (empty map area). Tiles now render, sidebar aligned below the full header stack:

map desktop fixed

Wotuu:
:robot: Follow-up per Wotuu: the header is now one opaque band — the wrapper carries `var(--theme-header)`, so page content no longer shows through the slivers between the bars' centered containers while scrolling. Probed every 12px across the full header band at 5 x-positions while scrolled: 0 leak points on darkly and lux. The map header explicitly stays transparent (floating look + click-through gutters unchanged).

unified header band

Wotuu:
:robot: Map header unified too (per Wotuu): a `#map_header::before` band, sized to the containers' 1400px max-width, paints `var(--theme-header)` behind the whole bar stack — the spacing between bars stays but shows the header color instead of the map. Gutters remain transparent and click-through (verified: left gutter hits the sidebar, right gutter hits the Leaflet map element, center hits the header). The faint map ghosting through the band is the pre-existing `map_fade_out` effect (0.8 opacity, 1.0 on hover).

map unified band

resources/assets/js/custom/inline/common/general/siteheader.js
```diff
@@ -0,0 +1,93 @@
...
+ * position, and a single threshold would let that nudge immediately re-cross it and flap
+ * the header back and forth (#3118, #3851).
+ *
+ * @param {number} scrollY
+ * @param {boolean} currentlyShrunk
+ * @param {number} shrinkAt
+ * @param {number} unshrinkAt
+ * @returns {boolean}
+ */
+function shouldShrinkHeader(scrollY, currentlyShrunk, shrinkAt = 140, unshrinkAt = 40) {
```
Wotuu, line 15: Does this function need to be outside of the class?

Wotuu: :robot: Yes — it's a deliberate pure-function extraction so the hysteresis logic (the shrink/unshrink threshold flap-prevention from #3118/#3851) is unit-testable without a DOM. `siteheader.test.js` requires it directly (`const {shouldShrinkHeader} = require('./siteheader')`) and exercises all the threshold/nudge cases against it. `CommonGeneralSiteheader.activate()` just calls it with `window.scrollY` and the current shrink state. No code change made — this was intentional, not an oversight.

Wotuu:
:robot: Per Wotuu's feedback: the site header's full-width band now paints the **page background color** (`--theme-darker`) instead of the header color, so the gutters left/right of the centered containers read as page again — and the navbar's vertical padding is halved (the theme's 1rem was fully visible now that the bars stack in flow, where the old fixed bars overlapped it away).

**Before** (2560px viewport — full-width light band, bulky navbar):

before

**After** (dark gutters, tighter navbar):

after

**Shrunk state** (scrolled — hysteresis unchanged, still one flip):

shrink

Map pages are unaffected: the `#map_header` band override stays transparent/centered, gutters still click through to the map, and `--ksg-header-height` re-measures automatically.

Wotuu:
:robot: Second round of Wotuu's feedback, three commits (verified in headless Chrome):

**Slim floating map header** (`f121e1414`) — the game version selection row is hidden on map pages (new `showGameVersionSelection` include flag; the dungeon context cards stay where shown), the title bar is skipped on explore-family pages for non-admins (admins keep it for the Edit mapping version button), and the vertical padding between the map header's bars is zeroed — those strips were pointer-events dead zones, so clicks fell through to the map and hovering them flickered the `map_fade_out` hover state. Route page header: 161px → 120px.

Before / after (route page):

before

after

Explore as anonymous (cards + navbar only, no title bar):

explore

**Site header transparent at rest** (`99519c47f`) — the always-on band regressed wallpaper pages (production lets the art bleed through the gutters). The opaque `--theme-darker` band is now keyed to the shrink class, so it only appears once the header is pinned over scrolling content. Below with a stand-in gradient wallpaper (the worktree has no wallpaper assets); the map header keeps its own transparent override:

transparent

Probes after the change: navbar↔route-bar gap 0px, cards↔navbar gap 0px, `elementFromPoint` in both former strips hits header content, shrunk band computes opaque `rgb(32,32,32)`, `--ksg-header-height` re-measures (120px route / 128px explore).

Wotuu:

PR #3858, opened 2026-08-06
#3664 Fix Mockery\Exception imported instead of \Exception

### LINKED ISSUE #3664: Mockery\Exception imported instead of \Exception in AjaxPathController and Dungeon
:robot: Found while auditing controllers for #170. Split out of #3663, which fixed the same defect in `AjaxKillZoneController` only.

Three files import `Mockery\Exception` instead of the global `\Exception` — almost certainly an IDE auto-import mistake:

- `app/Http/Controllers/Ajax/AjaxPathController.php:24`
- `app/Models/Dungeon.php:37`

`mockery/mockery` is in **`require-dev`**, so the class does not exist in production at all. PHP does not fatal on a `catch` clause naming a missing class — it just never matches — so each of these is silently dead or wrong:

### `AjaxPathController::store()`

```php
throw new \Exception(__('controller.path.error.unable_to_save_path'));  // L98, global
...
} catch (Exception) {                                                    // L102, Mockery
    $result = response(__('controller.generic.error.not_found'), Http::NOT_FOUND);
}
```

The catch never fires, in dev **or** production. A failed path save propagates out as a 500 instead of the intended 404.

**This is not a safe one-line fix.** The try/catch sits inside a `DB::transaction(...)` closure. Today the exception escapes the closure, so the transaction rolls back. Correcting the import makes the catch fire, which means the transaction **commits** whatever partial state was written before the failure. Whoever fixes this needs to decide whether the 404 response or the rollback is the intended behaviour — probably: catch, respond 404, and roll back explicitly.

### `Dungeon.php:191`

```php
} catch (Exception $exception) {
    dd($exception);
}
```

Dead code that would `dd()` in production if it ever fired. Should just be removed along with the try.

### Suggested check

`grep -rn "use Mockery\\\\Exception;" app/` should return nothing when this is done.

Wotuu:
:robot: Closes #3664

`Mockery\Exception` (require-dev only) was imported in place of the global `\Exception` in two catch clauses. PHP doesn't fatal on a `catch` naming a missing class - it just never matches - so both catches were silently dead/wrong in production.

## `AjaxPathController::store()`

The catch never fired, so a failed path save propagated out as a 500 instead of the intended 404. The try/catch sits inside a `DB::transaction(...)` closure - simply fixing the import would make the catch fire and swallow the exception *inside* the closure, which means the transaction would **commit** instead of rolling back. Moved the try/catch to wrap the whole `DB::transaction()` call so a thrown exception triggers Laravel's normal transaction rollback before being caught and turned into the 404 response. `Gate::authorize('edit', ...)` runs before the try, so an `AuthorizationException` is unaffected.

Note: the whole closure (including `Path::create()`/`update()`) is now inside the outer try, so a genuine DB failure there would also be converted to a 404 instead of surfacing as a 500 - this is inherent to fixing the transaction/rollback semantics correctly, flagged during review, and judged acceptable (matches the existing behaviour of the sibling `delete()` method and other Ajax controllers, which also 404 on `catch (\Exception)`).

## `Dungeon::getEnemyForcesMappedStatusAttribute()`

Dead code that would `dd()` in production if it ever fired. Removed the try/catch entirely, per the issue's suggestion. No functional impact - no test exercises this accessor, and the catch never matched in production anyway.

`grep -rn "use Mockery\\\\Exception;" app/` now returns nothing.

## Tests

Added `AjaxPathControllerTest::store_givenUpdateThatFailsToSave_shouldReturnNotFoundAndRollBackTransaction` - forces `Path::update()` to fail via an `updating` model-event listener, then asserts a 404 response and that the polyline created later in the same transaction was never persisted. Confirmed this test fails against the pre-fix code (uncaught error) and passes with the fix.

- `AjaxPathControllerTest` - 5 passed
- `composer run fix` / `composer run analyse` - clean

## Cold review

Independent review (agent, no shared context) confirmed the transaction/rollback restructuring is correct and `AuthorizationException` is unaffected. It found one real bug in the test: `Path::flushEventListeners()` in the `finally` block would have wiped `Path::boot()`'s `deleting` cascade listener for the rest of the PHPUnit process (listeners register once per class, not per test). Fixed by scoping cleanup to `Event::forget('eloquent.updating: '.Path::class)` instead.

### DIFF
--- a/app/Http/Controllers/Ajax/AjaxPathController.php
+++ b/app/Http/Controllers/Ajax/AjaxPathController.php
@@ -23,7 +23,6 @@ use Illuminate\Http\Response;
 use Illuminate\Support\Facades\Auth;
 use Illuminate\Support\Facades\DB;
 use Illuminate\Support\Facades\Gate;
-use Mockery\Exception;
 use Teapot\StatusCode\Http;
 
 class AjaxPathController extends Controller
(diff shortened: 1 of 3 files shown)


Wotuu:
:robot: Cold review: 2 findings, both addressed (test event-listener pollution fixed; DB-failure-now-404 scope broadening judged acceptable and noted in the PR body)

Issue #3852, opened 2026-08-05
Split SpellConstants into separate files per concern (follow-up to #3842)


Wotuu:
:robot: Follow-up from a review comment on PR #3842 (raised by Wotuu on `app/Models/Spell/SpellConstants.php`).

## Background

`app/Models/Spell/SpellConstants.php` is a single trait that has grown into a grab-bag of
unrelated constant groups. Each group represents its own concern and would be clearer as its own
enum/class:

- **Spell schools** (`SCHOOL_*`, `ALL_SCHOOLS`) — bitmask of damage schools (physical, holy, fire, ...).
- **Miss types** (`MISS_TYPE_*`, `ALL_MISS_TYPES`, `GUID_MISS_TYPE_MAPPING`) — bitmask of combat log
  miss types, plus a mapping from `App\Logic\CombatLog\Guid\MissType\*` classes to those bitmask values.
- **Counters** (`COUNTER_*`, `ALL_COUNTERS`) — bitmask of counter-ability types (vanish, shadowmeld,
  feign death, invisibility, cloak of shadows).
- **Dispel types** (`DISPEL_TYPE_*`, `ALL_DISPEL_TYPES`, `DISPEL_TYPE_TRANSLATION_KEY_PREFIX`) —
  string-keyed dispel type values that double as translation key suffixes.
- **Spell categories** (`CATEGORY_*`, `ALL_CATEGORIES`) — class/general categorization.
- **Cooldown groups** (`COOLDOWN_GROUP_*`, `ALL_COOLDOWN_GROUPS`).
- **Mechanics** (`MECHANIC_*`, `ALL_MECHANIC`) — Blizzard mechanic IDs (asleep, banished, stunned, ...).
- **Hard-coded exception spell IDs** (`SPELL_BLOODLUST`, `SPELL_HEROISM`, ..., `CHARM_SPELLS`,
  `BLOODLUSTY_SPELLS`, `EXCLUDE_MDT_IMPORT_SPELLS`) — a set of specific spell IDs the codebase treats
  as special cases, unrelated to any of the groups above.

That's 8 distinct concerns living in one 280-line trait, several of which are already awkwardly
named with "ALL_" prefixes to avoid collisions between groups.

## Remaining work

Split `SpellConstants` into separate files, one per concern, likely as native PHP enums (backed
where the group already carries meaningful scalar values, e.g. `int` for bitmasks, `string` for
[...]

PR #3859, opened 2026-08-06
#3857 Split DungeonConstants into DungeonKey, RaidKey and DungeonDifficulty enums

### LINKED ISSUE #3857: Split DungeonConstants into separate files per concern (follow-up to #3852)
:robot: Sibling of #3852, which does the same for `app/Models/Spell/SpellConstants.php`.

## Background

`app/Models/DungeonConstants.php` is a single 462-line trait, pulled into `App\Models\Dungeon` via
`use DungeonConstants;`, that mixes several unrelated concerns:

- **Speedrun difficulties** (`DIFFICULTY_10_MAN`/`_20_MAN`/`_25_MAN`/`_40_MAN`, `DIFFICULTY_ALL`) —
  a slug → database-id map (`'10_man' => 1`, ...). The slug doubles as the translation key suffix
  in `Dungeon::getDifficultyName()` (`view_common.dungeon.difficulty.<slug>`), the id is what is
  stored in `dungeon_routes.difficulty` and `dungeon_speedrun_difficulties.difficulty`.
- **Dungeon keys** (`DUNGEON_*`) — 153 string constants, one per dungeon, grouped by expansion.
- **Raid keys** (`RAID_*`) — 30 string constants, one per raid, grouped by expansion.
- **The expansion groupings** (`ALL`, `ALL_RAID`) — `array<expansionKey, list<dungeonKey>>`, which
  most consumers immediately `array_merge_recursive()` back together.

The two groups have nothing to do with each other: difficulty is a small int-keyed lookup, the
dungeon/raid keys are a flat catalogue of instances. The `ALL_` prefixes exist purely to avoid
collisions inside the one trait.

Verified facts about the data (checked by reflection over the trait):

- 153 `DUNGEON_*` constants, all 153 present in `ALL`; 30 `RAID_*` constants, all 30 present in
  `ALL_RAID`. No constant is missing from its grouping and no grouping entry lacks a constant.
- No duplicate *values* across all 183 constants.
- Exactly one *name* collision between the two groups: `DUNGEON_ZUL_GURUB` (`zul_gurub`, Cataclysm)
  vs `RAID_ZUL_GURUB` (`zulgurub`, Classic). This is why dungeons and raids need to stay two
  separate enums rather than being merged into one.

## Remaining work

Split the trait into separate files, one per concern, as native PHP enums:

- **`DungeonDifficulty`** — `int`-backed enum (backed by the database id, i.e. `TEN_MAN = 1`,
  `TWENTY_FIVE_MAN = 2`, `TWENTY_MAN = 3`, `FORTY_MAN = 4`), with a `slug(): string` method
  returning `10_man`/`25_man`/... for the translation key, and the translated-name helper currently
  living in `Dungeon::getDifficultyName()`. Note `name` is reserved on enums (it is the case-name
  property), so the translated getter needs another name (e.g. `translatedName()`).
- **`DungeonKey`** — `string`-backed enum of the 153 dungeon keys, carrying its expansion (replacing
  `Dungeon::ALL`), e.g. `expansionKey(): string` plus `casesForExpansion(string $expansionKey)`.
- **`RaidKey`** — `string`-backed enum of the 30 raid keys, same shape (replacing `Dungeon::ALL_RAID`).

`Dungeon::findExpansionByKey()` then becomes
`DungeonKey::tryFrom($key)?->expansionKey() ?? RaidKey::tryFrom($key)?->expansionKey()`.

Naming and exact file locations are up to the implementer. `app/Models/Dungeon.php` is *not* in an
`app/Models/Dungeon/` sub-namespace (unlike `Spell`), so the new files most likely sit flat in
`app/Models/` rather than creating an `App\Models\Dungeon` namespace that shadows the
`App\Models\Dungeon` class name.

- **Blast radius**: ~480 references to `Dungeon::DUNGEON_*`/`Dungeon::RAID_*` across app, tests,
  seeders, migrations and Blade — 183 inside the trait itself, 161 in
  `app/Logic/MDT/Conversion.php` (`DUNGEON_NAME_MAPPING`), the rest spread over ~60 files
  (`tests/Feature/Controller/Api/V1/APICombatLogController/**` alone has dozens of one-line
  `getDungeonKey()` overrides). Wide, mechanical, but easy to get subtly wrong.
- Almost every consumer compares against `$dungeon->key`, a plain `string` column, so migrating to
  a backed enum means adding `->value` at most call sites (or casting `Dungeon::$key`, which is a
  much larger change and should **not** be attempted here).
- `DIFFICULTY_ALL` is referenced from **migrations** (`2026_06_24_120000_create_dungeon_speedrun_difficulties_table.php`
  and a legacy one) — migrations must keep working, so whatever replaces it has to stay resolvable
  from those files.
- The difficulty ids are duplicated in JavaScript (`resources/assets/js/custom/constants.js`,
  `DUNGEON_DIFFICULTY_10_MAN = 1`) and in JS test fixtures which reference `DungeonConstants::` by
  name in comments — those comments need updating so they don't point at a deleted file.
- `Conversion::DUNGEON_NAME_MAPPING` contains large commented-out blocks referencing these
  constants; they should be rewritten consistently rather than left pointing at a removed trait.

## Recommended approach

One safe change at a time, running `composer run analyse` between steps:

1. Add the three new enums (no deletions yet).
2. Move `Dungeon::getDifficultyName()`/`findExpansionByKey()` logic onto the enums, leaving the
   `Dungeon` methods as thin delegates if callers are widespread.
3. Mechanically rewrite `Dungeon::DUNGEON_X` → `DungeonKey::X->value` and
   `Dungeon::RAID_X` → `RaidKey::X->value` across the codebase (a scripted `sed` over the
   constant list, then review the diff by hand — watch for the `ZUL_GURUB` name collision).
4. Rewrite `Dungeon::ALL`/`ALL_RAID` consumers (`DungeonController`, `DungeonFormRequest`,
   `Console\Commands\Dungeon\CreateMissing`, `Dungeon::findExpansionByKey`).
5. Rewrite `DIFFICULTY_*` consumers, including the two migrations and the Blade views.
6. Delete `app/Models/DungeonConstants.php` and its `use` in `Dungeon`.

## Key files

- `app/Models/DungeonConstants.php`, `app/Models/Dungeon.php`
- `app/Logic/MDT/Conversion.php`
- `app/Http/Controllers/Dungeon/DungeonController.php`, `app/Http/Requests/DungeonFormRequest.php`
- `app/Console/Commands/Dungeon/CreateMissing.php`
- `database/migrations/2026_06_24_120000_create_dungeon_speedrun_difficulties_table.php`
- `resources/views/common/dungeonroute/create/dungeondifficultyselect.blade.php`,
  `resources/views/admin/dungeon/edit.blade.php`, `resources/views/admin/floor/speedrunrequirednpcs.blade.php`
- `tests/Feature/Controller/Api/V1/APICombatLogController/**`

## Definition of done

- Each concern lives in its own file/enum; `app/Models/DungeonConstants.php` is gone.
- All prior usages across PHP, Blade, seeders, migrations and tests updated.
- `php artisan test --compact` green (no new failures).

Wotuu:
:robot: Closes #3857

Sibling of #3852 (the same split for `SpellConstants`).

## What changed

`app/Models/DungeonConstants.php` was a single 462-line trait, pulled into `Dungeon` via `use DungeonConstants;`, mixing four unrelated concerns. It is gone, replaced by three native enums in `app/Models`:

| New file | Replaces | Backed by |
| --- | --- | --- |
| `DungeonDifficulty` | `DIFFICULTY_*`, `DIFFICULTY_ALL`, `Dungeon::getDifficultyName()` | `int` — the database ID stored in `dungeon_routes`.`difficulty` |
| `DungeonKey` | the 153 `DUNGEON_*` constants + `ALL` | `string` — the `dungeons`.`key` column |
| `RaidKey` | the 30 `RAID_*` constants + `ALL_RAID` | `string` — the `dungeons`.`key` column |

- Each key enum carries its own expansion: `expansionKey()` (a `match`, so PHPStan enforces exhaustiveness when a case is added) and `casesByExpansionKey()`, which is what `ALL`/`ALL_RAID` were.
- `DungeonDifficulty` carries `slug()` (`10_man`, ... — also the translation-key suffix) and `translatedName()`, which is what `Dungeon::getDifficultyName()` did.
- **Dungeons and raids stay two enums** rather than one: `DUNGEON_ZUL_GURUB` (`zul_gurub`, Cataclysm) and `RAID_ZUL_GURUB` (`zulgurub`, Classic) collide by name, so a merged enum could not name both cases. For the consumers that used `array_merge_recursive(Dungeon::ALL, Dungeon::ALL_RAID)`, `Dungeon::allKeysByExpansionKey()` and `Dungeon::allKeys()` do the merge; `Dungeon::findExpansionByKey()` now resolves through both enums.
- The `//blackfanthomdeeps`-style trailing comments (MDT/Blizzard names) were carried over onto the enum cases.

## Verification that nothing shifted

The enums were generated from the old trait and then diffed against it at runtime: `ALL`, `ALL_RAID`, `array_merge_recursive(ALL, ALL_RAID)`, its flattened form, and `DIFFICULTY_ALL` all reproduce **identically, including ordering** (which matters — `ALL` order drives the admin "create dungeon" key dropdown, and the flattened order feeds `Rule::in`).

`DungeonResource`'s public `speedrunDifficulties` output (a list of difficulty slugs) is unchanged.

## Blast radius

110 files: ~480 constant references rewritten to `DungeonKey::X->value` / `RaidKey::X->value` across app, tests, seeders, migrations and Blade, plus the two migrations and the JS comments/fixtures that named `DungeonConstants::` by hand.

## Checks

- `composer run analyse` clean, `composer run fix` clean.
- Backend-only — no rendered output changed apart from the admin speedrun-NPC tab DOM ids, which still resolve to the same slugs. No screenshots.


routes/breadcrumbs.php
```diff
@@ -385,7 +386,7 @@
     $trail->parent('admin.floor.edit', $dungeon, $floor);
     $trail->push(
         __('breadcrumbs.home.admin.dungeonspeedrunrequirednpc.new_dungeonspeedrunrequirednpc', [
-            'difficulty' => Dungeon::getDifficultyName($difficulty),
+            'difficulty' => DungeonDifficulty::from($difficulty)->translatedName(),
```
Wotuu, line 389: :robot: `DungeonDifficulty::from()` on an unvalidated route parameter now throws instead of degrading.

The `{difficulty}` segment of `admin.dungeonspeedrunrequirednpc.*` (routes/web.php:387-389) has no route constraint and no Form Request covers the path parameter — `DungeonSpeedrunRequiredNpcsController::create()` just takes `int $difficulty` straight from the URL. The old `Dungeon::getDifficultyName()` degraded gracefully on an out-of-range value (`array_search()` returned `false`, so it rendered the bare translation key). `DungeonDifficulty::from()` throws an uncaught `ValueError` → 500, both here and in `admin/dungeonspeedrunrequirednpc/new.blade.php` (lines 19 and 24).

Admin-gated and not a user-facing regression, so not a blocker — but since the value is attacker/typo-controlled it would be cheaper to fail cleanly: either constrain the route (`->whereIn('difficulty', DungeonDifficulty::values())`) or use `tryFrom()` + `abort(404)` in the controller and pass the enum instance to the view.

app/Models/DungeonKey.php
```diff
@@ -0,0 +1,366 @@
+<?php
+
+namespace App\Models;
+
+/**
+ * The `key` column of every dungeon known to the application. Previously `Dungeon::DUNGEON_*`.
+ */
+enum DungeonKey: string
```
Wotuu, line 8: :robot: Nothing enforces the disjointness invariant the split now depends on.

Splitting one constant list into two enums that both back the same `dungeons`.`key` column introduces a new invariant: `DungeonKey` and `RaidKey` values must never overlap. `Dungeon::findExpansionByKey()` (`DungeonKey::tryFrom(...) ?? RaidKey::tryFrom(...)`) would silently return the dungeon expansion for a colliding key, and `Dungeon::allKeys()` would emit a duplicate into the `Rule::in` list in `DungeonFormRequest`. Previously the two lists lived in one file next to each other; now they are two files 300 lines apart and the collision is invisible at review time (the near-miss `zul_gurub`/`zulgurub` is exactly the kind of pair that makes this plausible).

I verified today that the sets are disjoint and that `casesByExpansionKey()` reproduces `ALL`/`ALL_RAID` byte-for-byte including ordering — but that verification was a throwaway script, not something the suite keeps checking. Worth a small unit test asserting `array_intersect(DungeonKey::values-ish, RaidKey::...)` is empty and that `Dungeon::allKeys()` has no duplicates, so a future key added to the wrong enum fails in CI rather than in prod.

Wotuu:

Both are non-blocking. Verification I did beyond reading the diff:

- Reconstructed `ALL`/`ALL_RAID` from the deleted trait and compared them to `DungeonKey`/`RaidKey` grouped by **declaration order** (which is what `casesByExpansionKey()` actually iterates, not the `match` arm order): identical, including per-expansion ordering, 153 + 31 cases, no case unmapped, no value duplicated within or across the two enums.
- Resolved every `Dungeon::DUNGEON_*`/`RAID_*` removed and every `DungeonKey::*`/`RaidKey::*` added, per file, across all 110 files, and compared the resolved string sequences — no mis-substituted key anywhere.
- Diffed out every changed line that contains no constant/enum reference, to catch changes riding along with the rename: the residue is exactly the restructurings described in the PR body (`allKeysByExpansionKey`/`allKeys`, `DungeonResource`, the speedrun blades, one `->values()` drop in `DungeonRouteSaveServiceSaveTest`). Nothing unaccounted for.
- `DungeonResource.speedrunDifficulties` and the `Rule::in` difficulty lists come out identical, and the `empty($difficultiesWithData)` guard in `speedrunrequirednpcs.blade.php` is unaffected by the assoc→list key change.

routes/breadcrumbs.php
Wotuu: :robot: Fixed by constraining the route instead of restructuring the controller/breadcrumb signatures.

`routes/web.php` now applies `->whereIn('difficulty', DungeonDifficulty::values())` to the whole `speedrunrequirednpcs` route group, so all three routes (`new`, `savenew`, `delete`) 404 at the routing layer for any out-of-range or typo'd `difficulty` before dispatch ever reaches a controller. That makes the three `DungeonDifficulty::from()` call sites (`routes/breadcrumbs.php:389`, `new.blade.php:19`, `new.blade.php:24`) unreachable with an invalid value — left them untouched since the ValueError can no longer occur.

Verified the constraint actually merged onto all three named routes (not just the GET) via `Route::getRoutes()->getByName(...)->wheres`, and that it survives a `route:cache`/`route:clear` round-trip.

Added `tests/Feature/Controller/Admin/DungeonSpeedrunRequiredNpcsControllerTest.php`:
- `create_givenValidDifficulty_returnsOk` — happy path, exercises the breadcrumb/view `from()` calls end-to-end.
- `create_givenDifficultyOutsideDeclaredValues_returnsNotFound` — guards the 404 behavior directly.

app/Models/DungeonKey.php
Wotuu: :robot: Added a defensive unit test in `tests/Unit/App/Models/DungeonKeyTest.php`:

- `dungeonKeyAndRaidKey_givenBothEnums_haveDisjointValueSets` — asserts `array_intersect()` of the two enums' backing values is empty.
- `allKeys_givenEveryDungeonAndRaidKey_containsNoDuplicates` — asserts `Dungeon::allKeys()` has no duplicate entries (surfaces the actual colliding key(s) on failure rather than just a count mismatch).

No production code changes were needed — both invariants already held, this just guards them in CI going forward.

app/Http/Requests/DungeonFormRequest.php
```diff
@@ -45,7 +46,7 @@ public function rules(): array
             'key' => [
                 'required',
                 Rule::unique(Dungeon::class, 'key')->ignore($this->get('key'), 'key'),
-                Rule::in(collect(array_merge_recursive(Dungeon::ALL, Dungeon::ALL_RAID))->flatten()),
+                Rule::in(Dungeon::allKeys()),
```
Wotuu, line 49: Does this function even exist? Should it not be `DungeonKey::allKeys()` or something?

app/Models/Dungeon.php
```diff
@@ -579,21 +578,47 @@ public function hasImageWallpaper(): bool
...
+     * All known dungeon and raid keys.
+     *
+     * @return list<string>
      */
-    public static function getDifficultyName(int $difficulty): string
+    public static function allKeys(): array
     {
-        return __(sprintf('view_common.dungeon.difficulty.%s', array_search($difficulty, self::DIFFICULTY_ALL, true)));
+        return array_merge(...array_values(self::allKeysByExpansionKey()));
     }
```
Wotuu, line 618: So it does exist, my bad. But why is this here and not somewhere in the DungeonKey class? I don't think it should live here.

app/Http/Requests/DungeonFormRequest.php
Wotuu: :robot: Confirmed — `Dungeon::allKeys()` is defined in `app/Models/Dungeon.php` and merges `DungeonKey::casesByExpansionKey()` + `RaidKey::casesByExpansionKey()`, so this validation rule correctly accepts both dungeon and raid keys.

app/Models/Dungeon.php
Wotuu: :robot: Left it on `Dungeon` rather than `DungeonKey` because `allKeys()`/`allKeysByExpansionKey()` merge both `DungeonKey` and `RaidKey` cases — putting it on `DungeonKey` would mean that enum needs to know about `RaidKey`, which felt backwards. `Dungeon` is the model that validates against both (its `key` column can hold either), so it seemed like the natural home. Open to moving it if you have a specific place in mind — didn't want to guess at a bigger restructure during this pass.

PR #3860, opened 2026-08-06
#3666 Fix hasKilledAllRequiredEnemies() and restore the publish guard

### LINKED ISSUE #3666: DungeonRoutePolicy::publish guards on hasKilledAllRequiredEnemies(), which always returns true
:robot: Found while auditing policies for #170. Left untouched by #3665 because it is a product decision, not cleanup.

`DungeonRoutePolicy::publish()` opens with:

```php
public function publish(User $user, DungeonRoute $dungeonroute): Response
{
    if (!$dungeonroute->hasKilledAllRequiredEnemies()) {
        return $this->deny(__('policy.publish_not_all_required_enemies_killed'));
    }
    ...
}
```

But `DungeonRoute::hasKilledAllRequiredEnemies()` (`app/Models/DungeonRoute/DungeonRoute.php:1122`) has its entire body commented out and unconditionally returns `true`:

```php
public function hasKilledAllRequiredEnemies(): bool
{
    $result = true;

    //        foreach ($this->dungeon->enemies as $enemy) {
    //            if ($enemy->required && ...) {
    //                if (!$this->isEnemyKilled($enemy->id)) {
    //                    $result = false;
    //                    break;
    //                }
    //            }
    //        }

    return $result;
}
```

So the guard never fires, and the `policy.publish_not_all_required_enemies_killed` translation string is unreachable in every locale.

Two ways out:

1. **Restore the check** — decide what "required enemy" means under the current seasonal/teeming rules and reimplement it. `Enemy::$required` and `isEnemyKilled()` still exist.
2. **Remove it** — drop the guard from the policy, delete `hasKilledAllRequiredEnemies()` and `isEnemyKilled()` if nothing else uses them, and remove the `policy.publish_not_all_required_enemies_killed` key from `lang/en_US/policy.php`.

Either is fine; the current state (a guard that reads as validation but validates nothing) is the worst of the three.

Wotuu:
:robot: Closes #3666

## The bug

`DungeonRoutePolicy::publish()` guarded on `DungeonRoute::hasKilledAllRequiredEnemies()`, whose body was entirely commented out and unconditionally returned `true`. The issue offered "restore it" or "remove it" as equally fine options — but the reason it was commented out is that **restoring it as written makes routes unpublishable**, so this MR fixes the underlying bug and restores the guard.

The original implementation walked `$this->dungeon->enemies` — a `hasManyThrough` over floors that is **not scoped to a mapping version**. It therefore demanded the route kill required enemies from *every* mapping version the dungeon ever had, including enemies that are not on the route's own map and can never be pulled. Measured on the seeded data:

```
BROKEN dungeons.bfa.atal_dazar.name              all=20  current=4
BROKEN dungeons.bfa.freehold.name                all=49  current=0
BROKEN dungeons.bfa.kings_rest.name              all=104 current=28
BROKEN dungeons.bfa.siege_of_boralus.name        all=5   current=0
BROKEN dungeons.bfa.temple_of_sethraliss.name    all=41  current=8
BROKEN dungeons.bfa.the_motherlode.name          all=8   current=0
BROKEN dungeons.bfa.the_underrot.name            all=48  current=6
BROKEN dungeons.bfa.waycrest_manor.name          all=8   current=0
BROKEN dungeons.bfa.mechagon_workshop.name       all=9   current=0
BROKEN dungeons.sl.tazavesh_streets_of_wonder.name all=1 current=0
10 dungeons publish-blocked by stale required enemies, 3 have legit current required enemies, 142 clean
```

Waycrest Manor is the clearest case: 8 required enemies exist, **all** in an old mapping version, and its current mapping version has **zero** — so every Waycrest route was permanently unpublishable.

## Changes

- **Scope the check to the route's own mapping version** (`$this->mappingVersion->enemies`). This is the actual fix.
- **Fix the teeming comparison.** Both the commented-out PHP and the *live* JS tested `teeming === 'invisible'`, but the column only ever holds `null`, `'visible'` or `'hidden'` (`Enemy::TEEMING_ALL`) — the branch meant to excuse a teeming-despawned enemy was dead code, so such an enemy was wrongly demanded of teeming routes.
- **Always allow making a route *less* visible.** The `publish` gate also covers unpublishing, so the guard as-written would trap an already-published route whose mapping later gained a required enemy — its author could never take it down. The policy now takes the target state and only blocks moves *to* a published state.
- **Apply the guard to scheduled publishes**, which bypass the policy entirely (`DungeonRouteService::publishScheduledDungeonRoutes()`), following the existing skip-and-log pattern there.
- **Restore the editor UI.** Rather than disabling the whole select (which would also block unpublishing), the publish select now disables only its *publishing* options and explains why via a tooltip. This lives in `publish.js`, which owns the select and builds its options, instead of the commented-out `edit.js` code that raced option construction.
- `isEnemyKilled()` now shares one killed-enemy-id map instead of rescanning every kill zone per enemy.
- `mapping_version_id` is a nullable column; the PHPDoc claimed non-null `int`. Corrected (PHPStan stays clean, no cascade).

## Verification

- New `DungeonRouteHasKilledAllRequiredEnemiesTest` (7 tests) and 2 new policy tests; **confirmed they fail against the original `dungeon->enemies` implementation** and pass with the fix.
- New Vitest suite for the JS method (8 tests) pinning all three teeming values; **confirmed the teeming-hidden test fails with the old `'invisible'` string**.
- Full JS suite green (372 tests). PHPStan clean. Ran the check over 300 real seeded routes: 35 publishable / 1 legitimately blocked (a demo route that genuinely skips one required enemy), where before the fix every one of them returned `true` unconditionally.
- Three `--group=DungeonRoute` failures ("given inactive dungeon") are pre-existing and reproduce on a clean tree — the local seed has no inactive dungeon.

### Visual verification

Share modal with an unkilled required enemy, dropdown open. Before (master JS) — everything selectable; after — publishing options greyed out, **Unpublished still selectable**, tooltip explains why.

| Before | After |
| --- | --- |
| before | after |

### DIFF
--- a/app/Models/DungeonRoute/DungeonRoute.php
+++ b/app/Models/DungeonRoute/DungeonRoute.php
@@ -112,7 +112,7 @@ use Override;
  * @property Carbon               $published_at
  * @property Carbon|null          $expires_at
  *
- * @property MappingVersion                    $mappingVersion
+ * @property MappingVersion|null               $mappingVersion
  * @property Dungeon                           $dungeon
  * @property Path                              $route
  * @property Season|null                       $season
@@ -1148,12 +1148,15 @@ class DungeonRoute extends Model implements TracksPageViewInterface
      */
     public function hasKilledAllRequiredEnemies(): bool
     {
-        // Sandbox/unmigrated routes have no mapping version to compare against - nothing can be required of them.
-        if ($this->mapping_version_id === null) {
+        $this->loadMissing('mappingVersion.enemies');
+
+        // Sandbox/unmigrated routes have no mapping version to compare against - and since this schema carries no
+        // foreign keys, a route can also point at a mapping version that has since been deleted. Nothing can be
+        // required of either, and a hard failure here would 500 the publish endpoint and abort the whole
+        // scheduled-publish cron run.
+        if ($this->mappingVersion === null) {
             return true;
         }
-
-        $this->loadMissing('mappingVersion.enemies');
 
         $requiredEnemies = $this->mappingVersion->enemies
             ->filter(fn(Enemy $enemy) => $enemy->required && $this->isEnemyVisibleForTeeming($enemy));
--- a/app/Models/DungeonRoute/DungeonRoute.php
+++ b/app/Models/DungeonRoute/DungeonRoute.php
@@ -1150,9 +1150,9 @@ class DungeonRoute extends Model implements TracksPageViewInterface
     {
         $this->loadMissing('mappingVersion.enemies');
 
-        // Sandbox/unmigrated routes have no mapping version to compare against - and since this schema carries no
-        // foreign keys, a route can also point at a mapping version that has since been deleted. Nothing can be
-        // required of either, and a hard failure here would 500 the publish endpoint and abort the whole
+        // Sandbox routes always have a mapping version, but this schema carries no foreign keys, so a route can
+        // point at a mapping version that has since been deleted. Nothing can be required against a missing
+        // mapping version, and a hard failure here would 500 the publish endpoint and abort the whole
         // scheduled-publish cron run.
         if ($this->mappingVersion === null) {
             return true;
--- a/app/Models/DungeonRoute/DungeonRoute.php
+++ b/app/Models/DungeonRoute/DungeonRoute.php
@@ -1137,37 +1137,34 @@ class DungeonRoute extends Model implements TracksPageViewInterface
      */
     public function isEnemyKilled(int $enemyId): bool
     {
-        $result = false;
-
-        foreach ($this->killZones as $killZone) {
-            if ($killZone->getEnemies()->filter(static fn($enemy) => $enemy->id === $enemyId)->isNotEmpty()) {
-                $result = true;
-                break;
-            }
-        }
-
-        return $result;
+        return isset($this->getKilledEnemyIds()[$enemyId]);
     }
 
     /**
-     * Checks if this route has killed all required enemies.
+     * Checks if this route has killed all enemies that are marked as required in the route's mapping version.
+     *
+     * Enemies are only considered when they are visible for this route's teeming setting - an enemy that only spawns
+     * on teeming keys cannot be required on a non-teeming route and vice versa.
      */
     public function hasKilledAllRequiredEnemies(): bool
     {
-        $result = true;
+        // Sandbox/unmigrated routes have no mapping version to compare against - nothing can be required of them.
+        if ($this->mapping_version_id === null) {
+            return true;
+        }
 
-        //        foreach ($this->dungeon->enemies as $enemy) {
-        //            if ($enemy->required &&
-        //                ($enemy->teeming === null || ($enemy->teeming === 'visible' && $this->teeming) || ($enemy->teeming === 'invisible' && $this->teeming))) {
-        //
-        //                if (!$this->isEnemyKilled($enemy->id)) {
-        //                    $result = false;
-        //                    break;
-        //                }
-        //            }
-        //        }
+        $this->loadMissing('mappingVersion.enemies');
 
-        return $result;
+        $requiredEnemies = $this->mappingVersion->enemies
+            ->filter(fn(Enemy $enemy) => $enemy->required && $this->isEnemyVisibleForTeeming($enemy));
+
+        if ($requiredEnemies->isEmpty()) {
+            return true;
+        }
+
+        $killedEnemyIds = $this->getKilledEnemyIds();
+
+        return $requiredEnemies->every(static fn(Enemy $enemy) => isset($killedEnemyIds[$enemy->id]));
     }
 
     public function hasUniqueAffix(string $affix): bool
@@ -1390,6 +1387,39 @@ class DungeonRoute extends Model implements TracksPageViewInterface
             // Make sure the relation should be reloaded
             $this->unsetRelation('affixGroups');
         }
+    }
+
+    /**
+     * All enemy ids killed by any of this route's kill zones, keyed by enemy id for O(1) lookups.
+     *
+     * @return array<int, true>
+     */
+    private function getKilledEnemyIds(): array
+    {
+        $this->loadMissing('killZones.enemies');
+
+        $result = [];
+
+        foreach ($this->killZones as $killZone) {
+            foreach ($killZone->getEnemies() as $enemy) {
+                $result[$enemy->id] = true;
+            }
+        }
+
+        return $result;
+    }
+
+    /**
+     * Whether the enemy spawns at all given this route's teeming setting. Enemies that are not teeming-specific
+     * (teeming === null) always spawn.
+     */
+    private function isEnemyVisibleForTeeming(Enemy $enemy): bool
+    {
+        return match ($enemy->teeming) {
+            Enemy::TEEMING_VISIBLE => (bool)$this->teeming,
+            Enemy::TEEMING_HIDDEN  => !$this->teeming,
+            default                => true,
+        };
     }
 
     /**
--- a/app/Models/DungeonRoute/DungeonRoute.php
+++ b/app/Models/DungeonRoute/DungeonRoute.php
@@ -77,7 +77,7 @@ use Override;
  * @property string               $public_key
  * @property int                  $author_id
  * @property int                  $dungeon_id
- * @property int                  $mapping_version_id
+ * @property int|null             $mapping_version_id
  * @property int                  $season_id
  * @property int                  $faction_id
  * @property int|null             $team_id
--- a/app/Policies/DungeonRoutePolicy.php
+++ b/app/Policies/DungeonRoutePolicy.php
@@ -4,6 +4,7 @@ namespace App\Policies;
 
 use App\Models\DungeonRoute\DungeonRoute;
 use App\Models\Laratrust\Role;
+use App\Models\PublishedState;
 use App\Models\User;
 use Illuminate\Auth\Access\HandlesAuthorization;
 use Illuminate\Auth\Access\Response;
@@ -65,10 +66,14 @@ class DungeonRoutePolicy
 
     /**
      * Determine whether the user can publish dungeon routes.
+     *
+     * @param string|null $publishedState The state the route is being changed to, if known. Unpublishing is always
+     *                                    allowed - the required enemies guard would otherwise trap an already
+     *                                    published route in a state its author can no longer undo.
      */
-    public function publish(User $user, DungeonRoute $dungeonroute): Response
+    public function publish(User $user, DungeonRoute $dungeonroute, ?string $publishedState = null): Response
     {
-        if (!$dungeonroute->hasKilledAllRequiredEnemies()) {
+        if ($publishedState !== PublishedState::UNPUBLISHED && !$dungeonroute->hasKilledAllRequiredEnemies()) {
             return $this->deny(__('policy.publish_not_all_required_enemies_killed'));
         }
 
--- a/app/Policies/DungeonRoutePolicy.php
+++ b/app/Policies/DungeonRoutePolicy.php
@@ -67,9 +67,10 @@ class DungeonRoutePolicy
     /**
      * Determine whether the user can publish dungeon routes.
      *
-     * @param string|null $publishedState The state the route is being changed to, if known. Unpublishing is always
-     *                                    allowed - the required enemies guard would otherwise trap an already
-     *                                    published route in a state its author can no longer undo.
+     * @param string|null $publishedState The state the route is being changed to, if known. The required enemies
+     *                                    check is not executed when you are unpublishing the route - otherwise the
+     *                                    guard would trap an already published route in a state its author can no
+     *                                    longer undo.
      */
     public function publish(User $user, DungeonRoute $dungeonroute, ?string $publishedState = null): Response
     {
--- a/app/Service/DungeonRoute/DungeonRouteService.php
+++ b/app/Service/DungeonRoute/DungeonRouteService.php
@@ -282,7 +282,12 @@ readonly class DungeonRouteService implements DungeonRouteServiceInterface
             /** @var Collection<int, DungeonRouteScheduledPublish> $scheduledPublishes */
             $scheduledPublishes = DungeonRouteScheduledPublish::query()
                 ->where('publish_at', '<=', now())
-                ->with(['dungeonRoute.dungeon', 'dungeonRoute.author'])
+                ->with([
+                    'dungeonRoute.dungeon',
+                    'dungeonRoute.author',
+                    'dungeonRoute.mappingVersion.enemies',
+                    'dungeonRoute.killZones.enemies',
+                ])
                 ->get();
 
             foreach ($scheduledPublishes as $scheduledPublish) {
@@ -298,6 +303,12 @@ readonly class DungeonRouteService implements DungeonRouteServiceInterface
 
                 if ($publishedState === PublishedState::WORLD && !$dungeonRoute->dungeon->active) {
                     $this->log->publishScheduledDungeonRouteSkippedInactiveDungeon($dungeonRoute->id, $dungeonRoute->dungeon_id, $scheduledPublish->id);
+                    $scheduledPublish->delete();
+                    continue;
+                }
+
+                if (!$dungeonRoute->hasKilledAllRequiredEnemies()) {
+                    $this->log->publishScheduledDungeonRouteSkippedRequiredEnemiesNotKilled($dungeonRoute->id, $scheduledPublish->id);
                     $scheduledPublish->delete();
                     continue;
                 }
--- a/app/Service/DungeonRoute/DungeonRouteService.php
+++ b/app/Service/DungeonRoute/DungeonRouteService.php
@@ -286,7 +286,8 @@ readonly class DungeonRouteService implements DungeonRouteServiceInterface
                     'dungeonRoute.dungeon',
                     'dungeonRoute.author',
                     'dungeonRoute.mappingVersion.enemies',
-                    'dungeonRoute.killZones.enemies',
+                    // .npc because KillZone::getEnemies() loads it - without it that is a query per kill zone
+                    'dungeonRoute.killZones.enemies.npc',
                 ])
                 ->get();
 
--- a/resources/assets/js/custom/inline/common/dungeonroute/publish.js
+++ b/resources/assets/js/custom/inline/common/dungeonroute/publish.js
@@ -64,8 +64,11 @@ class CommonDungeonroutePublish extends InlineCode {
         let killZoneMapObjectGroup = this._getKillZoneMapObjectGroup();
         if (killZoneMapObjectGroup !== null) {
             this._refreshRequiredEnemiesState();
+            // KillZone.setEnemies() suppresses the per-enemy signals and emits only enemieschanged - that is the
+            // path remote/live-session updates take, so all three are needed to keep the state fresh
             killZoneMapObjectGroup.register('killzone:enemyadded', this, this._refreshRequiredEnemiesState.bind(this));
             killZoneMapObjectGroup.register('killzone:enemyremoved', this, this._refreshRequiredEnemiesState.bind(this));
+            killZoneMapObjectGroup.register('killzone:enemieschanged', this, this._refreshRequiredEnemiesState.bind(this));
         }
     }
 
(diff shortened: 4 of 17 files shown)


resources/assets/js/custom/inline/common/dungeonroute/publish.js
```diff
@@ -59,6 +59,65 @@ class CommonDungeonroutePublish extends InlineCode {
...
+     * @returns {KillZoneMapObjectGroup|null}
+     * @private
+     */
+    _getKillZoneMapObjectGroup() {
+        // The share modal this select lives in is also rendered on pages that have no map at all
+        if (typeof getState !== 'function') {
+            return null;
+        }
+
+        return getState().getDungeonMap()?.mapObjectGroupManager?.getByName(MAP_OBJECT_GROUP_KILLZONE) ?? null;
```
Wotuu: :robot: This null-guard doesn't hold for either of the two "no map" cases it's written for:

1. `getState()` is not undefined on mapless pages — `custom/util.js` defines a fallback `function getState() { return false; }`, so `typeof getState === 'function'` is always true and `getState().getDungeonMap()` throws `TypeError: getState(...).getDungeonMap is not a function`. The established pattern in this repo is `typeof getState === 'function' && getState() !== false` (see `custom/util.js:166`).
2. `MapObjectGroupManager.getByName()` returns `false` (not `undefined`/`null`) when the group isn't registered — `?? null` only normalises `null`/`undefined`, so `false` is returned here, passes the `!== null` checks in `activate()`/`cleanup()`, and `false.register(...)` throws.

Today the publish select is only rendered from `common/maps/controls/header.blade.php` so neither branch is reachable, but the guard doesn't do what its comment claims. A truthiness check (`const group = ...; return group || null;`) plus the `getState() !== false` test covers both.

resources/assets/js/custom/inline/common/dungeonroute/publish.js
```diff
@@ -59,6 +59,65 @@ class CommonDungeonroutePublish extends InlineCode {
         $select.bind('change', function () {
             self._setPublished($(this).val());
         });
+
+        // Only the route editor has a map (and thus kill zones) to check the required enemies against
+        let killZoneMapObjectGroup = this._getKillZoneMapObjectGroup();
+        if (killZoneMapObjectGroup !== null) {
+            this._refreshRequiredEnemiesState();
+            killZoneMapObjectGroup.register('killzone:enemyadded', this, this._refreshRequiredEnemiesState.bind(this));
+            killZoneMapObjectGroup.register('killzone:enemyremoved', this, this._refreshRequiredEnemiesState.bind(this));
```
Wotuu, line 70: :robot: These are the only two refresh hooks, but they miss the bulk path: `KillZone.setEnemies()` sets `_isBulkUpdating` (killzone.js:929-961), which suppresses the per-enemy `killzone:enemyadded`/`killzone:enemyremoved` signals and emits only `killzone:enemieschanged`. `setEnemies()` is what `_setEnemiesFromRemote()` calls, i.e. the initial remote load and every echo/live-session update of a pull's enemies — after such an update the disabled state is stale (options can stay enabled for a route that no longer kills a required enemy, so the user gets a 403 from `publishedState` instead of the tooltip).

Both other consumers of these signals register all three: `killzonessidebar.js:346` and `editkillzoneenemyselection.js:112` use `['killzone:enemyadded', 'killzone:enemyremoved', 'killzone:enemieschanged']`. Adding `killzone:enemieschanged` here (and to `cleanup()`) closes the gap.

resources/assets/js/custom/mapobjectgroups/killzonemapobjectgroup.js
```diff
@@ -383,7 +383,7 @@ class KillZoneMapObjectGroup extends MapObjectGroup {
             // If this enemy SHOULD have been killed by the user
             if (enemy.required &&
                 // If not teeming, OR if enemy is teeming AND we're teeming, or inverse that. THEN this enemy counts, otherwise it does not
-                (enemy.teeming === null || (enemy.teeming === 'visible' && mapContext.getTeeming()) || (enemy.teeming === 'invisible' && !mapContext.getTeeming()))
+                (enemy.teeming === null || (enemy.teeming === 'visible' && mapContext.getTeeming()) || (enemy.teeming === 'hidden' && !mapContext.getTeeming()))
```
Wotuu, line 386: :robot: Good catch on `'invisible'` — but the same typo still lives in `MapObject.shouldBeVisible()`, `resources/assets/js/custom/models/mapobject.js:774`:

```js
else if (mapContext.getTeeming() && this.teeming === 'invisible') {
```

`teeming` is loaded from the base `MapObject` attribute at mapobject.js:122 and only ever holds `null`/`'visible'`/`'hidden'`, so that branch is dead there too. After this PR the two disagree: on a teeming route a `teeming = 'hidden'` required enemy is no longer demanded by `hasKilledAllRequiredEnemies()` (correct), but it is still *rendered* on the map as an unkilled required enemy. Worth fixing in the same PR since it's the identical bug one file over.

app/Service/DungeonRoute/DungeonRouteService.php
```diff
@@ -282,7 +282,12 @@ public function publishScheduledDungeonRoutes(): int
             /** @var Collection<int, DungeonRouteScheduledPublish> $scheduledPublishes */
             $scheduledPublishes = DungeonRouteScheduledPublish::query()
                 ->where('publish_at', '<=', now())
-                ->with(['dungeonRoute.dungeon', 'dungeonRoute.author'])
+                ->with([
+                    'dungeonRoute.dungeon',
+                    'dungeonRoute.author',
+                    'dungeonRoute.mappingVersion.enemies',
+                    'dungeonRoute.killZones.enemies',
```
Wotuu: :robot: N+1: this eager load stops at `killZones.enemies`, but `getKilledEnemyIds()` goes through `KillZone::getEnemies()`, which does `loadMissing('enemies.npc')` (KillZone.php:145). So for every due scheduled publish you get one extra `npcs` query per kill zone inside the cron loop, and the `npc` relation isn't used by the required-enemies check at all.

Either add `'dungeonRoute.killZones.enemies.npc'` here, or (better) have `getKilledEnemyIds()` read `$killZone->enemies` directly instead of `getEnemies()` so it doesn't drag the npc relation along.

app/Models/DungeonRoute/DungeonRoute.php
```diff
@@ -1139,37 +1139,34 @@ public function isOwnedByUser(?User $user = null): bool
...
-        //                if (!$this->isEnemyKilled($enemy->id)) {
-        //                    $result = false;
-        //                    break;
-        //                }
-        //            }
-        //        }
+        $this->loadMissing('mappingVersion.enemies');
 
-        return $result;
+        $requiredEnemies = $this->mappingVersion->enemies
```
Wotuu, line 1161: :robot: `mapping_version_id` being non-null doesn't guarantee the row still exists — this schema has no foreign keys (per CLAUDE.md) and `MappingVersionController::delete()` lets an admin delete a mapping version that dungeon routes still point at. In that case `$this->mappingVersion` is `null` and this line fatals with "Attempt to read property 'enemies' on null" — a 500 on `POST /ajax/{publicKey}/publishedState` and a hard failure of the whole `publishScheduledDungeonRoutes()` cron run rather than a skip.

Cheapest fix: fold it into the existing early return, e.g. `if ($this->mappingVersion === null) { return true; }` after the `loadMissing()` (the `mapping_version_id === null` check above then becomes redundant).

Wotuu:

resources/assets/js/custom/inline/common/dungeonroute/publish.js
Wotuu: :robot: Fixed in 92671f5 — you are right on both counts. Now resolves `getState` through a truthiness check rather than `typeof`, and normalises the group with `||` instead of `??` so `getByName()`'s `false` is caught.

resources/assets/js/custom/inline/common/dungeonroute/publish.js
Wotuu: :robot: Fixed in 92671f5 — `killzone:enemieschanged` is now registered (and unregistered) alongside the other two, matching the sibling consumers.

resources/assets/js/custom/mapobjectgroups/killzonemapobjectgroup.js
Wotuu: :robot: Fixed in 92671f5 — `mapobject.js:774` now checks `'hidden' too, so the map and the publish check agree on which enemies a teeming route actually has.

app/Service/DungeonRoute/DungeonRouteService.php
Wotuu: :robot: Fixed in 92671f5 — the eager load is now `killZones.enemies.npc`.

app/Models/DungeonRoute/DungeonRoute.php
Wotuu: :robot: Fixed in 92671f5 — now null-checks the `mappingVersion` relation after `loadMissing()` instead of the id. Typing the property `MappingVersion|null` also made two pre-existing `@phpstan-ignore identical.alwaysFalse` comments redundant (DungeonRouteThumbnailRepository, ThumbnailService) — both were already null-checking it defensively, which confirms the case is real. Covered by a new test that deletes the mapping version out from under a route.

Wotuu:
:robot: Addressed all 5 cold review findings in 92671f5, plus a flaky-test fix of my own: the new tests built their mapping version with `MappingVersion::create()`, whose `created` hook clones the entire previous mapping into it — dragging the dungeon's seeded (sometimes `required`) enemies along, so the result depended on which random dungeon the factory picked. They now insert it quietly, as `MappingService::copyMappingVersionToDungeon()` does. Verified deterministic over repeated local runs; all 9 checks green.

Wotuu:

app/Models/DungeonRoute/DungeonRoute.php
```diff
@@ -1137,37 +1137,37 @@ public function isOwnedByUser(?User $user = null): bool
...
-        //                if (!$this->isEnemyKilled($enemy->id)) {
-        //                    $result = false;
-        //                    break;
-        //                }
-        //            }
-        //        }
+        // Sandbox/unmigrated routes have no mapping version to compare against - and since this schema carries no
+        // foreign keys, a route can also point at a mapping version that has since been deleted. Nothing can be
+        // required of either, and a hard failure here would 500 the publish endpoint and abort the whole
+        // scheduled-publish cron run.
```
Wotuu: Sandbox routes do have a mapping version. The only difference is that a sandbox route is automatically removed after a day if it isn't claimed by anyone. Before that time, a sandbox route's only security is the URL if you share the URL with anyone else. Any other user can claim the route for their profile, but it still has a mapping version. Otherwise, it wouldn't be able to function, so this comment is wrong in that sense, but the check below is still fine. If the route, for some reason, doesn't have a mapping version, then we need to catch that.

app/Policies/DungeonRoutePolicy.php
```diff
@@ -65,10 +66,14 @@ public function embed(?User $user, DungeonRoute $dungeonroute): Response
 
     /**
      * Determine whether the user can publish dungeon routes.
+     *
+     * @param string|null $publishedState The state the route is being changed to, if known. Unpublishing is always
+     *                                    allowed - the required enemies guard would otherwise trap an already
+     *                                    published route in a state its author can no longer undo.
```
Wotuu: This should perhaps be rephrased a bit. Now it says, "Unpublishing is always allowed," but this isn't true, obviously. It's only allowed when you want to bypass the required enemies check, so it should read something like, "The required enemies check is not executed when you are unpublishing the route." Something like that.

resources/assets/js/custom/inline/common/dungeonroute/publish.js
```diff
@@ -59,6 +59,70 @@ class CommonDungeonroutePublish extends InlineCode {
...
+     */
+    _getKillZoneMapObjectGroup() {
+        // The share modal this select lives in is also rendered on pages that have no map. Both fallbacks on that
+        // path yield `false` rather than a nullish value - util.js defines a no-op getState() and
+        // MapObjectGroupManager.getByName() returns false for an absent group - so normalise with ||, not ??.
+        let state = typeof getState === 'function' ? getState() : null;
+        let dungeonMap = state ? state.getDungeonMap() : null;
+
+        return (dungeonMap?.mapObjectGroupManager?.getByName(MAP_OBJECT_GROUP_KILLZONE)) || null;
+    }
```
Wotuu, line 99: I've seen similar functions pop up, and I'm not entirely sure what to think. I think this function is fine, except that we should maybe put it in a different or better place.
Right now, there are a lot of uncertainties when you want to get the map object group:
- You have to get the state, and the state can be null if you're not in a map context.
- When you get the state, you have to get the dungeon map, which can be null.
- The dungeon map itself should always have a map object group manager, but I guess it also can be null.
- You call the `get by name` function, and the result of that can also be null.
There are a lot of chances for it to go wrong. This is true every time you request a map object group. This is true for the map object group KillZone, but it's also true for enemies, enemy packs, enemy patrols, or map icons. All of these should have a helper function to "safely" get their map object group.
Right now, it's hidden away in an inline common Dungeon Route publish file, but it would be nice if all map object groups had access to a similar helper function. That's one thing.
The second thing is, I've seen this sort of function pop up. It may have been in enemy.js or in any other model. The way that I intended it is that the map object group manager can know about the models themselves. It can reference them, etc., but the other way around, there shouldn't be a dependency per se. By saying this, I know it's not always going to be true. If you have an enemy and you want to attach it to a KillZone, you need to fetch the map object group for the KillZone. There's already knowledge about map object groups being accessed, etc., but the idea is that they don't depend on it.
To the point of my story is that if you have a KillZone object, then inside that object, it should not know what map object group it's part of. You can correct me on this architectural decision if you want to. Maybe there should be a helper function inside KillZone.js which safely gets the map object group that it is part of, but I kind of don't think so. If there are helper functions to safely get a map object group, I think we should define them globally and accessibly for everyone and not hide them away in a file like this.
I know that util.js has functions called `get kill zones` or `get enemies`, etc., and I've historically only used them for getting access to map object groups while in the console of the browser. If we want to reuse those functions across the site, we maybe can, but let's think about this. Maybe we should file a follow-up function issue. Maybe you want to tackle this now. I'll leave that up to you.

resources/assets/js/custom/inline/common/dungeonroute/publish.js
```diff
@@ -59,6 +59,70 @@ class CommonDungeonroutePublish extends InlineCode {
...
+     * A route that does not kill all required enemies may not be published - but it must always remain possible to
+     * _unpublish_ it, so only the publishing options are disabled rather than the select as a whole.
+     * @private
+     */
+    _refreshRequiredEnemiesState() {
+        let hasKilledAllRequiredEnemies = this._getKillZoneMapObjectGroup().hasKilledAllRequiredEnemies();
+        let $select = $(this.options.publishSelector);
+
+        $select.find('option').each((index, option) => {
+            if (option.value === 'unpublished') {
```
Wotuu: Can this magic string be put in constants.js instead? The different publishing states should also live in there. Apparently they don't, but let's add them there and reference them at this point.

resources/assets/js/custom/models/mapobject.test.js
```diff
@@ -103,7 +103,7 @@ function makeFakeMap({sandbox = false, bounds = null} = {}) {
         ],
         teemingOptions: [
             {key: 'visible', description: 'Visible'},
-            {key: 'invisible', description: 'Invisible'},
+            {key: 'hidden', description: 'Hidden'},
```
Wotuu: Can you also move these keys to the constants.js? I know this isn't something that you changed or introduced here, but while we are at it, let's fix it properly.

Wotuu: Oh, so this is the test and not the actual map object. Up to you in that case. Some people argued that the constants should not be referenced during the test, so if you agree, don't do that. I merely thought that we were looking at the map objects' attributes, and in there I do want constants to be referenced.

resources/assets/js/custom/models/mapobject.js
```diff
@@ -771,7 +771,7 @@ class MapObject extends Signalable {
                     return false;
                 }
                 // If the map is teeming, but the enemy shouldn't be there for teeming maps..
-                else if (mapContext.getTeeming() && this.teeming === 'invisible') {
+                else if (mapContext.getTeeming() && this.teeming === 'hidden') {
```
Wotuu: Yeah, so at this point, can you reference the constants.js and have the teaming options listed there?

app/Models/DungeonRoute/DungeonRoute.php
Wotuu: :robot: You're right, sandbox routes always have a mapping version - I corrected the comment. The
actual guard stays (this schema has no foreign keys, so a route can still point at a mapping
version that has since been deleted, and that's what the null-check protects against):

```php
// Sandbox routes always have a mapping version, but this schema carries no foreign keys, so a
// route can point at a mapping version that has since been deleted. Nothing can be required
// against a missing mapping version, and a hard failure here would 500 the publish endpoint
// and abort the whole scheduled-publish cron run.
if ($this->mappingVersion === null) {
    return true;
}
```

app/Policies/DungeonRoutePolicy.php
Wotuu: :robot: Reworded the docblock to describe the actual behavior:

```php
* @param string|null $publishedState The state the route is being changed to, if known. The required enemies
*                                    check is not executed when you are unpublishing the route - otherwise the
*                                    guard would trap an already published route in a state its author can no
*                                    longer undo.
```

resources/assets/js/custom/inline/common/dungeonroute/publish.js
Wotuu: :robot: Agreed this is a real pattern - a grep for `mapObjectGroupManager.getByName(MAP_OBJECT_GROUP_KILLZONE)`
alone turns up ~25 duplicated call sites across the codebase, most without the null-safety this
one needs (since `publish.js` can render on pages that have no map at all, e.g. the share modal).

Centralizing that is out of scope for this fix, so I filed a follow-up:
https://github.com/RaiderIO/keystone.guru/issues/3874

resources/assets/js/custom/inline/common/dungeonroute/publish.js
Wotuu: :robot: Added `PUBLISHED_STATE_UNPUBLISHED`/`TEAM`/`WORLD_WITH_LINK`/`WORLD` to `constants.js`
(mirroring `App\Models\PublishedState`, which didn't have JS-side constants at all yet) and
switched `publish.js` to use them - both the `'unpublished'` check here and the `icons` map
keyed by the other publish-state strings a bit further up.

resources/assets/js/custom/models/mapobject.test.js
Wotuu: :robot: Since `mapobject.js` itself now references `TEEMING_VISIBLE`/`TEEMING_HIDDEN` as bare
globals (see the mapobject.js thread), the test needs those same globals defined to even load the
module - so I added `global.TEEMING_VISIBLE`/`global.TEEMING_HIDDEN` next to the existing
`global.ENEMY_SEASONAL_TYPE_*` assignments this file already uses for the same reason, and
switched every `'visible'`/`'hidden'` teeming literal in the test (the `teemingOptions` fixture
and all the assertion data) to reference them. Ran `npx vitest run` on the file afterwards -
still 34/34 green.

Wotuu: :robot: Ended up making the change anyway (see my reply just above) - but not purely for
"reference constants in tests" style reasons. Once `mapobject.js` switched to the
`TEEMING_VISIBLE`/`TEEMING_HIDDEN` bare globals, the test *had* to define those same globals to
load the module at all (this file already follows that "define the globals the module touches
before require()-ing it" pattern for the `ENEMY_SEASONAL_TYPE_*` constants). Once they existed as
globals, using them in the fixture/assertion data too was the natural next step rather than
keeping a second, disconnected copy of the literal strings. So: not a blanket "tests must
reference constants.js" stance, just following what this file was already doing.

resources/assets/js/custom/models/mapobject.js
Wotuu: :robot: `TEEMING_VISIBLE`/`TEEMING_HIDDEN` already existed in `constants.js` (used by
`admin/adminenemy.js`) - switched both branches here to reference them instead of the
`'visible'`/`'hidden'` string literals. Had to update `mapobject.test.js` to define these as
bare globals too, since it loads `mapobject.js` directly as a script (see that thread).

PR #3856, opened 2026-08-06
#3822 Drop the combat_log_parse_failures table and its model

Wotuu:
:robot: Closes #3822

Contract half of the expand/contract started in #3821. Rows are discarded deliberately — Sentry has carried the same information since #3820, and the data was never recoverable from seeders anyway.

- Add the drop migration for `combat_log_parse_failures` under `database/migrations_combatlog/`.
- Delete `app/Models/CombatLog/CombatLogParseFailure.php`.
- Delete `database/factories/CombatLog/CombatLogParseFailureFactory.php`.
- The repository test file named in the issue (`tests/Feature/App/Repository/CombatLog/CombatLogParseFailureRepositoryTest.php`) no longer exists — it was already removed in #3821.

**Holding this in draft**: #3822 must not ship in the same release as #3821 (per `.claude/CLAUDE.md` expand/contract rule — see #3497). #3821's MR (#3844) merged to master 2026-08-06 08:22, but the latest cut release (v15.9.2) predates that merge, so #3821 hasn't actually shipped yet. This MR will be marked ready for review only once a release containing #3821 has gone out.

## Verification

- `php artisan test --compact --group=CombatLog` — 303 passed.
- `php artisan migrate --database=combatlog --path=database/migrations_combatlog` on a fresh isolated worktree DB applies cleanly.
- `composer run fix` — no changes needed.
- `composer run analyse` — no errors.

PR #3861, opened 2026-08-06
#3567 Switch thumbnail reads from custom to variant

### LINKED ISSUE #3567: Drop the legacy `custom` column from dungeon_route_thumbnails (contract phase for #3447)
:robot: Follow-up to #3447 (PR that reworked thumbnails into variants).

#3447 made `variant` the source of truth for thumbnail kind (`standard`/`hero`/`custom`) but, to stay deploy-safe (expand/contract), it kept the pre-existing `custom` boolean column: it is **dual-written** and reads still use it during the deploy window (old containers create custom rows as `custom=1, variant='standard'`, so `variant` alone isn't reliable mid-deploy).

Once #3447 is released, this contract-phase ticket should:
- Switch the remaining `custom`-based reads to `variant` (`ThumbnailService::hasThumbnailsGenerated`/`copyThumbnails`, `DungeonRoute::thumbnails()`/`heroThumbnails()`, `hasFreshThumbnailForVariant`).
- Drop the dual-write of `custom` in `ThumbnailService::attachThumbnailToDungeonRoute`.
- Add a migration dropping the `custom` column.
- Remove the `custom` `@property`/fillable from the model.

Wotuu:
:robot: Refs #3567

Phase 1 of 2 for the #3567 contract phase (#3447 introduced `variant` as the new source of truth for thumbnail kind, keeping the legacy `custom` boolean dual-written for deploy safety).

A cold review caught that dropping `custom` in the same release that removes its last code readers is an expand/contract violation (independent migrate cron vs. ECS rollout race - the #3497 failure mode), so this PR only does the safe half:

## Summary
- Switched the remaining `custom`-based reads to `variant`: `ThumbnailService::hasThumbnailsGenerated`/`copyThumbnails`, `DungeonRoute::thumbnails()`/`heroThumbnails()`, and `RepairBrokenThumbnails`'s custom-thumbnail detection.
- Left the `custom` column, its dual-write in `ThumbnailService::attachThumbnailToDungeonRoute`, and the model's fillable/`@property` untouched - a follow-up PR will drop those once a release containing this PR has actually deployed.

## Test plan
- [x] `php artisan test --compact` on the affected thumbnail test files (46 passed)
- [x] `composer run analyse` (PHPStan, no errors)
- [x] `composer run fix` (no changes needed)
- [x] Confirmed the currently-failing `DungeonRouteControllerSaveNewTest`/`DungeonRouteControllerSaveNewTemporaryTest` `ModelNotFoundException` failures are pre-existing on master (reproduced with this change stashed), unrelated to this MR.

Cold review: `:robot: Cold review (fable): 2 findings posted.` - both addressed by splitting the migration out; threads resolved.


database/migrations/2026_08_06_000000_drop_custom_from_dungeon_route_thumbnails_table.php
```diff
@@ -0,0 +1,28 @@
...
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration {
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::table('dungeon_route_thumbnails', static function (Blueprint $table): void {
+            $table->dropIndex('dungeon_route_thumbnails_custom_index');
```
Wotuu: :robot: **Expand/contract violation — this drop must ship one release later than the code changes.** The currently-deployed release (v15.9.0, master at `50b6838dd`) still *reads* `dungeon_route_thumbnails.custom` in `DungeonRoute::thumbnails()`/`heroThumbnails()` and `ThumbnailService::hasThumbnailsGenerated()`/`copyThumbnails()`, and still *writes* it in `attachThumbnailToDungeonRoute()` (`'custom' => $isCustom`). Deploys are not atomic (the migrate cron runs independently of the ECS rollout — see the #3497 incident), so when the release containing this PR deploys, this migration can drop the column while old containers are still serving: every page eager-loading `thumbnails()` 500s with `Unknown column 'custom'`, and every thumbnail insert fails. Per the repo migration convention, release N removes the last code consumer while leaving the schema in place; release N+1 ships the drop. Suggest removing this migration from this PR and shipping it in a follow-up PR that only merges after a release containing the code-side changes has been deployed (same hold-in-draft pattern as #3856).

database/migrations/2026_08_06_000000_drop_custom_from_dungeon_route_thumbnails_table.php
```diff
@@ -0,0 +1,28 @@
...
+        });
+    }
+
+    /**
+     * Reverse the migrations.
+     */
+    public function down(): void
+    {
+        Schema::table('dungeon_route_thumbnails', static function (Blueprint $table): void {
+            $table->boolean('custom')->default(false)->after('file_id')->index();
```
Wotuu: :robot: Minor: `down()` re-adds the column with `default(false)` but does not backfill `custom = true` for `variant = 'custom'` rows, so after a rollback all custom thumbnails are misflagged as non-custom — pre-#3861 code would then treat them as regenerable: `attachThumbnailToDungeonRoute()` deletes them on the next standard render pass and `copyThumbnails()` copies them on route clone. Since the backfill is trivially derivable from `variant`, add after the column re-add: `DB::table('dungeon_route_thumbnails')->where('variant', 'custom')->update(['custom' => true]);` (mirrors the up() of the 2026_07_16 backfill migration).

Wotuu:

database/migrations/2026_08_06_000000_drop_custom_from_dungeon_route_thumbnails_table.php
Wotuu: :robot: Agreed — I split the migration (and the dual-write removal / model cleanup that depends on it) out of this PR entirely. This PR now only switches reads from `custom` to `variant`; the column and its dual-write stay in place. The column drop will go out in a follow-up PR that I'll open once this one has actually deployed.

database/migrations/2026_08_06_000000_drop_custom_from_dungeon_route_thumbnails_table.php
Wotuu: :robot: Moot for now — the migration moved out of this PR (see reply on the other thread), but I'll fold this backfill into the follow-up PR's `down()` when I add the drop migration back.

PR #3862, opened 2026-08-06
#3650 Batch per-pull enemy-forces queries on discover leaderboard cards

### LINKED ISSUE #3650: Batch per-pull enemy-forces queries on discover leaderboard cards (follow-up to #3447)

Wotuu:
:robot: Closes #3650

## Summary
- Follow-up to the N+1 flagged in #3447's cold review. `getEnemyForcesPerKillZone()` fired one raw
  aggregate query per rendered card (up to 18 leaderboard rows + up to 3 hero cards).
- Adds a batched `getEnemyForcesPerKillZoneForRoutes()` repository method that queries all routes
  on a page at once via a `dungeon_routes.id IN (...)` clause. Shrouded and non-shrouded routes are
  batched into two separate queries (the shrouded enemy-forces expression differs), so a page fires
  at most 2 forces queries regardless of card count. The original single-route method now delegates
  to the batched one (single source of truth for the SQL).
- Adds `DungeonRouteEnemyForcesPageResolver`, a small page-scoped, lazily-memoizing resolver: it
  only fires the batched query on the first cache miss encountered while rendering the page, so a
  fully-cached page still fires zero forces queries (preserving the existing "cache hit pays no
  container-resolution cost" invariant called out in `cardrow.blade.php`/`cardhero.blade.php`).
- `cardrow.blade.php` and `cardhero.blade.php` accept an optional `pullForcesResolver`; when absent
  they fall back to the old per-route lookup (kept for the existing direct-render tests).
  `leaderboard.blade.php` and `overview.blade.php` build one shared resolver for the whole page
  (hero band + leaderboard rows together).

## Test plan
- [x] `getEnemyForcesPerKillZoneForRoutes_givenRoutesWithKillZones_matchesPerRouteResults` -
  differential test against real seeded data confirming the batched query produces identical
  per-pull results to the original per-route query.
- [x] `discoverDungeon_givenReworkFlagActiveAndMultipleRoutes_batchesEnemyForcesQueryAcrossCards` -
  controller-level regression asserting exactly 1 forces query fires for a 5-route cold-cache page
  (previously would have fired one per card).
- [x] Existing `CardRowTest`/`CardHeroTest`/`PullGraphTest`/`DungeonRouteRepositoryTest`/
  `DungeonRouteDiscoverCategoryTest` suites all green.

### DIFF
--- a/app/Repositories/Database/DungeonRoute/DungeonRouteRepository.php
+++ b/app/Repositories/Database/DungeonRoute/DungeonRouteRepository.php
@@ -218,11 +218,39 @@ class DungeonRouteRepository extends DatabaseRepository implements DungeonRouteR
 
     public function getEnemyForcesPerKillZone(DungeonRoute $dungeonRoute): Collection
     {
-        if (!$dungeonRoute->exists) {
+        return $this->getEnemyForcesPerKillZoneForRoutes(collect([$dungeonRoute]))->get($dungeonRoute->id, collect());
+    }
+
+    /**
+     * Batched variant of {@see self::getEnemyForcesPerKillZone()} - fetches the per-pull enemy forces
+     * for an entire collection of routes (e.g. a leaderboard page) in O(1) queries instead of one
+     * query per route, keyed by dungeon route id.
+     *
+     * @param  Collection<int, DungeonRoute>                         $dungeonRoutes
+     * @return Collection<int, Collection<int, KillZoneEnemyForces>>
+     */
+    public function getEnemyForcesPerKillZoneForRoutes(Collection $dungeonRoutes): Collection
+    {
+        $existingRoutes = $dungeonRoutes->filter(static fn(DungeonRoute $dungeonRoute) => $dungeonRoute->exists);
+
+        // The shrouded enemy-forces expression differs from the non-shrouded one (it reads from
+        // mapping_versions.enemy_forces_shrouded* instead of npc_enemy_forces), so shrouded and
+        // non-shrouded routes are batched into two separate queries rather than one per route.
+        $isShrouded = static fn(DungeonRoute $dungeonRoute) => $dungeonRoute->getSeasonalAffix()?->key === Affix::AFFIX_SHROUDED;
+
+        return $this->queryEnemyForcesPerKillZoneForRouteIds($existingRoutes->reject($isShrouded)->pluck('id'), false)
+            ->union($this->queryEnemyForcesPerKillZoneForRouteIds($existingRoutes->filter($isShrouded)->pluck('id'), true));
+    }
+
+    /**
+     * @param  Collection<int, int>                                  $dungeonRouteIds
+     * @return Collection<int, Collection<int, KillZoneEnemyForces>>
+     */
+    private function queryEnemyForcesPerKillZoneForRouteIds(Collection $dungeonRouteIds, bool $isShrouded): Collection
+    {
+        if ($dungeonRouteIds->isEmpty()) {
             return collect();
         }
-
-        $isShrouded = $dungeonRoute->getSeasonalAffix()?->key === Affix::AFFIX_SHROUDED;
 
         // Ignore the shrouded query if we're not shrouded (make it fail)
         $ifIsShroudedEnemyForcesQuery = $isShrouded ? '
@@ -247,19 +275,25 @@ class DungeonRouteRepository extends DatabaseRepository implements DungeonRouteR
             NpcClassification::ALL[NpcClassification::NPC_CLASSIFICATION_FINAL_BOSS],
         );
 
+        $dungeonRouteIdList = implode(',', $dungeonRouteIds->map(static fn($id) => (int)$id)->all());
+
         // Mirrors DungeonRoute::getEnemyForces()'s accounting, scoped per kill zone instead of summed for
-        // the whole route: the inner query groups by (kill zone, enemy identity) to cancel join fan-out
-        // per enemy (the `/ COUNT(...)` normalization - an enemy matching >1 npc_enemy_forces/enemies row
-        // must not be counted more than once), then the outer query sums those per-identity values back up
-        // per kill zone. The `enemies.mapping_version_id = dungeon_routes.mapping_version_id` WHERE filter
-        // (also mirrored from getEnemyForces()) turns the `enemies` LEFT JOIN into an effective INNER JOIN,
-        // dropping kill_zone_enemies rows with no matching enemies row for the route's mapping version.
+        // the whole route: the inner query groups by (route, kill zone, enemy identity) to cancel join
+        // fan-out per enemy (the `/ COUNT(...)` normalization - an enemy matching >1 npc_enemy_forces/enemies
+        // row must not be counted more than once), then the outer query sums those per-identity values back
+        // up per route and kill zone. The `enemies.mapping_version_id = dungeon_routes.mapping_version_id`
+        // WHERE filter (also mirrored from getEnemyForces()) turns the `enemies` LEFT JOIN into an effective
+        // INNER JOIN, dropping kill_zone_enemies rows with no matching enemies row for the route's mapping
+        // version. `dungeon_routes`.`id` is repeated in the inner GROUP BY (functionally redundant given
+        // `kill_zones`.`id`) purely so ONLY_FULL_GROUP_BY doesn't reject it through the LEFT JOIN chain.
         $queryResult = DB::select("
-            select `per_identity`.`kill_zone_index` as `index`,
+            select `per_identity`.`dungeon_route_id` as `dungeon_route_id`,
+                   `per_identity`.`kill_zone_index` as `index`,
                    MAX(`per_identity`.`has_boss`) as has_boss,
                    SUM(`per_identity`.`enemy_forces`) as enemy_forces
             from (
-                select `kill_zones`.`id` as `kill_zone_id`,
+                select `dungeon_routes`.`id` as `dungeon_route_id`,
+                       `kill_zones`.`id` as `kill_zone_id`,
                        `kill_zones`.`index` as `kill_zone_index`,
                        MAX(IF(`npcs`.`classification_id` IN ({$bossClassificationIds}), 1, 0)) as has_boss,
                        CAST(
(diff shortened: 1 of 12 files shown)


app/Repositories/Database/DungeonRoute/DungeonRouteRepository.php
```diff
@@ -218,12 +218,40 @@ public function findSimilarRoutes(DungeonRoute $dungeonRoute, int $limit = 5): C
...
+    {
+        $existingRoutes = $dungeonRoutes->filter(static fn(DungeonRoute $dungeonRoute) => $dungeonRoute->exists);
+
+        // The shrouded enemy-forces expression differs from the non-shrouded one (it reads from
+        // mapping_versions.enemy_forces_shrouded* instead of npc_enemy_forces), so shrouded and
+        // non-shrouded routes are batched into two separate queries rather than one per route.
+        $isShrouded = static fn(DungeonRoute $dungeonRoute) => $dungeonRoute->getSeasonalAffix()?->key === Affix::AFFIX_SHROUDED;
+
+        return $this->queryEnemyForcesPerKillZoneForRouteIds($existingRoutes->reject($isShrouded)->pluck('id'), false)
+            ->union($this->queryEnemyForcesPerKillZoneForRouteIds($existingRoutes->filter($isShrouded)->pluck('id'), true));
```
Wotuu: :robot: `reject($isShrouded)` and `filter($isShrouded)` each evaluate the predicate for **every** route in `$existingRoutes`, so `getSeasonalAffix()` runs twice per route on every page render.

That isn't a cheap accessor: for a current-season TWW route it takes the Xal'Atath's Guile branch and calls `$affixGroup->load(['affixGroupCouplings'])`, which is an unconditional re-query (`load()`, not `loadMissing()`) against a per-route `AffixGroup` instance. So the batching work itself scales as `2 × N` `getSeasonalAffix()` calls.

`partition()` evaluates the predicate exactly once per item and gives you both halves:

```php
[$shroudedRoutes, $nonShroudedRoutes] = $existingRoutes->partition($isShrouded);

return $this->queryEnemyForcesPerKillZoneForRouteIds($nonShroudedRoutes->pluck('id'), false)
    ->union($this->queryEnemyForcesPerKillZoneForRouteIds($shroudedRoutes->pluck('id'), true));
```

Related: the docblock above claims "O(1) queries instead of one query per route". That holds for the aggregate SQL, but the shrouded classification is still O(N) relation work, so the claim is a bit stronger than what the method delivers — worth softening to "a constant number of aggregate queries".

Trade-off note (flagging, not asking for a redesign): because `DungeonRouteEnemyForcesPageResolver` batches the *whole* page collection on the first cache miss, a page where only one card misses now pays this classification for all ~21 routes instead of 1. That's inherent to not knowing which cards will miss ahead of time; halving it via `partition()` is the cheap mitigation.

tests/Feature/App/Repository/DungeonRoute/DungeonRouteRepositoryTest.php
```diff
@@ -174,4 +175,58 @@ public function getEnemyForcesPerKillZone_givenRouteWithNoKillZones_returnsEmpty
...
+
+            // Assert
+            $this->assertTrue($result->isEmpty());
+        } finally {
+            $dungeonRoutes->each(fn(DungeonRoute $dungeonRoute) => $dungeonRoute->delete());
+        }
+    }
+
+    #[Test]
+    public function getEnemyForcesPerKillZoneForRoutes_givenRoutesWithKillZones_matchesPerRouteResults(): void
```
Wotuu, line 208: :robot: The shrouded half of the new batched path has no test coverage: `queryEnemyForcesPerKillZoneForRouteIds($ids, isShrouded: true)` is never exercised with a **non-empty** id list, and `union()` is never exercised with two non-empty sides.

That's exactly the branch where the shrouded/non-shrouded split could drop or duplicate a route, and it's the only part of the new SQL the differential test doesn't reach (the `mapping_versions` join plus the shrouded forces expression now run under a multi-row `IN (...)` instead of `= :id`). A test that mixes at least one shrouded route with non-shrouded ones and asserts both are present in the merged result would close it.

Same gap on the controller side: `discoverDungeon_givenReworkFlagActiveAndMultipleRoutes_batchesEnemyForcesQueryAcrossCards` asserts the query *count* but nothing about rendered correctness — a resolver that returned `collect()` for every route would still fire exactly 1 query and return 200. An assertion that the rendered page contains the pull-forces bar markup would make that test failure-sensitive to a threading regression.

tests/Feature/App/Repository/DungeonRoute/DungeonRouteRepositoryTest.php
```diff
@@ -174,4 +175,58 @@ public function getEnemyForcesPerKillZone_givenRouteWithNoKillZones_returnsEmpty
...
+    {
+        // Arrange - a handful of seeded routes that already carry kill zones/enemies, batched together
+        $dungeonRoutes = DungeonRoute::query()
+            ->has('killZones')
+            ->with(['affixes', 'season.expansion'])
+            ->limit(5)
+            ->get();
+
+        $this->assertNotEmpty($dungeonRoutes, 'Expected some seeded routes with kill zones');
```
Wotuu, line 216: :robot: This differential test can pass vacuously. `->has('killZones')` guarantees the routes have kill zones, but not that any `kill_zone_enemies` row resolves to an `enemies` row on the route's own `mapping_version_id` — and that WHERE filter is what makes the query return rows at all. If all five seeded routes yield zero rows, every assertion degenerates to `assertEquals([], [])` and the test proves nothing about the batched-vs-per-route equivalence.

Worth asserting that at least one route produced a non-empty forces collection, e.g. `$this->assertTrue($batchedResult->contains(fn(Collection $forces) => $forces->isNotEmpty()), 'Expected at least one route to produce per-pull forces');` before the per-route loop.

resources/views/dungeonroute/discover/dungeon/overview.blade.php
```diff
@@ -58,6 +60,12 @@
         // The routes already promoted into the hero band are excluded from the leaderboard below.
         $heroRouteIds = collect();
         $startRank    = ($page - 1) * $perPage + 1;
+        // Batches the per-pull enemy forces query across every card on this page - hero band and
+        // leaderboard rows alike - into a single grouped query instead of one query per card.
+        $pullForcesResolver = new DungeonRouteEnemyForcesPageResolver(
+            app(DungeonRouteKillZoneServiceInterface::class),
+            $weeklyRoutes->pluck('dungeonRoute')->merge($popularItems)->unique('id')->values(),
```
Wotuu: :robot: The resolver collection unconditionally merges `$weeklyRoutes`, but the hero band that renders those routes is inside `@if($page === 1)` further down. On page 2+ the weekly routes are never rendered, yet they're still classified by `getSeasonalAffix()` (twice each — see the repository comment) and padded into the batched query's `IN (...)` list for nothing.

Scoping the merge to page 1 keeps the batch to what the page actually renders:

```php
$pullForcesResolver = new DungeonRouteEnemyForcesPageResolver(
    app(DungeonRouteKillZoneServiceInterface::class),
    ($page === 1 ? $weeklyRoutes->pluck('dungeonRoute') : collect())->merge($popularItems)->unique('id')->values(),
);
```

Wotuu:

app/Repositories/Database/DungeonRoute/DungeonRouteRepository.php
Wotuu: :robot: Fixed in 910bfcad6: `getEnemyForcesPerKillZoneForRoutes()` now uses `Collection::partition()` instead of separate `reject()`/`filter()` calls, so the shrouded predicate (and its `getSeasonalAffix()`/`AffixGroup::load()` cost) runs exactly once per route instead of twice. Also softened the docblock claim to describe a constant number of aggregate queries rather than implying the classification itself is O(1).

tests/Feature/App/Repository/DungeonRoute/DungeonRouteRepositoryTest.php
Wotuu: :robot: Fixed in 910bfcad6: added `getEnemyForcesPerKillZoneForRoutes_givenShroudedAndNonShroudedRoutes_matchesPerRouteResultsForBoth` in `tests/Feature/App/Repository/DungeonRoute/DungeonRouteRepositoryTest.php`, which splits the seeded routes into shrouded/non-shrouded buckets, asserts both buckets are non-empty, batches them together in one call (exercising `union()` with two non-empty sides), and checks every route still matches its single-route result.

tests/Feature/App/Repository/DungeonRoute/DungeonRouteRepositoryTest.php
Wotuu: :robot: Fixed in 910bfcad6: `getEnemyForcesPerKillZoneForRoutes_givenRoutesWithKillZones_matchesPerRouteResults` now tracks `$sawNonZeroForces`/`$sawBoss` while iterating the single-route comparison and asserts both are true at the end, so the test fails if the seeded data (and therefore the batched query) degenerates to empty results on both sides instead of passing vacuously.

resources/views/dungeonroute/discover/dungeon/overview.blade.php
Wotuu: :robot: Fixed in 910bfcad6: the resolver collection in `overview.blade.php` now only merges `$weeklyRoutes` when `$page === 1` (`($page === 1 ? $weeklyRoutes->pluck(...) : collect())->merge($popularItems)...`), matching the `@if($page === 1)` guard around the hero band that actually renders them. Page 2+ no longer classifies or batches weekly routes that are thrown away.

Issue #3673, opened 2026-07-25
Audit sweep: write paths reachable without a Gate::authorize


Wotuu:
:robot: **Narrowed 2026-08-07.** This issue originally proposed four searches over `app/` for defects that only reading finds. The sweep ran on 2026-08-06 ([coverage comment](https://github.com/RaiderIO/keystone.guru/issues/3673#issuecomment-5208343386)) and three and a half of them are closed for good. What is left is one check, and it is worth keeping because it is the only one that actually yielded findings — twice.

## The check

**A route-bound model parameter reaching a write path with no `Gate::authorize`.**

PHPStan cannot see this. A `Policy::foo(): return true`, or a controller that simply never calls the gate, is *semantically* always-true while every type in sight is honest — so `identical.alwaysTrue` / `notIdentical.alwaysTrue` / `deadCode.unreachable` (all live at level 6 with no baseline) have nothing to fire on.

Enumerate it mechanically rather than by reading:

```
php artisan route:list --path=ajax --json   # filter: non-GET, middleware lacks Authenticate
```

then grep each target method for `Gate::authorize`.

The `ajax` route group deliberately has no `auth` middleware — guests must be able to edit sandbox routes — so authorization is per-object and the **gate** is what matters, not the middleware. Absence of `auth` is therefore not the finding; absence of `Gate::authorize` on a write path is.

This pass found both of the 2026-08-06 findings, and previously #3663.

## Retired — do not re-run

| Search | Why it is closed |
|---|---|
| Dev-only imports (`Mockery\`, `PHPUnit\`, `Faker\`, `Tests\`) in `app/` | Exactly 2 hits, both were #3664. Worth a one-line grep if ever suspected again, but there is nothing left to sweep. |
| Unreachable `catch` clauses | Closed **by CI**. `catch.neverThrown` is live at level 6 with no baseline; the lone inline `@phpstan-ignore` at `LiveSessionController.php:130` proves it fires and is itself a correct ignore. Zero empty `catch` blocks in `app/`. |
[...]

PR #3864, opened 2026-08-06
Authorize tag deletion against TagPolicy

Wotuu:
:robot: `AjaxTagController::delete()` was the only write method in the class that did not consult `TagPolicy` before acting on the route-bound `$tag`. Its two siblings do:

- `updateAll()` — `Gate::authorize('edit', $tag)`
- `deleteAll()` — `Gate::authorize('delete', $tag)`
- `delete()` — nothing

This adds the missing `Gate::authorize('delete', $tag)`, matching `deleteAll()`. `TagPolicy::delete()` already existed and already delegates to `TagPolicy::edit()`, so no policy changes were needed — the ability was simply never invoked from this route.

The `ajax` route group deliberately carries no `auth` middleware, because sandbox routes have to work for guests; authorization there is done per object rather than by middleware. That makes the per-object gate the only check on this endpoint, not a redundant second one.

## No behaviour change for legitimate use

Tags reaching this endpoint are created by `AjaxTagController::store()`, which sets `context_class` to either `User::class` or `Team::class`. Both are handled by `TagPolicy::edit()` — the owner passes via the first branch, a team member via the second. The tag manager (`tabletags.js:153`) is the only caller.

## Tests

New `tests/Feature/Controller/Ajax/AjaxTagControllerTest.php`:

| Test | Asserts |
|---|---|
| `delete_givenOwnTag_deletesIt` | 204, row gone |
| `delete_givenAnotherUsersTag_returnsForbidden` | 403, row still present |
| `delete_givenGuest_returnsForbidden` | 403, row still present |

Verified the tests actually pin the fix: with the `Gate::authorize` line stashed, the two forbidden tests fail and the happy-path test still passes.

Found while working #3673.

### DIFF
--- a/app/Http/Controllers/Ajax/AjaxTagController.php
+++ b/app/Http/Controllers/Ajax/AjaxTagController.php
@@ -154,10 +154,13 @@ class AjaxTagController extends Controller
     /**
      * @return array<string, mixed>|ResponseFactory|Response
      *
+     * @throws AuthorizationException
      * @throws Exception
      */
     public function delete(Request $request, Tag $tag): array|ResponseFactory|Response
     {
+        Gate::authorize('delete', $tag);
+
         if ($tag->delete()) {
             $result = response()->noContent();
         } else {
--- a/app/Models/Tags/Tag.php
+++ b/app/Models/Tags/Tag.php
@@ -19,8 +19,8 @@ use Illuminate\Support\Collection;
  * @property int         $context_id
  * @property string      $context_class
  * @property int         $tag_category_id
- * @property int         $model_id
- * @property string      $model_class
+ * @property int|null    $model_id
+ * @property string|null $model_class
  * @property string      $name
  * @property string|null $color
  *
--- a/app/Policies/TagPolicy.php
+++ b/app/Policies/TagPolicy.php
@@ -24,18 +24,25 @@ class TagPolicy
         if ($tag->context_class === User::class && $tag->context_id === $user->id) {
             $result = true;
         } elseif ($tag->context_class === Team::class) {
-            // If we're editing a team tag, and the user is part of this team, we can edit it
-            if (Team::findOrFail($tag->context_id)->isUserMember($user)) {
-                $result = true;
-            }
-        } elseif ($tag->model_id !== null) {
+            // If we're editing a team tag, and the user is part of this team, we can edit it.
+            // find() rather than findOrFail(): a team tag can outlive its team. Team::removeMember()
+            // unassigns a leaving author's routes without dropping their team tags, and the team's
+            // deleting hook only clears tags for routes still assigned to it - so context_id can
+            // dangle, and findOrFail() would 404 every caller instead of denying them.
+            $result = Team::find($tag->context_id)?->isUserMember($user) ?? false;
+        }
+
+        // Falling back to the tagged route's own permissions keeps a tag whose team membership no
+        // longer holds - or whose team is gone entirely - manageable by the people who own the route
+        // it is stuck on, rather than leaving it undeletable by anyone
+        if (!$result && $tag->model_id !== null) {
             switch ($tag->tagCategory->name) {
                 case TagCategory::DUNGEON_ROUTE_PERSONAL:
                 case TagCategory::DUNGEON_ROUTE_TEAM:
-                    /** @var DungeonRoute $dungeonRoute */
+                    /** @var DungeonRoute|null $dungeonRoute */
                     $dungeonRoute = $tag->model;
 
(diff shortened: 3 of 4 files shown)


Wotuu:
:robot: Cold review (fable): no findings. Verified the gate matches deleteAll(), policy auto-discovery resolves TagPolicy for App\Models\Tags\Tag, guests are denied via the non-nullable User type-hint, store() only creates User/Team-context tags (no legitimate-use regression), and the tests correctly avoid AjaxPublicTestCase (which force-authenticates user 1 and would defeat the guest test).

app/Http/Controllers/Ajax/AjaxTagController.php
```diff
@@ -154,10 +154,13 @@ public function deleteAll(Request $request, Tag $tag): Response
     /**
      * @return array<string, mixed>|ResponseFactory|Response
      *
+     * @throws AuthorizationException
      * @throws Exception
      */
     public function delete(Request $request, Tag $tag): array|ResponseFactory|Response
     {
+        Gate::authorize('delete', $tag);
```
Wotuu, line 162: :robot: The gate itself is correct and consistent with `deleteAll()`, but it newly exposes this endpoint to two pre-existing holes in `TagPolicy::edit()` that the ungated `delete()` used to paper over — and `delete()` is the primary UI path (the tag chip in `tabletags.js:147` calls `ajax.tag.delete`), so the blast radius grows with this PR:

1. **Ex-member's leftover team tags become undeletable (403 instead of 204).** `Team::removeMember()` (`app/Models/Team.php:375`) unassigns the leaving author's routes via a raw `update(['team_id' => null])` *without* deleting their `DUNGEON_ROUTE_TEAM` tags (unlike `Team::removeRoute()` at `:161`, which does). Those tags keep `context_class = Team::class`, so `TagPolicy::edit()` requires current team membership — which the route author no longer has — and the `model_id`/`mayUserEdit` fallback branch is unreachable because it's an `elseif` on the context class. There is no `Gate::before` admin bypass either, so the tag is permanently stuck on the author's own route.

2. **Dangling `context_id` turns delete into a 404/exception.** If that team is later deleted, the `deleting` hook (`app/Models/Team.php:480`) only removes team tags for routes *still on the team*, so the leftover tag's `context_id` dangles. `TagPolicy::edit()` then hits `Team::findOrFail($tag->context_id)` and throws `ModelNotFoundException` (rendered 404) for every authenticated caller — a previously-working delete now can't succeed for anyone.

Both are latent in the policy already (`deleteAll()`/`updateAll()` share them), so I wouldn't drop the gate — but consider hardening `TagPolicy::edit()` in this PR: `Team::find($tag->context_id)` instead of `findOrFail`, and let a `Team`-context tag fall through to the `mayUserEdit($user)` check on the tagged route when the team is gone or the user isn't a member. A test for the ex-member/orphaned-team shapes would pin it.

Wotuu:

app/Http/Controllers/Ajax/AjaxTagController.php
Wotuu: :robot: Confirmed against the source and fixed in 751f9de66 — both holes were real, and the second one is worse than a denial, so thanks for catching it before this shipped.

Verified: `removeRoute()` deletes team tags at `Team.php:161` while `removeMember()` only does `update(['team_id' => null])`; the `deleting` hook filters on `$team->dungeonRoutes`, which no longer contains a route that `removeMember()` already unassigned — so the tag survives with a dangling `context_id`.

`TagPolicy::edit()` now uses `Team::find()` and falls through to the tagged route's own `mayUserEdit()` when the team is gone or membership no longer holds. One thing worth noting on the "unreachable `elseif`" framing: it was unreachable for `Team`-context tags specifically, but a `User`-context tag not owned by the caller already fell through to that same `mayUserEdit()` check, so the fallback widens only the `Team` branch and leaves the rest of the matrix as it was.

Two new tests pin it — a team tag stranded on the author's own route after they leave, and one whose team has been deleted. Both fail with the policy change stashed.

Also fixed `Tag::$model_id`/`$model_class`, which were annotated non-nullable while both columns are nullable; that made the new null guard read as `notIdentical.alwaysTrue` to PHPStan. Master's version escaped the same rule only because its check was the last condition in an if/elseif chain, which `reportAlwaysTrueInLastCondition` exempts by default.

Not fixed here: `Team::removeMember()` should probably drop team tags the way `removeRoute()` does. That is the root cause, but it is a data-mutation change that does nothing for tags already orphaned, so I have kept it out of an authorization PR — happy to file it as a follow-up.

Wotuu:

PR #3865, opened 2026-08-06
Authorize profile update and updatePrivacy against UserPolicy

Wotuu:
:robot: `ProfileController::update()` and `::updatePrivacy()` both act on the route-bound `{user}` — implicit binding by id — without checking that it is the acting user's own account.

`UserPolicy::update()` already exists for exactly this, and says so:

```php
/** Determine whether the user can update the target user's profile settings. */
public function update(User $user, User $target): bool
{
    // A user may only ever update their own settings
    return $user->id === $target->id;
}
```

It was never called from these two routes. The only `Gate::authorize('update', $user)` in `app/` was in `AjaxUserController`, a different endpoint. `tests/Feature/Policy/UserPolicyTest.php` covers the ability thoroughly in isolation, which is why the gap did not show up as a failing test.

This adds `Gate::authorize('update', $user)` to both methods. `ProfileController::delete()` already did the equivalent (`Gate::authorize('delete', $user)`), as does `TeamController` on `edit`/`update`/`delete`, so this brings the two outliers in line rather than introducing a new pattern.

`ProfileFormRequest::authorize()` kept its `return true` — the project does authorization in controllers, and `profile.updateprivacy` takes a plain `Request`, so the check has to live in the controller to cover both routes with one mechanism. The stale `//return Auth::user()?->hasRole(Role::ROLE_ALL) ?? false;` line underneath it was removed and replaced with a note pointing at the controller gate, so the next reader does not restore a check that was never right for this route (`ROLE_ALL` is not owner-scoping).

`updateCreatorProfile()` was checked and needs no change — it resolves the user from `Auth::user()` and takes no route parameter.

## Tests

New `tests/Feature/Controller/ProfileControllerTest.php`:

| Test | Asserts |
|---|---|
| `update_givenSelf_updatesTheProfile` | redirect, fields persisted |
| `update_givenAnotherUser_returnsForbidden` | 403, victim's email and timezone unchanged |
| `updatePrivacy_givenSelf_updatesTheSetting` | redirect, setting persisted |
| `updatePrivacy_givenAnotherUser_returnsForbidden` | 403, victim's setting unchanged |

Two things worth knowing if you touch this test:

- The helper attaches `Role::ROLE_USER`. The profile routes sit behind `role:user|admin`, so a plain `User::factory()->create()` would be turned away by that middleware and the forbidden assertions would pass for the wrong reason.
- Baselines are read after `refresh()`. `analytics_cookie_opt_out` has a schema default, so the in-memory value straight after `create()` is `null` while the persisted value is `0`.

Verified the tests pin the fix: with the two `Gate::authorize` lines stashed, both forbidden tests fail and both happy-path tests still pass.

Found while working #3673.

### DIFF
--- a/app/Http/Controllers/ProfileController.php
+++ b/app/Http/Controllers/ProfileController.php
@@ -143,6 +143,8 @@ class ProfileController extends Controller
         User                          $user,
         ReverbHttpApiServiceInterface $reverbHttpApiService,
     ): RedirectResponse {
+        Gate::authorize('update', $user);
+
         $validated = $request->validated();
 
         // Allow username change once!
@@ -227,6 +229,8 @@ class ProfileController extends Controller
 
     public function updatePrivacy(Request $request, User $user): RedirectResponse
     {
+        Gate::authorize('update', $user);
+
         $user->analytics_cookie_opt_out = $request->get('analytics_cookie_opt_out');
 
         if (!$user->save()) {
--- a/app/Http/Requests/ProfileFormRequest.php
+++ b/app/Http/Requests/ProfileFormRequest.php
@@ -11,12 +11,14 @@ use Illuminate\Validation\Rules\File;
 class ProfileFormRequest extends FormRequest
 {
     /**
-     * Determine if the user is authorized to make this request.
+     * Authorization is done in ProfileController::update() via Gate::authorize('update', $user),
+     * matching how the rest of the application gates its write endpoints. It has to live there
+     * rather than here because the sibling profile.updateprivacy route takes a plain Request and
+     * needs the same check.
      */
     public function authorize(): bool
     {
         return true;
-        //return Auth::user()?->hasRole(Role::ROLE_ALL) ?? false;
     }
 
     /** @return array<string, mixed> */
--- a/app/Http/Requests/ProfileFormRequest.php
+++ b/app/Http/Requests/ProfileFormRequest.php
@@ -5,20 +5,30 @@ namespace App\Http\Requests;
 use App\Models\GameServerRegion;
 use App\Models\User;
 use Illuminate\Foundation\Http\FormRequest;
+use Illuminate\Support\Facades\Gate;
 use Illuminate\Validation\Rule;
 use Illuminate\Validation\Rules\File;
 
 class ProfileFormRequest extends FormRequest
 {
     /**
-     * Authorization is done in ProfileController::update() via Gate::authorize('update', $user),
-     * matching how the rest of the application gates its write endpoints. It has to live there
-     * rather than here because the sibling profile.updateprivacy route takes a plain Request and
-     * needs the same check.
+     * Deliberately authorized here and not only in ProfileController::update(): authorize() runs
+     * BEFORE validation, whereas a controller-side gate runs after it. rules() below applies
+     * `Rule::unique('users', 'email')->ignore($user)` against the route-bound user, so gating only
+     * in the controller would let someone probing another account tell "this address is already
+     * taken by the account I am probing" (403) from "taken by a different account" (redirect with
+     * errors) - an email-to-account oracle. Failing here keeps an unauthorized request away from
+     * the validator entirely.
+     *
+     * ProfileController::update() still calls Gate::authorize('update', $user) as well, so the
+     * write is gated even if this request class is ever swapped out, and so it matches its sibling
+     * profile.updateprivacy route, which takes a plain Request and can only be gated there.
      */
     public function authorize(): bool
     {
-        return true;
+        $user = $this->route()->parameter('user');
+
+        return $user instanceof User && Gate::allows('update', $user);
     }
 
     /** @return array<string, mixed> */
--- a/tests/Feature/Controller/ProfileControllerTest.php
+++ b/tests/Feature/Controller/ProfileControllerTest.php
@@ -1,0 +1,135 @@
+<?php
+
+namespace Tests\Feature\Controller;
+
+use App\Models\Laratrust\Role;
+use App\Models\User;
+use PHPUnit\Framework\Attributes\Group;
+use PHPUnit\Framework\Attributes\Test;
+use Tests\TestCases\PublicTestCase;
+
+#[Group('Controller')]
+#[Group('Profile')]
+final class ProfileControllerTest extends PublicTestCase
+{
+    #[Test]
+    public function update_givenSelf_updatesTheProfile(): void
+    {
+        // Arrange
+        $user = $this->userWithUserRole();
+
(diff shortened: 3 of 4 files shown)


app/Http/Controllers/ProfileController.php
```diff
@@ -143,6 +143,8 @@ public function update(
         User                          $user,
         ReverbHttpApiServiceInterface $reverbHttpApiService,
     ): RedirectResponse {
+        Gate::authorize('update', $user);
```
Wotuu, line 146: :robot: Minor / non-blocking: because `ProfileFormRequest` is resolved (and validated) during controller method injection, validation still runs *before* this new authorization check for `update()`. An authenticated attacker PATCHing another user's profile therefore exercises the `Rule::unique(...)->ignore($victim)` name/email rules and the avatar file-upload validation before the 403 — e.g. submitting the victim's own email passes validation (ignored id) then 403s, while a third user's email 422s, a small enumeration side channel. If you ever want the gate to run first for this route, `ProfileFormRequest::authorize()` can do `return Gate::allows('update', $this->route('user'));` (FormRequest `authorize()` runs before `rules()`), keeping the controller gate only for `updatePrivacy`. Fine to leave as-is given the consistency rationale in the PR body.

Wotuu:
:robot: Cold review (fable): 1 finding posted (minor, non-blocking: validation-before-authorization ordering on `update()`). Verified separately: no schema/deploy-window risk, `Gate` import and `UserPolicy::update()` both present, no `Gate::before` admin override, and no admin/other flow reuses these routes for a different user — the new gate cannot lock out a legitimate flow.

app/Http/Requests/ProfileFormRequest.php
```diff
@@ -11,12 +11,14 @@
...
     /**
-     * Determine if the user is authorized to make this request.
+     * Authorization is done in ProfileController::update() via Gate::authorize('update', $user),
+     * matching how the rest of the application gates its write endpoints. It has to live there
+     * rather than here because the sibling profile.updateprivacy route takes a plain Request and
+     * needs the same check.
      */
     public function authorize(): bool
     {
         return true;
```
Wotuu: :robot: Validation runs before this authorization: with the gate in the controller, `ProfileFormRequest` validation executes first, and its `Rule::unique('users', 'email')->ignore($user)` ignores the **victim** bound from the route. An authenticated attacker PATCHing `/profile/{victimId}` with a candidate email X gets a 302-with-errors when X belongs to some *other* user, but a 403 when X is either unregistered or the victim's own email. Combined with the public register form's `unique:users` oracle ("is X registered at all"), this distinguishes "email X belongs to user id N" — linking an email address to a specific account. Same shape for `name`, though usernames are public anyway.

The docblock's claim that the check "has to live" in the controller isn't quite right: `FormRequest::authorize()` runs *before* validation, so `update()` can be gated here (e.g. `return Gate::allows('update', $this->route()->parameter('user'));`) which closes the oracle, while `updatePrivacy()` keeps its controller-side `Gate::authorize` since it takes a plain `Request`. Keeping the controller gate as well is harmless defense in depth.

tests/Feature/Controller/ProfileControllerTest.php
```diff
@@ -0,0 +1,135 @@
...
+        } finally {
+            $user->delete();
+        }
+    }
+
+    #[Test]
+    public function update_givenAnotherUser_returnsForbidden(): void
+    {
+        // Arrange
+        $attacker = $this->userWithUserRole();
```
Wotuu: :robot: Minor, per the `writing-tests` skill ("Create fixtures **inside** the `try`, not before it"): the users are created before the `try`, so a failure halfway through arrange leaks rows into the shared schema — `userWithUserRole()` is create + `addRole`, so an `addRole` throw leaks the user, and in the two-user tests a failure creating `$victim` leaks `$attacker`. Null-init before the `try`, create inside it, and null-guard the deletes in `finally`.

Wotuu:

app/Http/Requests/ProfileFormRequest.php
Wotuu: :robot: You are right and the docblock was wrong — fixed in 13d48b6ce. `FormRequest::authorize()` does run before validation, so my stated reason ("has to live in the controller") only justified the gate for `updatePrivacy`, not for `update`.

`ProfileFormRequest::authorize()` now resolves the ability through the same policy:

```php
$user = $this->route()->parameter('user');

return $user instanceof User && Gate::allows('update', $user);
```

The controller keeps its `Gate::authorize` too, as you suggested — it costs nothing, keeps the write gated if the request class is ever swapped out, and matches `updatePrivacy`.

Added a test that pins the oracle rather than just the fix: an attacker PATCHes the victim with an email that definitely belongs to a third user and must get a 403 with `assertSessionHasNoErrors()`, i.e. indistinguishable from probing with a free address. With `ProfileFormRequest` stashed and only the controller gate in place, that test fails with validation errors present — so the side channel was live, not theoretical.

app/Http/Controllers/ProfileController.php
Wotuu: :robot: Same finding as the thread on `ProfileFormRequest.php` — addressed there and in 13d48b6ce: the gate now runs in `ProfileFormRequest::authorize()`, before validation, so the enumeration side channel is closed. The controller gate stays as defense in depth.

tests/Feature/Controller/ProfileControllerTest.php
Wotuu: :robot: Fixed in 13d48b6ce — all five tests now null-init before the `try`, create fixtures inside it, and null-guard the deletes with `?->delete()`. You were right that the two-user tests were the real exposure: a throw while creating `$victim` would have leaked `$attacker` into the shared schema.

Wotuu:
:robot: CI is green now. The `php-tests (feature-controller)` failure was real, not outage fallout: `ProfileController::update()` injects `ReverbHttpApiServiceInterface`, and `ReverbHttpApiService` builds its Guzzle client **in its constructor** from `reverb.apps.apps.0.options.host/port`. `.env.ci.example` ships `REVERB_HOST=` empty, so the base_uri came out as `:6001` and Guzzle threw `MalformedUriException` while the container was still resolving the controller's dependencies — outside the service's own try/catch. This PR's happy-path test just happened to be the first test ever to reach that controller.

Fixed in `phpunit.xml` (d0ab5155d) rather than by mocking the interface in this one test: the constructor is what explodes, so a test-local mock would leave the same trap armed for the next controller test that injects it. There is no Stub/Dev binding for this interface to fall back on.

PR #3867, opened 2026-08-06
#3866 Drop a team's tags when a member leaves and when the team is deleted

### LINKED ISSUE #3866: Team::removeMember() leaves DUNGEON_ROUTE_TEAM tags behind on the routes it unassigns
:robot: Follow-up from the cold review on #3864.

`Team::removeMember()` unassigns a leaving member's routes from the team but does not delete the `DUNGEON_ROUTE_TEAM` tags on those routes. `Team::removeRoute()` — the other path by which a route leaves a team — does delete them. The two should behave the same.

## Where

`app/Models/Team.php`, `removeRoute()`:

```php
if ($dungeonRoute->team_id !== null) {
    // Delete all existing team tags from this route
    $dungeonRoute->tags(TagCategory::ALL[TagCategory::DUNGEON_ROUTE_TEAM])->delete();

    $dungeonRoute->team_id = null;
    $dungeonRoute->save();
```

`app/Models/Team.php`, `removeMember()`:

```php
$this->dungeonRoutes()->where('team_id', $this->id)->where('author_id', $member->id)->update(['team_id' => null]);
$result = TeamUser::where('team_id', $this->id)->where('user_id', $member->id)->delete();
```

No tag cleanup, and the raw `update()` bypasses model events, so nothing else picks it up either.

## Consequence

The leftover tag keeps `context_class = Team::class` and `context_id = <the team>` while sitting on a route that is no longer part of that team.

There is a second, related gap in the team `deleting` hook (`Team.php:480`), which clears team tags by route:

```php
Tag::where('tag_category_id', TagCategory::ALL[TagCategory::DUNGEON_ROUTE_TEAM])
    ->whereIn('model_id', $team->dungeonRoutes->pluck('id')->toArray())
    ->delete();
```

A route that `removeMember()` already unassigned is no longer in `$team->dungeonRoutes`, so its tag survives the team's deletion with a `context_id` pointing at a team row that no longer exists.

Both together mean the tags table accumulates rows whose team context is stale or dangling. #3864 made `TagPolicy::edit()` tolerate those rows (`Team::find()` plus a fallback to the tagged route's own permissions) so they stay manageable and no longer 404, but that is damage control — the rows should not be created in the first place.

## Suggested fix

1. In `removeMember()`, delete the `DUNGEON_ROUTE_TEAM` tags for exactly the routes being unassigned — the same `team_id`/`author_id` set that the `update()` targets, not all of the team's tags.
2. In the `deleting` hook, clear by `context_id = $team->id` rather than by current route membership, so tags stranded by an earlier `removeMember()` are caught too.
3. Consider a one-off cleanup for rows already orphaned in production. Low urgency — #3864's policy fallback means they are deletable through the UI rather than stuck.

Worth a test for each path: a member leaving with a team-tagged route of their own, and a team being deleted after such a departure.

## Note

Do not revert #3864's `TagPolicy::edit()` fallback when this lands. It is still needed for rows orphaned before this fix ships, and it also covers the "no longer a member but the tag is on your own route" case, which fixing `removeMember()` prevents going forward but does not retroactively repair.

Wotuu:
:robot: Closes #3866

`Team::removeMember()` unassigned a leaving member's routes but left their `DUNGEON_ROUTE_TEAM` tags behind, while `Team::removeRoute()` — the other way a route leaves a team — drops them. The raw `update(['team_id' => null])` bypasses model events, so nothing downstream cleaned them up either, and the tag stayed on a route that was no longer part of the team.

The `deleting` hook had the matching gap: it cleared tags by the routes the team *currently* held, and a route unassigned by an earlier `removeMember()` is no longer among them — so its tag outlived the team row its `context_id` pointed at.

## Changes

**`removeMember()`** resolves the affected route ids *before* nulling `team_id` — once it is null the routes can no longer be found by that query — then deletes exactly those routes' tags for this team:

```php
Tag::where('tag_category_id', TagCategory::ALL[TagCategory::DUNGEON_ROUTE_TEAM])
    ->whereIn('model_id', $dungeonRouteIds)
    ->delete();
```

Scoped to `context_id` as well as the route ids, so a tag belonging to a *different* team that the route happens to carry is left alone.

**The `deleting` hook** now clears by `context_id` rather than by current route membership, which also catches tags stranded by a `removeMember()` that happened before this fix.

## Tests

New `tests/Feature/App/Models/TeamTest.php`:

| Test | Asserts |
|---|---|
| `removeMember_givenTheirOwnTaggedRoute_dropsTheTeamTagWithIt` | route unassigned, tag gone |
| `removeMember_givenAnotherMembersTaggedRoute_leavesThatRouteAlone` | leaver's tag gone, staying member's tag and route untouched |
| `removeMember_givenAnotherTeamsTagOnTheSameRoute_leavesItAlone` | leaver's tag gone, a second team's tag on the same route untouched |
| `deleting_givenATagStrandedByAnEarlierRemoveMember_stillDropsIt` | stranded tag removed with the team |

The stranded-tag test builds the tag directly rather than by calling `removeMember()` first, so it keeps pinning the `deleting` hook even if `removeMember()`'s own cleanup is ever changed. The other-team test exists because cold review found the `context_class`/`context_id` scoping was otherwise unpinned — dropping both `where()` clauses left the rest of the suite green.

Each test fails with the change it covers reverted.

## Scope note

This does **not** make #3864's `TagPolicy::edit()` fallback redundant, and that fallback should not be reverted when this merges. It does nothing for rows already orphaned in production, which the policy fallback keeps manageable. A one-off cleanup for those is still open in #3866 as a low-urgency item.

It also does not stop *every* new orphan, which is a correction to what this PR body originally claimed. Cold review found that `User`'s deleting hook mass-deletes the account's routes (`$user->dungeonRoutes()->delete()`) **before** calling `removeMember()`, so this cleanup finds no routes to act on and account deletion still leaves tags behind — with a dangling `model_id` rather than a stale `context_id`. That is a separate and wider bug (the mass delete skips `DungeonRoute`'s whole `deleting` hook, so thumbnails, their files, and the challenge mode run are orphaned too), filed as #3869.

### DIFF
--- a/app/Models/Team.php
+++ b/app/Models/Team.php
@@ -19,6 +19,7 @@ use Illuminate\Database\Eloquent\Relations\HasMany;
 use Illuminate\Support\Carbon;
 use Illuminate\Support\Facades\App;
 use Illuminate\Support\Facades\Auth;
+use Illuminate\Support\Facades\DB;
 use Illuminate\Support\Str;
 use Override;
 
@@ -373,8 +374,34 @@ class Team extends Model
                     $member->patreonAdFreeGiveaway->delete();
                 }
 
-                $this->dungeonRoutes()->where('team_id', $this->id)->where('author_id', $member->id)->update(['team_id' => null]);
-                $result = TeamUser::where('team_id', $this->id)->where('user_id', $member->id)->delete();
+                // Unassigning the routes and dropping their tags has to be all-or-nothing: this used
+                // to be a single UPDATE, and the catch below swallows failures, so without a
+                // transaction a throw partway through would leave the tags permanently deleted
+                // while the routes stayed on the team - a state nothing else can repair
+                $result = DB::transaction(function () use ($member): bool {
+                    // Resolve the routes before unassigning them - once team_id is nulled they can
+                    // no longer be found by this query
+                    $dungeonRouteIds = $this->dungeonRoutes()
+                        ->where('team_id', $this->id)
+                        ->where('author_id', $member->id)
+                        ->pluck('dungeon_routes.id')
+                        ->toArray();
+
+                    DungeonRoute::whereIn('id', $dungeonRouteIds)->update(['team_id' => null]);
+
+                    // A route leaving the team loses that team's tags with it, the same way
+                    // removeRoute() drops them. The update() above bypasses model events, so
+                    // nothing else cleans these up and they would be stranded on a route that is no
+                    // longer part of this team. Scoped to this team's own tags so a tag belonging
+                    // to a different team is left alone.
+                    Tag::where('tag_category_id', TagCategory::ALL[TagCategory::DUNGEON_ROUTE_TEAM])
+                        ->where('context_class', self::class)
+                        ->where('context_id', $this->id)
+                        ->whereIn('model_id', $dungeonRouteIds)
+                        ->delete();
+
+                    return (bool)TeamUser::where('team_id', $this->id)->where('user_id', $member->id)->delete();
+                });
             } catch (Exception) {
                 logger()->error('Unable to remove member from team', [
                     'team' => $this,
@@ -483,9 +510,12 @@ class Team extends Model
                     $teamMember->patreonAdFreeGiveaway->delete();
                 }
             }
-            // Delete all tags team tags belonging to our routes
+            // Delete this team's tags by context rather than by the routes it currently holds: a
+            // route that an earlier removeMember() unassigned is no longer in dungeonRoutes, so
+            // filtering on that left its tag behind pointing at a team row about to disappear
             Tag::where('tag_category_id', TagCategory::ALL[TagCategory::DUNGEON_ROUTE_TEAM])
-                ->whereIn('model_id', $team->dungeonRoutes->pluck('id')->toArray())
+                ->where('context_class', self::class)
+                ->where('context_id', $team->id)
                 ->delete();
             // Remove all users associated with this team
             TeamUser::where('team_id', $team->id)->delete();
--- a/app/Policies/TagPolicy.php
+++ b/app/Policies/TagPolicy.php
@@ -26,9 +26,10 @@ class TagPolicy
         } elseif ($tag->context_class === Team::class) {
             // If we're editing a team tag, and the user is part of this team, we can edit it.
             // find() rather than findOrFail(): a team tag can outlive its team. Team::removeMember()
-            // unassigns a leaving author's routes without dropping their team tags, and the team's
-            // deleting hook only clears tags for routes still assigned to it - so context_id can
-            // dangle, and findOrFail() would 404 every caller instead of denying them.
+            // and the team's deleting hook now clean up a leaving/removed member's team tags going
+            // forward (#3866), but rows orphaned before that fix shipped can still have a
+            // context_id that dangles, and findOrFail() would 404 every caller instead of denying
+            // them.
             $result = Team::find($tag->context_id)?->isUserMember($user) ?? false;
         }
 
--- a/tests/Feature/App/Models/TeamTest.php
+++ b/tests/Feature/App/Models/TeamTest.php
@@ -137,6 +137,207 @@ final class TeamTest extends PublicTestCase
         }
     }
 
+    #[Test]
+    public function removeMember_givenTheirOwnTaggedRoute_dropsTheTeamTagWithIt(): void
+    {
+        $author = null;
+        $team   = null;
+        $route  = null;
+        $tag    = null;
+
+        try {
+            // Arrange
+            $author = User::factory()->create();
+            $team   = $this->createTeam();
+            $team->addMember($author, TeamUser::ROLE_MEMBER);
+            $route = DungeonRoute::factory()->create([
+                'author_id' => $author->id,
+                'team_id'   => $team->id,
+            ]);
+            $tag = $this->createTeamTag($team, $route);
+
+            // Act
+            $team->removeMember($author);
+
+            // Assert - the route leaves the team, and the team's tag leaves with it
+            $this->assertNull($route->fresh()->team_id);
+            $this->assertDatabaseMissing('tags', ['id' => $tag->id]);
+        } finally {
+            $this->cleanUp($tag, $route, $team, $author);
+        }
+    }
+
+    #[Test]
+    public function removeMember_givenAnotherMembersTaggedRoute_leavesThatRouteAlone(): void
+    {
+        // removeMember() only unassigns routes the leaving member authored, so the tag cleanup has
+        // to be scoped to exactly those routes rather than to everything the team holds.
+        $leaver    = null;
+        $staying   = null;
+        $team      = null;
+        $ownRoute  = null;
+        $themRoute = null;
+        $ownTag    = null;
+        $themTag   = null;
+
+        try {
+            // Arrange
+            $leaver  = User::factory()->create();
+            $staying = User::factory()->create();
+            $team    = $this->createTeam();
+            $team->addMember($leaver, TeamUser::ROLE_MEMBER);
+            $team->addMember($staying, TeamUser::ROLE_MEMBER);
+
+            $ownRoute  = DungeonRoute::factory()->create(['author_id' => $leaver->id, 'team_id' => $team->id]);
+            $themRoute = DungeonRoute::factory()->create(['author_id' => $staying->id, 'team_id' => $team->id]);
+            $ownTag    = $this->createTeamTag($team, $ownRoute);
+            $themTag   = $this->createTeamTag($team, $themRoute);
+
+            // Act
+            $team->removeMember($leaver);
+
+            // Assert
+            $this->assertDatabaseMissing('tags', ['id' => $ownTag->id]);
+            $this->assertDatabaseHas('tags', ['id' => $themTag->id]);
+            $this->assertSame($team->id, $themRoute->fresh()->team_id);
+        } finally {
+            $this->cleanUp($ownTag, $ownRoute, null, $leaver);
+            $this->cleanUp($themTag, $themRoute, $team, $staying);
+        }
+    }
+
+    #[Test]
+    public function removeMember_givenAnotherTeamsTagOnTheSameRoute_leavesItAlone(): void
+    {
+        // The delete is scoped by context_class/context_id as well as by route, so a tag left on the
+        // route by a team the route used to belong to must survive this team's cleanup. Without both
+        // where() clauses the other team's tag would be collateral.
+        $author    = null;
+        $team      = null;
+        $otherTeam = null;
+        $route     = null;
+        $ownTag    = null;
+        $otherTag  = null;
+
+        try {
+            // Arrange
+            $author    = User::factory()->create();
+            $team      = $this->createTeam();
+            $otherTeam = $this->createTeam();
+            $team->addMember($author, TeamUser::ROLE_MEMBER);
+            $route = DungeonRoute::factory()->create([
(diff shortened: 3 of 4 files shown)


app/Models/Team.php
```diff
@@ -372,7 +372,25 @@ public function removeMember(User $member): bool
...
+
+                // A route leaving the team loses that team's tags with it, the same way
+                // removeRoute() drops them. The raw update() below bypasses model events, so
+                // nothing else cleans these up and they would be stranded on a route that is no
+                // longer part of this team.
+                Tag::where('tag_category_id', TagCategory::ALL[TagCategory::DUNGEON_ROUTE_TEAM])
+                    ->where('context_class', self::class)
+                    ->where('context_id', $this->id)
+                    ->whereIn('model_id', $dungeonRouteIds)
+                    ->delete();
```
Wotuu: :robot: The tag `delete()` now runs before the route `update()` with no transaction around them, and the `catch (Exception)` swallows any failure. If the `update()` or the `TeamUser` delete throws after the tags are gone, `removeMember()` returns `false` ("nothing happened") — but the team tags have been permanently destroyed while the routes are still assigned to the team, an inconsistent state neither the old code (single atomic conditional UPDATE) nor the `deleting` hook can repair. Suggest wrapping the pluck/delete/update/TeamUser-delete in `DB::transaction()`, or at minimum reordering so the destructive tag delete happens *after* the update succeeds — a failure then leaves stranded tags (the recoverable, pre-existing failure class that the deleting hook now cleans) instead of destroyed data.

app/Models/Team.php
```diff
@@ -372,7 +372,25 @@ public function removeMember(User $member): bool
...
+                    ->pluck('dungeon_routes.id')
+                    ->toArray();
+
+                // A route leaving the team loses that team's tags with it, the same way
+                // removeRoute() drops them. The raw update() below bypasses model events, so
+                // nothing else cleans these up and they would be stranded on a route that is no
+                // longer part of this team.
+                Tag::where('tag_category_id', TagCategory::ALL[TagCategory::DUNGEON_ROUTE_TEAM])
+                    ->where('context_class', self::class)
+                    ->where('context_id', $this->id)
```
Wotuu: :robot: Test gap: the `context_class`/`context_id` scoping here — the exact property the PR body calls out ("a tag belonging to a *different* team that the route happens to carry is left alone") — is not pinned by any of the three new tests. Dropping both `where()` clauses leaves all tests green: `removeMember_givenAnotherMembersTaggedRoute_leavesThatRouteAlone` protects the other tag via `model_id` only. Worth adding a case where the leaver's route carries a stale tag from a different team and asserting it survives `removeMember()`.

app/Models/Team.php
```diff
@@ -372,7 +372,25 @@ public function removeMember(User $member): bool
...
+                    ->where('team_id', $this->id)
+                    ->where('author_id', $member->id)
+                    ->pluck('dungeon_routes.id')
+                    ->toArray();
+
+                // A route leaving the team loses that team's tags with it, the same way
+                // removeRoute() drops them. The raw update() below bypasses model events, so
+                // nothing else cleans these up and they would be stranded on a route that is no
+                // longer part of this team.
+                Tag::where('tag_category_id', TagCategory::ALL[TagCategory::DUNGEON_ROUTE_TEAM])
```
Wotuu: :robot: Informational (pre-existing, not introduced here, but relevant to the PR body claim that this "stops new orphans being created"): on account deletion, `User`'s deleting hook runs `$user->dungeonRoutes()->delete()` — a bulk query-builder delete that bypasses `DungeonRoute`'s own deleting hook (which would have run `$dungeonRoute->tags()->delete()`) — *before* it calls `$team->removeMember($user)`. By the time this cleanup runs, the pluck finds zero routes and it is a no-op, so the deleted user's team tags still become orphans (dangling `model_id` this time, rather than dangling `context_id`). Worth folding into the one-off cleanup tracked in #3866, or reordering/using evented deletes in the User hook.

Wotuu:

app/Models/Team.php
Wotuu: :robot: Agreed and fixed in 0b6766334 — you are right that this made the failure mode strictly worse than the single `UPDATE` it replaced, which is the part that matters on a destructive path.

Both changes applied rather than picking one:

- The three writes now run inside `DB::transaction()`, so a throw rolls the whole thing back and the existing `catch` still logs and returns `false` — but now against a consistent state.
- The route `update()` also moved ahead of the tag `delete()`, so even if the transaction were ever removed the worst case is stranded-but-recoverable tags rather than destroyed ones.

The `pluck()` stays first, since the routes cannot be found once `team_id` is null.

app/Models/Team.php
Wotuu: :robot: Confirmed and fixed in 0b6766334. I verified your claim exactly as you framed it: stripping both `where()` clauses left the suite green, so the scoping was advertised in the PR body and pinned by nothing.

Added `removeMember_givenAnotherTeamsTagOnTheSameRoute_leavesItAlone` — a second team's tag on the leaver's own route, which must survive. It fails with the two `where()` clauses removed and passes with them, so the property is now actually covered rather than incidentally true.

app/Models/Team.php
Wotuu: :robot: Good catch — the PR body claim was wrong and I have corrected it rather than leaving it to stand.

Following it through, the problem is wider than the tag angle: `$user->dungeonRoutes()->delete()` is a mass delete on a relation, so `DungeonRoute::deleting` never fires at all. That hook is also what deletes thumbnails one-by-one specifically so their files leave disk/S3, and what switches connection to drop the `challenge_mode_runs` row. Account deletion currently orphans all of that, not just tags.

Filed as #3869 rather than folded into #3866 — different mechanism, much wider blast radius, and #3866's fix is already in this PR. Kept out of scope here deliberately: the fix is a per-model loop in `User`'s hook, which changes account-deletion behaviour well beyond team tags and wants its own tests.

app/Models/Team.php
```diff
@@ -476,9 +503,12 @@ protected static function boot(): void
...
             }
-            // Delete all tags team tags belonging to our routes
+            // Delete this team's tags by context rather than by the routes it currently holds: a
+            // route that an earlier removeMember() unassigned is no longer in dungeonRoutes, so
+            // filtering on that left its tag behind pointing at a team row about to disappear
             Tag::where('tag_category_id', TagCategory::ALL[TagCategory::DUNGEON_ROUTE_TEAM])
-                ->whereIn('model_id', $team->dungeonRoutes->pluck('id')->toArray())
+                ->where('context_class', self::class)
+                ->where('context_id', $team->id)
                 ->delete();
```
Wotuu, line 519: So let me get this straight, you are removing a user from a team, and as a result you're removing ALL tags of this team? That doesn't sound right - the previous query needed a `->where('context_class', DungeonRoute::class)` but this sounds like an introduced bug? Correct me if I'm wrong.

Wotuu (commented):
!

Wotuu:

app/Models/Team.php
Wotuu: :robot: You're reading this line right, but it's a different method than you think — this hunk is in `Team::boot()`'s `deleting` hook, not `removeMember()`. GitHub anchored the comment on line 512 of the new file, which is the team-deletion path. So yes: when the **team itself** is deleted, all of that team's tags go. That was the old code's intent too — it just filtered by the routes the team currently held, which missed any it had already let go of.

`removeMember()` is the other hunk, and it is scoped:

```php
Tag::where('tag_category_id', TagCategory::ALL[TagCategory::DUNGEON_ROUTE_TEAM])
    ->whereIn('model_id', $dungeonRouteIds)   // <- only the leaving member's own routes
    ->delete();
```

`$dungeonRouteIds` is exactly the set `removeMember()` unassigns (`team_id = $this->id AND author_id = $member->id`), so a route belonging to a member who stays keeps its tags. `removeMember_givenAnotherMembersTaggedRoute_leavesThatRouteAlone` pins that.

On `context_class`: for a `DUNGEON_ROUTE_TEAM` tag it's `Team::class`, not `DungeonRoute::class` — `context_*` is the tag's **owner**, `model_*` is what's tagged. From `AjaxTagController::store()`:

```php
'context_id'    => $context->getKey(),    // the Team
'context_class' => $context::class,       // Team::class
'model_id'      => $model->getKey(),      // the DungeonRoute
'model_class'   => $tagCategory->model_class,  // DungeonRoute::class
```

So `->where('context_class', DungeonRoute::class)` would match zero rows. The old query had **no** context scoping at all, which is a real pre-existing bug in the other direction: two teams tagging the same route meant one team's cleanup took the other team's tags with it. `removeMember_givenAnotherTeamsTagOnTheSameRoute_leavesItAlone` covers that.

**One genuine behaviour change is worth calling out**, and I've just pushed tests for it (a7852b80c). `TeamController::createTag()` stores a team's tag *definitions* via `Tag::saveFromRequest()`, which writes `model_id = NULL`. `NULL` never matches a SQL `IN` list, so the old hook left every definition row behind, orphaned against a deleted team. The context-scoped query picks them up. Both sides are now pinned:

- `deleting_givenAnUnassignedTagDefinition_dropsItToo` — deleting the team drops them (fails against the old query, I checked)
- `removeMember_givenAnUnassignedTagDefinition_leavesItAlone` — removing a member does not, since the team lives on

Unrelated and deliberately untouched here: `Team::getAvailableTags()` (line 456) still uses the same unscoped `whereIn('model_id', ...)`, so it can surface another team's tags on a shared route. Happy to split that off into its own issue if you'd like.

Wotuu: Yeah split it off please. Thank you for the explanation.

Wotuu:
:robot: Filed as #3875 (the `Team::getAvailableTags()` unscoped-tag-leak follow-up you asked to split off). Note: I could not reply directly on the review thread — GitHub is rejecting new thread replies on this PR with "user_id can only have one pending review per pull request", because there is an unsubmitted pending review draft on this PR. Left it alone since it looks like your own in-progress draft.

Wotuu:
:robot: Fixed the red CI (`AjaxTagControllerTest::delete_givenTeamTagStrandedOnOwnRouteAfterLeavingTheTeam_deletesIt` expected 204, got 404).

Root cause: that test (added by the already-merged #3864) built its "stranded team tag" scenario by leaving a team via `Team::removeMember()`. This PR's own fix (`b1a511dc6`) made `removeMember()` delete exactly that tag inside the same transaction that unassigns the route — so by the time the test's `DELETE` request ran, the tag was already gone and the endpoint correctly 404'd. The two PRs' intents were in direct conflict: #3864 proved a fallback path that #3866 now closes off via `removeMember()`.

`TagPolicy::edit()`'s route-ownership fallback is still needed for rows that were already orphaned before this fix shipped, so rather than weakening either the fallback or the new cleanup, the test (`9574ad056`) is repurposed to arrange that orphaned state directly — a team tag on a route whose author was never actually a member of the tag's team — instead of manufacturing it via `removeMember()`. Also updated stale docblock comments in `TagPolicy` and the test itself that described `removeMember()`/the deleting hook as never cleaning up team tags (`f3f27abfe`) — true before this PR, false now.

Verified green: all 9 checks passing after rebase onto current master.

Wotuu:
:robot: Re https://github.com/RaiderIO/keystone.guru/pull/3867#discussion_r3735198318 — split off as #3875, already has an MR up ([#3880](https://github.com/RaiderIO/keystone.guru/pull/3880)). (Posting as a top-level comment rather than an inline reply — there's a pending, unsubmitted review draft on this PR blocking new inline replies; leaving that alone rather than touching it.)

PR #3870, opened 2026-08-06
#3868 Add Season.active admin flag and drop the 1-year cap on upcoming season lookahead

### LINKED ISSUE #3868: Season 2 preview leaks early (header/Create Route), and King's Rest shows the Awakened enemy set
## Summary

Two regressions surfaced while browsing production, both triggered by Season 2's start date being
pushed out (~a year), but with **unrelated root causes**:

1. The header top-nav already shows a "Season 2" link, and the Create Route dungeon-selection
   popup already lists Season 2 dungeons — even though Season 2's mapping isn't ready and it
   shouldn't be advertised yet.
2. Creating a route for **King's Rest** (a reused BfA dungeon) shows the "Awakened" enemy set in
   the Create Route popup. Awakened was only ever active during a specific BfA season and should
   never appear for a modern King's Rest route.

## Bug 1 — premature Season 2 UI leak

An existing "don't show an upcoming season until it's close enough to start" gate
(`HeaderComposer::getUpcomingSeason()`, `app/Http/View/Composers/HeaderComposer.php:76-85`, built
for #3761/#3746) is applied to only one of three UI surfaces that read the next season:

- `resources/views/common/layout/header.blade.php:39-56` — the top-nav "Season 2" link reads the
  raw, ungated `$nextSeason` directly.
- `app/Http/View/Composers/CreateRouteFormComposer.php:43` → `resources/views/common/dungeon/select.blade.php:48-61`
  — the Create Route dungeon `<optgroup>`, same raw value.

Both regressed independently of the #3761/#3746 fixes, which only covered
`DungeonService::getDungeonsForGameVersion()` and the header's dungeon-context card.

**Planned fix**: rather than extending the existing date-arithmetic gate
(`upcoming_visible_days` vs. `start`) to the missed call sites, add an explicit admin-controlled
`active` boolean on `Season` — mirroring the existing `Dungeon::active` precedent
(`app/Models/Dungeon.php`, admin checkbox in `resources/views/admin/dungeon/edit.blade.php:35`).
This keeps `start` as a pure calendar fact and makes "ready to reveal publicly" an explicit,
admin-set, seeder-persisted fact (round-tripped through `database/seeders/seasondata/seasons.json`
via `mapping:save`/`SeasonsSeeder`) instead of inferred date math duplicated across composers.

## Bug 2 — King's Rest shows the Awakened enemy set

`SeasonRepository::getUpcomingSeasonForDungeon()` (`app/Repositories/Database/SeasonRepository.php:39-62`)
caps its lookahead at `now()->addYear()`. King's Rest already has a real, curated `season_dungeons`
row for Season 2 — but once Season 2's start date is pushed further than a year out, the query
returns null, and callers (`Dungeon::getActiveSeason()`,
`DungeonRouteSaveService::resolveSeasonForEdit()`) fall back to
`getMostRecentSeasonForDungeon()` — King's Rest's last active BfA season, which carries the
seasonal `Awakened` affix. That affix flows into the new route's season assignment and unhides all
`seasonal_type = awakened` enemies client-side.

Reproduced via tinker: with `now()` set more than a year before Season 2's start,
`getUpcomingSeasonForDungeon()` returns `null` and `getActiveSeason()` resolves to the BfA season
carrying the Awakened affix.

**Planned fix**: drop the `now()->addYear()` upper bound in `getUpcomingSeasonForDungeon()`. A
`season_dungeons` row is a deliberate, curated assignment, not speculative noise — the "don't
advertise it too early" concern is Bug 1's job, not this query's. This also benefits
`app/Logic/MDT/Conversion.php:423-424` for any similarly far-future-seeded dungeon.

## Related issues

- #3761 — "Dungeon selector fully switches to next season's dungeons as soon as it's seeded, not
  when it starts" — same problem class as Bug 1, fixed a different subset of call sites.
- #3746 — "Seeding a season before it starts takes every page down via the site header" —
  companion issue from the same Season 2 seeding episode.
- #1306, #1285, #535 — prior recurrences of "Awakened enemies leak into the wrong context", none
  of which are this specific `getUpcomingSeasonForDungeon()` fallback.

None of the above cover these two specific regressions.

Wotuu:
:robot: Closes #3868

## Summary

Two independent fixes for the same underlying trigger (Season 2's start date being pushed out):

**1. Season 2 preview leak (header nav + Create Route dungeon list)**

`HeaderComposer`'s "don't advertise an upcoming season until it's ready" gate (built for #3761)
was only applied to the dungeon-context card, not the raw `nextSeason` used by the header nav
link or `CreateRouteFormComposer`'s dungeon dropdown - so both leaked Season 2 as soon as it was
seeded.

Rather than extending the existing date-arithmetic gate (`upcoming_visible_days` vs. `start`) to
the missed call sites, this adds an explicit admin-controlled `active` boolean on `Season`,
mirroring the existing `Dungeon::active` precedent. `start` stays a pure calendar fact used for
real game-rotation timing; `active` becomes the single "ready to reveal publicly" switch, set
through the admin panel and round-tripped through `database/seeders/seasondata/seasons.json` via
the existing `mapping:save`/`SeasonsSeeder` mechanism. Season 2 (id 18) is set `active: false` as
part of this change, so the fix actually suppresses today's leak.

**2. King's Rest shows the Awakened enemy set in Create Route**

`SeasonRepository::getUpcomingSeasonForDungeon()` capped its lookahead at `now()->addYear()`.
King's Rest already has a real, curated `season_dungeons` row for Season 2, but once Season 2's
start date is pushed further than a year out, the query returned null and callers
(`Dungeon::getActiveSeason()`, `DungeonRouteSaveService::resolveSeasonForEdit()`) fell back to
`getMostRecentSeasonForDungeon()` - King's Rest's last active BfA season, which carries the
seasonal `Awakened` affix. That affix flowed into new routes and unhid the Awakened enemy set.

Drops the 1-year cap: a `season_dungeons` row is a deliberate, curated assignment, not
speculative noise - the "don't advertise it too early" concern is bug 1's job, not this query's.

Verified via tinker: `getUpcomingSeasonForDungeon()` and `getActiveSeason()` now both correctly
resolve King's Rest to Season 2 (no seasonal affix) instead of the BfA Awakened season.

## Test plan

- [x] `SeasonRepositoryTest` (incl. new test for a season >1 year out being honored)
- [x] `HeaderComposerTest` (rewritten for the `active` flag: active season shows everywhere,
      inactive season shows nowhere, game-version-without-seasons still hides the card)
- [x] `CreateRouteFormComposerTest` (new: active/inactive upcoming season gating)
- [x] `AdminSeasonControllerTest` (new: `active` checkbox round-trips through create/update,
      unchecked persists as `false`)
- [x] Full suite green except the pre-existing, worktree-environment-only `MapTilesExistenceTest`
      failure (assets only mounted on the main stack, CI excludes it)


app/Repositories/Database/SeasonRepository.php
```diff
@@ -53,8 +53,6 @@ public function getUpcomingSeasonForDungeon(Dungeon $dungeon): ?Season
             ->join('season_dungeons', 'seasons.id', 'season_dungeons.season_id')
             ->where('season_dungeons.dungeon_id', $dungeon->id)
             ->where('seasons.start', '>', now())
-            // Don't go overboard with trying to fetch seasons in the future
-            ->where('seasons.start', '<', now()->addYear())
             ->orderBy('seasons.start', 'desc')
```
Wotuu, line 61: :robot: **Dropping the 1-year cap regresses the 13 Legion timewalking dungeons into the exact bug this PR fixes for King's Rest.**

`Season::SEASON_LEGION_TW_S1` (id 7) is seeded with `start = 2050-12-28` and `seasonal_affix_id = 25` (`Infernal`), and it has `season_dungeons` rows for dungeon ids 1–13. The 1-year cap is the only thing that kept that placeholder date out of this query. Verified against the live seeded DB (main stack, master):

```
# with the cap (current master)
1  The Arcway              oldUpcoming=NULL mostRecent=NULL currentSeasonOfExp=NULL   -> getActiveSeason() = null
2  Black Rook Hold         oldUpcoming=NULL mostRecent=12   currentSeasonOfExp=NULL   -> getActiveSeason() = season 12 (no seasonal affix)

# without the cap (this PR)
2  Black Rook Hold         upcoming = 7, start=2050-12-28, seasonal_affix_id=25 (Infernal)
```

`hasMappingVersionWithSeasons()` returns `true` for these dungeons, so `SeasonService` doesn't filter them out. So after this change `Dungeon::getActiveSeason()` and `DungeonRouteSaveService::resolveSeasonForEdit()` resolve every Legion dungeon to a season carrying the `Infernal` seasonal affix — the same failure mode (stale seasonal affix flowing into new routes and unhiding a seasonal enemy set) that bug 2 is fixing for King's Rest, just pointed at Legion instead of BfA.

Two remedies that look obvious but don't work, to save you the round trip:
- `->where('seasons.active', true)` — season 18 is `"active": false` in this very diff, and King's Rest resolving to season 18 *is* the fix, so this breaks bug 2.
- `->orderBy(..., 'asc')` — season 7 is the *only* future season for dungeons 1–13, so ascending returns it too.

Separate observation on the same line: `orderBy('seasons.start', 'desc')` returns the *farthest*-future season, not the nearest. That was bounded by the cap before; it now isn't. If a dungeon is ever curated into two future seasons, "upcoming" will be the later one.

app/Http/View/Composers/CreateRouteFormComposer.php
```diff
@@ -37,9 +37,14 @@ public function compose(View $view): void
             ])
             ->makeVisible(['seasonDungeons']);
 
+        // A season is seeded weeks - sometimes months - before it starts so its mapping can be
+        // reviewed; until an admin marks it `active` it must not be offered here (#3868).
+        $nextSeason = $this->viewService->getNextSeasonForRegion($gameServerRegion);
+        $nextSeason = $nextSeason?->active ? $nextSeason : null;
```
Wotuu: :robot: **This gate probably doesn't fix the Create Route dungeon-list leak — that list is fed by a different composer, which this PR doesn't touch.**

The season groups in the Create Route dungeon dropdown come from `common.dungeon.select` (included from `createroute.blade.php` with `showSeasons => true`), and that view has its **own** composer, `DungeonSelectComposer`, which does an ungated `$view->with('nextSeason', $this->viewService->getNextSeasonForRegion($gameServerRegion))`. Composers run in `View::renderContents()` and `View::with()` overwrites `$this->data[$key]`, so the composer's ungated value wins over anything inherited from the parent view — and `nextSeason` isn't passed into the `@include` anyway. `select.blade.php` then builds both the "Seasons" selector group and the per-season dungeon optgroups from `$nextSeason`.

Empirically, on master the rendered page confirms which path drives it — `GET /new` returns `<optgroup label="Season 2">` (Midnight S2, id 18, the upcoming season) in the dungeon select. `CreateRouteFormComposer`'s `nextSeason` only reaches the inline JS options block, not that dropdown.

Worth noting the new `CreateRouteFormComposerTest` asserts on `$view->getData()['nextSeason']` straight off the composer and never renders the view, so it passes either way — it can't distinguish "the dropdown is fixed" from "the dropdown is untouched".

Other ungated consumers of `getNextSeasonForRegion()` with the same "seeded but not ready" exposure: `DungeonSelectComposer`, `DungeonGridTabsComposer`, `AffixesComposer`, `DiscoverSearchComposer`, `RouteCoverageAffixGroupComposer`, and `ViewService::getGameServerRegionViewVariables()` (`'nextSeason' => ...`). Since every one of them wants the same rule, applying the `active` gate once inside `ViewService::getNextSeasonForRegion()` would cover all call sites instead of leaving the next one to be found in production too.

Wotuu:

1. `SeasonRepository::getUpcomingSeasonForDungeon()` — dropping the 1-year cap makes the 13 Legion timewalking dungeons resolve to `SEASON_LEGION_TW_S1` (id 7, `start = 2050-12-28`, seasonal affix `Infernal`), i.e. the same stale-seasonal-affix regression this PR fixes for King's Rest, pointed at Legion. Verified against the seeded DB.
2. `CreateRouteFormComposer` — the Create Route dungeon dropdown's season groups come from `DungeonSelectComposer` (`common.dungeon.select`), which this PR leaves ungated, so bug 1's Create Route leak looks unfixed; the new composer test can't detect that because it never renders. Several other `getNextSeasonForRegion()` consumers are ungated too — `ViewService::getNextSeasonForRegion()` would be the single choke point.

Non-blocking operational caveat (not a defect): `SeasonsSeeder` rebuilds `seasons` from `seasons.json` on every deploy, so an `active` flipped to true in the admin panel on reveal day is silently reverted by the next deploy unless someone runs `mapping:save` and commits. True of every season field, but `active` is the one whose whole purpose is a time-sensitive manual flip — worth a line in the admin help text or the release checklist.

Checked and found clean: migration is additive with a `true` default (old code tolerates it, existing rows stay visible), `SeasonsSeeder`'s `?? true` handles pre-`active` exports, `mapping:save` round-trips the field, the `??= 0` checkbox handling and blade `?->active ?? 1` match the `Dungeon::active` precedent, and `config('keystoneguru.season.upcoming_visible_days')` has no remaining consumers.

app/Repositories/Database/SeasonRepository.php
Wotuu: :robot: Good catch. Fixed in 4d48bf296: widened the cap from 1 to 3 years instead of dropping it — comfortably covers a real season being pushed out (the reported case) while staying far short of the Legion TW placeholder's 2050 date. Added a regression test (`getUpcomingSeasonForDungeon_givenDungeonWithOnlyTheFarFuturePlaceholderSeason_returnsNull`) asserting a Legion TW dungeon still resolves to null rather than the placeholder season.

app/Http/View/Composers/CreateRouteFormComposer.php
Wotuu: :robot: Confirmed. Fixed in 4d48bf296 by moving the gate to its single source, `ViewService::getNextSeasonForRegion()`, instead of gating each composer individually — that method has six independent composer consumers (HeaderComposer, CreateRouteFormComposer, DungeonSelectComposer, DungeonGridTabsComposer, AffixesComposer, DiscoverSearchComposer, RouteCoverageAffixGroupComposer), all of which now get the gated value automatically. Removed the now-redundant per-composer gating in HeaderComposer/CreateRouteFormComposer. Added an HTTP-level regression test (`DungeonRouteControllerCreateSeasonGateTest`) hitting the actual `dungeonroute.new` route and asserting the season's dungeon optgroup is absent/present based on `active`, so this can't silently regress again the way the per-composer approach did.

Wotuu:

PR #3872, opened 2026-08-07
#3869 Delete a user's dungeon routes one at a time so DungeonRoute::deleting fires

### LINKED ISSUE #3869: Deleting a user mass-deletes their routes, skipping DungeonRoute's deleting hook (orphans thumbnails, files, challenge mode runs, tags)
:robot: Found during the cold review of #3867.

Deleting a user account removes their dungeon routes with a **mass delete**, which does not fire per-model Eloquent events — so `DungeonRoute`'s `deleting` hook never runs and none of the cleanup it performs happens.

## Where

`app/Models/User.php`, in the `deleting` hook:

```php
$user->dungeonRoutes()->delete();
```

`$user->dungeonRoutes()` returns a `HasMany`, and `delete()` on a relation proxies to the query builder. `DungeonRoute` has no soft deletes, so the rows are removed directly and `DungeonRoute::deleting` is skipped entirely.

## What gets orphaned

Everything in `DungeonRoute`'s `deleting` hook (`app/Models/DungeonRoute/DungeonRoute.php:1442`):

- **Thumbnails** — `dungeonRouteThumbnails` are deleted individually precisely so each one also removes its file from disk/S3. A mass delete leaves both the `files` rows and the stored objects behind.
- **Thumbnail jobs** — `dungeonRouteThumbnailJobs` are never expired.
- **The challenge mode run on the `combatlog` connection** — the hook switches connection specifically to delete it; skipped entirely, so combat-log data outlives the route.
- **Tags** — `$dungeonRoute->tags()->delete()`, which is how this surfaced.
- `affixgroups`, `routeattributesraw`, `playerclasses`, `playerraces`, `playerspecializations`, `enemyRaidMarkers`, `brushlines`, `paths`, `arrows`, `killZones`, `mapicons`, `pridefulEnemies`, `ratings`, `favorites`, `pinnedByUsers`.

Brushlines, paths, arrows and kill zones are looped and deleted one at a time in the hook, which suggests they have their own cascading `deleting` hooks — so the orphaning is potentially more than one level deep.

## Why it matters beyond tags

The tag half is what #3867 ran into: `User`'s hook calls `$user->dungeonRoutes()->delete()` *before* `$team->removeMember($user)`, so by the time the team cleanup runs there are no routes left to find, and it plucks zero ids. That means account deletion still produces orphaned tags even after #3867 — but with a **dangling `model_id`** (the route is gone) rather than a stale `context_id`.

The thumbnail and challenge-mode-run cases look more consequential than the tags: orphaned files are storage that is never reclaimed, and an orphaned `challenge_mode_runs` row points at a `dungeon_route_id` that no longer exists.

## Suggested fix

Replace the mass delete with a per-model loop so the hook fires:

```php
foreach ($user->dungeonRoutes as $dungeonRoute) {
    $dungeonRoute->delete();
}
```

Worth checking the same pattern elsewhere — `$user->reports()->delete()` on the next line has the same shape, and any other `->relation()->delete()` on a model that owns a `deleting` hook is suspect. Account deletion is not a hot path, so the per-row cost is not a concern.

A test asserting that deleting a user with a fully-populated route leaves no thumbnails, tags or challenge mode run behind would pin it.

## Scope note

Do not fold this into #3866 — that one is specifically about team tag cleanup and its fix is already in #3867. This is a different mechanism (mass delete skipping model events) with a much wider blast radius.

Wotuu:
:robot: Closes #3869

`User::deleting` called `$user->dungeonRoutes()->delete()`. On a `HasMany` that proxies straight to the query builder, so no per-model events fire — and since `DungeonRoute` has no soft deletes the rows went out directly and `DungeonRoute::deleting` never ran. That hook is what removes the route's thumbnails (and, through `DungeonRouteThumbnail` → `File`, the stored objects), thumbnail jobs, the challenge mode run on the `combatlog` connection, tags, and every mapping object.

## Changes

- `$user->dungeonRoutes()->delete()` → a loop that deletes one route at a time, so the hook fires.
- `$user->patreonUserLink()->delete()` → same bug, two lines below: `PatreonUserLink::deleting` drops the user's patreon benefits and was being skipped. Fixed in the same way.
- `$user->reports()` left as a mass delete on purpose — `UserReport` has no `deleting` hook. Noted in a comment so it doesn't read as an oversight.

Account deletion is not a hot path, so the per-row cost is irrelevant.

## Two notes on the implementation

**Not the loop shape suggested in the issue.** The issue proposed `foreach ($user->dungeonRoutes as $dungeonRoute)`. That's a lazy collection load, and `AppServiceProvider` calls `Model::preventLazyLoading()` — so if the user was hydrated as part of a collection (an admin sweep, a bulk command) the hook would throw in dev/test and log a violation in production. Using the relation query `$user->dungeonRoutes()->get()` is an explicit fetch and can't trip it.

**The ordering concern in the issue resolves itself.** #3869 notes that routes are deleted before `$team->removeMember($user)`, so team-tag cleanup plucks zero ids and leaves tags with a dangling `model_id`. `Taggable::tags()` matches on `model_id` + `model_class` with no `context_*` scoping, so the route's own hook deletes *all* tags pointing at it, team tags included. No reorder needed.

## Test

`UserTest::delete_givenUserWithPopulatedDungeonRoute_firesTheDungeonRouteDeletingHook` builds a route carrying a thumbnail (+ its `File`), a tag and a challenge mode run, deletes the user, and asserts all four are gone — the challenge mode run asserted against the `combatlog` connection explicitly.

Verified it fails on the unfixed code: `dungeon_routes` is empty (the mass delete worked) but `dungeon_route_thumbnails` still has the row. It asserts DB rows rather than files on disk, since `File::isRemoteDiskProtectedFromLocalMutation()` means a local environment never physically deletes off a remote disk.

## Deliberately out of scope

I audited every `->relation()->delete()` call site in `app/` against the models that define `static::deleting`. Two more real occurrences exist, both with a wider blast radius than user deletion, so they're filed as **#3873** rather than widened into this diff:

- `DungeonRoute::deleting`'s own `challengeModeRun()->delete()` skips `ChallengeModeRun::deleting` and orphans `challenge_mode_run_data` on *every* route deletion — and it drags cross-connection handling along with it.
- `AjaxKillZoneController:452` skips `KillZone::deleting` (`killZoneEnemies`, `killZoneSpells`).

The cold review found a third that this audit missed, now also on #3873: `mapicons()->delete()` (`DungeonRoute.php:1493`, `MappingVersion.php:695`) skips `MapIcon`'s hook and orphans `map_object_to_awakened_obelisk_links`. `MapIcon` defines no `boot()` of its own — it inherits one from the `HasLinkedAwakenedObelisk` trait — so grepping model files for `static::deleting` doesn't surface it.

Everything else checked out: the remaining mass deletes target hook-less models, and `MappingVersion::deleting` already loops the one child that has a hook (`enemyPatrols`). #3869's stated goal — the `challenge_mode_runs` row no longer outliving the route — is met here.

## Cold review

`cold-reviewer-fable`, 2 findings, both low severity, both addressed:

1. The `MapIcon` audit gap above — added to #3873 rather than widened into this diff, consistent with how the other two sites were handled.
2. `->get()` hydrated every route at once inside a synchronous web request; a user at the cap (999 free, unlimited for patrons) made that unbounded. Switched to `lazyById()`, which pages on `id > lastSeen` and stays correct while the rows are deleted underneath the iteration.

It independently verified the `Taggable::tags()` claim above, that the test genuinely pins the bug and cleans up on both connections, and that the partial-failure risk introduced by looping is not a meaningful regression (both call sites catch, and the loop is resumable on retry).

## Interaction with #3867

#3867 (open draft) changes `Team::removeMember()`, which `User::deleting` calls. This branch is off `master` without it and its test doesn't assume that cleanup exists. Whoever merges second should re-check the combination.

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -441,9 +441,18 @@ class User extends Authenticatable implements LaratrustUser
             // never runs - drop the user's stored feature values explicitly instead
             Feature::forgetAllForUser($user);
 
-            $user->dungeonRoutes()->delete();
+            // Deleted one at a time on purpose: a mass delete on the relation goes straight to the query
+            // builder and never fires DungeonRoute::deleting, which is what cleans up the route's thumbnails
+            // (and their files on disk), thumbnail jobs, challenge mode run, tags and all mapping objects
+            foreach ($user->dungeonRoutes()->get() as $dungeonRoute) {
+                $dungeonRoute->delete();
+            }
+
+            // UserReport has no deleting hook, so a mass delete is fine here
             $user->reports()->delete();
-            $user->patreonUserLink()->delete();
+
+            // Same as the dungeon routes above - PatreonUserLink::deleting drops the user's patreon benefits
+            $user->patreonUserLink()->first()?->delete();
             foreach ($user->teams as $team) {
                 // Remove ourselves from the team
                 $team->removeMember($user);
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -443,8 +443,10 @@ class User extends Authenticatable implements LaratrustUser
 
             // Deleted one at a time on purpose: a mass delete on the relation goes straight to the query
             // builder and never fires DungeonRoute::deleting, which is what cleans up the route's thumbnails
-            // (and their files on disk), thumbnail jobs, challenge mode run, tags and all mapping objects
-            foreach ($user->dungeonRoutes()->get() as $dungeonRoute) {
+            // (and their files on disk), thumbnail jobs, challenge mode run, tags and all mapping objects.
+            // lazyById() rather than get() so a user at the route cap doesn't hydrate every route at once -
+            // it pages on id > lastSeen, which stays correct while the rows are being deleted underneath it
+            foreach ($user->dungeonRoutes()->lazyById() as $dungeonRoute) {
                 $dungeonRoute->delete();
             }
 
--- a/tests/Feature/App/Models/UserTest.php
+++ b/tests/Feature/App/Models/UserTest.php
@@ -2,7 +2,14 @@
 
 namespace Tests\Feature\App\Models;
 
+use App\Models\CombatLog\ChallengeModeRun;
+use App\Models\DungeonRoute\DungeonRoute;
+use App\Models\DungeonRoute\DungeonRouteThumbnail;
+use App\Models\DungeonRoute\DungeonRouteThumbnailVariant;
+use App\Models\File;
 use App\Models\Laratrust\Role;
+use App\Models\Tags\Tag;
+use App\Models\Tags\TagCategory;
 use App\Models\User;
 use PHPUnit\Framework\Attributes\Group;
 use PHPUnit\Framework\Attributes\Test;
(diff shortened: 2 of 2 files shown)


app/Models/User.php
```diff
@@ -441,9 +441,18 @@ protected static function boot(): void
             // never runs - drop the user's stored feature values explicitly instead
             Feature::forgetAllForUser($user);
 
-            $user->dungeonRoutes()->delete();
+            // Deleted one at a time on purpose: a mass delete on the relation goes straight to the query
+            // builder and never fires DungeonRoute::deleting, which is what cleans up the route's thumbnails
+            // (and their files on disk), thumbnail jobs, challenge mode run, tags and all mapping objects
+            foreach ($user->dungeonRoutes()->get() as $dungeonRoute) {
+                $dungeonRoute->delete();
```
Wotuu, line 450: :robot: **Audit gap (low, pre-existing — belongs in #3873, not this diff):** the PR body claims "the remaining mass deletes target models with no `deleting` hook", but `MapIcon` *does* have one — `HasLinkedAwakenedObelisk` defines `boot()` and `MapIcon` declares no `boot()` of its own, so the trait boot (registering `linkedawakenedobelisks()->delete()` on `deleting`) becomes `MapIcon::boot()` (verified via reflection: it resolves to `HasLinkedAwakenedObelisk.php:80`). That means `DungeonRoute::deleting` (`DungeonRoute.php:1493`, `$dungeonRoute->mapicons()->delete()`) and `MappingVersion::deleting` (`MappingVersion.php:695`) both skip it and orphan `map_object_to_awakened_obelisk_links` rows. Easy to miss because the hook lives in a trait rather than the model file. Suggest adding it to the #3873 list rather than widening this diff.

app/Models/User.php
```diff
@@ -441,9 +441,18 @@ protected static function boot(): void
             // never runs - drop the user's stored feature values explicitly instead
             Feature::forgetAllForUser($user);
 
-            $user->dungeonRoutes()->delete();
+            // Deleted one at a time on purpose: a mass delete on the relation goes straight to the query
+            // builder and never fires DungeonRoute::deleting, which is what cleans up the route's thumbnails
+            // (and their files on disk), thumbnail jobs, challenge mode run, tags and all mapping objects
+            foreach ($user->dungeonRoutes()->get() as $dungeonRoute) {
```
Wotuu: :robot: **Note (low): unbounded synchronous loop in a web request.** Both deletion entry points (`UserController::delete()`, `ProfileController::delete()`) run this hook inline. `dungeonRoutes()->get()` hydrates every route at once, and `DungeonRoute::deleting` then does an 8-relation `load()` plus ~20 delete queries per route (including a cross-connection hop to `combatlog`). The free cap is 999 routes and patrons are unlimited, so a prolific author can push this into tens of thousands of queries in one request — the old mass delete was O(1) here. Failure mode is acceptable (both call sites catch, the loop is resumable on retry since deleted routes fall out of the relation, and `Auth::logout()` only runs after success), so no change demanded — but consider `->lazyById()` instead of `->get()` to bound memory (safe while deleting: iterated rows are behind the id cursor), and be aware a timeout mid-hook can leave one route stripped of thumbnails/tags but not deleted, since `DungeonRoute::delete()` is not transactional.

Wotuu:
:robot: Cold review (fable): 2 findings posted. Verified: `Taggable::tags()` is indeed unscoped by `context_*` (so team tags die with the route and no reorder vs `removeMember()` is needed); the new test pins the bug, asserts the `combatlog` connection explicitly, and its `finally` cleanup covers every created row on the correct connections; `PatreonUserLink::deleting` and the ordering vs the teams loop check out. Findings: (1) low, audit gap — `MapIcon` has a `deleting` hook via `HasLinkedAwakenedObelisk::boot()`, so `DungeonRoute::deleting`/`MappingVersion::deleting` mass-deleting `mapicons()` skip it (orphaned `map_object_to_awakened_obelisk_links`); should join the #3873 list. (2) low, note — the per-route loop is unbounded in a synchronous request (999+/unlimited routes × ~25 queries each); resumable on failure, but `lazyById()` would bound memory.

app/Models/User.php
Wotuu: :robot: Confirmed and already tracked — filed as #3873, which lists the `MapIcon`/`HasLinkedAwakenedObelisk` orphan (and the other `->relation()->delete()` sites) explicitly. Leaving this diff scoped to `User::deleting` as suggested.

app/Models/User.php
Wotuu: :robot: Acknowledged — no change made here since the finding says none is required. Noting the lazyById() suggestion for future reference if this ever needs to scale beyond the current free/patron route caps.

PR #3876, opened 2026-08-07
#3873 Fire deleting hooks for ChallengeModeRun, KillZone and MapIcon on mass deletes

### LINKED ISSUE #3873: Mass deletes skip deleting hooks in ChallengeModeRun and KillZone cleanup (follow-up to #3869)
:robot: Follow-up to #3869, which fixed the two occurrences inside `User::deleting`. The same mechanism exists elsewhere and was deliberately left out of #3872 to keep that diff scoped.

## The pattern

`$model->relation()->delete()` on a `HasMany`/`HasOne` proxies to the query builder, so **no per-model Eloquent events fire**. When the related model has its own `deleting` hook, everything that hook is responsible for is silently orphaned. None of the models involved use soft deletes, so the rows go out directly.

## Confirmed sites

Cross-referencing every `->relation()->delete()` call site in `app/` against the models that define `static::deleting`:

| Call site | Relation target | Hook that gets skipped |
|---|---|---|
| `app/Models/DungeonRoute/DungeonRoute.php:1446` | `ChallengeModeRun` | `challengeModeRunData()` — orphans `challenge_mode_run_data` on **every** route deletion, not just user deletion |
| `app/Http/Controllers/Ajax/AjaxKillZoneController.php:452` | `KillZone` | `killZoneEnemies()`, `killZoneSpells()` |
| `app/Models/DungeonRoute/DungeonRoute.php:1493` and `app/Models/Mapping/MappingVersion.php:695` (`mapicons()->delete()`) | `MapIcon` | `linkedawakenedobelisks()` — orphans `map_object_to_awakened_obelisk_links` |

**The `MapIcon` case is the easiest one to miss** and was itself missed on the first pass of this audit (caught in the cold review of #3872). `MapIcon` defines no `boot()` of its own — it gets one from the `HasLinkedAwakenedObelisk` trait, so grepping the model files for `static::deleting` does not turn it up. Any future audit of this pattern has to expand traits to their users: `grep -rl HasLinkedAwakenedObelisk app/Models/`. (`Path` also uses the trait but defines its own `boot()` and is already handled.)

The `ChallengeModeRun` one is the most consequential: it fires on every single route deletion, and it drags cross-connection semantics along with it (`setConnection('combatlog')` … `setConnection(null)`), which is why it wants its own change and its own test rather than being folded into #3872.

## Checked and safe

These call sites target models with **no** `deleting` hook, so a mass delete is correct and should be left alone: `$user->reports()` (`UserReport`), `affixgroups`, `routeattributesraw`, `playerclasses`, `playerraces`, `playerspecializations`, `tags`, `enemyRaidMarkers`, `mapicons`, `pridefulEnemies`, `ratings`, `favorites`, `mdtImport`, `metrics`, `metricAggregations`, `dungeonSpeedrunDifficulties`, `npcDungeons`, `npcbolsteringwhitelists`, `npcSpells`, `npcCharacteristics`, `npcEnemyForces`, `enemyActiveAuras`, `floorcouplings`, `patreonUserBenefits`, `overpulledenemies`, `linkedawakenedobelisks`, `polyline`, `affixGroupEaseTiers`, and `MappingVersion`'s child relations.

## Suggested approach

Same as #3872 — replace the mass delete with a loop over `->get()` (not the lazy collection property, which trips `Model::preventLazyLoading()`), and pin each with a test asserting the second-level rows are gone.

Worth considering as part of this: a guard that makes the mistake harder to repeat, since it has now been made at least three times independently.

Wotuu:
Closes #3873

:robot: Follow-up to #3869/#3872, confirmed-sites audit. `$model->relation()->delete()` proxies to the query builder, so it never fires the target model's own `deleting` hook. Fixed the three confirmed sites:

- `DungeonRoute::boot()`'s `deleting` listener: `challengeModeRun()->delete()` → `->first()?->delete()`, so `ChallengeModeRun::deleting` (cleans up `challenge_mode_run_data`) fires on every route deletion, not just user deletion.
- `DungeonRoute::boot()`'s `deleting` listener and `MappingVersion::boot()`'s `deleting` listener: `mapicons()->delete()` / `mapIcons()->delete()` → loop over `->get()`, so `MapIcon::deleting` (added by the `HasLinkedAwakenedObelisk` trait, cleans up `map_object_to_awakened_obelisk_links`) fires.
- `AjaxKillZoneController::deleteAll()`: `killZones()->delete()` → loop over the already-fetched `$killZones` collection, so `KillZone::deleting` (cleans up `killZoneEnemies`/`killZoneSpells`) fires.

`pridefulEnemies`/`enemyPatrols`/etc. were left as-is — confirmed in the issue's audit as targeting models with no `deleting` hook.

## Test plan
- [x] New tests assert the previously-orphaned child rows are gone after each cascade
- [x] `php artisan test --group=DungeonRoute --group=KillZone --group=MappingVersion` green (3 pre-existing unrelated failures confirmed present on master too)

### DIFF
--- a/app/Http/Controllers/Ajax/AjaxKillZoneController.php
+++ b/app/Http/Controllers/Ajax/AjaxKillZoneController.php
@@ -449,7 +449,11 @@ class AjaxKillZoneController extends Controller
                 $killZones       = $dungeonRoute->killZones;
                 $pridefulEnemies = $dungeonRoute->pridefulEnemies;
 
-                $dungeonRoute->killZones()->delete();
+                // Deleted one at a time - a mass delete on the relation skips KillZone::deleting, which
+                // is what cleans up the kill zone's enemies and spells
+                foreach ($killZones as $killZone) {
+                    $killZone->delete();
+                }
                 $dungeonRoute->pridefulEnemies()->delete();
 
                 if (Auth::check()) {
--- a/app/Models/DungeonRoute/DungeonRoute.php
+++ b/app/Models/DungeonRoute/DungeonRoute.php
@@ -1481,7 +1481,9 @@ class DungeonRoute extends Model implements TracksPageViewInterface
                 'livesessions',
             ]);
 
-            $dungeonRoute->setConnection('combatlog')->challengeModeRun()->delete();
+            // A mass delete on the relation skips ChallengeModeRun::deleting, which is what cleans up
+            // challenge_mode_run_data - fetch and delete the single row instead so the hook fires
+            $dungeonRoute->setConnection('combatlog')->challengeModeRun()->first()?->delete();
             $dungeonRoute->setConnection(null);
 
             // Delete thumbnails
@@ -1523,7 +1525,11 @@ class DungeonRoute extends Model implements TracksPageViewInterface
                 $killZone->delete();
             }
 
-            $dungeonRoute->mapicons()->delete();
+            // A mass delete on the relation skips MapIcon::deleting (via HasLinkedAwakenedObelisk),
+            // which is what cleans up map_object_to_awakened_obelisk_links
+            foreach ($dungeonRoute->mapicons()->get() as $mapIcon) {
+                $mapIcon->delete();
+            }
             $dungeonRoute->pridefulEnemies()->delete();
             // External
             $dungeonRoute->ratings()->delete();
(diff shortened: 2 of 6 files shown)


Wotuu:
:robot: Cold review: 1 finding (fixed: ChallengeModeRun test only asserted on `challenge_mode_runs`, which is deleted either way regardless of the hook — now asserts on `challenge_mode_run_data`, the row the hook actually cleans up. Also aligned the two `mapicons`/`mapIcons` loops to the bare relation property, matching every sibling foreach in the same `deleting` listeners.)

PR #3878, opened 2026-08-07
#3874 Centralize the "safely get map object group" helpers in util.js

### LINKED ISSUE #3874: Centralize "safely get map object group" helper functions (follow-up to #3860)
:robot: Follow-up spun off from a review comment on PR #3860 (issue #3666).

## Background

`resources/assets/js/custom/inline/common/dungeonroute/publish.js` has a private helper,
`_getKillZoneMapObjectGroup()`, that "safely" fetches the killzone `MapObjectGroup` off the
current `DungeonMap` - falling back to `null` when there is no map at all (e.g. the share modal
that hosts the publish select also renders on pages that have no map):

```js
_getKillZoneMapObjectGroup() {
    let state = typeof getState === 'function' ? getState() : null;
    let dungeonMap = state ? state.getDungeonMap() : null;

    return (dungeonMap?.mapObjectGroupManager?.getByName(MAP_OBJECT_GROUP_KILLZONE)) || null;
}
```

Wotuu pointed out on that PR that this exact "safely get a map object group by name" shape is
duplicated all over the codebase - a plain `grep` for
`mapObjectGroupManager.getByName(MAP_OBJECT_GROUP_KILLZONE)` alone turns up ~25 call sites
across `models/killzone.js`, `models/enemy.js`, `enemyforces/enemyforcesmanager.js`,
`mapcontrols/*.js`, `inline/common/maps/killzonessidebar.js`,
`inline/common/maps/editsidebar/*.js`, `echo/messagehandler/listen/models/killzone/*.js`, and
more - most without the `getState`/`dungeonMap` null-guards that `publish.js`'s version has
(they can get away with it because they run in a context where a map is guaranteed to exist;
`publish.js` can't assume that).

Wotuu's comment was explicitly a "thinking out loud" observation, not a requirement for #3666 -
see the review thread on PR #3860 for the original discussion. He left it up to whoever picks
this up whether/how to centralize it.

## Suggested approach

- Add a small set of "safely get object group by name" helpers (likely on `util.js` or
  `MapObjectGroupManager` itself) that fold in the same null-safety `publish.js`'s version has,
  and let call sites that need the defensive version opt into it explicitly rather than
  reimplementing the `?.`/`|| null` chain each time.
- This is a pure refactor - no behavior change. Do it incrementally (e.g. one map object group
  type or one directory at a time) rather than as one giant diff, and keep JS test coverage
  green throughout.
- Not all ~25 call sites necessarily need to move - many run in contexts where the map object
  group is guaranteed non-null and a defensive helper would just hide a bug. Judgement call per
  site.

## Definition of done

- A shared helper (or small helper set) exists for safely fetching a map object group by name.
- `publish.js`'s `_getKillZoneMapObjectGroup()` and at least a meaningful subset of the other
  duplicated call sites use it.
- `npx vitest run` stays green.

Wotuu:
:robot: Closes #3874

Follow-up to the review thread on #3860. `publish.js` had grown a private `_getKillZoneMapObjectGroup()` that null-guarded the whole `getState() → getDungeonMap() → mapObjectGroupManager → getByName()` chain; this moves that safety into `util.js` where every caller can reach it.

## What's here

**`getMapObjectGroup(name)`** in `util.js` — the one place that knows how each step of the chain can be absent:

- pages outside a map context have no state manager, and `util.js`'s own fallback `getState()` returns `false` — so `typeof getState === 'function'` is never enough on its own;
- `StateManager.getDungeonMap()` returns `null` until a map registers itself;
- `MapObjectGroupManager.getByName()` returns `false` (not a nullish value) for a group the page suppressed via `hiddenMapObjectGroups`, so the result is normalised with `||`, not `??`.

**Fifteen named accessors** on top of it — `getKillZoneMapObjectGroup()`, `getEnemyMapObjectGroup()`, … — one per entry of `MAP_OBJECT_GROUP_NAMES`, per the review comment:

> Right now, it's hidden away in an inline common Dungeon Route publish file, but it would be nice if all map object groups had access to a similar helper function. […] If there are helper functions to safely get a map object group, I think we should define them globally and accessibly for everyone and not hide them away in a file like this.

Only two of them get a caller in this PR; the rest exist deliberately so the set is complete and usable from a browser console session. The existing `getEnemies()` / `getKillZones()` / … console shorthands are kept and now delegate to the accessors, which makes them null-safe too — plus `getEnemyForcesCheckpoints()`, `getArrows()`, `getFloorUnions()` and `getFloorUnionAreas()`, which were missing.

**Call sites converted** — the ones that already resolved the map through global state, so each is a like-for-like swap:

| File | Now uses |
| --- | --- |
| `inline/common/dungeonroute/publish.js` | `getKillZoneMapObjectGroup()`; the private helper is gone |
| `inline/common/dungeonroute/simulate.js` | `getKillZoneMapObjectGroup()` |
| `inline/common/maps/editsidebar/pullworkbench.js` | `getKillZoneMapObjectGroup()` |
| `inline/common/maps/editsidebar/settingstabpull.js` | `getKillZoneMapObjectGroup()` |
| `mapcontrols/enemyforcescontrols.js` | `getKillZoneMapObjectGroup()` |
| `inline/common/maps/dungeonroutesearchsidebar.js` | `getEnemyMapObjectGroup()` |
| `inline/common/maps/map.js` | `getMapObjectGroup(mapObjectGroupName)` — dynamic name; its `instanceof MapObjectGroup` guard already handled a non-group return |
| `mapcontext/mapcontextdungeonroutesearch.js` | `getMapObjectGroup(name)`, dropping the hoisted manager local (see note below) |

None of these had a `=== false` / `!== false` guard that the switch from `false` to `null` would break; checked per site.

To be precise about what this does and does not buy: only `publish.js` and `map.js` actually *act* on the null — everywhere else the result is dereferenced on the next line, so a missing group still throws, just one line later and with a clearer message. The value here is that the chain is now written once, correctly, in a place any future caller can reach; the remaining five call sites are unchanged in behaviour rather than newly safe.

## What's deliberately *not* here

The other ~145 `getByName()` call sites hold a local map reference (`this.map`, `self.map`, `this.echo.map`, `this.manager`) and are left alone, for two different reasons:

- **Genuinely unsafe to convert**: anything on the `DungeonMap` construction path — `dungeonmap.js`, `mapobjectgroups/*`, `MapObjectGroupManager` itself. `dungeonmap.js:83` calls `state.setDungeonMap(this)` from inside the constructor, so re-resolving through global state there would find the previous map or none.
- **Merely deferred**: the rest (echo message handlers, `killzonessidebar.js`, `mapcontrols/*`, `admin/*`) run long after construction and would convert safely. They are out of scope because the right primitive for a caller that already holds a map is named getters **on `MapObjectGroupManager` itself** — matching the "the map object group manager can know about the models […] but the other way around, there shouldn't be a dependency" point from the same review — not a global that re-resolves through state. A 145-site diff also isn't reviewable in one go. Filed as #3879.

## Notes for review

- **`globalThis.getState` rather than the bare identifier.** `util.js` declares its own no-op `getState()` fallback at the top of the file; that declaration is hoisted, and when `util.js` is loaded as a module (which the Vitest suite does) it shadows the real `getState()` from `statemanager.blade.php`, so the helper would always see `false`. In the browser both are properties of the global object, so the two forms are equivalent there — checked in both build modes, since `concat.mjs` only minifies for production and esbuild rewrites that block: the shipped bundle contains `typeof getState!="function")var getState=function(){return!1}`, and a top-level `var` is a global-object property, so `globalThis.getState` resolves either way and regardless of which script defines it last.
- **`mapcontextdungeonroutesearch.js` re-resolves the map per iteration** instead of hoisting the manager once. Five iterations, so the cost is irrelevant, but it does move where a crash would surface: previously a missing map threw on `.mapObjectGroupManager` before the loop, now each iteration gets `null` and throws on `null.reset()`. Neither is reachable from the one page that builds this map context — `dungeon/dungeonroute/search/gameversion/dungeon.blade.php` hides only `floorunion`/`floorunionarea`, none of the five groups the loop resets.
- **`models/enemy.js:735`** guards with `if (enemyPackMapObjectGroup !== null)` while `getByName()` returns `false`, and its comment claims the group "may be null in certain view modes". It can't be — `enemypack` appears in no page's `hiddenMapObjectGroups`, so the group is always constructed and the guard is unreachable either way. Changing it to `!== false` would turn a currently-impossible crash into a silent skip, which is a behaviour change; left alone and flagged here instead.
- **Test naming.** The new `describe` block uses the project's `[function]_given[Condition]_returns[Result]` convention while the rest of `util.test.js` is prose-style — the inconsistency is deliberate, following CLAUDE.md for new tests rather than retro-fitting the whole file.
- No `MAP_OBJECT_GROUP_MAPICON_AWAKENED_OBELISK` accessor: it is an alias resolved through `MapIconMapObjectGroup`'s multi-name registration, not a group of its own. `PolygonMapObjectGroup` / `PolylineMapObjectGroup` have no constant and no factory entry, so they get none either.

## Verification

- `npx vitest run` — 433 tests across 43 files pass, including the 21 new ones.
- The accessor tests assert against the **real** `MAP_OBJECT_GROUP_*` values (now exported from `constants.js` for the test runner) rather than values the test installs itself, and one test asserts the accessor table covers `MAP_OBJECT_GROUP_NAMES` exactly, so a new group without an accessor fails. Both checks mutation-tested: re-pointing `getArrowMapObjectGroup()` at `MAP_OBJECT_GROUP_BRUSHLINE` fails the pairing test, and dropping an accessor from the table fails the coverage test.
- `npm run dev` builds cleanly (`util.js` still concatenates as entry #2 of `custom-scripts-order.mjs`).
- No PHP touched, so no PHP tests, PHPStan or PHP-CS-Fixer run applies.
- No rendered output changes, so no screenshots.
- Headless-Chrome smoke test of the converted paths (route edit, route view, and a mapless page), logged in as the route's author:
  - **route edit** - `getKillZoneMapObjectGroup()` &rarr; `KillZoneMapObjectGroup`, `getEnemyMapObjectGroup()` &rarr; `EnemyMapObjectGroup`, `getArrowMapObjectGroup()` &rarr; `ArrowMapObjectGroup`, `getMapObjectGroup(MAP_OBJECT_GROUP_MAPICON)` &rarr; `MapIconMapObjectGroup`, console alias `getKillZones()` &rarr; `KillZoneMapObjectGroup`. `getFloorUnionMapObjectGroup()` returns `null` - correct, `edit.blade.php` hides `floorunion`, and that is exactly the case where `getByName()` answers `false`. The publish select's required-enemies refresh ran (all options enabled, tooltip cleared), so the `publish.js` conversion works end to end.
  - **mapless page** - every accessor returns `null` rather than throwing.
  - No uncaught JS errors on any page; the only console error is the pre-existing benign `"undefined" is not valid JSON` from `map.js:658`, which every map page logs over plain http.


resources/assets/js/custom/util.js
```diff
@@ -329,48 +329,205 @@ function removeStrayTooltips() {
...
+ *
+ * getState is reached through globalThis rather than as a bare identifier: the fallback declaration
+ * at the top of this file is hoisted, which shadows the real getState() from statemanager.blade.php
+ * when util.js is loaded as a module (as the Vitest suite does). In the browser both are properties
+ * of the global object, so the two forms are equivalent there.
+ *
+ * @param {string} name One of the MAP_OBJECT_GROUP_* constants.
+ * @returns {MapObjectGroup|null} Null when this page has no such map object group.
+ */
+function getMapObjectGroup(name) {
```
Wotuu, line 351: :robot: The `@returns {X|null}` contract these accessors now advertise is honoured by only two of the eight converted call sites, so the PR body's claim that "Six of the eight gain null-safety they did not have before" doesn't hold.

`getByName()` previously returned `false` for an absent group, and the old expressions dereferenced that `false` directly — so those sites already threw a `TypeError`. Normalising to `null` doesn't make them safe, it just moves the same crash one line later:

- `simulate.js:103` → `killZoneMapObjectGroup.objects`
- `pullworkbench.js:205` → `.setLayerToMapObject(...)`
- `settingstabpull.js:69` → `.applyPullGradient(...)`
- `enemyforcescontrols.js:79` → `.hasKilledAllRequiredEnemies()`
- `mapcontextdungeonroutesearch.js:19` → `getMapObjectGroup(name).reset()`

(Only `publish.js` and `map.js` actually branch on the result.) Worth noting that the one *behavioural* difference is at `simulate.js:104`: `_.values(false.objects)` yields `[]` while `_.values(null.objects)` throws. It's unreachable today — the simulate modal is inside the `@isset($dungeonroute)` block at `common/maps/controls/header.blade.php:350`, and no view that renders a dungeon route hides `killzone` — but it is the one shape where `false` was quietly more forgiving than `null`.

Nothing to change in the code; the ask is to correct the PR body, because the next person adding a caller will reasonably read `{KillZoneMapObjectGroup|null}` plus "six of the eight gain null-safety" as meaning the existing callers already handle `null`.

resources/assets/js/custom/util.test.js
```diff
@@ -242,3 +246,100 @@ describe('isElementFullyVisible', () => {
...
+        // Arrange
+        let killZoneMapObjectGroup = {name: 'killzone'};
+        let getByName = stubMapWithGroup(killZoneMapObjectGroup);
+
+        // Act
+        let result = getKillZoneMapObjectGroup();
+
+        // Assert
+        expect(result).toBe(killZoneMapObjectGroup);
+        expect(getByName).toHaveBeenCalledWith('killzone');
```
Wotuu: :robot: These two accessor tests can only catch a killzone↔enemy swap: they assert `getByName` was called with the very value the test itself installed into `globalThis.MAP_OBJECT_GROUP_*`. The other thirteen accessors have no test at all, and a mis-paired one (e.g. `getFloorUnionMapObjectGroup()` returning `MAP_OBJECT_GROUP_FLOOR_UNION_AREA`) would pass silently — the whole point of adding 15 near-identical wrappers is that a copy-paste slip in one of them is invisible.

I checked all 15 pairings against `constants.js` by hand and they are correct, so this is a suite gap rather than a bug. A single data-driven test over the accessor→constant pairs (exporting the accessors and looping) would cover the set for roughly the cost of the two hand-written cases.

Wotuu:

resources/assets/js/custom/util.js
Wotuu: :robot: Correct, and the PR body overstated it — fixed. Only `publish.js` and `map.js:669` actually branch on the result; the other five dereference on the next line, so a missing group still throws, just one line later. Reworded to say the five converted sites are behaviour-unchanged rather than newly safe, and that the value is having the chain written once in a reachable place. Also confirmed your reachability point about `simulate.js:104` independently: the simulate modal is inside `@isset($dungeonroute)` and no dungeon-route view hides `killzone`.

resources/assets/js/custom/util.test.js
Wotuu: :robot: Fair — the tests were asserting against values they had installed themselves, which proves nothing. Fixed in b6e18f939: `constants.js` now exports the `MAP_OBJECT_GROUP_*` constants and `MAP_OBJECT_GROUP_NAMES` through its existing test-runner export block, and the accessor tests are data-driven over an `[accessor, realConstant]` table, so all 15 are covered with the real values. Added `getMapObjectGroup_givenEveryRegisteredGroupName_hasAnAccessor`, which asserts the table matches `MAP_OBJECT_GROUP_NAMES` exactly, so adding a group without an accessor fails.

Mutation-tested both: re-pointing `getArrowMapObjectGroup()` at `MAP_OBJECT_GROUP_BRUSHLINE` fails the pairing test, and removing an accessor from the table fails the coverage test. Suite is 433 green.

PR #3880, opened 2026-08-07
#3875 Scope Team::getAvailableTags() by owning team context

### LINKED ISSUE #3875: Team::getAvailableTags() can leak another team's tags on a shared route
:robot: Follow-up split off from [PR #3867](https://github.com/RaiderIO/keystone.guru/pull/3867#discussion_r3735198318) at Wotuu's request.

`Team::getAvailableTags()` builds its query without scoping by `context_class`/`context_id`:

```php
public function getAvailableTags(): EloquentCollection
{
    return Tag::where('tag_category_id', TagCategory::ALL[TagCategory::DUNGEON_ROUTE_TEAM])
        ->whereIn('model_id', $this->dungeonRoutes->pluck('id'))
        ->get();
}
```

For a `DUNGEON_ROUTE_TEAM` tag, `context_id`/`context_class` identify the *owning* team and `model_id`/`model_class` identify the tagged `DungeonRoute`. Because this query only filters by `model_id` (the route), if the same route is shared by/tagged from two different teams, `getAvailableTags()` can surface the *other* team's tags as if they were this team's own.

The fix (established in #3867 for the `Team::boot()` deleting hook and `removeMember()`) is to also scope on `context_class = Team::class` and `context_id = $this->id`:

```php
public function getAvailableTags(): EloquentCollection
{
    return Tag::where('tag_category_id', TagCategory::ALL[TagCategory::DUNGEON_ROUTE_TEAM])
        ->whereIn('model_id', $this->dungeonRoutes->pluck('id'))
        ->get();
}
```

Deliberately left untouched in #3867 since it's a separate call site with its own test coverage to write.

Wotuu:
:robot: Closes #3875

`Team::getAvailableTags()` filtered tags only by `model_id` (the tagged `DungeonRoute`), not by the tag's owning-team context (`context_class`/`context_id`). If a route is shared by or moves between teams, another team's `DUNGEON_ROUTE_TEAM` tag on that same route would leak into this team's available list.

Fixes it the same way #3867 fixed the sibling `boot()`/`removeMember()` call sites: also scope by `context_class = Team::class` and `context_id = $this->id`.

## Test plan
- [x] New `TeamTest::getAvailableTags_givenOwnTeamsTaggedRoute_returnsTheTag`
- [x] New `TeamTest::getAvailableTags_givenAnotherTeamsTagOnASharedRoute_excludesIt`

## Cold review
1 finding — pre-existing gap where `whereIn('model_id', ...)` never matches tag *definitions* (`model_id IS NULL`), unrelated to this fix's `context_class`/`context_id` scoping. Filed as follow-up #3881 rather than fixed here.

### DIFF
--- a/app/Models/Team.php
+++ b/app/Models/Team.php
@@ -454,6 +454,8 @@ class Team extends Model
     public function getAvailableTags(): EloquentCollection
     {
         return Tag::where('tag_category_id', TagCategory::ALL[TagCategory::DUNGEON_ROUTE_TEAM])
+            ->where('context_class', self::class)
+            ->where('context_id', $this->id)
             ->whereIn('model_id', $this->dungeonRoutes->pluck('id'))
             ->get();
     }
--- a/tests/Feature/App/Models/TeamTest.php
+++ b/tests/Feature/App/Models/TeamTest.php
@@ -1,0 +1,126 @@
+<?php
+
+namespace Tests\Feature\App\Models;
+
+use App\Models\DungeonRoute\DungeonRoute;
+use App\Models\Tags\Tag;
+use App\Models\Tags\TagCategory;
+use App\Models\Team;
+use App\Models\TeamUser;
+use App\Models\User;
+use PHPUnit\Framework\Attributes\Group;
+use PHPUnit\Framework\Attributes\Test;
+use Tests\TestCases\PublicTestCase;
+
+#[Group('Team')]
+final class TeamTest extends PublicTestCase
+{
+    #[Test]
+    public function getAvailableTags_givenOwnTeamsTaggedRoute_returnsTheTag(): void
+    {
+        $author = null;
+        $team   = null;
+        $route  = null;
+        $tag    = null;
+
+        try {
+            // Arrange
+            $author = User::factory()->create();
+            $team   = $this->createTeam();
+            $team->addMember($author, TeamUser::ROLE_MEMBER);
+            $route = DungeonRoute::factory()->create([
+                'author_id' => $author->id,
+                'team_id'   => $team->id,
+            ]);
+            $tag = $this->createTeamTag($team, $route);
+
+            // Act
+            $availableTags = $team->getAvailableTags();
+
+            // Assert
+            $this->assertTrue($availableTags->contains('id', $tag->id));
+        } finally {
+            $this->cleanUp($tag, $route, $team, $author);
+        }
+    }
+
+    #[Test]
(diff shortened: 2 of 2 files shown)


app/Models/Team.php
```diff
@@ -454,6 +454,8 @@ public static function getRaiderIOTeam(): ?Team
     public function getAvailableTags(): EloquentCollection
     {
         return Tag::where('tag_category_id', TagCategory::ALL[TagCategory::DUNGEON_ROUTE_TEAM])
+            ->where('context_class', self::class)
+            ->where('context_id', $this->id)
```
Wotuu, line 458: :robot: The context scoping itself is correct — `AjaxTagController::store()` is the only writer of `DUNGEON_ROUTE_TEAM` tags and always sets `context_class = Team::class`, and a check against the dev DB confirms all 322 `tag_category_id = 2` rows carry exactly `App\Models\Team` (no legacy namespace / `User::class` strays), so nothing legitimate gets hidden by the new clauses.

One question about completeness while this query is being touched: `TeamController::createTag()` → `Tag::saveFromRequest()` creates team tag *definitions* with `model_id = NULL`, and `whereIn('model_id', ...)` never matches NULL — so tags a team admin creates in the tag manager never show up in this "available tags" autocomplete. The same dev DB has 6 such rows today, so it's not hypothetical. Now that `context_class`/`context_id` scope ownership properly, the `model_id` clause is doing much less work; dropping it, or relaxing it to

```php
->where(fn($q) => $q->whereNull('model_id')->orWhereIn('model_id', $this->dungeonRoutes()->select('id')))
```

would surface those definitions. Is that intentionally out of scope? If so, worth a line in the PR body.

(Drive-by on the same clause, non-blocking: `$this->dungeonRoutes->pluck('id')` hydrates every `DungeonRoute` model of the team just to get ids, and produces an unbounded `IN (...)` list — a `->select('id')` subquery on the relation would avoid both. Pre-existing, not introduced here.)

Wotuu:

app/Models/Team.php
Wotuu: :robot: Good catch, and confirmed pre-existing (unrelated to this PR's `whereIn('model_id', ...)` clause — it was there before and is unchanged here). Filed as #3881 to fix separately, since it's a distinct bug (definitions never surfacing) from the leak this issue/PR addresses, and needs its own look at call sites + test coverage rather than a drive-by change here.

PR #3882, opened 2026-08-07
AjaxProfileController 500s for guests: guard with auth middleware

### LINKED ISSUE #3863: AjaxProfileController 500s for guests: Auth::user() passed into non-nullable User parameters on unauthenticated routes
:robot: Found during the #3673 audit sweep.

`AjaxProfileController` passes `Auth::user()` straight into non-nullable `User` parameters on two routes that carry no `Authenticate` middleware, so an unauthenticated request produces a `TypeError` (HTTP 500) instead of a 401/403.

## Where

`app/Http/Controllers/Ajax/AjaxProfileController.php:17`

```php
public function legalAgree(Request $request): Response
{
    $time = $request->get('time', -1);

    /** @var User $user */
    $user = Auth::user();
    $user->update([...]);   // $user is null for a guest
```

`app/Http/Controllers/Ajax/AjaxProfileController.php:31`

```php
public function addAdFreeGiveaway(Request $request, User $user): PatreonAdFreeGiveaway
{
    /** @var User $currentUser */
    $currentUser = Auth::user();

    if (PatreonAdFreeGiveaway::getCountLeft($currentUser) <= 0) {
```

`PatreonAdFreeGiveaway::getCountLeft()` is declared `getCountLeft(User $user): int` (`app/Models/Patreon/PatreonAdFreeGiveaway.php:39`), so a `null` argument is a `TypeError`, not a falsy-check.

## Why it is reachable

`POST ajax/profile/legal` and `POST ajax/profile/adfree/{user}` sit in the `ajax` prefix group (`routes/web.php:614-616`), which has no `Authenticate` middleware:

```
POST ajax/profile/legal      web, ViewCacheBuster, SetLocaleFixed, DebugBarMessageLogger,
                             ReadOnlyMode, DebugInfoContextLogger, TracksUserIpAddress, OnlyAjax
```

That group being open is deliberate — sandbox routes need to work for guests — but these two methods assume a user is present.

The `/** @var User $user */` annotations are what keep PHPStan quiet about it: they assert the non-null type the analyser would otherwise flag from `Auth::user(): ?Authenticatable`.

## Impact

Fails closed — nothing is written and nothing leaks. The cost is a 500 where a 401 belongs, plus Sentry noise from any crawler or stale tab that hits either endpoint after a session expires (`legalAgree` in particular is fired from the cookie-consent banner, so an expired session on an open tab is enough).

## Suggested fix

Either add `auth` middleware to these two routes, or return a 401 explicitly when `Auth::user()` is `null` — whichever matches how the cookie-consent banner is expected to behave for guests. `legalAgree` recording consent for a logged-out visitor may not be meaningful at all, in which case middleware is the simpler answer.

Worth a test asserting a guest gets a non-500 response on both.

Wotuu:
:robot: Closes #3863

## Summary
- `legalAgree` and `addAdFreeGiveaway` passed `Auth::user()` straight into non-nullable `User` parameters on routes with no auth middleware, so a guest request produced a `TypeError` (500) instead of a 401. Both routes are now gated behind `['auth', 'role:user|admin']`, matching the convention already used elsewhere in `routes/web.php`. `removeAdFreeGiveaway` is left alone — its `Gate::authorize()` call already denies guests cleanly.
- That alone still 500ed: the custom exception `Handler::render()` never special-cased `AuthenticationException`, so it fell through every `instanceof` branch to the generic 500 tail instead of `unauthenticated()`'s 401 JSON response. This is a latent bug affecting **any** auth-gated `/ajax/` or `/api/` route hit by a guest, not just these two — fixed at the root.

## Test plan
- [x] New feature test: guest requests to both routes return 401 instead of 500
- [x] `php artisan test --compact tests/Feature/Controller/Ajax/AjaxProfileControllerTest.php` passes
- [x] `php artisan test --compact tests/Feature/Controller/Ajax` — same 6 pre-existing failures as on master (Reverb connectivity in this worktree, unrelated to this change), no new failures
- [x] `composer run fix` — no changes

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -112,6 +112,12 @@ class Handler extends ExceptionHandler
                         'model' => $e->getModel(),
                     ]),
                 ], StatusCode::NOT_FOUND);
+            } elseif ($e instanceof AuthenticationException) {
+                // AuthenticationException is not an HttpExceptionInterface, so without this check it
+                // falls through every instanceof branch below and hits the generic 500 tail instead
+                // of the 401 unauthenticated() already builds - a guest hitting any auth-gated
+                // /ajax/ or /api/ route would 500 instead of 401 (#3863).
+                return $this->unauthenticated($request, $e);
             }
 
             // Normalize the same way parent::render() would (AuthorizationException -> 403,
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -167,7 +167,10 @@ class Handler extends ExceptionHandler
     protected function unauthenticated($request, AuthenticationException $exception)
     {
         if ($this->shouldReturnJson($request, $exception)) {
-            return response()->json(['error' => __('exceptions.handler.unauthenticated')], StatusCode::UNAUTHORIZED);
+            // defaultAjaxErrorFn() (resources/assets/js/custom/inline/layouts/app.js) only reads
+            // responseJSON.errors then responseJSON.message - 'error' was silently dropped, falling
+            // back to the generic "An error occurred" toast.
+            return response()->json(['message' => __('exceptions.handler.unauthenticated')], StatusCode::UNAUTHORIZED);
         }
 
         return redirect()->guest('login');
--- a/routes/web.php
+++ b/routes/web.php
@@ -619,8 +619,10 @@ Route::middleware(['viewcachebuster', 'language', 'debugbarmessagelogger', 'read
             Route::post('/mdt/details', new MDTImportController()->details(...))->name('mdt.details');
         });
 
-        Route::post('/profile/legal', new AjaxProfileController()->legalAgree(...));
-        Route::post('/profile/adfree/{user:public_key}', new AjaxProfileController()->addAdFreeGiveaway(...));
+        Route::middleware(['auth', 'role:user|admin'])->group(static function () {
+            Route::post('/profile/legal', new AjaxProfileController()->legalAgree(...));
+            Route::post('/profile/adfree/{user:public_key}', new AjaxProfileController()->addAdFreeGiveaway(...));
+        });
         Route::delete('/profile/adfree/{user:public_key}', new AjaxProfileController()->removeAdFreeGiveaway(...));
 
         // Metrics
--- a/routes/web.php
+++ b/routes/web.php
@@ -619,8 +619,13 @@ Route::middleware(['viewcachebuster', 'language', 'debugbarmessagelogger', 'read
             Route::post('/mdt/details', new MDTImportController()->details(...))->name('mdt.details');
         });
 
-        Route::middleware(['auth', 'role:user|admin'])->group(static function () {
+        // legalAgree only mutates the caller's own record, so plain 'auth' is enough - a role-less
+        // authenticated user (e.g. the seeded Internal Team account) must still be able to clear the
+        // legal modal, which re-shows on every page until legal_agreed is set.
+        Route::middleware('auth')->group(static function () {
             Route::post('/profile/legal', new AjaxProfileController()->legalAgree(...));
+        });
+        Route::middleware(['auth', 'role:user|admin'])->group(static function () {
             Route::post('/profile/adfree/{user:public_key}', new AjaxProfileController()->addAdFreeGiveaway(...));
         });
         Route::delete('/profile/adfree/{user:public_key}', new AjaxProfileController()->removeAdFreeGiveaway(...));
--- a/tests/Feature/Controller/Ajax/AjaxProfileControllerTest.php
+++ b/tests/Feature/Controller/Ajax/AjaxProfileControllerTest.php
@@ -27,6 +27,27 @@ final class AjaxProfileControllerTest extends PublicTestCase
     }
 
     #[Test]
+    public function legalAgree_givenAuthenticatedUserWithoutRole_isAllowedToAgree(): void
+    {
+        // Arrange - e.g. the seeded Internal Team account has no 'user'/'admin' role, but must
+        // still be able to clear its own legal modal (it re-shows on every page until agreed)
+        $roleLessUser = User::factory()->create();
+
+        try {
+            $this->actingAs($roleLessUser);
+
+            // Act
+            $response = $this->post('/ajax/profile/legal', ['time' => 123], self::AJAX_HEADERS);
+
+            // Assert
+            $response->assertNoContent();
+            $this->assertSame(1, $roleLessUser->fresh()->legal_agreed);
+        } finally {
+            $roleLessUser->delete();
+        }
+    }
+
+    #[Test]
     public function addAdFreeGiveaway_givenGuest_returnsUnauthorizedInsteadOf500(): void
     {
         // Arrange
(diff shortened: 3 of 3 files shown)


routes/web.php
```diff
@@ -619,8 +619,10 @@
             Route::post('/mdt/details', new MDTImportController()->details(...))->name('mdt.details');
         });
 
-        Route::post('/profile/legal', new AjaxProfileController()->legalAgree(...));
-        Route::post('/profile/adfree/{user:public_key}', new AjaxProfileController()->addAdFreeGiveaway(...));
+        Route::middleware(['auth', 'role:user|admin'])->group(static function () {
```
Wotuu: :robot: **`role:user|admin` is too narrow for `/profile/legal`, and can permanently brick the legal modal.**

`legalAgree()` only mutates `Auth::user()`'s own record — the role adds no authorization value here, it only narrows who can call it. And there are authenticated users that fail `role:user|admin`:

- The seeded `Internal Team` user (id 2) has **only** the `internal_team` role — I checked the local DB: `1 Admin [admin]`, `2 Internal Team [internal_team]`, `3 User [user]`.
- `UserFactory` attaches no roles at all, so `User::factory()->create()` produces a role-less user.
- `OAuthLoginController:118` backfills `ROLE_USER` onto *existing* users — that backfill only exists because role-less accounts exist in production.

Consequence for this specific route: `sitescripts.blade.php` shows the legal modal on every page while `!legal_agreed`, and `_agreeLegalBtnClicked()` has **no `error` handler** — a 403 just re-enables the button. Such a user is stuck behind the modal forever with no way to ever set `legal_agreed`. The `role:user|admin` convention elsewhere in this file gates *features* (live sessions, cloning), where being locked out is recoverable; here it isn't.

Suggest `Route::middleware('auth')` for `/profile/legal` and keeping `['auth', 'role:user|admin']` for `/profile/adfree/{user}` (where `PatreonAdFreeGiveaway::getCountLeft($currentUser)` is the real gate anyway, so the role is harmless).

Related: the new test only covers the guest path, so the newly-added role gate is entirely unexercised. Worth adding a happy-path test asserting an authenticated user still succeeds — and note it must call `addRole(Role::ROLE_USER)` explicitly, because a plain `actingAs(User::factory()->create())` will 403 and look like a controller bug.

app/Exceptions/Handler.php
```diff
@@ -112,6 +112,12 @@ public function render($request, Throwable $e)
                         'model' => $e->getModel(),
                     ]),
                 ], StatusCode::NOT_FOUND);
+            } elseif ($e instanceof AuthenticationException) {
```
Wotuu, line 115: :robot: **The 401 body this now routes to uses `error`, not `message` — the frontend can't read it.**

`unauthenticated()` (just below) returns `['error' => __('exceptions.handler.unauthenticated')]`, while every other JSON path in `render()` returns `['message' => ...]`. That inconsistency was mostly unreachable before; this branch newly routes the *entire* auth-gated `/ajax/` + `/api/` surface through it, which is exactly the surface parsed by JS.

`defaultAjaxErrorFn()` (`resources/assets/js/custom/inline/layouts/app.js:137-153`) reads `responseJSON.errors` then `responseJSON.message`, and its status switch has cases for 403/404/419 but not 401. So a guest now gets the toast `401: An error occurred while performing your request. Please try again.` — the actual "you are not logged in" string is silently dropped. Not a crash, but the fix loses the one piece of information that makes the 401 actionable.

Cheapest fix is to emit both keys (or rename to `message`) in `unauthenticated()`; optionally add a `case 401` to the JS switch. Either way it's worth doing in this PR since this diff is what makes the path live.

Wotuu:

routes/web.php
Wotuu: :robot: Fixed — split into two groups: `legalAgree` now only requires `auth` (it only mutates the caller's own record), `addAdFreeGiveaway` keeps `['auth', 'role:user|admin']`. Added tests for both: a role-less authenticated user can still agree to the legal terms, and is forbidden from the ad-free giveaway. See 101d32880.

app/Exceptions/Handler.php
Wotuu: :robot: Fixed — changed the key from `error` to `message` to match every other JSON error response in the Handler (and what `defaultAjaxErrorFn()` actually reads). See 101d32880.

Wotuu:
:robot: Cold review: 2 findings

tests/Feature/Controller/Ajax/AjaxProfileControllerTest.php
```diff
@@ -0,0 +1,91 @@
...
+
+            // Assert
+            $response->assertStatus(StatusCode::UNAUTHORIZED);
+        } finally {
+            $target->delete();
+        }
+    }
+
+    #[Test]
+    public function addAdFreeGiveaway_givenAuthenticatedUserWithoutRole_isForbidden(): void
```
Wotuu: A roleless user does not exist, so remove this test.

tests/Feature/Controller/Ajax/AjaxProfileControllerTest.php
```diff
@@ -0,0 +1,91 @@
...
+
+            // Assert
+            $response->assertNoContent();
+            $this->assertSame(1, $roleLessUser->fresh()->legal_agreed);
+        } finally {
+            $roleLessUser->delete();
+        }
+    }
+
+    #[Test]
```
Wotuu, line 52: There is now no happy path for when you are allowed to do an ad-free giveaway. I'd like to see a test that returns 200. Remember that for an ad-free giveaway to work, you need to have your Patreon account connected and be able to do this action in the first place by being in the higher tier.

tests/Feature/Controller/Ajax/AjaxProfileControllerTest.php
Wotuu: :robot: Removed the `addAdFreeGiveaway_givenAuthenticatedUserWithoutRole_isForbidden` test entirely — confirmed `User::factory()->create()` never auto-assigns a role (only registration/login controllers attach roles), so this scenario has no reachable real-world counterpart. Remaining tests in the file pass (3 passed, 5 assertions).

tests/Feature/Controller/Ajax/AjaxProfileControllerTest.php
Wotuu: :robot: Added `addAdFreeGiveaway_givenAdminGiver_returnsCreatedAndCreatesGiveaway`. It acts as an admin user (satisfies the route's `role:user|admin` middleware, and `User::hasPatreonBenefit()` grants `AD_FREE_TEAM_MEMBERS` automatically to admins, which is what `PatreonAdFreeGiveaway::getCountLeft()` checks) and posts to the route for a fresh target user.

One correction on the expected status: the controller returns the created `PatreonAdFreeGiveaway` model directly, so Laravel auto-responds `201 Created`, not `200`. The test asserts `assertCreated()` (201) plus that the `PatreonAdFreeGiveaway` row was actually created with the right giver/receiver and that `$target->hasAdFreeGiveaway()` is true afterward.

Verified locally: `php artisan test tests/Feature/Controller/Ajax/AjaxProfileControllerTest.php` — 4 passed. Pushed as `b865dbb94`.

PR #3883, opened 2026-08-07
#3551 Rewrite seed-dev-routes for the current thumbnail flow

### LINKED ISSUE #3551: Update seed-dev-routes skill for the #3548 thumbnail flow (follow-up to #3548, blocked on #3447)
:robot: ## Background

`.claude/skills/seed-dev-routes/SKILL.md` only exists on the still-open #3447 branch
(`3349-route-poster-card`) — it was never on `master`, so it wasn't touched by #3550 (which
closed #3548). That skill documents local thumbnail generation using the *old*, now-removed
workflow:

- Step 2 says to hard-disable the env guard by editing source (`false &&` prefix on
  `app()->environment('local')`) and `git checkout --` it afterwards.
- Step 2 overrides `APP_URL=http://nginx` per-run to make puppeteer's preview URL reachable.
- Step 3 ("Make the thumbnails actually serve") symlinks `public/storage ->
  storage/app/private`, `chmod -R a+rX`s the thumbnail dirs after every generation batch, and
  works around the `local` disk's signed-URL requirement.

#3550 replaced all of this:

- The guard is now a disk-safety check (`ThumbnailService::isRemoteDiskUnsafeForLocalGeneration`)
  that only blocks local+S3, not local+public/local — no more editing source to generate
  locally.
- Local dev now uses the framework `public` disk (`FILESYSTEM_DISK=public` in
  `.env.docker.example`), served via the standard `storage:link` symlink
  (`public/storage -> storage/app/public`) — no signed-URL problem, no chmod hack, no private-disk
  symlink.
- The preview URL is now `config('keystoneguru.thumbnail.preview_base_url')`
  (`THUMBNAIL_PREVIEW_BASE_URL=http://nginx` in `.env.docker.example`) instead of an `APP_URL`
  override.
- New dev-only command `php artisan dungeonroute:migratethumbnaildisk` migrates any existing
  `disk=local` thumbnail rows to `public`.

## Remaining work

Once #3447 merges (so the skill file actually exists on `master`), rewrite
`.claude/skills/seed-dev-routes/SKILL.md` step 2 and step 3:

- Drop the "temporarily edit the guard" instructions entirely.
- Drop the `APP_URL=http://nginx` per-run override; reference `THUMBNAIL_PREVIEW_BASE_URL`
  instead (already set correctly in `.env.docker.example` after #3550, so likely nothing to do
  per-run at all).
- Drop step 3 ("Make the thumbnails actually serve") — the symlink/chmod workaround is no longer
  needed with the `public` disk. Just needs `storage:link`, which worktrees already run at init.
- Keep the environmental prereqs that are unrelated to this fix: node_modules hardlink-copy,
  chrome-headless-shell + apt deps, and the `ASSETS_BASE_URL_INTERNAL`/`DEBUGBAR_ENABLED`
  per-run toggles (still needed — separate, pre-existing env gap, see #3548's verification
  notes).

## Why it's blocked, not just deferred

`.claude/skills/seed-dev-routes/SKILL.md` doesn't exist on `master` yet — it only lives on the
`3349-route-poster-card` branch (#3447). There's nothing to edit on `master` until that branch
merges.

## Definition of done

- SKILL.md step 2/3 rewritten per above, verified against a real worktree by seeding a route and
  generating its thumbnail with no source edits and no `APP_URL` override.
- `generate_thumbnails.php` (same directory) reviewed for any hardcoded assumptions tied to the
  old flow (e.g. disk name, guard workaround) and updated if needed.

Related: #3548, #3550, #3447

Wotuu:
:robot: Closes #3551

## What changed

`.claude/skills/seed-dev-routes` documented a thumbnail workflow that no longer exists. Steps 2
and 3 told the agent to edit the `ThumbnailService` guard in source, copy `chrome-headless-shell`
into the `app` container with an `apt-get` dependency block, override `APP_URL=http://nginx`,
`sed -i` the `.env`, symlink `public/storage` at the private disk and `chmod -R a+rX` the output.
Every one of those is now explicitly listed as forbidden by the `generating-thumbnails` skill.

- **SKILL.md**: step 2 rewritten to delegate to `generating-thumbnails` (Path B —
  `dungeonroute:queuethumbnail`), step 3 deleted entirely, plus a new "Where to run it" section.
- **`generate_thumbnails.php`: deleted.** It calls `ThumbnailService::createThumbnail()`
  synchronously via tinker in the `app` container — which has deliberately had no Chrome since
  the render moved to the worker image, so the script cannot work at all. This goes beyond the
  issue's "reviewed and updated if needed"; there is nothing left to update it into that
  `dungeonroute:queuethumbnail` doesn't already do.
- **`seed_routes.php`**: docblock now says where to run it (see below); no behaviour change.

## Beyond / against the issue text

- The issue only asked to drop the guard edit, the `APP_URL` override and step 3. The
  `.chrome-tmp` install block and `generate_thumbnails.php` itself are equally dead, so they went
  too.
- The issue asked to **keep** the `ASSETS_BASE_URL_INTERNAL`/`DEBUGBAR_ENABLED` per-run toggles.
  They are dropped as a recipe and kept as prose. Two reasons: the mechanism was `sed -i` on
  `.env`, which I'm instructed never to touch; and under Path B the render happens in the *main
  stack's* Horizon reading the *main checkout's* `.env`, so a `sed` run from a worktree edits the
  wrong file and does nothing. The knowledge (dev renders show the DebugBar and black tiles)
  survives as a gotcha bullet.
- **New**: seeding must happen in the main checkout or a `--shared-db` worktree. A default
  worktree has its own schema and redis prefix, so the shared Horizon sees neither the route rows
  nor the dispatch — you'd get routes that can never get a thumbnail. This wasn't in the issue and
  is the trap the old runbook's local-generation hack used to paper over.
- **New**: a wedged-Horizon gotcha (queue grows, nothing renders, `docker compose restart horizon`
  drains it) — hit during verification below.

## Verification

Full end-to-end run of the rewritten runbook in a `--shared-db` worktree, no source edits, no
`APP_URL` override, no `.env` changes:

1. `seed_routes.php` with `ROUTE_COUNT=2` on `pitofsaron` → `CREATED: 8JxT3PE,pw7FaEt`.
2. `dungeonroute:queuethumbnail <key> --force` per key. Nothing rendered — Horizon was wedged on a
   stale redis connection with 3 jobs on `local-thumbnail`; `docker compose restart horizon`
   drained it. (This is the new gotcha bullet.)
3. Both routes got a `standard` thumbnail on the **`public`** disk, and both served `200` from the
   main stack (`:8008`) *and* the worktree (`:8112`).
4. Seeded routes deleted afterwards (shared DB).

No code changed, so no test applies and there is no rendered output — visual verification is N/A.

## Unrelated test fix riding along

`47b0dba` fixes a flaky test that has nothing to do with this docs diff, per the "fix red CI in the
current MR, don't defer it" rule. `DungeonRoutePolicyTest::publish_givenOwner_returnsAllowed` built
a route with no pulls on whatever dungeon `DungeonRouteFactory`'s `inRandomOrder()` picked; 7 of the
75 eligible seeded dungeons carry required enemies, so ~1 run in 11 the publish guard denied and the
assertion failed — on **every** PR, not just this one. `publish_givenNonOwnerNonAdmin_returnsDenied`
had the same dependency in the passing direction (denied for the required-enemy reason rather than
the ownership one it tests). Both now use the empty mapping version the sibling helper already built.
Verified by pinning the factory to `kingsrest`: the old test fails at the line CI reported, the new
one passes.

Both cold reviews are done and every finding is resolved (3 + 1).


.claude/skills/seed-dev-routes/SKILL.md
```diff
@@ -21,82 +28,44 @@ docker compose exec -T -e DUNGEON_KEY=pitofsaron -e ROUTE_COUNT=12 app \
...
+  has no tiles). That is a known dev env gap affecting every render path, not a seeding
+  problem — fine for exercising card layout.
 
 ## Cleanup / gotchas
 
-- Revert the ThumbnailService guard edit (`git checkout -- app/Service/DungeonRoute/ThumbnailService.php`).
-- Routes are shared-DB data visible to every stack; delete via
-  `DungeonRoute::whereIn('public_key', [...])` teardown if they should go away again.
+- Routes live in the **shared** database and are visible on every stack — seed a small batch,
+  and delete them when done: `DungeonRoute::whereIn('public_key', [...])->delete()`.
```
Wotuu: :robot: `DungeonRoute::whereIn("public_key", [...])->delete()` is a **mass delete** — it goes through the query builder and never fires the `static::deleting` hook at `app/Models/DungeonRoute/DungeonRoute.php:1473`. That hook is what deletes the route's `DungeonRouteThumbnail` + `File` rows (and the files on disk), kill zones, brushlines, paths, arrows, affix groups, tags, live sessions and the combatlog `challengeModeRun`.

Following this cleanup line verbatim therefore leaves orphaned rows in the **shared** DB and orphaned JPGs on the **shared `public` disk** that is bind-mounted into every worktree — exactly the cross-stack pollution the rest of this rewrite is built to prevent. The old text stopped at `DungeonRoute::whereIn(...)` "teardown"; this diff is what makes the broken form concrete.

Suggest: `DungeonRoute::whereIn("public_key", [...])->get()->each->delete()` (`->get()` explicitly, so the models are hydrated and each `delete()` fires the hook).

.claude/skills/seed-dev-routes/SKILL.md
```diff
@@ -21,82 +28,44 @@ docker compose exec -T -e DUNGEON_KEY=pitofsaron -e ROUTE_COUNT=12 app \
...
-sed -i 's|^ASSETS_BASE_URL_INTERNAL=.*|ASSETS_BASE_URL_INTERNAL=https://assets.keystone.guru|; s|^DEBUGBAR_ENABLED=.*|DEBUGBAR_ENABLED=false|' .env
-
-docker compose exec -T \
-  -e APP_URL=http://nginx \
-  -e PUPPETEER_EXECUTABLE_PATH=/var/www/.chrome-tmp/chrome-headless-shell-linux64/chrome-headless-shell \
-  -e THUMBNAIL_KEYS=<keys from step 1> \
-  app php artisan tinker .claude/skills/seed-dev-routes/generate_thumbnails.php
-
-sed -i 's|^ASSETS_BASE_URL_INTERNAL=.*|ASSETS_BASE_URL_INTERNAL=http://app-assets|; s|^DEBUGBAR_ENABLED=.*|DEBUGBAR_ENABLED=true|' .env
+for key in <keys from step 1's CREATED: line>; do
```
Wotuu: :robot: The copy-paste path here is broken: step 1 prints `CREATED: 8JxT3PE,pw7FaEt` (comma-separated — see the `CREATED:` line in `seed_routes.php`), but `for key in ...` splits on whitespace. Pasting the keys as printed makes the loop run **once** with `key=8JxT3PE,pw7FaEt`, and `dungeonroute:queuethumbnail` errors with "No dungeon route found with public key ...".

It fails loudly rather than silently, but it is the literal documented invocation. Either say "space-separated" explicitly, or make the loop tolerant, e.g. `for key in $(echo "<CREATED list>" | tr "," " "); do`.

.claude/skills/seed-dev-routes/SKILL.md
```diff
@@ -1,14 +1,21 @@
...
-map thumbnails for them, entirely against the local dev stack. Proven end-to-end on
-PR #3447's worktree (12 Pit of Saron routes, 2026-07-14).
+`seed_routes.php` in this directory is a tinker script that creates demo `DungeonRoute`s
+(published `world`, `expires_at = null`, random pack-based pulls, varied views/rating/forces).
+Thumbnails are **not** this skill's job — once the routes exist, hand them to the
+[[generating-thumbnails]] skill.
+
+## Where to run it
+
+Seed from the **main checkout**, or from a worktree created with `sh/worktree.sh create ...
```
Wotuu: :robot: The restriction is stated one scope too wide: it is a *thumbnail* constraint, but it is written as a *seeding* constraint (and repeated in the `seed_routes.php` docblock).

Seeding into a default isolated worktree is perfectly fine when the task only needs route/list demo data — the private schema has enemy packs and mapping versions, and a thumbnail-less route just falls back to the dungeon image (`generating-thumbnails` says so). Only step 2 actually requires the main checkout or `--shared-db`.

As written it also directly contradicts `.claude/skills/worktree-docker/SKILL.md:118`, which points isolated-worktree users at *this* skill: "A freshly seeded schema has **no dungeon routes** — use the `seed-dev-routes` skill when a task needs route/demo data." An agent in an isolated worktree now gets told to use this skill and then told it cannot.

Suggest rewording to something like: seeding works anywhere; **thumbnails** require the main checkout or a `--shared-db` worktree, because the shared Horizon sees neither the worktree schema nor its redis prefix (confirmed: `sh/worktree.sh` only sets `REDIS_PREFIX`/`DB_DATABASE` when `shared_db == 0`).

Wotuu:

.claude/skills/seed-dev-routes/SKILL.md
Wotuu: :robot: Fixed in 37769ff — the cleanup snippet is now `->get()->each->delete()`, with a sentence explaining that a query-builder mass delete fires no model events and orphans the pulls/paths/tags/thumbnail rows and the JPGs on the shared public disk.

.claude/skills/seed-dev-routes/SKILL.md
Wotuu: :robot: Fixed in 37769ff — the loop now splits the comma-separated `CREATED:` line with `tr ',' ' '`.

.claude/skills/seed-dev-routes/SKILL.md
Wotuu: :robot: Fixed in 37769ff — step 1 now says it runs anywhere, and the main-checkout/`--shared-db` requirement moved into step 2, with an explicit note that seeding in an isolated worktree is fine (routes just fall back to the dungeon image). The `seed_routes.php` docblock was rescoped the same way.

Wotuu:
:robot: **Back to draft** — I undrafted this too early and then kept pushing to it, which is exactly the window where `babysit-prs` can grab a PR mid-flight. My mistake; it stays a draft until the pipeline is green on the final commit.

**Scope note:** `47b0dba` fixes a flaky test that is unrelated to this MR's docs-only diff, per the "fix red CI in the current MR, don't defer it" rule.

`php-tests (feature-other)` failed on `DungeonRoutePolicyTest::publish_givenOwner_returnsAllowed`. Root cause: `DungeonRouteFactory` picks its dungeon with `inRandomOrder()`, and 7 of the 75 eligible seeded dungeons carry required enemies (`ataldazar`, `kingsrest`, `shrineofthestorm`, `templeofsethraliss`, `theunderrot`, `toldagor`, `mechagonjunkyard`). The test publishes a route with no pulls, so whenever the dice landed on one of those the publish guard denied and the assertion failed — roughly a 1-in-11 flake on **every** PR, not just this one.

`publish_givenNonOwnerNonAdmin_returnsDenied` had the same dependency in the passing direction: on those dungeons it was denied for the required-enemy reason rather than the ownership one it claims to test.

Both now run on the empty mapping version that `giveRouteAnUnkilledRequiredEnemy()` already built, extracted as `giveRouteAnEmptyMappingVersion()`. Verified by pinning the factory to `kingsrest`: the old test fails at the same line CI reported (`:220`), the new one passes; the full file is 27/27 green.

.claude/skills/seed-dev-routes/SKILL.md
```diff
@@ -21,82 +24,57 @@ docker compose exec -T -e DUNGEON_KEY=pitofsaron -e ROUTE_COUNT=12 app \
...
-sed -i 's|^ASSETS_BASE_URL_INTERNAL=.*|ASSETS_BASE_URL_INTERNAL=https://assets.keystone.guru|; s|^DEBUGBAR_ENABLED=.*|DEBUGBAR_ENABLED=false|' .env
-
-docker compose exec -T \
-  -e APP_URL=http://nginx \
-  -e PUPPETEER_EXECUTABLE_PATH=/var/www/.chrome-tmp/chrome-headless-shell-linux64/chrome-headless-shell \
-  -e THUMBNAIL_KEYS=<keys from step 1> \
-  app php artisan tinker .claude/skills/seed-dev-routes/generate_thumbnails.php
-
-sed -i 's|^ASSETS_BASE_URL_INTERNAL=.*|ASSETS_BASE_URL_INTERNAL=http://app-assets|; s|^DEBUGBAR_ENABLED=.*|DEBUGBAR_ENABLED=true|' .env
+for key in $(echo '8JxT3PE,pw7FaEt' | tr ',' ' '); do
```
Wotuu: :robot: This swaps one copy-paste trap for a subtler one. `8JxT3PE,pw7FaEt` are two real, valid-looking public keys from the verification run in the PR body — an agent skimming a runbook copies a line that looks runnable far more readily than one containing angle brackets, and nothing in the snippet itself signals these are placeholders (the adjacent prose only warns about the `CREATED:` prefix). Suggest something self-evidently non-literal that still shows the shape, e.g. `for key in $(echo '<key1>,<key2>' | tr ',' ' '); do`.

Wotuu:
:robot: Cold review (opus) of the two commits after `37769ff` (`76a09461f`, `47b0dba97`): 1 finding posted.

.claude/skills/seed-dev-routes/SKILL.md
Wotuu: :robot: Fixed in 860cda8 — back to `<key1>,<key2>`, which keeps the comma-separated shape the `tr` split needs while staying obviously a placeholder.

PR #3884, opened 2026-08-07
#3881 Include unassigned tag definitions in Team::getAvailableTags()

### LINKED ISSUE #3881: Team::getAvailableTags() never returns tag definitions (model_id IS NULL)
:robot: Follow-up split off from [PR #3880](https://github.com/RaiderIO/keystone.guru/pull/3880) cold review, itself a follow-up of #3875.

`Team::getAvailableTags()` filters with `whereIn('model_id', $this->dungeonRoutes->pluck('id'))`. Tag *definitions* created via `TeamController::createTag()` / `Tag::saveFromRequest()` are stored with `model_id = NULL` (not yet applied to any route). SQL's `IN (...)` never matches `NULL`, so these tag definitions never appear in the team's available-tags list/autocomplete — the same NULL-vs-`whereIn` gap that was already fixed for the `Team::boot()` deleting hook in #3867 (see `deleting_givenAnUnassignedTagDefinition_dropsItToo`).

Now that `getAvailableTags()` is also scoped by `context_class`/`context_id` (#3875/#3880), the `model_id` filter could likely be relaxed to also include (or replace with) tags with `model_id IS NULL`, e.g.:

```php
Tag::where('tag_category_id', TagCategory::ALL[TagCategory::DUNGEON_ROUTE_TEAM])
    ->where(function ($query) {
        $query->whereNull('model_id')
            ->orWhereIn('model_id', $this->dungeonRoutes->pluck('id'));
    })
    ->get();
```

Needs a look at call sites to confirm whether definition-only tags (not yet on any route) should indeed show up here, plus test coverage.

Non-blocking drive-by from the same review: `$this->dungeonRoutes->pluck('id')` hydrates every route model for an unbounded `IN (...)` list where a `->select('id')` subquery would avoid the full hydration.

Wotuu:
:robot: Closes #3881

Stacked on #3880 — targets `master` per the stacked-PR convention (so CI actually runs), diff collapses to just this change once #3880 merges.

`Team::createTag()` (`Tag::saveFromRequest()`) stores a tag definition with `model_id = NULL` before it's ever applied to a route. `Team::getAvailableTags()` filtered with `whereIn('model_id', ...)`, and SQL's `IN (...)` never matches `NULL`, so a tag definition never showed up in the team's tag autocomplete (`resources/views/common/dungeonroute/table.blade.php`) until it happened to already be on a route — defeating the point of creating it standalone.

Now that #3880 scopes the query by `context_class`/`context_id`, it's safe to also include `model_id IS NULL` rows without reintroducing the cross-team leak #3875 fixed.

## Test plan
- [x] New `TeamTest::getAvailableTags_givenOwnTeamsUnassignedTagDefinition_returnsIt`
- [x] New `TeamTest::getAvailableTags_givenAnotherTeamsUnassignedTagDefinition_excludesIt`

### DIFF
--- a/tests/Feature/App/Models/TeamTest.php
+++ b/tests/Feature/App/Models/TeamTest.php
@@ -109,24 +109,31 @@ final class TeamTest extends PublicTestCase
     #[Test]
     public function getAvailableTags_givenAnotherTeamsUnassignedTagDefinition_excludesIt(): void
     {
+        // A team with no routes/tags of its own would make the negative assertion below pass
+        // vacuously (and wouldn't catch the whereNull()/orWhereIn() clause escaping its enclosing
+        // where() and collapsing the entire query into a no-op OR), so this team also gets its own
+        // definition to prove getAvailableTags() is actually finding *something*.
         $team      = null;
         $otherTeam = null;
+        $ownDef    = null;
         $otherDef  = null;
 
         try {
             // Arrange
             $team      = $this->createTeam();
             $otherTeam = $this->createTeam();
+            $ownDef    = $this->createTeamTagDefinition($team);
             $otherDef  = $this->createTeamTagDefinition($otherTeam);
 
             // Act
             $availableTags = $team->getAvailableTags();
 
             // Assert
+            $this->assertTrue($availableTags->contains('id', $ownDef->id));
             $this->assertFalse($availableTags->contains('id', $otherDef->id));
         } finally {
             $this->cleanUp($otherDef, null, $otherTeam, null);
-            $this->cleanUp(null, null, $team, null);
+            $this->cleanUp($ownDef, null, $team, null);
         }
     }
 
--- a/tests/Feature/App/Models/TeamTest.php
+++ b/tests/Feature/App/Models/TeamTest.php
@@ -82,6 +82,54 @@ final class TeamTest extends PublicTestCase
         }
     }
 
+    #[Test]
+    public function getAvailableTags_givenOwnTeamsUnassignedTagDefinition_returnsIt(): void
+    {
+        // TeamController::createTag() stores a team's tag definitions through Tag::saveFromRequest(),
+        // which writes model_id = NULL. A plain whereIn('model_id', ...) never matches NULL, so a
+        // tag definition never applied to a route would otherwise never show up in the autocomplete.
+        $team = null;
+        $def  = null;
+
+        try {
+            // Arrange
+            $team = $this->createTeam();
+            $def  = $this->createTeamTagDefinition($team);
+
+            // Act
+            $availableTags = $team->getAvailableTags();
+
+            // Assert
+            $this->assertTrue($availableTags->contains('id', $def->id));
+        } finally {
+            $this->cleanUp($def, null, $team, null);
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


tests/Feature/App/Models/TeamTest.php
```diff
@@ -0,0 +1,191 @@
...
+            // Arrange
+            $team      = $this->createTeam();
+            $otherTeam = $this->createTeam();
+            $otherDef  = $this->createTeamTagDefinition($otherTeam);
+
+            // Act
+            $availableTags = $team->getAvailableTags();
+
+            // Assert
+            $this->assertFalse($availableTags->contains('id', $otherDef->id));
```
Wotuu, line 133: :robot: This test is negative-only and currently near-vacuous: `$team` is created with no routes and no tags of its own, so `$availableTags` is an empty collection and the `assertFalse` passes for any reason at all — including `getAvailableTags()` being broken outright and returning nothing.

Concretely, it also wouldn't catch the classic precedence bug this PR is guarding against (dropping the closure, i.e. `... AND context_id = ? AND model_id IS NULL OR model_id IN (...)`): `AND` binds tighter, and with no routes the `IN ()` collapses to `0 = 1`, so the other team's definition still wouldn't match and this test would still pass.

The sibling test added just above (`getAvailableTags_givenAnotherTeamsTagOnASharedRoute_excludesIt`) gets this right by asserting both sides — please mirror it here: also create `$def = $this->createTeamTagDefinition($team);` in the Arrange and assert `$availableTags->contains('id', $def->id)` is true alongside the existing `assertFalse`. That turns "returned nothing" into a genuine discriminator, and makes the test fail loudly if the context scoping is ever loosened.

Wotuu:

Reviewed only the additional commit on top of #3880 (`d128d35` — `#3881 Include unassigned tag definitions in Team::getAvailableTags()`).

The production change is correct as written: the nested closure gives the right precedence, the `context_class`/`context_id` scoping from #3880 keeps the `model_id IS NULL` branch from reintroducing the #3875 cross-team leak, `tag_category_id` pins the category so no personal/other-category definitions sneak in, `Tag::$visible` (id/name/color) means the extra rows serialise fine for the autocomplete, and with a team that has no routes `orWhereIn('model_id', [])` correctly degrades to `0 = 1`. The single finding is a test-strengthening request, not a merge blocker.

Deliberate discards: duplicate autocomplete names (pre-existing — `AjaxTagController::store()` already writes one row per route, so a tag on N routes already yields N identical names), and the `cleanUp()` helper's query-builder deletes bypassing `Team::deleting()` (arrived with #3880, not this commit).

One carry-over note rather than a finding: the `$this->dungeonRoutes->pluck('id')` hydration nit recorded in #3881's body is still unaddressed on the very line this PR rewrites — already raised and consciously deferred on #3880, so not re-raised inline, but a `->select('id')` subquery on the relation would be a free win while the clause is open.

tests/Feature/App/Models/TeamTest.php
Wotuu: :robot: Good catch — fixed, the team under test now also gets its own definition so the assertion isn't vacuous.

PR #3891, opened 2026-08-08
#3595 Migrate away from unmaintained lightslider

### LINKED ISSUE #3595: Migrate away from unmaintained lightslider (follow-up to #3590)
:robot: 
### Background

bonus, checked which of them are still maintained upstream. `lightslider` is not:
last npm release **2016**, last GitHub push **2022** with **280 open issues**
([sachinchoolur/lightslider](https://github.com/sachinchoolur/lightslider)).

### Usage

Dungeon-route carousels (about page / image galleries), wired up in
`resources/assets/js/custom/inline/common/dungeonroute/handlers/carouselhandler.js`
via `.lightSlider(settings)`, including lazy-loading and slide-count-driven
prev/next button visibility.

### Task

Research an actively-maintained carousel library (e.g. Splide, Swiper) and migrate
`carouselhandler.js` to it, preserving the existing lazy-load-on-slide and
prev/next visibility behaviour. Update/remove the `lightslider` block in
`resources/assets/js/jquery4-plugin-audit.test.js` (added in #3590) once swapped.

### Related

Wotuu:
:robot: Closes #3595

`lightslider` was the last unowned jQuery plugin from the #3590 audit (last npm release 2016, last upstream push 2022, 280 open issues). Replaced with a self-owned `$.fn.thumbnailCarousel`, following the same template as #3593 (`star-rating`), #3597 (`password-strength`) and #3596 (`range-slider`) — **no new dependency**.

## What lightslider was actually doing

Much less than the issue assumed. The live config was `item:1, autoWidth:false, slideMargin:0, loop:true, controls:false, pager:false, gallery:false`, so the only real behaviour was *a 1-up viewport you could drag through*. Both features `carouselhandler.js` layered on top were dead:

| Feature | Reality on master |
| --- | --- |
| lazy-load-on-slide | **No-op.** The blades emit eager `src`, never `data-src`, so `img.attr('src', img.attr('data-src'))` is a jQuery getter that sets nothing. Measured on a master baseline build: a 28-thumbnail card loaded **30/30** images up front. |
| prev/next visibility | **Dead.** `controls: false` means lightslider never renders `.lSPrev`/`.lSNext`, so the show/hide logic in `onSliderLoad`/`onAfterSlide` targeted nothing that exists. |
| `.aboutSliderHidden` removal | **Dead.** That class exists nowhere in the repo. |

## Scope note

Per the issue this should "preserve the existing lazy-load-on-slide and prev/next visibility behaviour" — but since both were dead code there was nothing to preserve, so **they are reimplemented for real** rather than ported. That was an explicit call, not scope creep: without arrows, a route's other floor thumbnails are only reachable by guessing that the image can be dragged. Same card now loads **2/28** images instead of 30/30.

## The widget

`resources/assets/js/thumbnail-carousel.js` (+ `css/lib/thumbnail-carousel.css`), required from `bootstrap.js` as a bare `require('./thumbnail-carousel')` like the other self-owned plugins.

- **Slide count comes from the DOM, never from the `single`/`multiple` class.** That class is derived from the dungeon's *floor* count while the slides come from `$dungeonroute->thumbnails`; they disagree whenever thumbnail generation lags behind floors. Below 2 slides the widget **no-ops entirely** — no track, no arrows, no transform — so single-thumbnail and dungeon-image-fallback cards are untouched.
- **Percentage layout**: a flex track translated by whole multiples of 100%. No width measurement, so it can be called before images load — which removes the reason `searchhandlerdungeonroute.js` had a "wait for the images to render first" comment.
- Loop stays `true` (matching the old config), so both arrows show whenever there are ≥2 slides; wrapping around either end skips the animation rather than sweeping backwards past every slide. Arrows are hover-revealed on desktop and always visible on touch (`@media (hover: none)`) — happy to make them always-on if you'd prefer.
- Pointer/touch drag with the same 40px threshold lightslider used, `touch-action: pan-y` so page scroll still works, and a swallowed post-drag click so a swipe can't register as a tap on a surrounding handler.
- Arrow handlers `stopPropagation()`. The routes table's thumbnail column is `clickable: false` today so nothing is behind them, but the arrows are new DOM in click-sensitive containers.
- Keyboard + ARIA the vendor didn't have (`aria-roledescription`, per-slide `n / m` labels, arrow keys on the nav buttons). The carousel itself is deliberately *not* focusable — that would add two tab stops to every card in a list of them.

## Also in this MR

- Blades emit `data-src` for every slide after the first; the first keeps a real eager `src` so cards are never blank, including without JS. The `@else` fallback `` is now wrapped in an `<li>` (it was a direct child of the `<ul>` — invalid HTML).
- Clean-break rename `.light-slider` → `.thumbnail-carousel__track` / `.light-slider-container` → `.thumbnail-carousel`, carrying the sizing rules in `discover.css` and `datatables.css` with it.
- `cardhorizontalrow.blade.php` had a fully commented-out carousel block plus a live-but-unused `$owlClass`; both removed.
- `sass.test.mjs`'s bare-package-import fixture retargeted from `lightslider` to `nouislider` — it goes red the moment the dep leaves, and after this MR `app.scss` has no bare imports left, so that test is the only thing keeping the importer branch honest.

## Verification

- `npm run test`: 428 passing (44 files), including 16 new cases in `thumbnail-carousel.test.js`. The `lightslider` block in `jquery4-plugin-audit.test.js` is gone and its header ledger updated.
- `composer run fix` (0 files) and `composer run analyse` (no errors).
- Headless Chrome, A/B against a master baseline built in the same worktree. Home, `/routes` and `/dungeonroutes` all render to the **same pixel height** as master, and the carousel/card boxes measure identically (`246x164`; cards `420/434/396/434`).
  - This caught a real bug: each theme compiles its own Bootstrap into `theme-{version}.css`, which loads *after* `custom-{version}.css`, so `.darkly ul` (0,1,1) beat my single-class list reset and restored 2rem of list padding — shrinking every thumbnail to `214x143`. Fixed by scoping the whole sheet two classes deep (second commit).
- Mobile viewport: `350x233`, correct 3:2, arrows visible.


Wotuu:
:robot: ## Visual verification

Headless Chrome, 1600x1000, same worktree — the "before" build is master's version of the changed files rebuilt in place, so only the bundle contents differ.

**Home page — before (master) vs after.** Both render to exactly 1585x2882; `/routes` and `/dungeonroutes` are likewise pixel-height-identical. (The broken "Thumbnail" boxes are dev-only CDN 404s and appear identically on both sides.)

| before (master) | after |
| --- | --- |
| before | after |

**The one multi-thumbnail card on that page (28 floors), showing the new arrows and a slide change.** On master this card had no controls at all — you had to guess it could be dragged — and it eagerly loaded all 30 images; here it loads 2 and fetches the next one as you go.

| slide 1 | after clicking &rsaquo; |
| --- | --- |
| slide 1 | slide 2 |

That route's floors happen to look near-identical, so the two crops differ only subtly — the state is verified directly as well: `transform: translateX(-100%)`, `aria-hidden` moves to slide 2, loaded images 2 → 3.

resources/assets/js/thumbnail-carousel.js
```diff
@@ -0,0 +1,258 @@
...
+            }
+
+            keyEvent.preventDefault();
+            move(keyEvent.key === 'ArrowLeft' ? -1 : 1);
+        });
+
+        let dragStartX = null;
+        let dragDeltaX = 0;
+
+        $track.on('pointerdown', function (pointerEvent) {
```
Wotuu, line 292: :robot: `pointerdown` never calls `preventDefault()`, the ``s have no `draggable="false"`, and the CSS has no `user-select: none` / `-webkit-user-drag: none`. `` is natively draggable, so a desktop mouse drag on a thumbnail starts the browser's HTML5 drag-and-drop instead of (well, on top of) the swipe: a ghost image follows the cursor, and per the Pointer Events spec the UA fires `pointercancel` the moment a native drag begins (Chrome does). Combined with the `pointercancel` handling below, every desktop mouse drag both shows a drag ghost and commits a slide change from the cancel path.

Neither jsdom (no native DnD) nor a headless screenshot can observe this, which is why the verification pass came back clean. A `pointerEvent.preventDefault()` here suppresses both drag-start and the text/image selection that a swipe otherwise leaves behind.

resources/assets/js/thumbnail-carousel.js
```diff
@@ -0,0 +1,258 @@
...
+        $track.on('pointermove', function (pointerEvent) {
+            if (dragStartX === null) {
+                return;
+            }
+
+            dragDeltaX = pointerEvent.clientX - dragStartX;
+            track.style.transform = `translateX(calc(${currentIndex * -100}% + ${dragDeltaX}px))`;
+        });
+
+        $track.on('pointerup pointercancel pointerleave', function (pointerEvent) {
```
Wotuu, line 328: :robot: `pointercancel` is grouped with `pointerup` and therefore *commits* a slide change, but `pointercancel` means the opposite: the UA took the gesture over and the pointer input is no longer valid. It fires when the browser starts a native image drag (see the `pointerdown` comment above) and when a touch gesture is reclassified as a page scroll — so a diagonal touch swipe with >40px of horizontal travel that the browser turns into a vertical scroll will also flip the slide underneath the user.

It should snap back (`paint(true)`) rather than `move()`. Worth splitting the handler:

```js
$track.on("pointercancel", function () {
    if (dragStartX === null) { return; }
    dragStartX = null;
    dragDeltaX = 0;
    $carousel.removeClass("thumbnail-carousel--dragging");
    paint(true);
});
```

Note the `drag()` test helper only fires `pointerdown`/`pointermove`/`pointerup`, so nothing in `thumbnail-carousel.test.js` covers the cancel path.

resources/assets/js/thumbnail-carousel.js
```diff
@@ -0,0 +1,258 @@
...
+        let currentIndex = 0;
+
+        /**
+         * Loads the current slide's images plus the next one's, so a slide change never waits
+         * on a network request. This is what lightslider's onBeforeStart/onAfterSlide callbacks
+         * were meant to do before the blades stopped emitting `data-src`.
+         *
+         * @param {number} index
+         */
+        function loadAround(index) {
```
Wotuu, line 152: :robot: `loadAround` only preloads forward (`index` and `index + 1`), but the docblock claims "a slide change never waits on a network request" — that only holds going next. Every backwards move lands on a slide whose `src` is still the `//:0` placeholder.

With `loop: true` the prev arrow is enabled from the start, so the *very first* prev click on a fresh card jumps to the last slide, which is guaranteed unloaded — the user gets a broken-image frame (with the `alt` text) until the fetch completes. Same for a rightwards drag from slide 0.

Preloading the previous slide too is a one-liner and keeps the docblock true:

```js
loadSlideImages(slides[(index - 1 + slideCount) % slideCount]);
```

(Cost: 3 images instead of 2 on init — still far below the 30/30 the PR body measured on master.)

resources/assets/js/thumbnail-carousel.js
```diff
@@ -0,0 +1,258 @@
...
+        }
+        track.dataset.thumbnailCarousel = 'true';
+
+        const $track = $(track);
+        const $carousel = $track.parent();
+        const slideCount = slides.length;
+
+        $carousel.addClass('thumbnail-carousel--interactive').attr({
+            role: 'group',
+            'aria-roledescription': 'carousel',
```
Wotuu, line 96: :robot: `aria-roledescription` must not be used on an element without an accessible name (ARIA 1.2: "Authors MUST NOT use aria-roledescription on elements that do not have a role or that have an abstract role"; the APG carousel pattern and axe's `aria-roledescription` rule both require a name). This `role="group"` container gets no `aria-label`, so a screen reader announces a nameless "carousel" group. The per-slide labels below are done correctly — it's only the container that's missing one.

A translated label off the route/dungeon context, or even a static `lang.get(...)` string passed alongside `prevLabel`/`nextLabel`, would close it.

resources/views/common/dungeonroute/cardhorizontalrow.blade.php
```diff
@@ -54,7 +54,6 @@
     $enemyForcesPercentage = $dungeonroute->getEnemyForcesPercentage();
     $enemyForcesWarning    = $dungeonroute->enemy_forces < $dungeonroute->mappingVersion->enemy_forces_required || $enemyForcesPercentage >= 105;
     $activeFloors          = $dungeonroute->dungeon->floorsForMapFacade($dungeonroute->mappingVersion, true)->get();
```
Wotuu: :robot: Removing `$owlClass` leaves `$activeFloors` (this line) with no consumer at all in this file — `grep -n activeFloors resources/views/common/dungeonroute/cardhorizontalrow.blade.php` returns only line 56. It is now a dead `floorsForMapFacade($dungeonroute->mappingVersion, true)->get()` executed once per rendered row, and `cardhorizontalrow` is rendered in a loop from `cardlist.blade.php`, so this is a per-row query for a value nothing reads.

Since the commented-out carousel block that used it is gone too, the `$activeFloors` line should go with it.

Wotuu:

resources/assets/js/thumbnail-carousel.js
Wotuu: :robot: Fixed in 065f478. `pointerdown` now `preventDefault()`s, and the slide images get `-webkit-user-drag: none` / `user-select: none` in the stylesheet so the touch and selection paths are covered too. Added `thumbnailCarousel_givenPointerDown_preventsTheBrowsersNativeImageDrag`.

resources/assets/js/thumbnail-carousel.js
Wotuu: :robot: Fixed in 065f478. The handler now distinguishes the event type and only commits on a real `pointerup`; `pointercancel` resets `dragDeltaX` and snaps back via `paint(true)`. The `drag()` test helper takes the terminating event as a parameter, and `thumbnailCarousel_givenTheBrowserCancelsTheDrag_abandonsItInsteadOfChangingSlide` covers it.

resources/assets/js/thumbnail-carousel.js
Wotuu: :robot: Fixed in 065f478 — `loadAround` now also loads `(index - 1 + slideCount) % slideCount`, so at rest the first slide preloads the *last* one, which is exactly where a first prev click lands under `loop: true`. Docblock updated to say "its two neighbours". The existing preload test became `thumbnailCarousel_givenInitialised_loadsOnlyTheCurrentSlideAndItsNeighbours`.

resources/assets/js/thumbnail-carousel.js
Wotuu: :robot: Fixed in 065f478. The container now gets an `aria-label` from a new `label` option, wired to `js.thumbnail_carousel_label` ("Route thumbnails") in `carouselhandler.js` alongside the two arrow labels.

resources/views/common/dungeonroute/cardhorizontalrow.blade.php
Wotuu: :robot: Good catch — fixed in 065f478, the whole `$activeFloors` line is gone. It was a per-row `floorsForMapFacade(...)->get()` in a partial rendered once per route.

Wotuu:
:robot: All 5 cold-review findings addressed in 065f478 (threads replied to and resolved), plus a latent bug found alongside them: with `loop: false` a drag past the threshold at either end early-returned from `move()` without repainting, leaving the track stuck at its dragged-out offset.

Two of the five were things neither jsdom nor a screenshot could see — the native `` drag hijacking every desktop mouse swipe, and `pointercancel` counting as a completed swipe — so they are now covered by tests instead. 431 JS tests passing, CI green on the fix commit.

Also verified afterwards what the earlier screenshots had not: the **horizontal** card path (`.card_dungeonroute.horizontal`, the fixed 288x192 rule at >=1200px, reachable on `/search`) measures `288x192` for the container, track and slide with both arrows rendered — the vertical cards I had measured go through a different CSS rule.

resources/assets/js/thumbnail-carousel.js
```diff
@@ -0,0 +1,347 @@
...
+            const clone = target.cloneNode(true);
+            clone.setAttribute('aria-hidden', 'false');
+
+            if (forward) {
+                track.appendChild(clone);
+            } else {
+                track.insertBefore(clone, track.firstChild);
+                // Prepending shifts every real slide's visual position one slot to the right;
+                // silently correct for that before animating, so the insertion itself doesn't jump.
+                paint(false, currentIndex + 1);
```
Wotuu: :robot: **Blocking: the backward wrap still teleports, and it permanently bricks the carousel.**

`paint()` forces its reflow *before* the transform is written:

```js
track.classList.toggle('...--no-transition', !animate);
if (!animate) { void track.offsetHeight; }   // flush happens here...
track.style.transform = `translateX(${visualIndex * -100}%)`;   // ...then the write
```

So in `beginWrap(-1)`:
1. `paint(false, currentIndex + 1)` flushes with the transform still at its **old** value (`0%`), then sets `-100%` — dirty, never committed.
2. `paint(true, 0)` sets it back to `0%` in the same task, with no flush in between.
3. Next style recalc: before-change `0%`, after-change `0%` → **no computed change → no transition starts → `transitionend` never fires.**

Consequences: prev-from-the-first-slide still jumps instantly (the exact bug this commit was written to fix, in one of the two directions), the clone is never removed, and `wrapping` stays `true` forever — so after one prev click on slide 0 the arrows *and* dragging are dead on that card until a page reload.

Verified in real headless Chrome, replaying this exact call sequence on a `transition: transform 400ms` element:

```
backward wrap: events: []                              computed: matrix(1,0,0,1,0,0)   (never moved)
forward  wrap: events: ["start:transform","end:transform"]  computed: -1020px           (fine)
```

The jsdom suite can't see it: it has no style engine, the observable inline transform mid-wrap is `translateX(0%)` either way, and `dispatchTransitionEnd()` hand-fires an event a browser would never fire here.

Fix that works (re-verified in the same harness — backward then produced `start`+`end` and animated to 0): move the reflow to *after* the transform assignment in `paint()`:

```js
track.classList.toggle('thumbnail-carousel__track--no-transition', !animate);
track.style.transform = `translateX(${visualIndex * -100}%)`;
if (!animate) {
    void track.offsetHeight;
}
```

That still lands the class before the value is committed (its stated purpose — the `--no-transition` class is in the after-change style either way), and it commits the intermediate `-100%` so the following `paint(true, 0)` has a real starting value. The forward wrap and the teardown `paint(false)` stay correct under the move (both still see the class in their after-change style, so they snap without animating).

resources/assets/js/thumbnail-carousel.js
```diff
@@ -0,0 +1,347 @@
...
+                wrapping = false;
+                paint(false);
+                loadAround(currentIndex);
+
+                if (typeof settings.onAfterSlide === 'function') {
+                    settings.onAfterSlide.call(track, currentIndex, slideCount);
+                }
+            };
+
+            track.addEventListener('transitionend', onTransitionEnd);
```
Wotuu, line 225: :robot: **`transitionend` is not guaranteed to fire, and when it doesn't the widget stays broken forever.**

`thumbnail-carousel.css` ends with:

```css
@media (prefers-reduced-motion: reduce) {
    .thumbnail-carousel .thumbnail-carousel__track { transition: none; }
}
```

With reduced motion on, `transition-property` is gone, so **neither** wrap direction ever fires `transitionend`: the clone is never removed, `paint(false)` is never called (the track stays parked on the clone's slot with a stale index), `wrapping` never goes back to `false`, and `onAfterSlide` never fires. Arrows and drag are then dead on that card until reload. This is a regression from this commit specifically — before it, a wrap was a plain no-transition repaint, which was correct under reduced motion.

Same failure mode for any other case where the transition doesn't run or gets cancelled: the track being `display:none` when the click lands, another `transform` write cancelling the running transition, etc. There's no recovery path — the teardown is only reachable from the event.

Suggested mitigation, which also covers the reflow bug in my other comment: arm a `setTimeout` fallback (transition duration + slack) that runs the same teardown, made idempotent so whichever of the two arrives first wins and cancels the other. e.g.

```js
let settled = false;
const settle = function () {
    if (settled) { return; }
    settled = true;
    clearTimeout(fallbackTimer);
    track.removeEventListener('transitionend', onTransitionEnd);
    clone.remove();
    wrapping = false;
    paint(false);
    ...
};
```

resources/assets/js/thumbnail-carousel.test.js
```diff
@@ -0,0 +1,331 @@
...
+        expect(track.querySelectorAll('li')).toHaveLength(4);
+
+        dispatchTransitionEnd(track);
+
+        expect(track.style.transform).toBe('translateX(-200%)');
+        expect(track.querySelectorAll('li')).toHaveLength(3);
+        expect(track.querySelectorAll('li')[2].getAttribute('aria-hidden')).toBe('false');
+    });
+
+    test('thumbnailCarousel_givenNextClickedWhileAWrapIsStillAnimating_ignoresItUntilTheWrapSettles', () => {
```
Wotuu: :robot: **Test-coverage gap: nothing asserts that a wrap ever *starts*, or that it ever *unblocks*.**

`dispatchTransitionEnd()` hand-fires the event, so every wrap test asserts the teardown while assuming the browser reaches it — which is exactly the assumption that is false today (see my comments on `beginWrap`: the backward wrap starts no transition at all, and reduced motion kills both directions). The suite is green on a widget that permanently locks up after one prev click on the first slide.

jsdom can't observe a real transition, so the assertions that would have caught this are behavioural:

1. **`wrapping` unblocks.** `..._ignoresItUntilTheWrapSettles` only asserts the ignored click; it never calls `dispatchTransitionEnd(track)` and then clicks next again to prove the carousel is usable afterwards. As written, a widget that sets `wrapping = true` and never clears it passes this test.
2. **Keyboard nav during a wrap.** `ArrowRight` on a nav button goes through the same `move()` guard, but nothing covers it — and a future refactor that moves the guard out of `move()` into the click handler would go unnoticed.
3. **`slideCount === 2`,** where the forward and backward wrap targets are the same neighbouring slide, in both directions (I worked the index math through by hand and it's correct — worth pinning so it stays that way).
4. **`onAfterSlide` timing for a wrap:** that it does *not* fire when the wrap begins, fires exactly once after `dispatchTransitionEnd`, and gets `(newIndex, slideCount)`. Right now the shift from synchronous-on-click to 400ms-later is entirely untested.

Wotuu:

resources/assets/js/thumbnail-carousel.js
Wotuu: :robot: Fixed in fbb9b8830 (also caught live: the forced reflow janked the reveal's first frame, so this now uses a double requestAnimationFrame instead of `offsetHeight`, plus a 500ms fallback timer for when `transitionend` never fires).

resources/assets/js/thumbnail-carousel.js
Wotuu: :robot: Fixed in fbb9b8830 (also caught live: the forced reflow janked the reveal's first frame, so this now uses a double requestAnimationFrame instead of `offsetHeight`, plus a 500ms fallback timer for when `transitionend` never fires).

resources/assets/js/thumbnail-carousel.test.js
Wotuu: :robot: Added: a fake-timers test proving the fallback timer settles and clears `wrapping` when `transitionend` never fires, a slideCount===2 test pinning the index math both directions, a keyboard-during-wrap test, a click-after-settle test proving the guard only blocks mid-wrap (not after), and an onAfterSlide timing/argument test for a wrap. In fbb9b8830.

resources/assets/js/thumbnail-carousel.js
```diff
@@ -0,0 +1,359 @@
...
+                // Still repaint: a drag past the threshold at either end leaves the track sitting
+                // at its dragged-out offset, so it has to be snapped back to the current slide.
+                paint(true);
+
+                return;
+            }
+
+            // Computed from the *old* index: for a wrap this lands on the clone slot one past
+            // either end (0 or slideCount + 1); otherwise it's just the next real slide's own slot.
+            const visualSlot = currentIndex + 1 + delta;
```
Wotuu, line 255: :robot: **Interrupting a wrap makes the next move animate the wrong way, across the whole track.**

`cancelPendingSnap()` drops the pending snap but leaves the transform sitting on a decorative clone slot, and the new `paint(true, visualSlot)` then animates *from that clone slot*. Worked example on a 3-slide carousel (exactly what the new test at `thumbnail-carousel.test.js:223` exercises):

- `next` ×3 → wrap: track is at slot 4 (`-400%`), `currentIndex = 0`, snap pending.
- 4th `next` (before the 400ms transition settles) → `visualSlot = 0 + 1 + 1 = 2`, so `paint(true, 2)` animates `-400% → -200%`.

The track therefore moves **rightward by two slot widths, through slot 3 (real slide 2), while the user pressed "next"** — wrong direction, 2× the distance, and slide 2 renders as an intermediate frame. Symmetrically for repeated `prev` through the backward wrap (slot 0 → slot `slideCount - 1`). This is the same class of bad intermediate frame this commit exists to remove, just triggered by a double-click instead of by a DOM mutation; the old `wrapping` lock happened to mask it by ignoring the input. The test at line 223 asserts `-200%` and so currently codifies the glitch as expected behaviour.

Fix: settle onto the real slot before starting the new move — but **not** as a naive `paint(false); paint(true, visualSlot);` pair in the same task. `paint(false)` does its `void track.offsetHeight` flush *before* writing the transform, so the settled `-100%` never reaches flushed style; at end-of-task the before-change style is still `-400%` and the transition starts from there, reproducing the identical sweep. The settle needs its reflow *after* the transform write (e.g. fold "write real-slot transform + flush" into `cancelPendingSnap()` and drop the separate `paint(false)` from the two snap callbacks).

jsdom can't observe the origin transform of a transition, so a regression test has to observe the settle directly — e.g. after a wrap, assert the transform is already back at the real slot immediately after `cancelPendingSnap()` runs (fire `pointerdown`, or the interrupting click, and assert the intermediate state), not just the end value.

resources/assets/js/thumbnail-carousel.test.js
```diff
@@ -0,0 +1,452 @@
...
+        $next.trigger('click'); // a plain move from the now-current first slide, not ignored
+
+        expect(track.style.transform).toBe('translateX(-200%)');
+
+        // The wrap's pending snap was cancelled along with it - its transitionend listener is
+        // gone, so this is a no-op rather than yanking the transform back to slot 1.
+        const transformBeforeStaleEvent = track.style.transform;
+        dispatchTransitionEnd(track);
+        expect(track.style.transform).toBe(transformBeforeStaleEvent);
+    });
```
Wotuu, line 278: :robot: **Coverage gap: the `pointerdown` cancel path is untested.**

This test covers `cancelPendingSnap()` via `move()` only. The other call site — `thumbnail-carousel.js:298`, whose comment claims "A drag starting mid-wrap shouldn't let the pending snap-back yank the transform out from under the user's finger" — has no test at all. Worth adding: wrap, then `pointerdown`, then `dispatchTransitionEnd(track)` (and separately `vi.advanceTimersByTime(500)`), asserting the transform is untouched by the stale snap.

Noting for scope: that path is benign today, because slot 0 and slot `slideCount` show the same image, so cancelling without settling is visually invisible once the animation has completed — this is a missing test, not a second bug.

Wotuu:

Full fresh review of the reworked wrap mechanism (permanent decorative clones), not a diff-only pass.

Cleared on my read: slot arithmetic in both directions and at `slideCount === 2`; `scheduleSnap()` cannot double-fire (both callbacks route through `cancelPendingSnap()`, which clears the counterpart, and `paint(false)` is idempotent); no path leaves the transform on a clone slot with nothing scheduled; the extra `<li>`s don't affect layout (the track is a block-level flex container at 100% width, so `translateX(-100%)` is unchanged) and add no unique image requests; `role`/`aria-roledescription` left on the clones is inert under `aria-hidden="true"` with no focusable content; the reduced-motion fallback test is valid as simplified; nothing else in the repo indexes the track's children.

One doc nit outside the diff: the PR body still says wrapping "skips the animation rather than sweeping backwards past every slide", which describes the pre-rework behaviour.

resources/assets/js/thumbnail-carousel.js
Wotuu: :robot: Fixed in b7cb61a01: `cancelPendingSnap()` is now `settle()` and repaints the real slide's own slot immediately (content-identical to whatever it's currently showing, so nothing visibly jumps). Hit the same reflow-ordering trap as de72ab3cd along the way (reflow has to run after the transform write, not before, or a chained `paint(true, ...)` still animates from the stale value) - fixed that in `paint()` too. Re-verified with real captured browser frames, interrupting click landing mid-animation, in both directions.

resources/assets/js/thumbnail-carousel.test.js
Wotuu: :robot: Added: a test for the `pointerdown` settle() call, and rewrote the interrupt test to spy on the sequence of transform writes (not just the end value) since the old assertion couldn't distinguish the settle-then-redirect fix from the bug - both land on the same final transform. In b7cb61a01.

resources/assets/js/thumbnail-carousel.test.js
```diff
@@ -0,0 +1,503 @@
...
+        const $next = $('.thumbnail-carousel__nav--next');
+        $next.trigger('click');
+        $next.trigger('click');
+        $next.trigger('click'); // wraps: currentIndex 0, mid-animation onto the end clone (slot 4)
+
+        const writes = recordTransformWrites(track);
+        $next.trigger('click'); // interrupts the still-unsettled wrap
+
+        // First write: an instant, un-animated settle onto the real first slide's own slot (1) -
+        // content-identical to the clone it was just showing, so nothing visibly jumps. Second
```
Wotuu: :robot: Comment-accuracy nit (behaviour is correct, the parenthetical isn't).

"content-identical to the clone it was just showing, so nothing visibly jumps" only holds once the wrap's transition has actually finished. In the case this test models the wrap is explicitly *mid*-animation (slot 3 -> slot 4), so at the moment of the interrupting click the computed transform is somewhere around -350%: the viewport is showing the right half of slide C plus the left half of the `A'` clone. Snapping to -100% changes both halves, so it is a visible cut-to-destination, not an invisible snap. Slot 4 (`A'`) and slot 1 (`A`) are pixel-identical only at rest.

Cutting the in-flight wrap short is the right behaviour for a rapid click (and matches what every other carousel does), so no code change needed - but a future reader who trusts "nothing visibly jumps" will assume settling mid-flight is free and may build on that. Suggest rewording to something like "settling onto the real slide's own slot, cutting the in-flight wrap short (content-identical only once the wrap has settled)". The same claim is in the commit message.

Wotuu:

resources/assets/js/thumbnail-carousel.test.js
Wotuu: :robot: Fixed in 2421d88c8 - reworded to note the claim only holds once the reveal has finished, not for a genuinely mid-transition interrupt.

PR #3896, opened 2026-08-08
#3893 Let a single scroll gesture reach the top of the page

### LINKED ISSUE #3893: Scrolling to the top of a page requires two scroll actions
:robot: Follow-up to #3851.

On Staging, scrolling from the bottom of a long page to the top in a **single** scroll action
does not land at the top. The scrollbar reaches the top, the header expands, and the page ends
up roughly 40px down again — a second scroll action is needed to actually reach the top.

### Cause

#3851 moved the header bars into normal document flow (one sticky `.ksg-header` wrapper). That
means the `.ksg-header--shrink` toggle changes the **document height**, not just the size of a
fixed overlay:

| element | at rest | shrunk | delta |
| --- | --- | --- | --- |
| `.dungeon_context_header` height | 99px | 64px | 35px |
| `.navbar-second` y-padding (x2) | 0.5rem | 0.25rem | 8px |

Unshrinking therefore grows the document by ~43px *above the viewport*, and the browser's
scroll anchoring compensates by adding that ~43px to `scrollY` to keep the visible content
stable. The inertial scroll was heading for 0; anchoring bumps it back to ~43 and the gesture
is already over.

The same displacement in the *shrink* direction is already accounted for by the wide hysteresis
band in `siteheader.js` (`shrinkAt = 140` / `unshrinkAt = 40`) and its
`shouldShrinkHeader_givenShrinkInducedScrollNudge_neverUnshrinks` test. The unshrink direction
has no equivalent cover.

### Scope

Fix the scroll mechanics only — the header must keep compressing exactly as it does today.

Files: `resources/assets/js/custom/inline/common/general/siteheader.js`,
`resources/assets/css/sections/header.css`.

Wotuu:
:robot: Closes #3893

Scrolling from the bottom of a page to the top in a single gesture stopped short of the top,
needing a second scroll.

## Cause

Since #3851 the header bars stack in **normal document flow**, so the `.ksg-header--shrink`
toggle changes the *document* height:

| element | at rest | shrunk | delta |
| --- | --- | --- | --- |
| `.dungeon_context_header` | 99px | 64px | 35px |
| `.navbar-second` y-padding (x2) | 0.5rem | 0.25rem | 8px |

Unshrinking therefore grows the document by 43px **above the viewport**, and the browser's
scroll anchoring adds those 43px to `scrollY` to keep the visible content stable. That
compensation eats the tail of an upward fling, which is already decelerating, so the gesture
ends short of the top.

## Measurements

Measured in headless Chrome on a synthetic page reproducing the header shape (tall body, one
sticky header whose height changes in flow):

| condition | scroll position moved by |
| --- | --- |
| unshrink at `scrollY = 100` | **+43px** |
| unshrink at `scrollY = 0` | 0px — Chrome does not anchor while the scroller sits at the top |
| shrink at `scrollY = 141` | −43px |
| any of the above with `overflow-anchor: none` | 0px |

The `transition: height 0.2s ease` makes no difference to any of these, so it is kept.

## Fix

Suppress scroll anchoring for the duration of the expand transition, and hand it back
afterwards. The shrink/unshrink thresholds are untouched — `shouldShrinkHeader` is unchanged —
so nothing about when or how the header renders moves.

Suppression is deliberately **not** applied when the header shrinks: there the compensation is
wanted, since it is what keeps the content you are reading from jumping upwards as you scroll
down. An open suppression window is therefore closed again if a shrink follows an unshrink
within it.

## Verification

Real-page probe (`/affixes`), one **decelerating** wheel gesture from the bottom of the page.
The deceleration is the part that matters: a constant-speed gesture has leftover travel to
spend after the header expands, recovers from the bump, and hides the bug entirely.

| build | desktop (1600x1000) | mobile (375x700) |
| --- | --- | --- |
| master | **43px** | **8px** |
| this branch | **0px** | **0px** |

3 consecutive runs per cell on this branch, no page errors.

`npx vitest run .../siteheader.test.js` — 13 passed. Added jsdom coverage for the suppression
seam: applied on unshrink, not on shrink, restored after the window, window closed early when a
shrink follows, timer re-armed on a second unshrink.

No PHP changed. Header rendering is unchanged in both states (screenshots below).

### DIFF
--- a/resources/assets/js/custom/inline/common/general/siteheader.js
+++ b/resources/assets/js/custom/inline/common/general/siteheader.js
@@ -28,6 +28,13 @@ class CommonGeneralSiteheader extends InlineCode {
      */
     static MIN_SCROLLABLE_DISTANCE = 200;
 
+    /**
+     * How long scroll anchoring stays suppressed after the header starts growing back. Must
+     * outlast the CSS height/padding transition in header.css (0.2s), since the document keeps
+     * growing - and anchoring keeps re-evaluating - for its whole duration.
+     */
+    static SCROLL_ANCHOR_SUPPRESSION_MS = 400;
+
     activate() {
         super.activate();
 
@@ -81,8 +88,47 @@ class CommonGeneralSiteheader extends InlineCode {
             shouldShrinkHeader(window.scrollY, currentlyShrunk);
 
         if (shrunk !== currentlyShrunk) {
+            if (shrunk) {
+                // Shrinking wants the compensation - close any suppression window a recent
+                // unshrink left open, or the shrink runs without it
+                this._restoreScrollAnchoring();
+            } else {
+                this._suppressScrollAnchoring();
+            }
+
             this.shrinkTarget.classList.toggle('ksg-header--shrink', shrunk);
         }
+    }
+
+    /**
+     * Stop the browser from compensating for the header growing back (#3893).
+     *
+     * The bars stack in normal document flow since #3851, so unshrinking grows the document by
+     * the height the header regains - above the viewport. Scroll anchoring then adds that same
+     * height to the scroll position to keep the visible content stable, which eats the tail of
+     * an upward fling (already decelerating by then) and leaves the page short of the top.
+     *
+     * Deliberately not applied when the header shrinks: there the compensation is wanted, as it
+     * keeps the content the user is reading from jumping upwards mid-page.
+     */
+    _suppressScrollAnchoring() {
+        document.documentElement.style.overflowAnchor = 'none';
+
+        clearTimeout(this._scrollAnchorRestoreTimer);
+        this._scrollAnchorRestoreTimer = setTimeout(
+            this._restoreScrollAnchoring.bind(this),
+            CommonGeneralSiteheader.SCROLL_ANCHOR_SUPPRESSION_MS
+        );
+    }
+
+    /**
+     * Hand scroll anchoring back to the browser. Suppression is always temporary - leaving it on
+     * would stop the whole page from compensating for anything that resizes above the viewport,
+     * such as an image loading in.
+     */
+    _restoreScrollAnchoring() {
+        clearTimeout(this._scrollAnchorRestoreTimer);
+        document.documentElement.style.removeProperty('overflow-anchor');
     }
 }
 
--- a/resources/assets/js/custom/inline/common/general/siteheader.test.js
+++ b/resources/assets/js/custom/inline/common/general/siteheader.test.js
@@ -1,13 +1,14 @@
 // ---------------------------------------------------------------------------
 // CommonGeneralSiteheader is a global-script style class extending the bare
 // global `InlineCode`. The hysteresis decision is a pure function
-// (shouldShrinkHeader) so the flap-prevention logic is testable without DOM.
+// (shouldShrinkHeader) so the flap-prevention logic is testable without DOM;
+// the scroll-anchoring suppression needs one, but only a class on an element.
 // ---------------------------------------------------------------------------
 
 const {InlineCode}    = require('../../inlinecode');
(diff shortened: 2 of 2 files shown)


Wotuu:
:robot: Header rendering is unchanged in both states (`/affixes`, 1600x700 viewport):

**At the top (`scrollY = 0`), 201px**
top

**Scrolled/shrunk, 158px**
shrunk

resources/assets/js/custom/inline/common/general/siteheader.js
```diff
@@ -81,9 +94,34 @@ class CommonGeneralSiteheader extends InlineCode {
             shouldShrinkHeader(window.scrollY, currentlyShrunk);
 
         if (shrunk !== currentlyShrunk) {
+            if (!shrunk) {
+                this._suppressScrollAnchoring();
```
Wotuu, line 96: :robot: The shrink branch inherits a still-open suppression window, so the "compensation is wanted when shrinking" intent doesn't actually hold.

`_suppressScrollAnchoring()` arms a fire-and-forget 400ms timer and nothing ever cancels it except a *subsequent* unshrink. So: arrive at the top (unshrink -> suppression armed), then fling back down past 140px within 400ms -> the header shrinks while `overflow-anchor: none` is still on `documentElement`. The document loses its 43px above the viewport uncompensated and visible content jumps up — exactly what the docblock on `_suppressScrollAnchoring()` says is deliberately preserved:

> Deliberately not applied when the header shrinks: there the compensation is wanted [...]

The PR's own tests show it: in `updateShrink_givenSecondUnshrinkWithinTheWindow_restartsTheSuppressionTimer` the intermediate `scrollTo(500); siteheader._updateShrink();` runs a shrink with the style still `'none'`, and nothing asserts on it. `updateShrink_givenHeaderShrinks_leavesScrollAnchoringAlone` starts from a clean state, so it structurally can't catch this case.

Relatedly, because the timer is only ever re-armed and never cancelled, any repeated toggle keeps anchoring off document-wide for longer than the transition the suppression is scoped to.

Suggested fix: make the shrink branch the explicit inverse rather than relying on whether a stale timer happened to fire —

```js
if (!shrunk) {
    this._suppressScrollAnchoring();
} else {
    clearTimeout(this._scrollAnchorRestoreTimer);
    document.documentElement.style.removeProperty('overflow-anchor');
}
```

with a test that shrinks *within* the suppression window and asserts anchoring is back on.

Wotuu:

resources/assets/js/custom/inline/common/general/siteheader.js
Wotuu: :robot: Good catch — real gap, and more reachable than it looks: unshrink at 40px then scroll back down past 140px inside the 400ms window is an ordinary flick.

Fixed in `4adc4b8e6`: the shrink branch now calls a `_restoreScrollAnchoring()` that clears the pending timer and removes the property, so a shrink never inherits an open window. The timeout callback is the same method, so there is one restore path rather than two.

Covered by `updateShrink_givenShrinkWithinTheSuppressionWindow_restoresScrollAnchoringImmediately`, which fails against the previous commit.

Also dropped the `unshrinkAt` 40 → 1 change that was in the reviewed commit: re-probing with suppression only showed it reaches the top on its own (3/3 desktop, 3/3 mobile), so the threshold change was unnecessary and `shouldShrinkHeader` is now untouched.

Wotuu:
:robot: Cold review: 1 findings

Fixed in `4adc4b8e6` (shrink now closes an open suppression window; see the resolved thread). The reviewed commit also carried an `unshrinkAt` 40 → 1 threshold change — re-probing showed suppression alone reaches the top on desktop and mobile, so that was dropped and `shouldShrinkHeader` is untouched.

PR #3897, opened 2026-08-08
Fix share/map modal close button hidden behind the header

### LINKED ISSUE #3895: Share modal close button hidden behind the map header
:robot: Follow-up to #3851.

On a map page, opening the share modal (or any modal declared inside `#map_header` - simulate,
edit route settings/admin settings, live session start/stop, edit mapping version) makes its close
button unclickable/invisible, hidden behind the header.

### Cause

`resources/assets/css/sections/map-header.css` has:

```css
#map_header,
#map_header > div {
    z-index: 100020 !important;
}
```

These modals are all declared as **direct children of `#map_header`** (see
`resources/views/common/maps/controls/header.blade.php`, included inside
`<nav id="map_header">` in `common/maps/map.blade.php`), so this ID-selector rule also matches
them - beating `.theme .modal { z-index: 100021 !important; }` (theme.scss) on specificity
regardless of the numeric value. The modal ends up pinned to `z-index: 100020`, same as the
header, and loses the paint-order tie to `.ksg-header`.

Confirmed via headless Chrome on `/route/the-necrotic-wake/cjuFP3F`:
`getComputedStyle(document.querySelector('#share_modal')).zIndex` reports `100020`, and the
element under the close button's screen position is a header nav link.

### Fix

Out-specify the ID-selector rule for modals specifically, e.g.:

```css
#map_header > .modal {
    z-index: 100021 !important;
}
```

Wotuu:
:robot: Closes #3895

## Cause

Modals declared inside `#map_header` (share, simulate, live session, edit route/mapping-version
settings, ...) are direct children of it, so the existing rule

```css
#map_header,
#map_header > div {
    z-index: 100020 !important;
}
```

also matches them - an ID selector beats `.theme .modal { z-index: 100021 !important; }`
(theme.scss) on specificity regardless of value, pinning the modal to the header's own z-index
and losing the paint-order tie to `.ksg-header`, hiding the close button behind it.

## Fix

Out-specify the rule for modals specifically in `map-header.css`:

```css
#map_header > .modal {
    z-index: 100021 !important;
}
```

## Test plan
- [x] Headless Chrome, `#share_modal" on a published non-demo route: before the fix,
  `getComputedStyle(...).zIndex` reported `100020` and the element under the close button was a
  header nav link; after, it reports `100021` and the close button itself is hit-tested, and
  clicking it closes the modal.
- CSS-only change, no PHP touched.

### DIFF
--- a/resources/assets/css/sections/map-header.css
+++ b/resources/assets/css/sections/map-header.css
@@ -74,9 +74,16 @@
 /* The map header's modals (share, simulate, live session, ...) are declared inside #map_header,
    so the `#map_header > div` rule above also matches them - an ID selector beats `.theme .modal`'s
    two classes regardless of value, pinning the modal to 100020 and hiding its close button behind
-   the header. Out-specify it here so these modals float above the whole header again. */
+   the header. Out-specify it here so these modals float above the whole header again. One higher
+   than `.ksg-header` above (not equal) so it outright wins instead of merely tying and depending
+   on DOM order - a future bump of `.ksg-header` would otherwise silently re-hide the close button.
+   pointer-events also needs restoring: these modals inherit `none` from #map_header (below), which
+   silently breaks click-outside-the-dialog-to-dismiss - the backdrop is a sibling of #map_header
+   (Bootstrap appends it to <body>) so it stays clickable, but the click never reaches the modal's
+   own dismiss listener without this. */
 #map_header > .modal {
-    z-index: 100021 !important;
+    z-index: 100022 !important;
+    pointer-events: auto;
 }
 
 #map_header .rounded {
--- a/resources/assets/css/sections/map-header.css
+++ b/resources/assets/css/sections/map-header.css
@@ -71,6 +71,14 @@
     z-index: 100021 !important;
 }
 
+/* The map header's modals (share, simulate, live session, ...) are declared inside #map_header,
+   so the `#map_header > div` rule above also matches them - an ID selector beats `.theme .modal`'s
+   two classes regardless of value, pinning the modal to 100020 and hiding its close button behind
+   the header. Out-specify it here so these modals float above the whole header again. */
+#map_header > .modal {
+    z-index: 100021 !important;
+}
+
 #map_header .rounded {
     border-radius: 0 !important;
 }


resources/assets/css/sections/map-header.css
```diff
@@ -71,6 +71,14 @@
     z-index: 100021 !important;
 }
 
+/* The map header's modals (share, simulate, live session, ...) are declared inside #map_header,
+   so the `#map_header > div` rule above also matches them - an ID selector beats `.theme .modal`'s
+   two classes regardless of value, pinning the modal to 100020 and hiding its close button behind
+   the header. Out-specify it here so these modals float above the whole header again. */
+#map_header > .modal {
```
Wotuu, line 84: :robot: **The `.ksg-header` z-index above is _not_ a no-op — `#map_header` is a flex container, so this fix ties rather than wins.**

`#map_header` is `display: flex` (line 37-42), and per CSS Flexbox §5.4 *z-index applies to flex items even when they are `position: static`*, and creates a stacking context. So `#map_header .ksg-header { position: static }` does **not** neutralise `#map_header > .ksg-header { z-index: 100021 }` — that rule is live, and it is the only model that explains the empirical result in the PR body (computed `100020`, `elementFromPoint` → a header nav link). Under the "static ⇒ z-index ignored" reading, a modal at 100020 would already have painted above a static header, which contradicts what was measured.

Consequences for this rule:

1. The fix is correct and correctly scoped — `#map_header > .modal` (1,1,0) beats `#map_header > div` (1,0,1), and all seven modals go through `common.general.modal`, whose root is `<div class="modal fade">`, emitted by `common/maps/controls/header.blade.php` as direct children of `#map_header`. No coverage gap.
2. But at `100021` the modal **ties** `.ksg-header` and only wins on tree order (the modals are included after `common.layout.header`). The comment's "float above the whole header again" overstates what the value buys.
3. The comment at line 67-69 is itself evidence that bumping `.ksg-header`'s z-index is the local idiom for fixing overlay bugs. The next such bump silently re-breaks the close button, with no test and no obvious link back to here.

Suggestion (non-blocking): use `100022` and drop the DOM-order dependency. `#map_header`'s own stacking context clamps all child values relative to the rest of the page, so 100021 → 100022 changes nothing outside the header, and a closed modal is `display: none`, so the `.ksg-header` dropdown fix is undisturbed. Worth also correcting the "static ⇒ no-op" framing wherever it appears, since the flex-item rule is the actual reason this bug existed.

resources/assets/css/sections/map-header.css
```diff
@@ -71,6 +71,14 @@
...
+    z-index: 100021 !important;
```
Wotuu: :robot: **Pre-existing, non-blocking, but adjacent enough to be worth verifying while you're in here: these modals likely inherit `pointer-events: none`, breaking click-outside-to-close.**

Chain, each link confirmed:
- `#map_header { pointer-events: none }` (this rule), and `pointer-events` is an inherited property.
- Bootstrap 5.3 `_modal.scss` declares `pointer-events: none` on `.modal-dialog` (line 58) and `auto` on `.modal-content` (line 102) — but declares **nothing** on `.modal` itself.
- Nothing in `map-header.css` restores `auto` for `#map_header > .modal` (the `> .container` exceptions below only cover the bars).

So `.modal` inherits `none`. BS5 dismisses on backdrop click by listening on `this._element` (the `.modal`) and comparing `event.target === this._element`; with `pointer-events: none` that target can never be the `.modal`, so clicking the dimmed area outside the dialog should neither close it nor be swallowed — it falls through to the Leaflet map behind. None of these modals set `data-bs-backdrop="static"`, so dismiss-on-outside-click is the intended behaviour.

Cheap to check on the page you already had open: `getComputedStyle(document.getElementById('share_modal')).pointerEvents`. If it reads `none`, `#map_header > .modal { pointer-events: auto }` next to the new z-index rule is the one-liner. Not a blocker for this PR — it predates it — just flagging since it's the same "modal declared inside the click-through header" family.

Wotuu:

Wotuu:
:robot: Addressed both cold-review findings in 886560f2b:

- Bumped modal z-index from `100021` to `100022` - `.ksg-header`'s z-index is live (flex items honor z-index even with `position: static`), so `100021` only tied it and depended on DOM order; a future bump to `.ksg-header` would have silently re-broken this.
- Restored `pointer-events: auto` on these modals - they were inheriting `none` from `#map_header`, which silently broke click-outside-the-dialog-to-dismiss even though the backdrop itself stayed clickable (it's a sibling of `#map_header`, appended to `<body>` by Bootstrap).

Verified both with a real hit-tested `page.mouse.click()` (not `el.click()`, which bypasses `pointer-events`) in headless Chrome, plus a negative control confirming the test actually distinguishes before/after.

Wotuu:
:robot: Cold review: 2 findings (both fixed)

resources/assets/css/sections/map-header.css
Wotuu: :robot: Fixed exactly as suggested in commit 886560f2b: bumped `#map_header > .modal` z-index from 100021 to 100022 so it outright wins over `.ksg-header` instead of merely tying on DOM order, and corrected the comment's "static ⇒ no-op" framing to explain the flex-item z-index behavior.

resources/assets/css/sections/map-header.css
Wotuu: :robot: Fixed in commit 886560f2b: the reasoning chain checks out (`#map_header { pointer-events: none }` is inherited, Bootstrap 5.3 `.modal` itself gets no override, only `.modal-content` does). Added `pointer-events: auto` to the `#map_header > .modal` rule so the modal element (not just its content) is clickable again, restoring dismiss-on-backdrop-click.

PR #3922, opened 2026-08-08
Guard route-attributes filter against a non-array search value

### LINKED ISSUE #3910: /ajax/routes 500s: array_diff() TypeError when route attribute filter param is malformed
:robot: `DungeonRouteAttributesColumnHandler::applyFilter()` calls `array_diff($allRouteAttributeIds, $routeAttributeIds)` where `$routeAttributeIds` can be a string instead of an array, causing a `TypeError`.

- Occurrences: 256, Users impacted: 3, ongoing (highest-volume of this pass's findings)
- First/last seen: 2026-07-16 to 2026-08-02
- Endpoint: `GET /ajax/routes`

```
TypeError: array_diff(): Argument #2 must be of type array, string given
```
```php
42 │     ->pluck('id')
43 │     ->toArray();
44 │ // Compute the attribute IDs that the user does NOT want
45 │ $invalidAttributeIds = array_diff($allRouteAttributeIds, $routeAttributeIds);
```
Local variables show `$routeAttributeIds` as the string `"-1"` rather than an array in the failing case — likely a request param (`route_attributes` or similar) sent as a bare string instead of an array. Fix is to coerce/validate this to an array before the `array_diff()` call, matching the pattern already used a few lines below for `$requirements`/`$tags` (`if (!is_array($x)) { $x = []; }`).

Wotuu:
:robot: Closes #3910

## Summary
- `DungeonRouteAttributesColumnHandler::applyFilter()` passed the DataTables column's `search.value` straight into `array_diff()`'s second argument. When the front-end route-attributes select isn't rendered for the current view, jQuery's `.val()` on the missing element sends the literal string `'undefined'` instead of an array of attribute ids (same failure mode already handled for the `tags`/`requirements` query params in `AjaxDungeonRouteController`), which raised `TypeError: array_diff(): Argument #2 must be of type array, string given` on `GET /ajax/routes` (Sentry `PHP-LARAVEL-PD`, 256 events, 3 users, ongoing since 2026-07-16).
- Fix: coerce a non-array `$routeAttributeIds` to `[]` before use, matching the existing pattern for `tags`/`requirements`.
- Front-end fix (added after review): `DungeonrouteTable`'s filter-button click handler (`resources/assets/js/custom/inline/dungeonroute/table.js`) read `$(self.options.affixSelectId).val()` and `$(self.options.attributesSelectId).val()` with no fallback, so a missing select sends `undefined` as the search value in the first place — exactly the pattern the `requirementsSelectId`/`tagsSelectId` reads a few lines down already guard against (`.val() || []`), with a comment explaining why. Added the same `|| []` guard to `affixes`/`attributes` so the client stops sending `'undefined'` at all; the backend guard stays as defense in depth regardless (public-ish endpoint, and other future callers).

Fixes PHP-LARAVEL-PD

## Test plan
- [x] New regression test (`get_givenRouteAttributesColumnSearchValueIsTheStringUndefined_returnsOk`) reproduces the exact reported TypeError without the fix, and passes with it.
- [x] Full `AjaxDungeonRouteControllerTest.php` suite green (8 tests).
- [x] `composer run fix` — no drift.
- [x] Full Vitest suite green (440 tests) after the front-end guard.

---
:robot: Note: the second commit (`f14965c1e`, `DungeonRouteAffixesColumnHandler`) implements the cold reviewer's own suggestion from its finding on this PR and was not separately re-reviewed.

### DIFF
--- a/app/Logic/Datatables/ColumnHandler/DungeonRoutes/DungeonRouteAttributesColumnHandler.php
+++ b/app/Logic/Datatables/ColumnHandler/DungeonRoutes/DungeonRouteAttributesColumnHandler.php
@@ -28,6 +28,11 @@ class DungeonRouteAttributesColumnHandler extends DatatablesColumnHandler
                 $generalSearch,
     ): void {
         $routeAttributeIds = $columnData['search']['value'] ?? null;
+        // A select that isn't rendered for the current view sends the literal string 'undefined'
+        // instead of omitting the param
+        if (!is_array($routeAttributeIds)) {
+            $routeAttributeIds = [];
+        }
         // If filtering or ordering
         if (!empty($routeAttributeIds) || $order !== null) {
             // $builder->leftJoin('dungeon_route_attributes', 'dungeon_route_attributes.dungeon_route_id', '=', 'dungeon_routes.id');
--- a/app/Logic/Datatables/ColumnHandler/DungeonRoutes/DungeonRouteAffixesColumnHandler.php
+++ b/app/Logic/Datatables/ColumnHandler/DungeonRoutes/DungeonRouteAffixesColumnHandler.php
@@ -32,6 +32,11 @@ class DungeonRouteAffixesColumnHandler extends DatatablesColumnHandler
                 $generalSearch,
     ): void {
         $affixIds = $columnData['search']['value'] ?? [];
+        // A select that isn't rendered for the current view sends the literal string 'undefined'
+        // instead of omitting the param
+        if (!is_array($affixIds)) {
+            $affixIds = [];
+        }
         if (!empty($affixIds)) {
             $subBuilder->whereHas('affixes', static function ($query) use (&$affixIds) {
                 /** @var $query Builder */
--- a/resources/assets/js/custom/inline/dungeonroute/table.js
+++ b/resources/assets/js/custom/inline/dungeonroute/table.js
@@ -54,8 +54,12 @@ class DungeonrouteTable extends InlineCode {
         $(this.options.filterButtonSelector).unbind('click').bind('click', function () {
             // Build the search parameters
             let dungeonId = $(self.options.dungeonSelectId).val();
-            let affixes = $(self.options.affixSelectId).val();
-            let attributes = $(self.options.attributesSelectId).val();
+            // .val() on a zero-element jQuery collection (e.g. a select that doesn't exist for
+            // this table's view) returns undefined, which would otherwise be sent to the server
+            // as the literal search value 'undefined' instead of an empty array. See the same
+            // guard on requirements/tags below.
+            let affixes = $(self.options.affixSelectId).val() || [];
+            let attributes = $(self.options.attributesSelectId).val() || [];
 
             // Find wherever the columns are we're looking for, then filter using them
             // https://stackoverflow.com/questions/32598279/how-to-get-name-of-datatable-column
--- a/tests/Feature/Controller/Ajax/AjaxDungeonRouteControllerTest.php
+++ b/tests/Feature/Controller/Ajax/AjaxDungeonRouteControllerTest.php
@@ -88,6 +88,44 @@ final class AjaxDungeonRouteControllerTest extends AjaxPublicTestCase
     }
 
     #[Test]
+    public function get_givenRouteAttributesColumnSearchValueIsTheStringUndefined_returnsOk(): void
+    {
+        // Arrange - same failure mode as the tags/requirements parameters above, but for the
+        // routeattributes.name DataTables column's own search value: a route-attributes select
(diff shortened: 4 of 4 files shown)


app/Logic/Datatables/ColumnHandler/DungeonRoutes/DungeonRouteAttributesColumnHandler.php
```diff
@@ -28,6 +28,11 @@ protected function applyFilter(
                 $generalSearch,
     ): void {
         $routeAttributeIds = $columnData['search']['value'] ?? null;
+        // A select that isn't rendered for the current view sends the literal string 'undefined'
+        // instead of omitting the param
```
Wotuu, line 32: :robot: The identical hole exists one file over and is left open: `DungeonRouteAffixesColumnHandler::applyFilter()` (line 34, `$affixIds = $columnData['search']['value'] ?? [];`). The `?? []` doesn't help — `'undefined'` is non-empty, so it reaches `whereIn('affix_groups.id', 'undefined')`, and Laravel's `whereIn` does `count($values)`:

```
TypeError: count(): Argument #1 ($value) must be of type Countable|array, string given
```

(verified in tinker against this repo's Laravel 12).

Same source as the bug being fixed here: `resources/assets/js/custom/inline/dungeonroute/table.js` sets both column searches from `$(...).val()` inside the *same* click handler, and both selects live in the same blade filter container (`common/dungeonroute/table.blade.php` ~130-147) — so a table view that renders the `affixes`/`features` column (note `features` also uses `name: 'affixes.id'`) without that filter container hits it. It's also trivially reachable on this public, unauthenticated endpoint with a crafted `GET /ajax/routes?columns[0][name]=affixes.id&columns[0][search][value]=undefined&...` → uncaught 500.

I'm not claiming the current Sentry traffic hits it (the affixes handler is registered *before* the attributes one in `AjaxDungeonRouteController`, so if it were, it would have crashed first) — but PHP-LARAVEL-PD can come back under a different column name. Suggest adding the same `is_array()` guard there plus a sibling `get_givenAffixesColumnSearchValueIsTheStringUndefined_returnsOk` test while you're in here.

Wotuu:

app/Logic/Datatables/ColumnHandler/DungeonRoutes/DungeonRouteAttributesColumnHandler.php
Wotuu: :robot: Fixed in f14965c1e — added the same is_array() guard to DungeonRouteAffixesColumnHandler and a regression test (get_givenAffixesColumnSearchValueIsTheStringUndefined_returnsOk) that reproduces the exact count() TypeError you found before the fix, and passes after.

Wotuu (commented):
`the missing element sends the literal string 'undefined' instead of an array of attribute ids` should the front end not also have a fix for this, instead of just the back-end? Just checking, but I think sending `undefined` is wrong in the first place.

Wotuu:
:robot: Re: "should the front end not also have a fix for this, instead of just the back-end?"

Yes — you were right, and I found the concrete client-side bug. `DungeonrouteTable`'s filter-button click handler (`resources/assets/js/custom/inline/dungeonroute/table.js`) reads:

```js
let dungeonId = $(self.options.dungeonSelectId).val();
let affixes = $(self.options.affixSelectId).val();
let attributes = $(self.options.attributesSelectId).val();
```

with no fallback. `.val()` on a zero-element jQuery collection (the select not being rendered for the current table view) returns `undefined`, and that's exactly what ends up as the DataTables column search value — i.e. the client sends the literal string `'undefined'` in the first place. A few lines further down, in the same file, the `requirementsSelectId`/`tagsSelectId` reads already have this guard with a comment explaining the exact failure mode:

```js
d.requirements = $(self.options.requirementsSelectId).val() || [];
d.tags = $(self.options.tagsSelectId).val() || [];
```

So this was the same bug, just not yet patched for the `affixes`/`attributes` pair. Pushed a follow-up commit (`13d2b28ce`) adding the same `.val() || []` guard to both, so the client stops emitting `'undefined'` at all. Kept the backend `is_array()` guards from the earlier commits too — this is a public, unauthenticated endpoint, so it should stay defensive regardless of what the client sends.

Ran the full Vitest suite (440 tests, all green) and the `AjaxDungeonRouteControllerTest` suite (8 tests, all green) after the change. Didn't add a dedicated new JS test for the click handler itself: the existing `table.test.js` explicitly only unit-tests the two pure helper methods that don't need a real jQuery/DataTables instance (see its top-of-file comment) — exercising the click handler would need constructing a full DataTables instance, which is a different scope than this fix. Happy to add that scaffolding separately if you'd like it.

PR #3923, opened 2026-08-08
Guard heatmap timer-fraction filter against timer_max_seconds=0

### LINKED ISSUE #3913: Heatmap data endpoint: DivisionByZeroError when a mapping version's timer_max_seconds is 0
:robot: `CombatLogEventFilter::fromHeatmapDataFilter()` divides by the mapping version's `timer_max_seconds` without guarding against zero, causing a `DivisionByZeroError`.

- Occurrences: 270, Users impacted: 2, ongoing (second-highest volume of this pass's findings)
- First/last seen: 2026-02-10 to 2026-08-01
- Endpoint: `POST /ajax/heatmap/data`

```
DivisionByZeroError: Division by zero
```
```php
517 │     throw new InvalidArgumentException('Mapping version does not have a timer max seconds value');
518 │ }
519 │
520 │ $combatLogEventFilter->setDurationMin((int)(($heatmapDataFilter->getTimerFractionMin() * 60) / $timerSeconds));
```
Note there's already a guard just above (line 517) that throws `InvalidArgumentException` when `timer_max_seconds` isn't set — but that check apparently doesn't catch every zero case (e.g. `timer_max_seconds` explicitly stored as `0` rather than null/unset would pass an `isset`-style check but still divide by zero here). Fix is to make the existing guard check for `<= 0`, not just unset/null.

Wotuu:
:robot: Closes #3913

## Summary
- `CombatLogEventFilter::fromHeatmapDataFilter()` divides by a mapping version's `timer_max_seconds` when the caller supplies a timer-fraction range, guarded by `if ($timerSeconds === null)`. `timer_max_seconds` is a non-nullable `int` column with a `NOT NULL DEFAULT 0` (already documented as a known divide-by-zero hazard in `MappingServiceCreateNewMappingVersionFromMDTMappingTest`), so the `=== null` check could never trip — PHPStan even had a `@phpstan-ignore identical.alwaysFalse` on it acknowledging the dead code. A mapping version that still carries the column default of `0` (e.g. one of the MoP Challenge Mode dungeons, per a recent event's payload) hit `DivisionByZeroError` on `POST /ajax/heatmap/data` instead of the intended `InvalidArgumentException` (Sentry `PHP-LARAVEL-8J`, 270 events, 2 users, ongoing since 2026-02-10 — the highest-volume unfixed crash found in this triage pass).
- Fix: check `$timerSeconds <= 0` instead of `=== null`, and drop the now-meaningful comparison's `@phpstan-ignore`.

Fixes PHP-LARAVEL-8J

## Test plan
- [x] New regression test (`CombatLogEventFilterTest::fromHeatmapDataFilter_givenMappingVersionTimerMaxSecondsIsZero_throwsInvalidArgumentException`) reproduces the exact reported `DivisionByZeroError` without the fix, and passes (asserting the intended `InvalidArgumentException`) with it.
- [x] `composer run fix` — no drift.

---
:robot: Cold review also caught that the block this PR hardens had an inverted fraction→minutes conversion — fixed in a second commit (`f2890d382`) with its own regression test. Separately, this PR intentionally still lets `InvalidArgumentException` surface as a 500 on `POST /ajax/heatmap/data` for affected dungeons (matching what #3913 asked for) rather than adding graceful degradation in the controller — that's a product call on desired behavior (silently drop the filter vs. surface an error), left as a follow-up rather than expanded scope here.

### DIFF
--- a/app/Service/CombatLogEvent/Dtos/CombatLogEventFilter.php
+++ b/app/Service/CombatLogEvent/Dtos/CombatLogEventFilter.php
@@ -513,7 +513,7 @@ class CombatLogEventFilter implements Arrayable
 
         if ($heatmapDataFilter->getTimerFractionMin() !== null && $heatmapDataFilter->getTimerFractionMax() !== null) {
             $timerSeconds = $heatmapDataFilter->getDungeon()->getCurrentMappingVersion()->timer_max_seconds;
-            if ($timerSeconds === null) { // @phpstan-ignore identical.alwaysFalse
+            if ($timerSeconds <= 0) {
                 throw new InvalidArgumentException('Mapping version does not have a timer max seconds value');
             }
 
--- a/app/Service/CombatLogEvent/Dtos/CombatLogEventFilter.php
+++ b/app/Service/CombatLogEvent/Dtos/CombatLogEventFilter.php
@@ -517,8 +517,11 @@ class CombatLogEventFilter implements Arrayable
                 throw new InvalidArgumentException('Mapping version does not have a timer max seconds value');
             }
 
-            $combatLogEventFilter->setDurationMin((int)(($heatmapDataFilter->getTimerFractionMin() * 60) / $timerSeconds));
-            $combatLogEventFilter->setDurationMax((int)(($heatmapDataFilter->getTimerFractionMax() * 60) / $timerSeconds));
+            // durationMin/Max are minutes (see toArray()'s duration_ms * 60000 below), and a
+            // fraction of the dungeon's timer converts to seconds as fraction * timerSeconds, so
+            // divide by 60 - not the other way around, which silently truncated every duration to 0
+            $combatLogEventFilter->setDurationMin((int)(($heatmapDataFilter->getTimerFractionMin() * $timerSeconds) / 60));
+            $combatLogEventFilter->setDurationMax((int)(($heatmapDataFilter->getTimerFractionMax() * $timerSeconds) / 60));
         }
 
         return $combatLogEventFilter;
--- a/tests/Feature/App/Service/CombatLogEvent/Dtos/CombatLogEventFilterTest.php
+++ b/tests/Feature/App/Service/CombatLogEvent/Dtos/CombatLogEventFilterTest.php
@@ -1,0 +1,63 @@
+<?php
+
+namespace Tests\Feature\App\Service\CombatLogEvent\Dtos;
+
+use App;
+use App\Models\CombatLog\CombatLogEventDataType;
+use App\Models\CombatLog\CombatLogEventEventType;
+use App\Service\CombatLogEvent\Dtos\CombatLogEventFilter;
+use App\Service\RaiderIO\Dtos\HeatmapDataFilter;
+use App\Service\Season\SeasonAffixGroupServiceInterface;
+use App\Service\Season\SeasonServiceInterface;
+use InvalidArgumentException;
+use PHPUnit\Framework\Attributes\Group;
+use PHPUnit\Framework\Attributes\Test;
+use Tests\Feature\Traits\ProvidesDungeon;
+use Tests\TestCases\PublicTestCase;
+
+#[Group('CombatLogEvent')]
+final class CombatLogEventFilterTest extends PublicTestCase
+{
+    use ProvidesDungeon;
+
+    /**
+     * A mapping version's `timer_max_seconds` column is NOT NULL DEFAULT 0 - fromHeatmapDataFilter()
+     * divides by it when the caller requested a timer-fraction range, so a mapping version that still
+     * carries the column's default of 0 must raise a clean InvalidArgumentException rather than a
+     * DivisionByZeroError (PHP-LARAVEL-8J).
+     */
+    #[Test]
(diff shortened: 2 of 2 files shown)


app/Service/CombatLogEvent/Dtos/CombatLogEventFilter.php
```diff
@@ -513,7 +513,7 @@ public static function fromHeatmapDataFilter(
 
         if ($heatmapDataFilter->getTimerFractionMin() !== null && $heatmapDataFilter->getTimerFractionMax() !== null) {
             $timerSeconds = $heatmapDataFilter->getDungeon()->getCurrentMappingVersion()->timer_max_seconds;
-            if ($timerSeconds === null) { // @phpstan-ignore identical.alwaysFalse
+            if ($timerSeconds <= 0) {
```
Wotuu, line 516: :robot: Pre-existing, but it lives in the exact block this PR is hardening and it is the only thing the new guard protects, so worth flagging here: **the fraction -> minutes conversion on the next two lines (520-521) is inverted.**

`durationMin`/`durationMax` are minutes — `toArray()`/`getQuery()` emits `Range::make('duration_ms', ['gte' => $this->durationMin * 60000, ...])`. And `timer_fraction` is defined as `durationMs / (timer_max_seconds * 1000)` (`CombatLogEvent::setTimeInterval()`). So fraction -> minutes is `fraction * $timerSeconds / 60`, but the code computes `($fraction * 60) / $timerSeconds`, which is dimensionally `1/seconds`.

Concretely, for a 30-minute dungeon (`timer_max_seconds = 1800`) with `minTimerFraction=0.5`, `maxTimerFraction=1.0`:
- current: `(int)(0.5*60/1800) = 0` and `(int)(1.0*60/1800) = 0` -> `duration_ms` range collapses to 0..0 and the heatmap silently returns nothing
- intended: 15 and 30

The `(int)` cast added in `bfac964f4` (phpstan level 6) is what turned a wrong-but-nonzero float into a hard `0`. Blast radius is the self-hosted OpenSearch path (`RaiderIOKeystoneGuruApiService::getHeatmapData()` -> `getGridAggregation()`), which actually queries on `duration_ms`; the Raider.IO path builds its URL from `HeatmapDataFilter::toArray()` and only uses the `CombatLogEventFilter` for response mapping, so it is unaffected.

Not blocking this PR, but since the guard is being made live, the divisor it guards is probably worth fixing (or filing) in the same pass.

app/Service/CombatLogEvent/Dtos/CombatLogEventFilter.php
```diff
@@ -513,7 +513,7 @@ public static function fromHeatmapDataFilter(
...
                 throw new InvalidArgumentException('Mapping version does not have a timer max seconds value');
```
Wotuu, line 517: :robot: Acknowledging that #3913 asks for exactly this change, so this is a question rather than a defect: `InvalidArgumentException` is **not** caught by `AjaxHeatmapController::getData()` (it only catches `InvalidApiResponseException`), so after this fix `POST /ajax/heatmap/data` still returns a 500 for the affected dungeons — the Sentry issue just moves from `DivisionByZeroError` (PHP-LARAVEL-8J) to a fresh `InvalidArgumentException` group with the same ~270 events/period.

The zeros have known producers, so they will keep occurring: `CombatLogMappingVersionService.php:129` writes `'timer_max_seconds' => 0` explicitly, and `MappingService::createNewMappingVersionFromMDTMapping()` only clones `timer_max_seconds` in when `$facadeSourceMappingVersion !== null`.

Was graceful degradation considered — i.e. skipping the duration range (leaving `durationMin`/`durationMax` null) when `$timerSeconds <= 0`, so the heatmap still renders unfiltered, or returning a 422 rather than a 500? Alternatively a follow-up to backfill `timer_max_seconds` for the mapping versions that carry the column default. Non-blocking either way, but worth stating in the PR body which of the two this PR is choosing so the new Sentry group is not a surprise.

Wotuu:

app/Service/CombatLogEvent/Dtos/CombatLogEventFilter.php
Wotuu: :robot: Fixed in f2890d382 (durationMin/Max now compute as fraction * timerSeconds / 60) plus a regression test asserting the exact 7/22-minute bounds for the 30-minute-timer example, confirmed to fail with the old formula (0/0) before the fix.

app/Service/CombatLogEvent/Dtos/CombatLogEventFilter.php
Wotuu: :robot: Leaving this PR narrow: it intentionally converts the unhandled DivisionByZeroError into an explicit, correctly-worded InvalidArgumentException (matching #3913 which asked for exactly this), rather than expanding scope into controller-level graceful degradation. The follow-up idea (catch and skip the duration filter, or return 422, when timer_max_seconds <= 0) needs a product call on the desired behavior that isn't mine to make, so leaving it as a note here rather than filing a new issue for it. Also noted in the PR body.

Wotuu:
:robot: Cold review: 2 findings, both addressed.

PR #3925, opened 2026-08-09
#3911 Fix /ajax/routes 500 when the datatables 'search' param is absent

### LINKED ISSUE #3911: /ajax/routes 500s when the datatables 'search' request param is absent
:robot: `DatatablesColumnHandler::applyToBuilder()` accesses `\$request->get('search')['value']` without checking that `search` was actually sent, causing "Trying to access array offset on null" when it's missing.

- Occurrences: 3, Users impacted: 2, ongoing
- First/last seen: 2026-07-27 to 2026-08-01
- Endpoint: `GET /ajax/routes`

```
ErrorException: Trying to access array offset on null
```
```php
83 │ if (is_array($order)) {
84 │     $order = $order[0] ?? null;
85 │ }
86 │ $generalSearch = ($request->get('search'))['value'];   // <-- null when 'search' absent
```
One event came from a `curl` client with no `search` param at all, so this looks like a caller (possibly an API consumer or an older/bare request) omitting the datatables `search` param entirely rather than sending `search[value]=""`. Fix is a null-safe read, e.g. `$request->input('search.value')` or `($request->get('search') ?? [])['value'] ?? null`.

Wotuu:
:robot: Closes #3911

`DatatablesColumnHandler::applyToBuilder()` indexed into `$request->get('search')['value']` unconditionally, which throws "Trying to access array offset on null" when a caller omits the `search` parameter entirely (e.g. a bare `curl` request) instead of sending `search[value]=""`.

Fix: read it via `$request->input('search.value')`, coerced to `null` when it isn't a string (e.g. `search[value][]=...`).

Also hardened the adjacent, unguarded `$value['name']` column lookup a few lines below (same method) to be null-safe, since a cold review found it 500s the same way for a column entry missing `name`. The `SimpleColumnHandler` `searchable`/`orderable` reads are a similar pattern repeated across several column handlers — left as a follow-up rather than widening this PR's scope.

Added regression tests:
- `get_givenSearchParameterIsMissing_returnsOk`
- `get_givenSearchValueIsAnArray_returnsOk`

Both reproduce the exact reported failure modes against the pre-fix code and pass after.

### DIFF
--- a/app/Logic/Datatables/ColumnHandler/DatatablesColumnHandler.php
+++ b/app/Logic/Datatables/ColumnHandler/DatatablesColumnHandler.php
@@ -84,13 +84,14 @@ abstract class DatatablesColumnHandler
             $order = $order[0] ?? null;
         }
         $generalSearch = $request->input('search.value');
+        $generalSearch = is_string($generalSearch) ? $generalSearch : null;
 
         // Find the column we should handle
         $column = null;
         // Find the index too; needed to handle sorting later on
         $columnIndex = -1;
         foreach ($columns as $index => $value) {
-            if ($value['name'] === $this->columnName) {
+            if (($value['name'] ?? null) === $this->columnName) {
                 $column      = $value;
                 $columnIndex = $index;
                 break;
--- a/app/Logic/Datatables/ColumnHandler/DatatablesColumnHandler.php
+++ b/app/Logic/Datatables/ColumnHandler/DatatablesColumnHandler.php
@@ -83,7 +83,7 @@ abstract class DatatablesColumnHandler
         if (is_array($order)) {
             $order = $order[0] ?? null;
         }
-        $generalSearch = ($request->get('search'))['value'];
+        $generalSearch = $request->input('search.value');
 
         // Find the column we should handle
         $column = null;
--- a/tests/Feature/Controller/Ajax/AjaxDungeonRouteControllerTest.php
+++ b/tests/Feature/Controller/Ajax/AjaxDungeonRouteControllerTest.php
@@ -59,6 +59,33 @@ final class AjaxDungeonRouteControllerTest extends AjaxPublicTestCase
     }
 
     #[Test]
+    public function get_givenSearchParameterIsMissing_returnsOk(): void
+    {
+        // Arrange - a caller (e.g. a bare curl request) that omits the datatables 'search'
+        // parameter entirely, rather than sending 'search[value]=""'
+        $query = http_build_query([
+            'draw'    => 1,
+            'start'   => 0,
+            'length'  => 25,
+            'columns' => [
+                [
+                    'data'       => 0,
+                    'name'       => 'title',
+                    'searchable' => 'true',
+                    'orderable'  => 'true',
+                    'search'     => ['value' => '', 'regex' => 'false'],
+                ],
+            ],
+        ]);
+
+        // Act
+        $response = $this->get(sprintf('/ajax/routes?%s', $query));
+
+        // Assert
+        $response->assertOk();
+    }
+
+    #[Test]
     public function get_givenRequirementsParameterIsTheStringUndefined_returnsOk(): void
     {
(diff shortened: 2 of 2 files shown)


app/Logic/Datatables/ColumnHandler/DatatablesColumnHandler.php
```diff
@@ -83,7 +83,7 @@ public function applyToBuilder(Builder $subBuilder, Builder $orderBuilder): self
         if (is_array($order)) {
             $order = $order[0] ?? null;
         }
-        $generalSearch = ($request->get('search'))['value'];
+        $generalSearch = $request->input('search.value');
```
Wotuu, line 86: :robot: The null case is fixed, but the *non-string* case on this same line still 500s: `input('search.value')` happily returns an **array** for `?search[value][]=abc`, and `applyFilter()` declares `?string $generalSearch`, so it dies with a `TypeError` instead.

Verified against the running app (master, i.e. unaffected by this diff):
```
GET /ajax/routes?...&search[value][]=abc  ->  500
```
Since this PR is specifically about hardening this read against callers that don't send a well-formed DataTables payload, it seems worth closing in the same line:

```php
$generalSearch = $request->input('search.value');
$generalSearch = is_string($generalSearch) ? $generalSearch : null;
```

(Note: `$request->string('search.value')` is *not* a safe shortcut here — it does `(string)` on the input, and "Array to string conversion" is a warning that this app's error handler rethrows as `ErrorException`, so you'd get the same 500.)

Wotuu, line 86: :robot: The PR body and the new test's comment both frame this as fixing "a bare `curl` request", but a genuinely bare caller still 500s — including **two lines below this one, in the same method**. `$value['name']` at line 93 is an unguarded array access, and an undefined-array-key warning is rethrown as `ErrorException` by Laravel's handler, exactly like the `search` warning was.

Verified against the running app (master):
```
GET /ajax/routes?...&columns[0][data]=0&search[value]=
  -> 500 ErrorException: Undefined array key "name"
     DatatablesColumnHandler.php:93
GET /ajax/routes?...&columns[0][name]=title&search[value]=
  -> 500 ErrorException: Undefined array key "searchable"   (SimpleColumnHandler.php:50)
GET /ajax/routes?...&columns[0][name]=title&columns[0][searchable]=true&search[value]=
  -> 500 ErrorException: Undefined array key "orderable"    (SimpleColumnHandler.php:58)
```
The new regression test doesn't catch this because its request is fully-formed apart from `search` — every column still carries `data`/`name`/`searchable`/`orderable`/`search`. So the Sentry issue as *described* ("a caller omitting datatables params") isn't fully closed; only the one reported line is.

Minimal in-scope fix for line 93: `if (($value['name'] ?? null) === $this->columnName) {`. The `SimpleColumnHandler` pair (`($columnData['searchable'] ?? null) === 'true'`, same for `orderable`) is arguably a follow-up, but it's the very next thing a bare caller hits. Either way, worth softening the "e.g. a bare curl request" wording in the test comment if these aren't fixed here.

Wotuu:

app/Logic/Datatables/ColumnHandler/DatatablesColumnHandler.php
Wotuu: :robot: Fixed as suggested — `$generalSearch` is now coerced to null when non-string, and covered by a new `get_givenSearchValueIsAnArray_returnsOk` test. Thanks for catching this.

Wotuu: :robot: Fixed line 93's `$value['name']` read to be null-safe. Left the `SimpleColumnHandler` searchable/orderable pair as a follow-up per your note — it's a different pattern repeated across several column handlers and widening scope there felt like it deserved its own pass rather than piggybacking on this fix. Softened the test/PR framing accordingly.

PR #3926, opened 2026-08-09
#3912 Fall back to enemy_forces when a teeming NPC has no enemy_forces_teeming override

### LINKED ISSUE #3912: MDT import crashes with TypeError when a pull's enemy-forces amount is null
:robot: `PullImporter::parseMdtNpcClonesInPull()` calls `ImportStringPulls::addEnemyForces(int $amount)` with a value that can be `null`, raising an uncaught `TypeError` since the parameter is typed `int` (no `?`).

Two Sentry clusters with an identical message, seemingly grouped separately (different stack fingerprint) — likely the same underlying bug hit via two slightly different call sites/inputs:

**PHP-LARAVEL-RV** — 3 occurrences, 1 user, 2026-08-06
**PHP-LARAVEL-RT** — 3 occurrences, 1 user, 2026-08-06 (same time window as RV; probably a duplicate cluster of the same bug)

```
TypeError: App\Service\MDT\Models\ImportStringPulls::addEnemyForces(): Argument #1 ($amount) must be of type int, null given, called in /var/www/html/app/Service/MDT/Import/PullImporter.php on line 292
```
```php
78 │ return $this->mdtPulls;
79 │ }
80 │
81 │ public function addEnemyForces(int $amount): self
82 │ {
83 │     $this->enemyForces += $amount;
```
Endpoint: `POST /ajax/mdt/details`. This means some MDT import strings encode a pull's enemy-forces contribution as `null` (rather than omitting it or `0`) — either widen `addEnemyForces()`'s parameter to accept `null` and coalesce to `0`, or fix the caller in `PullImporter::parseMdtNpcClonesInPull()` (around line 292) to coalesce before calling.

Wotuu:
Closes #3912

:robot: `PullImporter::parseMdtNpcClonesInPull()` passed `NpcEnemyForces::enemy_forces_teeming` directly into `ImportStringPulls::addEnemyForces(int $amount)` for teeming routes. That column is `int|null` (no teeming-specific override for the npc) - confirmed via Sentry (`PHP-LARAVEL-RV`/`PHP-LARAVEL-RT`, both fingerprints of the same bug, all 6 events at `PullImporter.php:292`) that this null reaches production and raises an uncaught `TypeError` on `POST /ajax/mdt/details`.

Every other reader of this column already falls back to `enemy_forces` when it's null (`KillZone::getSkippableEnemyForces()`, `DungeonRoute`, `DungeonRouteRepository` all use `IFNULL(enemy_forces_teeming, enemy_forces)` in SQL) - `PullImporter` now does the same with `?? $npcEnemyForces->enemy_forces`.

## Test plan
- [x] Added `MDTImportStringServicePullsTest::getDungeonRoute_givenTeemingRouteAndEnemyWithoutEnemyForcesTeemingOverride_fallsBackToEnemyForces` - forces a real seeded `NpcEnemyForces` row's `enemy_forces_teeming` to `null`, exports/imports a teeming route through that npc, asserts it doesn't throw. Verified it reproduces the crash on the pre-fix code (`git stash` the fix, rerun - `TypeError`) and passes after.
- [x] Ran the new test 5x in a row to rule out flakiness from `getSafeMdtEnemies()`'s shuffle

### DIFF
--- a/app/Service/MDT/Import/PullImporter.php
+++ b/app/Service/MDT/Import/PullImporter.php
@@ -289,9 +289,12 @@ class PullImporter
                 $npcEnemyForces = $enemyForcesByNpcIds->get($enemy->npc->id);
 
                 if ($npcEnemyForces !== null) {
+                    // enemy_forces_teeming may be null (no teeming-specific override for this NPC) - fall back
+                    // to the regular value, matching the IFNULL(enemy_forces_teeming, enemy_forces) convention
+                    // used everywhere else this column is read (e.g. KillZone::getSkippableEnemyForces()).
                     $importStringPulls->addEnemyForces(
                         $importStringPulls->isRouteTeeming() ?
-                            $npcEnemyForces->enemy_forces_teeming :
+                            $npcEnemyForces->enemy_forces_teeming ?? $npcEnemyForces->enemy_forces :
                             $npcEnemyForces->enemy_forces,
                     );
                 } else {
--- a/tests/Feature/App/Service/MDT/MDTImportStringServicePullsTest.php
+++ b/tests/Feature/App/Service/MDT/MDTImportStringServicePullsTest.php
@@ -3,6 +3,7 @@
 namespace Tests\Feature\App\Service\MDT;
 
 use App\Models\KillZone\KillZone;
+use App\Models\Npc\NpcEnemyForces;
 use PHPUnit\Framework\Attributes\Group;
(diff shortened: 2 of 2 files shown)


Wotuu:
:robot: Cold review skipped under the trivial-change rule (4 net lines changed in production code, no UI/schema/auth impact, covered by a new passing test).

Issue #3917, opened 2026-08-08
KillZone path calculation crashes when a kill location lands on a facade floor


Wotuu:
:robot: `CoordinatesService::calculateIngameLocationForMapLocation()` throws when asked to convert a `latlng` that sits on a facade floor, and `KillZonePathService::loadAndBuildGraph()` doesn't guard against this before calling it — surfacing as an uncaught 500 on kill-zone save.

- Occurrences: 49, Users impacted: 2, ongoing
- First/last seen: 2026-05-14 to 2026-07-29
- Endpoint: `PUT /ajax/{dungeonRoute}/killzone/{killZone}`

```
InvalidArgumentException: Unable to convert latlng {"lat":-209.58,"lng":62.06,"floor_id":422} that is on facade floor!
```
```php
29 │ if ($floor === null) {
30 │     throw new InvalidArgumentException('No floor set for latlng!');
31 │ } elseif ($floor->facade) {
32 │     throw new InvalidArgumentException(
33 │         sprintf('Unable to convert latlng %s that is on facade floor!', ...),
```
Called from `KillZonePathService::loadAndBuildGraph()` when computing path nodes from kill zone locations. This throws whenever a kill zone's location was placed on a facade (combined multi-floor) view rather than a real floor — the path-graph builder should either skip facade-floor kill zones when computing ingame paths, or resolve the underlying real floor for the coordinate conversion instead of calling this with the facade floor directly.

Likely related to `PHP-LARAVEL-RY` (also filed this pass) — both originate from the same `getKillZonePaths()` call inside `AjaxKillZoneController`'s save flow.

Wotuu:
The real bug here is that the KillZone's LatLong ends up on a facade floor in the first place. The facade floor is just that: a facade. It should always land on the original floor that the facade is representing.

PR #3927, opened 2026-08-09
#3916 Widen AjaxKillZoneController::store()'s return type to allow its 404 fallback

### LINKED ISSUE #3916: AjaxKillZoneController closure return-type violation when saveKillZone's not-found fallback fires
:robot: The `AjaxKillZoneController` route closure declares a `: \App\Models\KillZone\KillZone` return type, but its `catch (\Exception)` fallback branch returns a plain `Illuminate\Http\Response` (a 404) instead — violating PHP's return type declaration and raising a `TypeError` rather than actually returning the intended 404.

- Occurrences: 1, Users impacted: 1
- First/last seen: 2026-08-07
- Endpoint: `PUT /ajax/{dungeonRoute}/killzone/{killZone}`

```
TypeError: ...(): Return value must be of type App\Models\KillZone\KillZone, Illuminate\Http\Response returned
```
```php
30 │     $result = $this->saveKillZone($coordinatesService, $dungeonRoute, $data, $killZone !== null);
31 │     $result->setAttribute('killzone_paths', $this->getKillZonePaths($killZonePathService, $dungeonRoute));
32 │ } catch (...) {
33 │     $result = response(__('controller.generic.error.not_found'), \Teapot\StatusCode\Http::NOT_FOUND);
34 │ }
35 │
36 │ return $result;   // <-- violates the closure's declared KillZone return type when the catch fired
```
Fix is to widen the closure's return type to `\App\Models\KillZone\KillZone|\Illuminate\Http\Response` (or return the 404 via `abort()`/a thrown `HttpException` instead of assigning it to `$result`), so the intended not-found response actually reaches the client instead of being masked by this TypeError.

Likely related to `PHP-LARAVEL-FQ` (also filed this pass) — both surface from `getKillZonePaths()` throwing inside this same try/catch, so this return-type bug may be the reason FQ's `InvalidArgumentException` reaches the client as a 500 instead of the intended 404 fallback.

Wotuu:
:robot: Closes #3916

`AjaxKillZoneController::store()` declared a non-nullable `KillZone` return type, but its `catch (Exception)` fallback assigns a plain `Response` (a 404) to `$result`. That fallback itself threw a `TypeError` instead of ever reaching the client as the intended 404 — masking whatever the real underlying exception was (e.g. #3917's facade-floor crash in `getKillZonePaths()`, filed the same triage pass) as an opaque 500.

Fix: widen the return type to `KillZone|Response`.

Following a cold review, also:
- `report()` the swallowed exception in the fallback before returning the 404, so a genuine failure still reaches Sentry instead of vanishing silently now that the fallback actually works.
- Isolate `getKillZonePaths()` into its own try/catch. By the time it runs, the kill zone is already saved and broadcast — a failure there now degrades to `killzone_paths: []` instead of discarding the whole request as a 404, so the client still gets its id back and a natural retry can't create a duplicate kill zone.

Regression test (`store_givenGetKillZonePathsThrows_savesTheKillZoneAndReturnsEmptyPathsInsteadOfFailingTheWholeRequest`) mocks `KillZonePathServiceInterface::calculateForRoute()` to throw and asserts the kill zone is saved with an empty `killzone_paths` in the response. Confirmed it reproduces the exact reported `TypeError: ... Return value must be of type App\Models\KillZone\KillZone, Illuminate\Http\Response returned` against the pre-fix code.

### DIFF
--- a/app/Http/Controllers/Ajax/AjaxKillZoneController.php
+++ b/app/Http/Controllers/Ajax/AjaxKillZoneController.php
@@ -253,7 +253,7 @@ class AjaxKillZoneController extends Controller
         APIKillZoneFormRequest       $request,
         DungeonRoute                 $dungeonRoute,
         ?KillZone                    $killZone = null,
-    ): KillZone {
+    ): KillZone|Response {
         // Outside the try/catch on purpose: an authorization failure must surface as a 403, not be
         // rewritten into the 404 below.
         $dungeonRoute = $this->authorizeKillZoneEdit($dungeonRoute, $killZone);
--- a/app/Http/Controllers/Ajax/AjaxKillZoneController.php
+++ b/app/Http/Controllers/Ajax/AjaxKillZoneController.php
@@ -273,12 +273,25 @@ class AjaxKillZoneController extends Controller
             $data['id'] = $killZone?->id ?? null; // @phpstan-ignore nullsafe.neverNull
 
             $result = $this->saveKillZone($coordinatesService, $dungeonRoute, $data, $killZone !== null);
-            $result->setAttribute('killzone_paths', $this->getKillZonePaths($killZonePathService, $dungeonRoute));
         } catch (AuthorizationException|HttpException $deliberateResponse) {
             // A 403 or a 422 we raised on purpose must not be rewritten into the 404 below
             throw $deliberateResponse;
-        } catch (Exception) {
-            $result = response(__('controller.generic.error.not_found'), Http::NOT_FOUND);
+        } catch (Exception $exception) {
+            report($exception);
+
+            return response(__('controller.generic.error.not_found'), Http::NOT_FOUND);
+        }
+
+        try {
+            // Deliberately isolated from the save above: the kill zone is already persisted and
+            // broadcast at this point, so a failure computing this cosmetic add-on must not be
+            // reported back to the client as a total failure - the caller would never learn the
+            // save actually succeeded and retry with an id-less payload, creating a duplicate.
+            $result->setAttribute('killzone_paths', $this->getKillZonePaths($killZonePathService, $dungeonRoute));
+        } catch (Exception $exception) {
+            report($exception);
+
+            $result->setAttribute('killzone_paths', []);
         }
 
         return $result;
--- a/tests/Feature/Controller/Ajax/AjaxKillZoneControllerTest.php
+++ b/tests/Feature/Controller/Ajax/AjaxKillZoneControllerTest.php
@@ -8,6 +8,8 @@ use App\Models\KillZone\KillZone;
 use App\Models\KillZone\KillZoneEnemy;
 use App\Models\PublishedState;
 use App\Models\User;
+use App\Service\KillZonePath\KillZonePathServiceInterface;
+use Mockery;
 use PHPUnit\Framework\Attributes\Group;
 use PHPUnit\Framework\Attributes\Test;
 use Teapot\StatusCode;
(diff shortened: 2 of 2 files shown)


app/Http/Controllers/Ajax/AjaxKillZoneController.php
```diff
@@ -253,7 +253,7 @@ public function store(
         APIKillZoneFormRequest       $request,
         DungeonRoute                 $dungeonRoute,
         ?KillZone                    $killZone = null,
-    ): KillZone {
+    ): KillZone|Response {
```
Wotuu, line 256: :robot: **Observability: after this fix the root-cause exception disappears from Sentry entirely.**

This PR's stated purpose is to stop the fallback from masking the underlying error. But the fix only changes *how* it is masked: before, the `catch (Exception)` swallowed the real exception and the subsequent `TypeError` at least produced a Sentry-visible 500. After the widening, `store()` returns a clean 404 and the original exception (e.g. the `getKillZonePaths()` facade-floor crash tracked in #3917) is swallowed silently — nothing reaches Sentry at all, so the next occurrence of a *different* underlying cause won't be triaged.

Suggest keeping the exception and reporting it in the fallback at line ~280:

```php
} catch (Exception $exception) {
    report($exception);

    $result = response(__('controller.generic.error.not_found'), Http::NOT_FOUND);
}
```

Scoped deliberately to this one site — the swallow-without-report pattern exists across the other Ajax controllers and changing those is not this PR's business; it just happens to matter here because this endpoint is the one that was under Sentry triage.

Wotuu, line 256: :robot: **Now that the fallback actually works, is a bare 404 the right answer for a request that already wrote?**

In the exact scenario the new test reproduces, the failure happens *after* the write succeeded: `saveKillZone()` has already persisted the kill zone (and broadcast `KillZoneChangedEvent` to collaborators); only the cosmetic `killzone_paths` add-on threw. The new test's own `finally { $this->dungeonRoute->killZones()->delete(); }` is evidence of this — the row exists.

Client-side consequence (`resources/assets/js/custom/models/mapobject.js` ~line 999): the `error:` branch calls `self.setSynced(false)` and never touches `self.id` — the id is only assigned in `success:` via `loadRemoteMapObject(json)`. So after this 404 the pull stays id-less client-side, and the user's natural retry POSTs again with no id, creating a **second** kill zone for the one they think failed. Additionally, `dataType: 'json'` against this plain-text 404 body means the intended `controller.generic.error.not_found` message can't even be surfaced.

Pre-fix this path crashed with the `TypeError` and had the same partial-write outcome, so this isn't a regression — but the diff is what turns it into a working, intended code path, which makes the semantics worth a second look. Degrading gracefully would keep client and DB in sync, e.g. only wrapping the paths call:

```php
try {
    $result->setAttribute('killzone_paths', $this->getKillZonePaths($killZonePathService, $dungeonRoute));
} catch (Exception $exception) {
    report($exception);
    $result->setAttribute('killzone_paths', []);
}
```

so the save reports success (with no paths) rather than total failure. Non-blocking; the one-line type widening is correct either way.

Wotuu:

app/Http/Controllers/Ajax/AjaxKillZoneController.php
Wotuu: :robot: Fixed as suggested — the fallback now calls `report($exception)` before returning the 404, so the underlying cause reaches Sentry again.

Wotuu: :robot: Fixed per your suggestion, generalized slightly — `getKillZonePaths()` now has its own try/catch, isolated from the save. A failure there reports the exception and degrades to `killzone_paths: []` instead of discarding the whole request, so the client gets its id back and a natural retry can't create a duplicate. Covered by an updated regression test.

PR #3928, opened 2026-08-09
#3908 Stop paging Sentry for MDT export on an unsupported dungeon

### LINKED ISSUE #3908: MDT export 500s for dungeons not in the MDT dungeon map (Gate of the Setting Sun, Mogu'shan Palace)
:robot: `GET /ajax/{dungeonRoute}/mdtExport` throws an uncaught error for dungeons whose `mdt_supported` flag is `false`, instead of returning a clean "export not supported for this dungeon" response.

**PHP-LARAVEL-S1** — `gate_of_the_setting_sun`, 8 occurrences, 1 user, 2026-08-07
**PHP-LARAVEL-S2** — `mogu_shan_palace`, 4 occurrences, 1 user, 2026-08-07

```
MDT export error: Unsupported MDT dungeon for dungeon key gate_of_the_setting_sun!
```
Both events' `dungeonroute.dungeon.mdt_supported` is `false` in the payload, so the export service correctly detects the unsupported dungeon and raises — but the route just lets that surface as a 500/uncaught error rather than the controller catching it and returning a 400 with a friendly message (similar to how `MDTImportController` catches its own domain exceptions).

Wotuu:
:robot: Closes #3908

`GET /ajax/{dungeonRoute}/mdtExport` already caught `InvalidMDTDungeonException` via its generic `catch (Exception)` and correctly returned a 400 — the client never actually saw a 500, contrary to the issue title. But that generic branch also calls `Log::error()`, and the `sentry` log channel alerts on error-level logs (`config/logging.php`), so this expected, user-input-driven case (the dungeon simply isn't in MDT's conversion tables — confirmed via `dungeons.mdt_supported`, itself just `Conversion::hasMDTDungeonName($this->key)`) was paging Sentry as if it were an application defect.

Fix: add a dedicated `catch (InvalidMDTDungeonException|InvalidMDTExpansionException)` before the generic catch that returns the same 400 without logging — matching how `MDTImportController::details()` already skips logging its own known/expected domain exceptions (`MDTStringParseException`, `InvalidMDTStringException`, `CliWeakaurasParserNotFoundException`).

Added a regression test (`get_givenDungeonNotSupportedByMdt_returnsBadRequestWithoutLoggingAnError`) that confirms both: the 400 response, and (via `Log::spy()`) that `Log::error()` is no longer called for this case. Confirmed the log assertion fails against the pre-fix code and passes after.

### DIFF
--- a/app/Http/Controllers/Ajax/AjaxDungeonRouteController.php
+++ b/app/Http/Controllers/Ajax/AjaxDungeonRouteController.php
@@ -21,6 +21,8 @@ use App\Logic\Datatables\ColumnHandler\DungeonRoutes\TitleColumnHandler;
 use App\Logic\Datatables\ColumnHandler\DungeonRoutes\ViewsColumnHandler;
 use App\Logic\Datatables\DungeonRoutesDatatablesHandler;
 use App\Logic\MDT\Exception\ImportWarning;
+use App\Logic\MDT\Exception\InvalidMDTDungeonException;
+use App\Logic\MDT\Exception\InvalidMDTExpansionException;
 use App\Models\Dungeon;
 use App\Models\DungeonRoute\DungeonRoute;
 use App\Models\DungeonRoute\DungeonRouteFavorite;
@@ -805,6 +807,13 @@ class AjaxDungeonRouteController extends Controller
                 'mdt_string' => $dungeonRoute,
                 'warnings'   => $warningResult,
             ];
+        } catch (InvalidMDTDungeonException|InvalidMDTExpansionException $ex) {
+            // Expected, user-input-driven case (#3908): the dungeon/expansion is simply absent from
+            // MDT's conversion tables, not an application defect. Skip the Log::error() the generic
+            // catch below does - the sentry log channel alerts on error-level logs
+            // (config/logging.php), and this isn't worth paging on. Matches how MDTImportController
+            // skips logging its own known domain exceptions.
+            return abort(400, sprintf(__('controller.apidungeonroute.mdt_generate_error'), $ex->getMessage()));
         } catch (Exception $ex) {
             Log::error(sprintf('MDT export error: %s', $ex->getMessage()), ['dungeonroute' => $dungeonRoute]);
 
--- a/app/Http/Controllers/Ajax/AjaxDungeonRouteController.php
+++ b/app/Http/Controllers/Ajax/AjaxDungeonRouteController.php
@@ -22,7 +22,6 @@ use App\Logic\Datatables\ColumnHandler\DungeonRoutes\ViewsColumnHandler;
 use App\Logic\Datatables\DungeonRoutesDatatablesHandler;
 use App\Logic\MDT\Exception\ImportWarning;
 use App\Logic\MDT\Exception\InvalidMDTDungeonException;
-use App\Logic\MDT\Exception\InvalidMDTExpansionException;
 use App\Models\Dungeon;
 use App\Models\DungeonRoute\DungeonRoute;
 use App\Models\DungeonRoute\DungeonRouteFavorite;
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/Ajax/AjaxDungeonRouteController.php
```diff
@@ -805,6 +807,13 @@ public function mdtExport(
                 'mdt_string' => $dungeonRoute,
                 'warnings'   => $warningResult,
             ];
+        } catch (InvalidMDTDungeonException|InvalidMDTExpansionException $ex) {
```
Wotuu: :robot: `InvalidMDTExpansionException` shouldn't share this silent arm — its semantics are the opposite of the dungeon case.

`MDTDungeon::__construct()` only reaches the expansion throw *after* `Conversion::hasMDTDungeonName()` has already passed, and `Dungeon::mdt_supported` is exactly `hasMDTDungeonName()`. So whenever `InvalidMDTExpansionException` fires, `mdt_supported` is necessarily **true** — every UI entrypoint (`dungeonroute/view|edit|present|embed|livesession` blades, `map.js:114`) has therefore rendered the MDT export button, the user clicks it, gets a 400, and now there is no signal anywhere. That's the inverse of the dungeon case, which is safe to silence precisely *because* the UI hides the button when `mdt_supported` is false.

It is unreachable today (every expansion key present in `DUNGEON_NAME_MAPPING` has a non-null `EXPANSION_NAME_MAPPING` value; `TBC` is null but has no dungeon entries), so this is a semantics correction rather than a live bug — but if the two tables ever drift (new expansion added to `DUNGEON_NAME_MAPPING`, no MDT expansion name), the failure mode is a user-visible broken button that Sentry now stays quiet about.

Suggest dropping it from the union and letting it fall through to the generic `Log::error` arm: a dungeon mapped under an expansion with no MDT expansion name is an internal table inconsistency, which is exactly what Sentry is for. Doesn't affect the new test.

tests/Feature/Controller/Ajax/AjaxDungeonRouteControllerTest.php
```diff
@@ -199,6 +201,42 @@ private function createDungeonRouteForActiveDungeon(): DungeonRoute
         ]);
     }
 
+    /**
+     * Guards #3908: mdtExport() must return a clean 400 for a dungeon MDTDungeon doesn't recognize
+     * (mdt_supported === false), rather than letting InvalidMDTDungeonException surface uncaught -
```
Wotuu: :robot: This docblock re-introduces the exact misconception the PR body debunks: "rather than letting `InvalidMDTDungeonException` surface uncaught" — it never surfaced uncaught, the pre-existing generic `catch (Exception)` already returned a 400. A future reader will take `assertStatus(400)` as the load-bearing assertion, when the regression this test actually guards is the `shouldNotHaveReceived('error')` line (the 400 passed before the fix too).

Suggest rewording to something like "...must return a clean 400 *without* logging an error — the 400 was already correct, but the generic catch's `Log::error()` forwarded this expected case to Sentry (the sentry log channel alerts on error level; see `config/logging.php`)."

Wotuu:

app/Http/Controllers/Ajax/AjaxDungeonRouteController.php
Wotuu: :robot: Fixed as suggested — dropped `InvalidMDTExpansionException` from the silent catch; it now falls through to the generic, logged catch.

tests/Feature/Controller/Ajax/AjaxDungeonRouteControllerTest.php
Wotuu: :robot: Reworded — thanks for catching that, good point on which assertion is actually load-bearing.

PR #3929, opened 2026-08-09
#3917 Skip facade-floor kill zones when building the ingame path graph

### LINKED ISSUE #3917: KillZone path calculation crashes when a kill location lands on a facade floor

Wotuu:
:robot: Closes #3917

`CoordinatesService::calculateIngameLocationForMapLocation()` rejects a `latlng` on a facade floor, and `KillZonePathService::loadAndBuildGraph()` called it unconditionally for every kill zone's location — crashing with an uncaught `InvalidArgumentException` whenever a kill zone's location was placed on the facade (combined multi-floor) view rather than a real floor. This is the same call that #3916 (merged separately, PR #3927) improved the fallback handling for — this PR fixes the actual root cause.

Fix: skip facade-floor kill zones the same way `loadAndBuildGraph()` already skips facade-floor `DungeonFloorSwitchMarker`s a few lines above, rather than resolving them to a real floor — the facade view is a collaborative-editing convenience, not stored data the path visualization needs to trust.

Following a cold review, also guarded the dungeon-start node a few lines above the kill-zone loop, which had the exact same unguarded call — a stale `dungeon_start_map_icon_id` pointing at an old mapping version's facade floor would throw before the kill-zone loop is even reached, affecting map loads too, not just kill-zone save.

Added a regression test (`findPathsToKillZones_givenKillZoneOnFacadeFloor_skipsKillZoneInsteadOfThrowing`) that creates a kill zone on a facade floor alongside one on a real floor, and asserts the facade one is excluded from the graph while the real one is the only remaining entry. Confirmed it reproduces the exact reported `InvalidArgumentException: Unable to convert latlng ... that is on facade floor!` against the pre-fix code.

### DIFF
--- a/app/Http/Controllers/Ajax/AjaxKillZoneController.php
+++ b/app/Http/Controllers/Ajax/AjaxKillZoneController.php
@@ -165,6 +165,15 @@ class AjaxKillZoneController extends Controller
             // A location the user just placed is worth failing on, so they get feedback instead of
             // watching their marker silently disappear. 422 as a literal, like
             // abortIfDungeonRouteLimitReached() - Teapot's Http does not expose UNPROCESSABLE_ENTITY
+            //
+            // Floor union areas are expected to tile the entire facade map with no gaps, so this
+            // should never trigger - reported to diagnose which mapping version/floor unions do.
+            report(new Exception(sprintf(
+                'Facade location could not be converted to a real floor for mapping version %d, submitted floor %d',
+                $dungeonRoute->mappingVersion->id,
+                $submittedFloor->id,
+            )));
+
             abort(422, __('controller.killzone.error.facade_location_not_convertible'));
         }
 
--- a/app/Http/Controllers/Ajax/AjaxKillZoneController.php
+++ b/app/Http/Controllers/Ajax/AjaxKillZoneController.php
@@ -111,6 +111,7 @@ class AjaxKillZoneController extends Controller
     private function convertSubmittedFacadeLocation(
         CoordinatesServiceInterface $coordinatesService,
         DungeonRoute                $dungeonRoute,
+        KillZone                    $killZone,
         array                      &                       $data,
     ): ?LatLng {
         if (!isset($data['floor_id'], $data['lat'], $data['lng'])) {
@@ -147,12 +148,23 @@ class AjaxKillZoneController extends Controller
 
         // The location falls outside every floor union area, so it belongs to no real floor at all -
         // the dead space of the combined image (0.003% of it, measured across every facade dungeon).
-        // Refuse the save rather than persist a facade floor: every consumer downstream (ingame
-        // coordinate conversion, kill zone paths, MDT export) assumes a real floor, and silently
-        // dropping the location would throw away what the user just placed with no feedback.
+        // It may never be persisted: every consumer downstream (ingame coordinate conversion, kill
+        // zone paths, MDT export) assumes a real floor.
         if ($convertedFloor === null || $convertedFloor->facade) {
-            // 422 as a literal, like abortIfDungeonRouteLimitReached() - Teapot's Http does not
-            // expose UNPROCESSABLE_ENTITY
+            // A location the client is merely echoing back unchanged comes from a row corrupted
+            // before this fix, and refusing it would make that pull uneditable forever. Drop it and
+            // let the save land - the KillZonePathService guard covers the row until it is re-placed.
+            if (!$this->isNewLocationForKillZone($killZone, $data)) {
+                $data['floor_id'] = null;
+                $data['lat']      = null;
+                $data['lng']      = null;
+
+                return null;
+            }
+
+            // A location the user just placed is worth failing on, so they get feedback instead of
+            // watching their marker silently disappear. 422 as a literal, like
+            // abortIfDungeonRouteLimitReached() - Teapot's Http does not expose UNPROCESSABLE_ENTITY
             abort(422, __('controller.killzone.error.facade_location_not_convertible'));
         }
 
@@ -161,6 +173,24 @@ class AjaxKillZoneController extends Controller
         $data['lng']      = $convertedLatLng->getLng();
 
         return $submittedLatLng;
+    }
+
+    /**
+     * Whether the submitted location differs from the one the kill zone already holds, i.e. whether
+     * the user actually placed or moved the kill area rather than the client echoing back what it
+     * loaded. Compared with a tolerance because the coordinates round-trip through JSON as strings.
+     *
+     * @param array<string, mixed> $data
+     */
+    private function isNewLocationForKillZone(KillZone $killZone, array $data): bool
+    {
+        if (!$killZone->exists || $killZone->floor_id === null) {
+            return true;
+        }
+
+        return (int)$killZone->floor_id !== (int)$data['floor_id'] ||
+            abs((float)$killZone->lat - (float)$data['lat']) > 0.0001 ||
+            abs((float)$killZone->lng - (float)$data['lng']) > 0.0001;
     }
 
     /**
@@ -190,7 +220,7 @@ class AjaxKillZoneController extends Controller
         $this->abortIfDungeonRouteLimitReached($dungeonroute, DungeonRouteLimitType::KillZones);
 
         // Must happen before the row is written - the database may never hold a facade floor
-        $submittedFacadeLatLng = $this->convertSubmittedFacadeLocation($coordinatesService, $dungeonroute, $data);
+        $submittedFacadeLatLng = $this->convertSubmittedFacadeLocation($coordinatesService, $dungeonroute, $killZone, $data);
 
         $beforeModel = clone $killZone;
 
--- a/app/Http/Controllers/Ajax/AjaxKillZoneController.php
+++ b/app/Http/Controllers/Ajax/AjaxKillZoneController.php
@@ -12,9 +12,11 @@ use App\Http\Requests\KillZone\APIDeleteAllFormRequest;
 use App\Http\Requests\KillZone\APIKillZoneFormRequest;
 use App\Http\Requests\KillZone\APIKillZoneMassFormRequest;
 use App\Jobs\RefreshEnemyForces;
+use App\Logic\Structs\LatLng;
 use App\Models\DungeonRoute\DungeonRoute;
 use App\Models\DungeonRoute\DungeonRouteLimitType;
 use App\Models\Enemy;
+use App\Models\Floor\Floor;
 use App\Models\KillZone\KillZone;
 use App\Models\KillZone\KillZoneEnemy;
 use App\Models\KillZone\KillZoneSpell;
@@ -89,6 +91,66 @@ class AjaxKillZoneController extends Controller
     }
 
     /**
+     * Rewrites a submitted kill zone location that is expressed on a facade (combined multi-floor)
+     * map onto the real floor it belongs to, mutating $data so the conversion happens *before* the
+     * row is written rather than as a second update afterwards.
+     *
+     * Whether the submitted location is a facade location is derived from the submitted floor's own
+     * `facade` flag, deliberately not from the acting user's map facade style: that style is
+     * request-scoped state which the save request does not necessarily share with the request that
+     * rendered the map the location was placed on, and a mismatch used to persist the facade floor
+     * verbatim (#3917).
+     *
+     * @throws HttpException when the facade location cannot be expressed on any real floor
+     *
+     * @param  array<string, mixed> $data
+     * @return LatLng|null          the location exactly as submitted, so the response can echo it back in the
+     *                              plane the client is rendering, or null if no facade location was submitted
+     */
+    private function convertSubmittedFacadeLocation(
+        CoordinatesServiceInterface $coordinatesService,
+        DungeonRoute                $dungeonRoute,
+        array                      &                       $data,
+    ): ?LatLng {
+        if (!isset($data['floor_id'], $data['lat'], $data['lng'])) {
+            return null;
+        }
+
+        /** @var Floor|null $submittedFloor */
+        $submittedFloor = Floor::find($data['floor_id']);
+
+        if ($submittedFloor === null || !$submittedFloor->facade) {
+            return null;
+        }
+
+        $submittedLatLng = new LatLng((float)$data['lat'], (float)$data['lng'], $submittedFloor);
+
+        $convertedLatLng = $coordinatesService->convertFacadeMapLocationToMapLocation(
+            $dungeonRoute->mappingVersion,
+            $submittedLatLng,
+        );
+
+        $convertedFloor = $convertedLatLng->getFloor();
+
+        // convertFacadeMapLocationToMapLocation() is a no-op when the mapping version does not have
+        // facades enabled, and it hands back the facade floor unchanged when the location falls
+        // outside every floor union area. Refuse the save rather than persist a facade floor: every
+        // consumer downstream (ingame coordinate conversion, kill zone paths, MDT export) assumes a
+        // real floor, and silently dropping the location would throw away what the user just placed.
+        if ($convertedFloor === null || $convertedFloor->facade) {
+            // 422 as a literal, like abortIfDungeonRouteLimitReached() - Teapot's Http does not
+            // expose UNPROCESSABLE_ENTITY
+            abort(422, __('controller.killzone.error.facade_location_not_convertible'));
+        }
+
+        $data['floor_id'] = $convertedFloor->id;
+        $data['lat']      = $convertedLatLng->getLat();
+        $data['lng']      = $convertedLatLng->getLng();
+
+        return $submittedLatLng;
+    }
+
+    /**
      * @throws \Exception
      *
      * @param array<string, mixed> $data
@@ -113,6 +175,9 @@ class AjaxKillZoneController extends Controller
         $dungeonroute = $killZone->dungeonRoute ?? $dungeonroute;
 
         $this->abortIfDungeonRouteLimitReached($dungeonroute, DungeonRouteLimitType::KillZones);
+
+        // Must happen before the row is written - the database may never hold a facade floor
+        $submittedFacadeLatLng = $this->convertSubmittedFacadeLocation($coordinatesService, $dungeonroute, $data);
 
         $beforeModel = clone $killZone;
 
@@ -200,34 +265,13 @@ class AjaxKillZoneController extends Controller
                 $afterAttributes['enemies']  = $killZone->enemies->pluck('id');
             });
 
-            // If killzone has lat/lng set, convert it to facade location if it's not already
-            $useFacade = $dungeonroute->mappingVersion->facade_enabled &&
-                User::getCurrentUserMapFacadeStyle() === User::MAP_FACADE_STYLE_FACADE;
-
-            if ($killZone->hasKillArea()) {
-                // Track this latlng so we can re-echo it back to the user if we still want to use facades
-                $originalLatLng = $killZone->getLatLng();
-
-                // Only when we actually have a kill location set!
-                if ($useFacade && $originalLatLng->getFloor() !== null) {
-                    $latLng = $coordinatesService->convertFacadeMapLocationToMapLocation(
-                        $dungeonroute->mappingVersion,
-                        $originalLatLng,
-                    );
-
-                    // Save the killzone with converted lat/lngs in the database
-                    $killZone->update([
-                        'lat'      => $latLng->getLat(),
-                        'lng'      => $latLng->getLng(),
-                        'floor_id' => $latLng->getFloor()->id,
-                    ]);
-                }
-
-                // But echo it back as facade!
-                $killZone->setAttribute('lat', $originalLatLng->getLat());
-                $killZone->setAttribute('lng', $originalLatLng->getLng());
-                $killZone->setAttribute('floor_id', $originalLatLng->getFloor()->id);
-                $killZone->setRelation('floor', $originalLatLng->getFloor());
+            // The row now holds the real-floor location; echo the location back in the plane the
+            // client submitted it in, so the marker it just placed does not jump on its map
+            if ($submittedFacadeLatLng !== null) {
+                $killZone->setAttribute('lat', $submittedFacadeLatLng->getLat());
+                $killZone->setAttribute('lng', $submittedFacadeLatLng->getLng());
+                $killZone->setAttribute('floor_id', $submittedFacadeLatLng->getFloor()->id);
+                $killZone->setRelation('floor', $submittedFacadeLatLng->getFloor());
             }
 
             if (Auth::check()) {
--- a/app/Http/Controllers/Ajax/AjaxKillZoneController.php
+++ b/app/Http/Controllers/Ajax/AjaxKillZoneController.php
@@ -101,11 +101,12 @@ class AjaxKillZoneController extends Controller
      * rendered the map the location was placed on, and a mismatch used to persist the facade floor
      * verbatim (#3917).
      *
-     * @throws HttpException when the facade location cannot be expressed on any real floor
+     * @throws HttpException when a facade location on a facade-rendering route belongs to no real floor
      *
      * @param  array<string, mixed> $data
      * @return LatLng|null          the location exactly as submitted, so the response can echo it back in the
-     *                              plane the client is rendering, or null if no facade location was submitted
+     *                              plane the client is rendering, or null if no facade location was
+     *                              submitted or the submitted one was dropped
      */
     private function convertSubmittedFacadeLocation(
         CoordinatesServiceInterface $coordinatesService,
@@ -123,6 +124,18 @@ class AjaxKillZoneController extends Controller
             return null;
         }
 
+        // The route's map never renders a facade, so this cannot be a location the user just placed
+        // on one - it is a bad location echoed back by a client that loaded an already-corrupted row
+        // (convertFacadeMapLocationToMapLocation() would hand it straight back unconverted anyway).
+        // Drop it so the save still succeeds: failing here would make every such pull uneditable.
+        if (!$dungeonRoute->mappingVersion->facade_enabled) {
+            $data['floor_id'] = null;
+            $data['lat']      = null;
+            $data['lng']      = null;
+
+            return null;
+        }
+
         $submittedLatLng = new LatLng((float)$data['lat'], (float)$data['lng'], $submittedFloor);
 
         $convertedLatLng = $coordinatesService->convertFacadeMapLocationToMapLocation(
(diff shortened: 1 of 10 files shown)


app/Service/KillZonePath/KillZonePathService.php
```diff
@@ -126,7 +126,12 @@ private function loadAndBuildGraph(DungeonRoute $dungeonRoute): array
         $killZoneNodeIds = [];
         foreach ($killZones as $killZone) {
             $killLocation = $killZone->getKillLocation();
-            if ($killLocation === null || $killLocation->getFloor() === null) {
+            // A facade floor's coordinates aren't real ingame coordinates, so
+            // calculateIngameLocationForMapLocation() below rejects them - mirrors the floor-switch
+            // marker skip above (#3917). Excluding the kill zone from the graph entirely (rather than
+            // resolving its real floor) matches that same precedent: this is the collaborative-editing
+            // combined view, not stored data the path visualization needs to trust.
+            if ($killLocation === null || $killLocation->getFloor() === null || $killLocation->getFloor()->facade) {
```
Wotuu, line 141: :robot: **The same crash is still unguarded for the dungeon-start node a few lines above (L99-103).**

`$dungeonStart->getLatLng()` is passed straight into `calculateIngameLocationForMapLocation()` with no facade check, so a dungeon-start map icon on a facade floor throws the exact same `InvalidArgumentException` — and unlike the kill-zone case it throws for *every* route on that mapping version, before the kill-zone loop is even reached (so it hits `MapContextDungeonRoute` map loads as well as kill-zone save).

This isn't hypothetical: in the seeded mapping data there are **87 map icons on facade floors, 18 of them `dungeon_start`**, and for 3 mapping versions (235 Throne of the Tides, 251/253 Dawn of the Infinite) `getDungeonStartMapIcon()`'s `first()` fallback actually returns the facade-floor icon. Those 3 aren't the *current* mapping versions, but `DungeonRoute.mapping_version_id` pins old routes to old mapping versions, so any route on one of them would 500 (0 such routes in the dev DB, so I can't claim live breakage — but the path is real, and map icons keep being saved through `AjaxMapIconController`'s facade conversion, which falls back to the facade floor when the point sits outside every floor-union area).

The fix is provably safe as a one-liner because a missing start node is already an anticipated state everywhere downstream — `findPathsToKillZones()` does `isset($nodes['start']) ? ... : []` and `computePathSegments()` wraps its start segment in `if (isset($nodes['start']))`:

```php
if ($dungeonStart !== null && $dungeonStart->floor?->facade !== true) {
```

Worth doing here rather than as a follow-up (with its own regression test, per the repo's test-enforcement rule) — it's the same bug class this PR exists to fix, and the higher-blast-radius half of it.

tests/Feature/App/Service/KillZonePath/KillZonePathServiceTest.php
```diff
@@ -219,4 +222,57 @@ public function findPathsToKillZones_givenOneWayMarkerWithNullLinkedId_returnsEm
...
+
+    /**
+     * Guards #3917: a kill zone whose location was placed on a facade (combined multi-floor) view
+     * has a facade floor_id, which CoordinatesService::calculateIngameLocationForMapLocation()
+     * rejects - loadAndBuildGraph() must skip that kill zone instead of letting the exception
+     * surface uncaught on save.
+     */
+    #[Test]
+    public function calculateForRoute_givenKillZoneOnFacadeFloor_skipsItInsteadOfThrowing(): void
+    {
```
Wotuu, line 234: :robot: Test name says `calculateForRoute_...` but the test calls `findPathsToKillZones()` — the repo convention is `[functionname]_given[Condition]_returns[ExpectedResult]`, and both sibling tests in this file use `findPathsToKillZones_given...`. Suggest `findPathsToKillZones_givenKillZoneOnFacadeFloor_skipsKillZoneInsteadOfThrowing`.

Secondary: `assertArrayHasKey($realKillZone->id, $result)` is vacuous as a "still resolves" check — once the facade kill zone is skipped, the real one is the only entry left and `findPathsToKillZones()` creates a key for every non-skipped kill zone regardless of whether a path was found (it can be `[]`). Asserting `$this->assertCount(1, $result)` or that the entry is non-empty would actually pin the behaviour down.

Wotuu:

1. `loadAndBuildGraph()`'s dungeon-start node (L99-103) makes the identical unguarded `calculateIngameLocationForMapLocation()` call — same crash class, wider blast radius (throws before the kill-zone loop, so it also breaks map load via `MapContextDungeonRoute`). Seeded mapping data has 18 `dungeon_start` map icons on facade floors and 3 mapping versions where `getDungeonStartMapIcon()` returns one.
2. Test naming/assertion nit: `calculateForRoute_...` names a test that calls `findPathsToKillZones()`, and the "still resolves" assertion is vacuous.

Notes I deliberately did **not** raise as findings:
- **The skip approach is the right call**, not a compromise. The issue's alternative ("resolve the underlying real floor") would use `convertFacadeMapLocationToMapLocation()`, which *silently returns the facade floor* when the point falls outside every floor-union area — almost certainly how this row got persisted in the first place (`AjaxKillZoneController::saveKillZone()` already runs that conversion on save). So resolving isn't reliable and would still need this guard.
- Downside of skipping is real but small and already anticipated downstream: the drawn path connects pull n-1 straight to n+1, bypassing the skipped pull, and `RaidEventsCollection` gets `$pathsToKillZones[$killZone->id] ?? []` (no travel modelled for that pull in the SimC export). Both degrade gracefully.
- Genuine root cause worth a follow-up issue (out of scope here): the save path persists a `floor_id` that can never be converted back, with no feedback to the user, so the pull is silently dropped from the path graph forever. Clamping to a real floor (or rejecting) at save time is where that belongs.

app/Service/KillZonePath/KillZonePathService.php
Wotuu: :robot: Fixed as suggested — guarded the dungeon-start node with the same facade check. Good catch on the wider blast radius.

tests/Feature/App/Service/KillZonePath/KillZonePathServiceTest.php
Wotuu: :robot: Both fixed — renamed to `findPathsToKillZones_givenKillZoneOnFacadeFloor_skipsKillZoneInsteadOfThrowing` and swapped the vacuous `assertArrayHasKey` for `assertCount(1, $result)`.

Wotuu (commented):
Okay, this is a cute fix, but did you ever figure out why KillZone's are getting their lat/longs and their floor IDs set to a facade floor? That's the real issue that we need to fix. I fixed these coordinates by hand for some of them. There may even be a mapping for it, but yet it still keeps happening. I'd like you to figure out why this happens and fix that. You can leave this code change in there, but it should be a fail-safe and not the entire fix.

Wotuu:
:robot: You were right that the guard wasn't the fix. I went and found the write path. Root cause below, then the full inventory, then what changed, then the backfill question — which I'm handing back to you rather than deciding.

## Root cause

`AjaxKillZoneController::saveKillZone()` writes the request's `floor_id`/`lat`/`lng` to the row **first** (`KillZone::create($data)` / `$killZone->update($data)`), and only *afterwards* converts facade coordinates — and only when this held:

```php
$useFacade = $dungeonroute->mappingVersion->facade_enabled &&
    User::getCurrentUserMapFacadeStyle() === User::MAP_FACADE_STYLE_FACADE;
```

So the decision "is this a facade location?" was taken from **request-scoped session state**, not from the data. When that state disagreed with the map the client was actually rendering, the facade `floor_id` and facade-plane lat/lng were persisted verbatim and never touched again.

**Why the state disagrees — and why it keeps happening.** `User::forceMapFacadeStyle()` writes a **static** property (`User::$OVERRIDE_MAP_FACADE_STYLE`). Three controllers set it and none of them ever clear it:

- `DungeonRouteController::embed()` :566
- `DungeonExploreController` :214
- `DungeonHeatmapController` :216

We run Octane/Swoole (`OCTANE_SERVER=swoole`). A static is **worker** state there, not request state — `config/octane.php` only runs `prepareApplicationForNextRequest()` on `RequestReceived`, which flushes container bindings and `Context`, not our own statics. So one hit on `/embed?mapFacadeStyle=split_floors` poisons that worker until it recycles, and *every* later request that worker handles — including kill zone saves by completely unrelated users — sees `split_floors`.

The reason it's intermittent and unfixable by hand: `Floor::indexOrFacade()` reads the same poisoned static, so if the page render and the AJAX save land on the *same* poisoned worker they agree (split render, split save — no bug). The bad row needs the render and the save on **different** workers: a clean worker renders the facade map, the client places a pull and POSTs facade coordinates, a poisoned worker saves them raw.

Proof, both runnable: `ResetsMapFacadeStyleOverrideTest` fails on master (`split_floors` survives a whole request), and `AjaxKillZoneControllerTest::store_givenAFacadeLocationWhileTheMapFacadeStyleSaysSplitFloors_persistsTheLocationOnItsRealFloor` fails on master (facade `floor_id` persisted).

Corroboration that the codebase already knew this leaks: `HasLatLng::getCoordinatesData()` carries a *"If we for some reason currently have facade floor assigned to this map icon (shouldn't happen, but just in case)"* branch.

### The cold review's lead, checked and rejected as the primary cause

The suggestion was that `convertFacadeMapLocationToMapLocation()` silently returns the facade floor when the point falls outside every floor union area. It does — but I measured it. Sampling an 8×8 lat/lng grid across the facade map of all 49 facade-enabled dungeons in the seeded mapping (77,221 points): **2 unconvertible, 0.003%**. Floor union areas effectively tile the whole facade map. It's a real hole, and it's now closed, but it is not what produced your rows.

## Write-path inventory

| # | Path | Can it persist a facade `floor_id`? |
|---|------|--------------------------------------|
| 1 | `AjaxKillZoneController::saveKillZone()` :124/:138 — `create`/`update` with raw `$data` | **Yes — this is the one.** See above. |
| 2 | `AjaxKillZoneController::saveKillZone()` :222 — the post-write facade conversion | **Yes, secondary.** It re-persisted `$latLng->getFloor()->id`, which is still the facade floor for the 0.003% unconvertible case. The 2026_05_20 repair migration guards this (`if ($newFloor->facade) return;`); the controller did not. |
| 3 | `APIKillZoneFormRequest` :34 | No write, but **no guard either** — `floor_id` is only `Rule::exists(Floor::class,'id')`: any floor of any dungeon, facade included. |
| 4 | `AjaxKillZoneController::storeAll()` / `APIKillZoneMassFormRequest` | No — the mass request has no `floor_id`/`lat`/`lng` rules. It routes through `saveKillZone()`, whose old post-write block could still rewrite an existing row's coords; the new code touches nothing when no location is submitted. |
| 5 | `DungeonRouteKillZoneRelationParser` :67 (demo-route seeder) | Structurally yes — copies `floor_id` verbatim from `database/seeders/dungeondata/**`. **I checked: 0 seeded kill zones on facade floors** (and 0 with any `floor_id`), so the committed JSON is clean. |
| 6 | `DungeonRoute::cloneRelationsInto()` :955 (route clone / save-as) | Propagates only — copies whatever the source row holds. A bad source row spreads to every clone; it never creates one. |
| 7 | `KillZoneFactory` | Test-only, hardcoded `[null,1,2,3]`. |
| 8 | MDT import — `PullImporter::insert()` :315 | No. Inserts `dungeon_route_id`/`color`/`description`/`index` only; MDT has no kill-area concept. |
| 9 | Auto route creator — `DungeonRouteBuilder` :135 | No. `dungeon_route_id`/`color`/`index` only. |
| 10 | Live session / broadcast — `KillZoneChangedEvent`, `OverpulledEnemyService` | No. Read-only / never saved. |
| 11 | API v1 controllers, `mapping:copy`, `BackfillKillZoneEnemyId` | No kill zone coordinate writes. |
| 12 | JS — `addkillzonemapstate.js` :21 (`floor_id = getState().getCurrentFloor().id`) + `killzone.js` :136-151 (lat/lng from `layer.getLatLng()`) | This is the **source** of facade coordinates, and correctly so — the client submits the plane it renders. The server is what must normalise them. |

## What changed

1. **`saveKillZone()` now converts before it writes**, via a new `convertSubmittedFacadeLocation()` that decides facade-ness from **the submitted floor's own `facade` flag** — no session state involved at all. If the location can't be expressed on a real floor (mapping version has facades off, or the 0.003% dead-space case) the save is **refused with a 422** instead of silently persisting a facade floor or silently dropping what the user placed. The response still echoes the location back in the submitted plane, so the marker doesn't jump.
2. **`ResetsMapFacadeStyleOverride` global middleware** nulls `User::forceMapFacadeStyle()` at the start of every request, making the override request-scoped like its three callers already assume. This kills the trigger — not just for kill zones.
3. **The `KillZonePathService` guard stays**, exactly as you asked — it's the fail-safe for rows that already exist.

Note that `AjaxMapIconController` :105 and `SavesPolylines` :43 carry the **identical `$useFacade` pattern**, so map icons and polylines could take facade coordinates the same way. Change 2 removes the trigger for them too, but their write paths still trust the session state. I'd rather do those in their own MR than widen this one — happy to file it.

## Backfill: out of scope here, and one part needs your call

Existing bad rows won't self-heal, and there are **two classes**, not one:

- **Mapping version has `facade_enabled = 1`** — re-running the 2026_05_20 logic would heal these cleanly. Mechanical, but I don't want to ship a data migration you didn't ask for on top of a behaviour fix.
- **Mapping version has `facade_enabled = 0`** — `convertFacadeMapLocationToMapLocation()` returns early, so the 2026_05_20 migration's `if ($newFloor->facade) return;` **permanently skips these**. They survived the migration and are almost certainly the ones you fixed by hand. Repairing them means deciding *which* mapping version's floor unions to project through — your call, not mine. For scale: 70 mapping versions in the seeded mapping have a facade floor while `facade_enabled = 0`.

I can't count the real rows from a worktree DB (mine has 0). The query for prod/staging:

```sql
SELECT mv.facade_enabled, COUNT(*)
FROM kill_zones kz
JOIN floors f ON f.id = kz.floor_id
JOIN dungeon_routes dr ON dr.id = kz.dungeon_route_id
JOIN mapping_versions mv ON mv.id = dr.mapping_version_id
WHERE f.facade = 1
GROUP BY mv.facade_enabled;
```

If you want the backfill in this MR rather than a follow-up, say so and I'll add it.

Wotuu:
:robot: Follow-up on my own comment above — I caught a regression in the first version of the fix before it could bite you.

The 422 I described was too broad. `convertFacadeMapLocationToMapLocation()` returns the facade floor **unchanged by design** when the route's mapping version has `facade_enabled = 0`, so *every* already-corrupted pull in that class (the 70-mapping-version group I mentioned — the ones the 2026_05_20 migration permanently skips, i.e. most likely the ones you fixed by hand) would have hard-failed on save the moment a client echoed its stored location back. Uneditable pulls would have been worse than the bug.

Split into two branches in `b2b094c1f`:

- **Mapping version doesn't render a facade** → the submitted facade location cannot be something the user just placed, so it's a stale value from a corrupted row. The location is **dropped** (`floor_id`/`lat`/`lng` → null) and the save succeeds. The pull stays editable and the marker can be re-placed; the `KillZonePathService` guard is the fail-safe for the row until then.
- **Mapping version does render a facade, but the point falls outside every floor union area** → a genuine placement in the dead space of the combined image (0.003% of it). Still a **422**, so the user gets feedback rather than a silently vanishing marker.

New test `store_givenAFacadeFloorOnARouteThatDoesNotRenderAFacade_dropsTheLocationInsteadOfFailingTheSave` pins the first branch.

I've also re-dispatched a cold review over the two new commits, since the earlier `pr cold reviewed` label only covered the original `KillZonePathService` guard and this adds a global middleware plus a new failure path on a write endpoint. Will report back and resolve whatever it raises.

tests/Feature/Controller/Ajax/AjaxKillZoneControllerTest.php
```diff
@@ -336,6 +341,123 @@ public function deleteAll_givenKillZoneWithEnemies_deletesItThroughItsOwnDeletin
...
+        $facadeLatLng = $coordinatesService->convertMapLocationToFacadeMapLocation($mappingVersion, $enemy->getLatLng());
+        $this->assertTrue((bool)$facadeLatLng->getFloor()->facade, 'Expected the enemy to live inside a floor union area');
+
+        $dungeonRoute = DungeonRoute::factory()->create([
+            'dungeon_id'         => $dungeon->id,
+            'mapping_version_id' => $mappingVersion->id,
+        ]);
+
+        // The state an /embed, /explore or /heatmap request handled by the same worker leaves behind
+        User::forceMapFacadeStyle(User::MAP_FACADE_STYLE_SPLIT_FLOORS);
```
Wotuu: :robot: This test is vacuous with respect to the controller fix — it passes against the pre-fix, style-gated code.

The new `ResetsMapFacadeStyleOverride` is registered globally in `bootstrap/app.php`, so it runs in feature tests too. `ResetsMapFacadeStyleOverrideTest` in this very PR proves it: after a request, `getCurrentUserMapFacadeStyle()` is back to `DEFAULT_MAP_FACADE_STYLE`. So the override arranged here is cleared before `saveKillZone()` ever reads it, and the controller sees the default `MAP_FACADE_STYLE_FACADE`.

Walk the old code with that: `$useFacade = facade_enabled && style === FACADE` is **true**, `hasKillArea()` is true (`floor_id`/`lat`/`lng` all set), so the old block converts and updates the row to the real floor and echoes the facade location back. Every assertion here (`floor_id !== facadeFloor`, `!floor->facade`, `floor_id === enemy->floor_id`, `assertJsonPath('floor_id', facadeFloor->id)`) passes without the `convertSubmittedFacadeLocation()` change.

To actually pin the fix, drive the style through a path the middleware can't clear — `actingAs()` a user whose persisted `map_facade_style` is `MAP_FACADE_STYLE_SPLIT_FLOORS` (`getCurrentUserMapFacadeStyle()` falls through to `Auth::user()?->map_facade_style`). That fails on the pre-fix controller and passes on this one. (Not the cookie route: `getCurrentUserMapFacadeStyle()` reads `$_COOKIE` directly, which `withCookie()` doesn't populate.)

While you're here: submit `$facadeLatLng->getFloor()->id` instead of the separately queried `$facadeFloor->id`, so the arrangement stays self-consistent if a dungeon ever has more than one facade floor — otherwise `getFloorUnionsOnFloor()` finds nothing for the submitted floor and the test would hit the new 422 branch instead.

app/Http/Controllers/Ajax/AjaxKillZoneController.php
```diff
@@ -88,6 +90,79 @@ private function authorizeKillZoneEdit(DungeonRoute $dungeonRoute, ?KillZone $ki
...
+
+        // The location falls outside every floor union area, so it belongs to no real floor at all -
+        // the dead space of the combined image (0.003% of it, measured across every facade dungeon).
+        // Refuse the save rather than persist a facade floor: every consumer downstream (ingame
+        // coordinate conversion, kill zone paths, MDT export) assumes a real floor, and silently
+        // dropping the location would throw away what the user just placed with no feedback.
+        if ($convertedFloor === null || $convertedFloor->facade) {
+            // 422 as a literal, like abortIfDungeonRouteLimitReached() - Teapot's Http does not
+            // expose UNPROCESSABLE_ENTITY
+            abort(422, __('controller.killzone.error.facade_location_not_convertible'));
```
Wotuu, line 177: :robot: This 422 can strand exactly the rows the bug produced, making them permanently unsaveable.

Dead space isn't only a fresh-misclick case — it is *how* a facade floor got persisted on a `facade_enabled` route in the first place. In the old code with `$useFacade === true`, `convertFacadeMapLocationToMapLocation()` returns its input unchanged when no floor union area contains the point, and the block below then wrote that facade floor to the row. So facade-enabled routes are precisely where this row shape exists in production.

Such a row is then re-submitted on every subsequent save: `HasLatLng::getCoordinatesData()` reports the facade floor for both planes, so the client holds it and echoes `floor_id`/`lat`/`lng` back even when the user only changes the colour or the enemy selection. With this branch, every one of those saves now 422s and the pull becomes uneditable — the exact failure mode b2b094c1f fixed for the non-facade branch, just on the other side of the split.

Suggestion: only refuse when the user is actually placing something new — `!$killZone->exists`, or the submitted `floor_id`/location differs from the persisted one (compare `floor_id` as an int rather than the floats) — and otherwise drop the location like the sibling branch does. That needs the existing `$killZone` (or its persisted location) passed into `convertSubmittedFacadeLocation()`, which it doesn't currently receive.

tests/Feature/Controller/Ajax/AjaxKillZoneControllerTest.php
```diff
@@ -336,6 +341,123 @@ public function deleteAll_givenKillZoneWithEnemies_deletesItThroughItsOwnDeletin
...
+        } finally {
+            User::forceMapFacadeStyle(null);
+            $dungeonRoute->killZones()->delete();
+            $dungeonRoute->delete();
+        }
+    }
+
+    /**
+     * Guards #3917: a kill zone that already carries a facade floor_id on a route whose mapping
+     * version does not render a facade cannot be converted - convertFacadeMapLocationToMapLocation()
```
Wotuu, line 460: :robot: The two new tests cover the convert branch and the drop branch, but not the 422 branch — which is the only one that can hard-fail a user's save, and the only one that returns a new user-visible string (`controller.killzone.error.facade_location_not_convertible`). CLAUDE.md asks for failure paths and edge cases too.

A test that submits a facade-plane location on a facade-enabled route that falls outside every floor union area (dead space of the combined image) and asserts `422` plus the message would lock in both the status code and the fact that nothing was written — and would also make the behaviour change suggested on `AjaxKillZoneController.php:156` (don't reject a save that merely echoes an already-persisted bad location) explicit rather than implicit.

Wotuu:
:robot: Cold review (opus), second pass over commits c9e7622 + b2b094c: 3 findings posted.

tests/Feature/Controller/Ajax/AjaxKillZoneControllerTest.php
Wotuu: :robot: Fixed in `8a1a3b40c` — you were right, and this was the most valuable finding of the three. Confirmed the vacuousness by checking out master's controller under the new tests: with `forceMapFacadeStyle()` the test passed against pre-fix code; it now fails (persists facade floor 422). Switched the driver to a persisted `map_facade_style = split_floors` on the acting user, which nothing clears — and which is the mismatch that genuinely survives the middleware fix (preference flipped in another tab while a map tab keeps rendering the facade).

app/Http/Controllers/Ajax/AjaxKillZoneController.php
Wotuu: :robot: Fixed in `8a1a3b40c`. Took your suggestion: `convertSubmittedFacadeLocation()` now receives the `$killZone` and only aborts when the location is genuinely new or moved (`!$killZone->exists`, or floor differs, or lat/lng differ beyond a 1e-4 tolerance — they round-trip through JSON as strings). An unchanged echo of an already-corrupted location is dropped instead, same as the `facade_enabled = 0` branch, so those pulls stay editable.

tests/Feature/Controller/Ajax/AjaxKillZoneControllerTest.php
Wotuu: :robot: Fixed in `8a1a3b40c` — added `store_givenANewFacadeLocationThatBelongsToNoRealFloor_refusesTheSave`, which posts a new facade location far outside the map (no floor union area can contain it) and asserts a 422 with nothing written. It also pins the branch split from your previous finding: the 422 stays reachable only for a location the user actually placed.

Wotuu:
:robot: Second cold review came back with 3 findings, all fixed in `8a1a3b40c`, all threads replied to and resolved. CI green (9/9).

1. **My main write-path test was vacuous.** The new `ResetsMapFacadeStyleOverride` middleware also runs in feature tests, so it wiped the `forceMapFacadeStyle()` I'd arranged before the controller ever read it — the pre-fix code converted too and every assertion passed. I confirmed this by running the tests against master's controller. Now driven by a **persisted `map_facade_style = split_floors`** on the acting user, which nothing clears — and which is the mismatch that genuinely survives the middleware fix: you flip the setting in another tab (or on the profile page) while a map tab keeps rendering the facade. Verified it now fails against master's controller (persists the facade floor).
2. **The 422 could strand the very rows the bug produced.** Dead space on a *facade-enabled* route is exactly how a facade floor got persisted there, and the client echoes that location back on every save — so a colour-only edit would have 422'd forever. The abort now only fires for a location the user genuinely **placed or moved** (new kill zone, different floor, or lat/lng differing beyond 1e-4); an unchanged echo is dropped instead, same as the `facade_enabled = 0` branch.
3. **Added the missing test for the 422 branch** — `store_givenANewFacadeLocationThatBelongsToNoRealFloor_refusesTheSave`.

Also fixed a red CI job that was **not** mine but showed up on this run rather than deferring it: `MDTImportStringServicePullsTest::getDungeonRoute_givenTeemingRouteAndEnemyWithoutEnemyForcesTeemingOverride` drew a random MDT dungeon and only *afterwards* asserted one of its safe enemies had an `NpcEnemyForces` row. One of the 24 MDT-compatible non-facade dungeons (`courtofstars`) has none at all, so it failed roughly 1 run in 24. `getMDTCompatibleDungeonRouteWithSafeEnemies()`/`getSafeMdtEnemies()` now take an optional enemy filter so the requirement is part of the draw — the shape `ProvidesDungeon`'s own docblock warns about.

The cold reviewer also noted, outside this diff, that `AjaxMapIconController` and `SavesPolylines` still gate their facade write on `User::getCurrentUserMapFacadeStyle()`. The middleware closes the Octane leak for them too, so only the narrow "preference changed mid-session" mismatch remains there — that's the follow-up MR I mentioned above.

Everything from your review is addressed; the backfill question in my first comment is still yours to call.

Wotuu:

Wotuu:
Your findings are correct - but please note that Octane is only used for a handful of endpoints - on purpose - so while the static variable would be set for subsequent requests, that error would never come up in production since the offending code is never called in the whitelisted Octane endpoints.

That said - I think your middleware is a good guard against this in the future, so it can't regress just in case.

app/Http/Controllers/Ajax/AjaxKillZoneController.php
```diff
@@ -88,6 +90,109 @@ private function authorizeKillZoneEdit(DungeonRoute $dungeonRoute, ?KillZone $ki
...
+                $data['lat']      = null;
+                $data['lng']      = null;
+
+                return null;
+            }
+
+            // A location the user just placed is worth failing on, so they get feedback instead of
+            // watching their marker silently disappear. 422 as a literal, like
+            // abortIfDungeonRouteLimitReached() - Teapot's Http does not expose UNPROCESSABLE_ENTITY
+            abort(422, __('controller.killzone.error.facade_location_not_convertible'));
```
Wotuu, line 177: This is an interesting failure I'd like to have `report()`'ed so we can figure out what mapping version/floor unions etc trigger this. There should always be 100% overlap in Floor Union Areas on the entire map surface.

Wotuu: :robot: Added in `180c773da` — `report()`s a diagnostic exception with the mapping version ID and submitted floor ID right before the `abort(422)`, so Sentry will tell us exactly which mapping version/floor is hitting this.

Wotuu:
:robot: Understood, thanks for the context — the middleware stays as the future-proofing guard against a regression, not a fix for a live production issue. No further action needed here from my side.

PR #3930, opened 2026-08-09
#3904 Don't let a floor without ingame coordinates fail combat log route submission

### LINKED ISSUE #3904: Combat-log route import crashes when a floor has no ingame coordinates set (Halls of Spite)
:robot: `CoordinatesService::calculateMapLocationForIngameLocation()` throws an uncaught `InvalidArgumentException` when a floor's `ingame_min_x/max_x/min_y/max_y` computes to a zero-size map (Halls of Spite, floor id 454), instead of failing gracefully.

- Occurrences: 16, Users impacted: 1, escalating substatus
- First/last seen: 2026-08-04 (single burst)
- Endpoint: `POST /api/v1/combatlog/route`

```
InvalidArgumentException: Floor Halls of Spite (454) does not have ingame coordinates set!
```
```php
62 │ $ingameMapSizeX = $targetFloor->ingame_max_x - $targetFloor->ingame_min_x;
63 │ $ingameMapSizeY = $targetFloor->ingame_max_y - $targetFloor->ingame_min_y;
64 │
65 │ if ((int)$ingameMapSizeX === 0 || (int)$ingameMapSizeY === 0) {
66 │     throw new InvalidArgumentException(...);
67 │ }
```
Call path: `CombatLogRouteDungeonRouteService::saveCombatLogRouteEnemyFailures()` → `CoordinatesService::calculateMapLocationForIngameLocation()`. This throws from a public API endpoint (500), and it looks like floor 454's ingame coordinate columns are simply unset in the mapping data — likely a data/seeder gap rather than a code bug, but the API should not 500 on it.

Wotuu:
Closes #3904

:robot: `CombatLogRouteDungeonRouteService::saveCombatLogRouteEnemyFailures()` records diagnostic bookkeeping (`CombatLogRouteEnemyFailure`) for npcs the combat log route couldn't resolve to a mapped enemy. It called `CoordinatesService::calculateMapLocationForIngameLocation()` unguarded, which throws an `InvalidArgumentException` when the npc's floor has unset (zero-size) ingame coordinates - confirmed via Sentry (`PHP-LARAVEL-RD`, 16 occurrences, Halls of Spite floor 454) that this is a real mapping data gap that reaches production and 500s `POST /api/v1/combatlog/route`.

The dungeon route itself is already built successfully by this point (`saveCombatLogRouteEnemyFailures()` runs after `->build()`) - this table is purely diagnostic triage data, so one unlocatable failure shouldn't fail the whole submission. Now it's caught, logged via a new `saveCombatLogRouteEnemyFailuresUnableToCalculateMapLocation` structured-logging method, and skipped.

## Test plan
- [x] Added `CombatLogRouteDungeonRouteServiceEnemyFailuresTest::saveCombatLogRouteEnemyFailures_givenUnresolvableNpcOnFloorWithoutIngameCoordinates_skipsItWithoutThrowing` - calls the private method directly via reflection (see the class docblock for why: the full API/builder pipeline resolves floors through `FloorRepositorySwoole`/`EnemyRepositorySwoole`, which cache `Floor`/`Enemy` instances for the process lifetime, so mutating a floor's coordinates mid-test wouldn't reliably be observed there). Reproduced the exact reported exception on the pre-fix code (`git stash` the service fix, rerun - `InvalidArgumentException: Floor The High Gate (16) does not have ingame coordinates set!`) and confirmed it passes after.

### DIFF
--- a/app/Service/CombatLog/CombatLogRouteDungeonRouteService.php
+++ b/app/Service/CombatLog/CombatLogRouteDungeonRouteService.php
@@ -508,6 +508,10 @@ class CombatLogRouteDungeonRouteService implements CombatLogRouteDungeonRouteSer
                 continue;
             }
 
+            // Track the floor regardless of whether a failure gets recorded below - later npcs that
+            // fall back to $previousFloor must still see this npc's floor even if this one is skipped.
+            $previousFloor = $currentFloor;
+
             if ($combatLogRouteNpc->getResolvedEnemy() === null) {
                 // This table is diagnostic bookkeeping only (unresolved-npc triage) - a floor with
                 // unset ingame coordinates (a mapping data gap, #3904) must not fail the whole combat
@@ -521,7 +525,7 @@ class CombatLogRouteDungeonRouteService implements CombatLogRouteDungeonRouteSer
                         ),
                     );
                 } catch (InvalidArgumentException) {
-                    $this->log->saveCombatLogRouteEnemyFailuresUnableToCalculateMapLocation($currentFloor->id);
+                    $this->log->saveCombatLogRouteEnemyFailuresUnableToCalculateMapLocation($dungeonRoute->id, $combatLogRouteNpc->npcId, $currentFloor->id);
 
                     continue;
                 }
@@ -536,8 +540,6 @@ class CombatLogRouteDungeonRouteService implements CombatLogRouteDungeonRouteSer
                     'updated_at'         => $now,
                 ], $latLng->toArray());
             }
-
-            $previousFloor = $currentFloor;
         }
 
         CombatLogRouteEnemyFailure::insert($failureAttributes);
@@ -569,13 +571,21 @@ class CombatLogRouteDungeonRouteService implements CombatLogRouteDungeonRouteSer
                 continue;
             }
 
-            $latLng = $this->coordinatesService->calculateMapLocationForIngameLocation(
-                new IngameXY(
-                    $combatLogRouteNpc->coord->x,
-                    $combatLogRouteNpc->coord->y,
-                    $currentFloor,
-                ),
-            );
+            // A floor with unset ingame coordinates (a mapping data gap, #3904) must not fail the
+            // whole request just because this one npc's icon can't be placed.
+            try {
+                $latLng = $this->coordinatesService->calculateMapLocationForIngameLocation(
+                    new IngameXY(
+                        $combatLogRouteNpc->coord->x,
+                        $combatLogRouteNpc->coord->y,
+                        $currentFloor,
+                    ),
+                );
+            } catch (InvalidArgumentException) {
+                $this->log->generateMapIconsUnableToCalculateMapLocation($combatLogRouteNpc->getUniqueId(), $currentFloor->id);
+
+                continue;
+            }
             $latLngs[] = $latLng;
 
             /** @var Npc|null $npc */
--- a/app/Service/CombatLog/CombatLogRouteDungeonRouteService.php
+++ b/app/Service/CombatLog/CombatLogRouteDungeonRouteService.php
@@ -73,6 +73,7 @@ use Exception;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Support\Carbon;
 use Illuminate\Support\Collection;
+use InvalidArgumentException;
 use Ramsey\Uuid\Uuid;
 
 /**
@@ -508,13 +509,22 @@ class CombatLogRouteDungeonRouteService implements CombatLogRouteDungeonRouteSer
             }
 
             if ($combatLogRouteNpc->getResolvedEnemy() === null) {
-                $latLng = $this->coordinatesService->calculateMapLocationForIngameLocation(
-                    new IngameXY(
-                        $combatLogRouteNpc->coord->x,
-                        $combatLogRouteNpc->coord->y,
-                        $currentFloor,
-                    ),
-                );
+                // This table is diagnostic bookkeeping only (unresolved-npc triage) - a floor with
+                // unset ingame coordinates (a mapping data gap, #3904) must not fail the whole combat
+                // log route submission just because it can't be recorded here.
+                try {
+                    $latLng = $this->coordinatesService->calculateMapLocationForIngameLocation(
+                        new IngameXY(
+                            $combatLogRouteNpc->coord->x,
+                            $combatLogRouteNpc->coord->y,
+                            $currentFloor,
+                        ),
+                    );
+                } catch (InvalidArgumentException) {
+                    $this->log->saveCombatLogRouteEnemyFailuresUnableToCalculateMapLocation($currentFloor->id);
+
+                    continue;
+                }
 
                 $failureAttributes[] = array_merge([
                     'dungeon_route_id'   => $dungeonRoute->id,
--- a/tests/Feature/App/Service/CombatLog/CombatLogRouteDungeonRouteServiceEnemyFailuresTest.php
+++ b/tests/Feature/App/Service/CombatLog/CombatLogRouteDungeonRouteServiceEnemyFailuresTest.php
@@ -1,0 +1,93 @@
+<?php
+
+namespace Tests\Feature\App\Service\CombatLog;
+
+use App\Dto\Request\CombatLog\Route\CombatLogRouteCoordRequestDto;
+use App\Dto\Request\CombatLog\Route\CombatLogRouteNpcRequestDto;
+use App\Dto\Request\CombatLog\Route\CombatLogRouteRequestDto;
+use App\Models\CombatLog\CombatLogRouteEnemyFailure;
+use App\Models\DungeonRoute\DungeonRoute;
+use App\Models\Enemy;
+use App\Models\Floor\Floor;
+use App\Service\CombatLog\CombatLogRouteDungeonRouteServiceInterface;
+use Illuminate\Support\Collection;
+use PHPUnit\Framework\Attributes\Group;
+use PHPUnit\Framework\Attributes\Test;
+use ReflectionClass;
+use Tests\TestCases\PublicTestCase;
+
+/**
+ * Exercises CombatLogRouteDungeonRouteService::saveCombatLogRouteEnemyFailures() directly via
+ * reflection rather than through the full API/builder pipeline: that pipeline resolves floors
+ * through FloorRepositorySwoole/EnemyRepositorySwoole, which cache Floor/Enemy instances for the
+ * lifetime of the process, so mutating a floor's ingame coordinates mid-test would not reliably be
+ * observed there. The method under test only ever reads floors via plain Eloquent relations, so a
+ * direct call is both faster and immune to that caching layer.
+ */
+#[Group('CombatLog')]
+#[Group('CombatLogRouteDungeonRouteService')]
+final class CombatLogRouteDungeonRouteServiceEnemyFailuresTest extends PublicTestCase
+{
+    /**
+     * Guards #3904: a floor with unset (zero-size) ingame coordinates must not fail the whole
+     * combat log route submission - CombatLogRouteEnemyFailure is diagnostic bookkeeping only, so a
+     * failure that can't be located should be skipped instead of throwing.
+     */
+    #[Test]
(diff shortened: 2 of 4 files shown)


app/Service/CombatLog/CombatLogRouteDungeonRouteService.php
```diff
@@ -508,13 +509,22 @@ private function saveCombatLogRouteEnemyFailures(
...
-                        $combatLogRouteNpc->coord->x,
-                        $combatLogRouteNpc->coord->y,
-                        $currentFloor,
-                    ),
-                );
+                // This table is diagnostic bookkeeping only (unresolved-npc triage) - a floor with
+                // unset ingame coordinates (a mapping data gap, #3904) must not fail the whole combat
+                // log route submission just because it can't be recorded here.
+                try {
+                    $latLng = $this->coordinatesService->calculateMapLocationForIngameLocation(
```
Wotuu, line 520: :robot: **The identical unguarded call in `generateMapIcons()` is left in place, and it is client-triggerable.**

`generateMapIcons()` (same file, ~line 562) does the exact same `calculateMapLocationForIngameLocation()` call with the exact same `getResolvedEnemy()?->floor ?? $previousFloor` fallback, unguarded. It runs in the same request (`convertCombatLogRouteToDungeonRoute()` line 137) whenever `$combatLogRoute->settings->debugIcons` is truthy — and `settings.debugIcons` is a client-supplied field validated in `CombatLogRouteRequest` (line 76), not a server-side flag. So the same Halls of Spite floor-454 data gap still 500s `POST /api/v1/combatlog/route` for any submitter that sets `debugIcons: true`, after this PR.

Either guard that call the same way (it already has a `generateMapIconsUnableToFindFloor()`-style skip pattern right above it for the null-floor case) or note in the PR body why the debug path is deliberately left throwing.

app/Service/CombatLog/CombatLogRouteDungeonRouteService.php
```diff
@@ -508,13 +509,22 @@ private function saveCombatLogRouteEnemyFailures(
...
+                        new IngameXY(
+                            $combatLogRouteNpc->coord->x,
+                            $combatLogRouteNpc->coord->y,
+                            $currentFloor,
+                        ),
+                    );
+                } catch (InvalidArgumentException) {
+                    $this->log->saveCombatLogRouteEnemyFailuresUnableToCalculateMapLocation($currentFloor->id);
```
Wotuu: :robot: The log event records only `floorId`, which is enough to know *a* floor is unmapped but not enough to triage the record that was dropped: neither the npc nor the dungeon route is recoverable from it. The sibling event right above (`generateMapIconsUnableToFindFloor(string $uniqueId)`) deliberately identifies the npc. Suggest `saveCombatLogRouteEnemyFailuresUnableToCalculateMapLocation(int $floorId, ?int $npcId, int $dungeonRouteId)` (or `$combatLogRouteNpc->getUniqueId()`), since `$combatLogRouteNpc` and `$dungeonRoute` are both in scope here.

tests/Feature/App/Service/CombatLog/CombatLogRouteDungeonRouteServiceEnemyFailuresTest.php
```diff
@@ -0,0 +1,93 @@
...
+            $method  = new ReflectionClass($service)->getMethod('saveCombatLogRouteEnemyFailures');
+
+            // Act
+            $method->invokeArgs($service, [$dungeonRoute->mappingVersion, $combatLogRoute, $dungeonRoute]);
+
+            // Assert - no exception, and no failure recorded for the floor that couldn't be located
+            $this->assertSame($countBefore, CombatLogRouteEnemyFailure::where('dungeon_route_id', $dungeonRoute->id)->count());
+        } finally {
+            $floor->update($originalCoords);
+            CombatLogRouteEnemyFailure::where('dungeon_route_id', $dungeonRoute?->id)->delete();
```
Wotuu: :robot: **This cleanup line can mass-delete rows when `$dungeonRoute` is null.**

`$dungeonRoute` stays `null` if anything between the start of the `try` and the factory call throws (e.g. the `$floor->update()`). The `finally` then runs `CombatLogRouteEnemyFailure::where('dungeon_route_id', null)`, and Laravel's query builder converts a null value into `whereNull()` — so this becomes `DELETE FROM combat_log_route_enemy_failures WHERE dungeon_route_id IS NULL`. That column is nullable (it was added a day after the table in `2026_06_12_000001_add_dungeon_route_id_to_combat_log_route_enemy_failures_table`), so every pre-existing row without a route id in the dev/CI combatlog DB gets wiped on the failure path.

Guard the whole cleanup block:

```php
if ($dungeonRoute !== null) {
    CombatLogRouteEnemyFailure::query()->where('dungeon_route_id', $dungeonRoute->id)->delete();
    $dungeonRoute->delete();
}
```

(also switches to `Model::query()->where()`, which is the convention the sibling `CombatLogRouteDungeonRouteServicePersistenceTest` uses.)

tests/Feature/App/Service/CombatLog/CombatLogRouteDungeonRouteServiceEnemyFailuresTest.php
```diff
@@ -0,0 +1,93 @@
...
+
+            /** @var CombatLogRouteDungeonRouteServiceInterface $service */
+            $service = app(CombatLogRouteDungeonRouteServiceInterface::class);
+            $method  = new ReflectionClass($service)->getMethod('saveCombatLogRouteEnemyFailures');
+
+            // Act
+            $method->invokeArgs($service, [$dungeonRoute->mappingVersion, $combatLogRoute, $dungeonRoute]);
+
+            // Assert - no exception, and no failure recorded for the floor that couldn't be located
+            $this->assertSame($countBefore, CombatLogRouteEnemyFailure::where('dungeon_route_id', $dungeonRoute->id)->count());
```
Wotuu: :robot: **This assertion is `assertSame(0, 0)` — it proves "didn't throw" and nothing else.**

`$dungeonRoute` was created two lines earlier, so `$countBefore` is 0, and the expected post state is also 0. The test therefore cannot distinguish "the unlocatable failure was skipped and everything else still recorded" from "nothing was recorded at all" — which is precisely the safety claim the PR rests on. It would still pass if someone widened the catch to `\Throwable`, hoisted the `try` outside the `getResolvedEnemy() === null` branch, or broke the `?? $previousFloor` fallback so no npc ever reached the insert.

Suggest adding a *third* npc: unresolved, but whose `$previousFloor` is a floor that does have ingame coordinates, then asserting exactly one row is inserted and that it is that npc (`npc_id` / `floor_id`). That also pins the `continue` in the new catch block, which currently skips the trailing `$previousFloor = $currentFloor;` — harmless today only because `$currentFloor === $previousFloor` inside that branch, but nothing in the suite notices if that stops holding.

Secondly: the new `saveCombatLogRouteEnemyFailuresUnableToCalculateMapLocation()` interface method has no coverage at all. `$log` is constructor-typed to `CombatLogRouteDungeonRouteServiceLoggingInterface` and bound by convention in `LoggingServiceProvider`, so `$this->app->instance(CombatLogRouteDungeonRouteServiceLoggingInterface::class, $mock)` before `app()`-resolving the service lets you assert it is called exactly once with the zeroed floor's id.

tests/Feature/App/Service/CombatLog/CombatLogRouteDungeonRouteServiceEnemyFailuresTest.php
```diff
@@ -0,0 +1,93 @@
...
+use PHPUnit\Framework\Attributes\Group;
+use PHPUnit\Framework\Attributes\Test;
+use ReflectionClass;
+use Tests\TestCases\PublicTestCase;
+
+/**
+ * Exercises CombatLogRouteDungeonRouteService::saveCombatLogRouteEnemyFailures() directly via
+ * reflection rather than through the full API/builder pipeline: that pipeline resolves floors
+ * through FloorRepositorySwoole/EnemyRepositorySwoole, which cache Floor/Enemy instances for the
+ * lifetime of the process, so mutating a floor's ingame coordinates mid-test would not reliably be
```
Wotuu: :robot: The stated reason is inaccurate, even though the approach is fine. `OctaneServiceProvider::register()` registers the Swoole repositories with `app()->instance(...)`, and Laravel's `Foundation\Testing\TestCase` calls `refreshApplication()` in `setUp()` — a brand new container (and therefore brand new `FloorRepositorySwoole`/`EnemyRepositorySwoole` instances with empty caches) per test, not per process. A full-pipeline test that zeroes the floor *before* invoking the pipeline would observe the mutation.

The reflection call is still defensible (targeted, fast, no fixture needed), so no rewrite needed — but please reword the docblock so a future reader doesn't inherit a wrong mental model of the Swoole cache lifetime and skip a pipeline-level test on that basis.

Wotuu:

Wotuu:
:robot: Addressed all 5 cold-review findings:
1. `generateMapIcons()` guarded the same way (was reachable via client-supplied `debugIcons: true`)
2. Log call now carries `dungeonRouteId`/`npcId` alongside `floorId`
3. Fixed the null-`dungeonRoute` cleanup path that would have become a `whereNull()` delete
4. Test now proves a *selective* skip (second unresolvable npc on a floor with real coordinates gets recorded, the zeroed one doesn't) and pins `$previousFloor` tracking correctly instead of by coincidence
5. Reworded the reflection-approach docblock per the correction (test-scoped, not process-scoped)

app/Service/CombatLog/CombatLogRouteDungeonRouteService.php
Wotuu: :robot: Fixed — `generateMapIcons()` now wraps the same `calculateMapLocationForIngameLocation()` call in a try/catch and skips just that npc's icon on `InvalidArgumentException`, logged via a new `generateMapIconsUnableToCalculateMapLocation()` event. Commit `cf147fdc8`.

app/Service/CombatLog/CombatLogRouteDungeonRouteService.php
Wotuu: :robot: Fixed — `saveCombatLogRouteEnemyFailuresUnableToCalculateMapLocation()` now takes `dungeonRouteId`/`npcId` alongside `floorId`. Also moved `$previousFloor = $currentFloor` tracking to before the try/catch so it can no longer be skipped by the new `continue` (was only correct by coincidence before). Commit `cf147fdc8`.

tests/Feature/App/Service/CombatLog/CombatLogRouteDungeonRouteServiceEnemyFailuresTest.php
Wotuu: :robot: Fixed — the `finally` cleanup is now guarded on `$dungeonRoute !== null` before deleting, and switched to `Model::query()->where()` to match the sibling test's convention. Commit `cf147fdc8`.

tests/Feature/App/Service/CombatLog/CombatLogRouteDungeonRouteServiceEnemyFailuresTest.php
Wotuu: :robot: Fixed — the test now adds a second unresolvable npc on a floor WITH real coordinates, and asserts exactly one failure is recorded (the one on the good floor, by npc_id/floor_id) rather than just "count unchanged". This proves a selective skip, not "nothing got recorded", and pins the `$previousFloor` tracking correctly. Commit `cf147fdc8`.

tests/Feature/App/Service/CombatLog/CombatLogRouteDungeonRouteServiceEnemyFailuresTest.php
Wotuu: :robot: Fixed — reworded the docblock: the Swoole repository caches are test-scoped (Laravel's `TestCase` rebuilds the container per test via `refreshApplication()`), not process-scoped as previously stated; they just don't reset between requests within a single test, which is why the reflection approach is still used. Commit `cf147fdc8`.

PR #3931, opened 2026-08-09
#3903 Fix Handler::isApiRequest() missing bare /api/, /ajax/ and /benchmark paths

### LINKED ISSUE #3903: header.blade.php crashes with undefined $nextSeason on some render paths
:robot: `resources/views/common/layout/header.blade.php` reads `$nextSeason` unconditionally, but on at least one render path it is never set, causing an `Undefined variable` `ViewException`.

**PHP-LARAVEL-RG** — `$nextSeason`, culprit `/{dungeonRoute}`
- Occurrences: 6, Users impacted: 4
- First/last seen: 2026-08-04 to 2026-08-08 (still recurring)
```
Illuminate\View\ViewException: Undefined variable $nextSeason (View: .../header.blade.php)
```
```php
32 │ $forceShrink         ??= false;
33 │ $dungeonContextLinks ??= null;
34 │
35 │ if ($currentUserGameVersion->key === GameVersion::GAME_VERSION_RETAIL) {
36 │     if ($nextSeason !== null) {   // <-- undefined here
```

Points at the same header partial reading variables a view composer normally injects; worth checking whether the render path in question skips that composer — see the project's own `error-page-ajax-gotchas` note that `Route::fallback` paths can skip view composers.

## Update: the other half of this issue is already fixed

This issue originally also covered **PHP-LARAVEL-54** (`$currentUserGameVersion` undefined on the same partial). That was a mistake — it's the same root cause as #3806, already fixed by #3810 (merged 2026-08-03T14:18Z). Cross-checked against the true Sentry history: PHP-LARAVEL-54 has had **zero events since 2026-08-03T18:49Z**, same day as the fix, out of 3,057 total occurrences going back to 2026-01-16. Dropped from this issue accordingly — only the still-recurring `$nextSeason` cluster remains here.

Wotuu:
Closes #3903

:robot: A GET to a path with no trailing route segment (e.g. `GET /api/`) falls through to the public short-link catch-all route (`{dungeonRoute}`, bound to the leftover segment `"api"`), which 404s. View composers are correctly skipped for that raw, slash-preserving path, but `Handler::isApiRequest()` used `Request::decodedPath()` (which trims the trailing slash to `"api"`) against a `'api/'` prefix check that then never matches — so the 404 rendered as an HTML error view with no composer data, crashing on `header.blade.php`'s unconditional `$nextSeason` read. The same gap also silently missed `/benchmark` entirely.

Root cause was confirmed against the real Sentry event: `GET http://keystone.guru/api/`, `sapi: fpm-fcgi`, route match verified via tinker (`{dungeonRoute}` bound to `"api"`).

## Fix

The API path prefixes (`/ajax/`, `/api/`, `/benchmark`) used to live in `ViewService` as `VIEW_VARIABLES_URL_BLACKLIST`, with `Handler` hand-rolling its own drifting copy. Per review, "is this an API request" is now owned by a service of its own and **both** consumers read from it — the view surface no longer determines what an API request is:

- **`App\Service\Request\ApiRequestService`** + `ApiRequestServiceInterface` — owns the prefixes, exposes `isApiRequest(Request)` / `isApiRequestPath(string $pathInfo)`. Bound with `bind()` in `KeystoneGuruServiceProvider::register()`; dependency-free by design, since it is resolved very early on every request.
- **`ViewService`** constructor-injects the interface; `shouldLoadViewVariables()` is now "whitelisted → true, otherwise → not an API request".
- **`Handler::shouldReturnJson()`** resolves the service directly (matching how it already resolves `HandlerLoggingInterface`); `Handler::isApiRequest()` is deleted and `ViewServiceInterface` is no longer imported there.

`VIEW_VARIABLES_URL_WHITELIST` (`/ajax/dungeonroute/search`, `/ajax/search`, `/ajax/view`) deliberately stays in `ViewService`: those routes *are* API requests, and "they additionally render Blade so composers must run" is a view fact with exactly one consumer. `ViewServiceTest` therefore passes unchanged.

Because `Handler` no longer routes through that whitelist, `/ajax/search` errors are forced to JSON again — i.e. identical to `master`. The earlier revision of this PR had (via the `shouldLoadViewVariables()` delegation) turned them into HTML; that unrequested response-shape change is gone.

The `#3903` fix itself is unchanged: `getPathInfo()` preserves the trailing slash `decodedPath()` trimmed, so bare `/api/`, `/ajax/` and `/benchmark` are all detected.

## Tests

New `tests/Feature/App/Service/Request/ApiRequestServiceTest.php` exercises the extracted service directly: both methods, the load-bearing `'/api/'`-vs-`'/api'` and `/benchmark` boundaries, the whitelisted `/ajax/` paths still classifying as API requests, and an `assertInstanceOf` on `app(ApiRequestServiceInterface::class)` so a missing container binding can't pass silently.

`ErrorResponseTest` keeps the end-to-end coverage of both call sites:
- bare `/api/` (trailing slash) returns a JSON 404 instead of crashing — dispatched via a raw kernel call since Laravel's own test helpers (`MakesHttpRequests::prepareUrlForRequest()`) always trim the trailing slash and can't reproduce this shape
- bare `/api` (no trailing slash) still renders the normal HTML 404 page — it was never an API path, so no behavior change there
- `/benchmark` with a disallowed method returns a JSON 405 instead of crashing
- `/ajax/search` with a disallowed method returns a JSON 405, same as `master`

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -4,7 +4,7 @@ namespace App\Exceptions;
 
 use App\Exceptions\Logging\HandlerLoggingInterface;
 use App\Models\User;
-use App\Service\View\ViewServiceInterface;
+use App\Service\Request\ApiRequestServiceInterface;
 use Auth;
 use Illuminate\Auth\Access\AuthorizationException;
 use Illuminate\Auth\AuthenticationException;
@@ -178,15 +178,21 @@ class Handler extends ExceptionHandler
     }
 
     /**
-     * No view composer runs for a request whose path ViewService::shouldLoadViewVariables()
-     * blacklists (KeystoneGuruServiceProvider::boot() skips registering them there entirely). An
-     * HTML error view rendered for one of those paths is guaranteed to crash on a missing view
-     * variable (#3806, #3903), so isApiRequest() forces JSON for their errors - except
-     * ValidationException, which render() above already carves out since it never renders an
-     * HTML error view in the first place. The blacklist has a handful of /ajax/ routes
-     * (search/view) explicitly whitelisted back out of it to deliberately render an HTML
-     * fragment on success - composers DO run there, so their errors are left alone too and may
-     * render their own HTML instead of JSON.
+     * An API request must never be answered with an HTML error view: no view composer runs for it
+     * (KeystoneGuruServiceProvider::boot() skips registering them for any path
+     * ViewService::shouldLoadViewVariables() rejects, which is every API path bar a few that render
+     * a view on purpose), so such a view is guaranteed to crash on a missing view variable
+     * (#3806, #3903). Force JSON for those - except ValidationException, which render() above
+     * already carves out since it never renders an HTML error view in the first place.
+     *
+     * Whether the request is an API request is ApiRequestService's call, not something derived from
+     * the view layer: it is resolved here rather than constructor-injected to match how this class
+     * already resolves HandlerLoggingInterface, keeping the exception handler constructible even
+     * when the container is in a degraded state. The previous hand-rolled check
+     * (str_starts_with($path, 'api/') || str_starts_with($path, 'ajax/')) that the service replaces
+     * had drifted in two ways: it never covered '/benchmark' at all, and Request::decodedPath()
+     * trims the trailing slash, so a bare "/api/" request decoded to "api" and silently failed
+     * "starts with 'api/'" (#3903).
      */
     #[Override]
     protected function shouldReturnJson($request, Throwable $e): bool
@@ -199,20 +205,8 @@ class Handler extends ExceptionHandler
             return parent::shouldReturnJson($request, $e);
         }
 
-        return $this->isApiRequest($request) || parent::shouldReturnJson($request, $e);
-    }
-
-    /**
-     * Delegates to the same blacklist check ViewService's composer registration already applies,
-     * rather than duplicating its path list by hand - the previous hand-rolled version
-     * (str_starts_with($path, 'api/') || str_starts_with($path, 'ajax/')) drifted from it in two
-     * ways: it never covered '/benchmark' at all, and Request::decodedPath() trims the trailing
-     * slash Request::path() strips, so a bare "/api/" or "/api" request decoded to "api" and
-     * silently failed "starts with 'api/'" (#3903).
-     */
-    private function isApiRequest(Request $request): bool
-    {
-        return !app(ViewServiceInterface::class)->shouldLoadViewVariables($request->getPathInfo());
+        return app(ApiRequestServiceInterface::class)->isApiRequest($request)
+            || parent::shouldReturnJson($request, $e);
     }
 
     /**
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -4,6 +4,7 @@ namespace App\Exceptions;
 
 use App\Exceptions\Logging\HandlerLoggingInterface;
 use App\Models\User;
+use App\Service\View\ViewServiceInterface;
 use Auth;
 use Illuminate\Auth\Access\AuthorizationException;
 use Illuminate\Auth\AuthenticationException;
@@ -201,9 +202,15 @@ class Handler extends ExceptionHandler
 
     private function isApiRequest(Request $request): bool
     {
-        $path = $request->decodedPath();
-
-        return str_starts_with($path, 'api/') || str_starts_with($path, 'ajax/');
+        // Any exception rendered as an HTML error view on a path ViewService blacklists is
+        // guaranteed to crash - no view composer ran there, so the view can't read the variables
+        // it unconditionally depends on (#3806, #3903). This used to duplicate that blacklist as
+        // a hand-rolled prefix check here (str_starts_with($path, 'api/')), which drifted from it
+        // in two ways: it never covered '/benchmark' at all, and decodedPath() trims the trailing
+        // slash Request::path() strips, so a bare "/api/" or "/api" request decodes to "api" and
+        // silently fails "starts with 'api/'". Delegating to the same check both sides already
+        // share removes the duplication instead of patching one more case of it.
+        return !app(ViewServiceInterface::class)->shouldLoadViewVariables($request->getPathInfo());
     }
 
     /**
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -178,13 +178,15 @@ class Handler extends ExceptionHandler
     }
 
     /**
-     * No view composers run on /ajax/ (KeystoneGuruServiceProvider::boot(), gated by
-     * ViewService::shouldLoadViewVariables()) - not even for the handful of /ajax/ routes
-     * whitelisted there to deliberately render an HTML fragment on success (search/view). So an
-     * HTML error view rendered for ANY /ajax/ request is guaranteed to crash on a missing view
-     * variable (#3806). Force JSON for every /ajax/ error (matching how /api/ is already treated
-     * here) except ValidationException, which render() above already carves out since it never
-     * renders an HTML error view in the first place.
+     * No view composer runs for a request whose path ViewService::shouldLoadViewVariables()
+     * blacklists (KeystoneGuruServiceProvider::boot() skips registering them there entirely). An
+     * HTML error view rendered for one of those paths is guaranteed to crash on a missing view
+     * variable (#3806, #3903), so isApiRequest() forces JSON for their errors - except
+     * ValidationException, which render() above already carves out since it never renders an
+     * HTML error view in the first place. The blacklist has a handful of /ajax/ routes
+     * (search/view) explicitly whitelisted back out of it to deliberately render an HTML
+     * fragment on success - composers DO run there, so their errors are left alone too and may
+     * render their own HTML instead of JSON.
      */
     #[Override]
     protected function shouldReturnJson($request, Throwable $e): bool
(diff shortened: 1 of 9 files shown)


app/Exceptions/Handler.php
```diff
@@ -201,9 +202,15 @@ protected function shouldReturnJson($request, Throwable $e): bool
...
-        return str_starts_with($path, 'api/') || str_starts_with($path, 'ajax/');
+        // Any exception rendered as an HTML error view on a path ViewService blacklists is
+        // guaranteed to crash - no view composer ran there, so the view can't read the variables
+        // it unconditionally depends on (#3806, #3903). This used to duplicate that blacklist as
+        // a hand-rolled prefix check here (str_starts_with($path, 'api/')), which drifted from it
+        // in two ways: it never covered '/benchmark' at all, and decodedPath() trims the trailing
+        // slash Request::path() strips, so a bare "/api/" or "/api" request decodes to "api" and
+        // silently fails "starts with 'api/'". Delegating to the same check both sides already
+        // share removes the duplication instead of patching one more case of it.
+        return !app(ViewServiceInterface::class)->shouldLoadViewVariables($request->getPathInfo());
```
Wotuu: :robot: The `shouldReturnJson()` PHPDoc directly above this method (lines 181-188, unchanged by this PR) is now false on two counts, and it's the first thing the next reader of `isApiRequest()` will hit:

1. *"No view composers run on /ajax/ ... **not even for the handful of /ajax/ routes whitelisted there** to deliberately render an HTML fragment on success (search/view)"* — this was already wrong before this PR. `KeystoneGuruServiceProvider::boot()` early-returns **only when `shouldLoadViewVariables()` returns false**, and it returns `true` for the whitelisted paths, so composers *do* register for `/ajax/search`, `/ajax/dungeonroute/search` and `/ajax/view`.
2. *"Force JSON for every /ajax/ error (matching how /api/ is already treated here)"* — newly wrong because of this diff: whitelisted `/ajax/*` paths are no longer forced to JSON (see separate comment).

Since the new inline block here is exactly the rationale for the method, and CLAUDE.md prefers PHPDoc over inline comments, consider merging the two into one corrected PHPDoc on `isApiRequest()`/`shouldReturnJson()` rather than leaving a stale block asserting the opposite of the new behaviour.

Wotuu: :robot: Behavioural change not mentioned in the PR body (which describes the change as purely removing duplication): the three `ViewService::VIEW_VARIABLES_URL_WHITELIST` prefixes — `/ajax/dungeonroute/search`, `/ajax/search`, `/ajax/view` — **lose** forced-JSON error responses. Before: `str_starts_with('ajax/search', 'ajax/')` → `isApiRequest()` true → any non-`ValidationException` error returned JSON. After: whitelisted → `shouldLoadViewVariables()` true → `isApiRequest()` false → falls through to `parent::shouldReturnJson()` (i.e. `expectsJson()`).

I believe this is *correct and not a crash regression*: composers genuinely do run on those paths (`KeystoneGuruServiceProvider::boot()` only early-returns when the check is false), so the rendered HTML error view can't hit the #3806/#3903 missing-variable crash, and `OnlyAjax` returns a bare `403` for non-XHR callers before the Handler is ever reached, so the crawler shape can't land here either.

But it is a real observable change for the live callers: `searchhandler.js` uses `$.ajax({dataType: 'html'})`, whose `Accept: text/html, */*; q=0.01` makes `expectsJson()` false — so a 429 from `throttle:search-dungeonroute` or a 500 on `/ajax/search` now returns a full HTML error page where it previously returned `{"message": ...}`. Two asks: (a) confirm this is intended and say so in the PR body, and (b) pin it with a test in `ErrorResponseTest` (e.g. a plain XHR-shaped GET to `/ajax/search` asserting the new shape) — the current suite has nothing covering the whitelisted prefixes, so a future revert of this semantic would go unnoticed.

tests/Feature/Controller/ErrorResponseTest.php
```diff
@@ -69,6 +73,62 @@ public function fallback_givenPlainGetRequestToUnmatchedAjaxRoute_returns404Json
...
+            'ids'   => 'api',
+            'model' => DungeonRoute::class,
+        ])]);
+    }
+
+    #[Test]
+    public function fallback_givenPlainGetRequestToBareApiPathWithoutTrailingSlash_returns404HtmlPage(): void
+    {
+        // Act - "/api" (no trailing slash) is a different shape from "/api/" above: Laravel's own
+        // Request::path()/getPathInfo() disagree on it, so ViewService::shouldLoadViewVariables()
```
Wotuu: :robot: This comment's central claim can't be verified by this test, because the composer gate is disabled under PHPUnit. In `KeystoneGuruServiceProvider::boot()` the `shouldLoadViewVariables()` early-return lives *inside* `if (!app()->runningUnitTests())` — so in the test environment **every view composer registers for every path**, blacklisted or not. "view composers run normally here and it never crashed" is therefore vacuously true in tests regardless of whether the path is blacklisted, and none of the tests in this PR can actually detect a composer-skip regression; they only pin response *shape* (JSON vs HTML).

Worth (a) correcting/removing that clause so a future reader doesn't take the green test as proof composers ran for the right reason, and (b) adding a direct assertion of the underlying predicate — e.g. `$this->assertFalse(app(ViewServiceInterface::class)->shouldLoadViewVariables('/api/'))` next to the `/api/` test — which is the part that actually differs between the old and new `isApiRequest()`.

Wotuu:

app/Exceptions/Handler.php
Wotuu: :robot: Fixed in 0517aa1e70af479b728dfaf3f17f73991eeb124e — corrected the docblock, added a test pinning the whitelisted-ajax response-shape change, and updated the PR body.

Wotuu: :robot: Fixed in 0517aa1e70af479b728dfaf3f17f73991eeb124e — corrected the docblock, added a test pinning the whitelisted-ajax response-shape change, and updated the PR body.

tests/Feature/Controller/ErrorResponseTest.php
Wotuu: :robot: Fixed in 0517aa1e70af479b728dfaf3f17f73991eeb124e — corrected the docblock, added a test pinning the whitelisted-ajax response-shape change, and updated the PR body.

app/Exceptions/Handler.php
```diff
@@ -199,11 +202,17 @@ protected function shouldReturnJson($request, Throwable $e): bool
...
+     * ways: it never covered '/benchmark' at all, and Request::decodedPath() trims the trailing
+     * slash Request::path() strips, so a bare "/api/" or "/api" request decoded to "api" and
+     * silently failed "starts with 'api/'" (#3903).
+     */
     private function isApiRequest(Request $request): bool
     {
-        $path = $request->decodedPath();
-
-        return str_starts_with($path, 'api/') || str_starts_with($path, 'ajax/');
+        return !app(ViewServiceInterface::class)->shouldLoadViewVariables($request->getPathInfo());
```
Wotuu: I think the code to check whether something is an API request should live somewhere else, and then the view service should read from that place. This handler should also read from that place. Right now, we have it backwards. The view surface does not determine whether something is an API request or not.

Wotuu: :robot: Agreed, and inverted in 0362b88f0.

**New home for the check** — `App\Service\Request\ApiRequestService` (+ `ApiRequestServiceInterface`), bound with `$this->app->bind(...)` in `KeystoneGuruServiceProvider::register()` under `// Internals`. It's dependency-free (like `CoordinatesService`) on purpose: `boot(ViewServiceInterface $viewService)` resolves it transitively on every request, and the exception handler resolves it on every errored one. It owns the API path prefixes verbatim — `'/ajax/'`, `'/api/'`, `'/benchmark'` — and exposes `isApiRequest(Request)` / `isApiRequestPath(string $pathInfo)`.

**Both call sites now read from it:**
- `ViewService` constructor-injects `ApiRequestServiceInterface`; `shouldLoadViewVariables()` is now "whitelisted → true, otherwise → not an API request". `VIEW_VARIABLES_URL_BLACKLIST` is gone from `ViewService`.
- `Handler::shouldReturnJson()` calls `app(ApiRequestServiceInterface::class)->isApiRequest($request)` directly — `Handler::isApiRequest()` is deleted, and `ViewServiceInterface` is no longer imported there at all. It resolves rather than constructor-injects to match how this class already resolves `HandlerLoggingInterface`, keeping the handler constructible if the container is degraded.

**What deliberately did *not* move:** `VIEW_VARIABLES_URL_WHITELIST` (`/ajax/search`, `/ajax/dungeonroute/search`, `/ajax/view`) stays in `ViewService`. Those routes *are* API requests; that they additionally render a Blade fragment is a view fact, and the view layer is its only consumer. Keeping it there means `ViewServiceTest` passes with zero edits.

Side effect worth flagging: this drops the behavior change my previous commit introduced. Because `Handler` no longer goes through the whitelist, `/ajax/search` errors are forced to JSON again — exactly as they were before this PR. I've flipped `fallback_givenPlainWrongMethodRequestToViewRenderingAjaxSearchRoute_returns405Json` back to assert JSON accordingly; a bugfix PR shouldn't have been reshaping that response.

**Tests:** new `tests/Feature/App/Service/Request/ApiRequestServiceTest.php` exercises the service directly (both methods, the load-bearing `'/api/'`-vs-`'/api'` and `'/benchmark'` boundaries, and an `assertInstanceOf` that the container binding actually resolves — otherwise a missing binding would silently pass). `ErrorResponseTest` keeps the end-to-end coverage of both call sites; its comments no longer describe the delegation that's been removed. 46 tests pass across `ApiRequestServiceTest` + `ViewServiceTest` + `ErrorResponseTest`; `composer run fix` and `composer run analyse` are clean.

One thing I left alone: `grep` finds no third hand-rolled `'/api/'`/`'/ajax/'` prefix check anywhere else in `app/`, so the diff stays on the two call sites you named.

Wotuu: :robot: Final state: `84b4963f2`, all 9 checks green. Two follow-ups on top of the extraction commit — `ViewComposerSmokeTest` builds a partial `ViewService` mock with an explicit `setConstructorArgs()` list and needed the new parameter (CI caught it; it was the only such consumer besides `ServiceFixtures`), and a docblock line in `Handler` that overclaimed degraded-container robustness the unguarded `app()` call does not actually provide.

Issue #2800, opened 2025-04-01
Fix MDT string note imports


Wotuu:
[...]

PR #3932, opened 2026-08-09
#3906 Stop logging garbage MDT import strings routed to the Legacy codec as errors

### LINKED ISSUE #3906: /ajax/mdt/details returns uncaught 500s for malformed MDT import strings (base64/compression errors)
:robot: Three related Sentry clusters on `POST /ajax/mdt/details`: user-supplied MDT import strings that fail to base64-decode or decompress raise an uncaught `Error`/`Exception` instead of a clean validation response. All are attacker/user-input-shaped (malformed strings from the wild, including addon-sourced ones based on the `string` field in a couple of events), not server bugs — but they currently 500 instead of returning a friendly "invalid MDT string" response.

**PHP-LARAVEL-RC** — `Failed to decode base64: Invalid length`, 13 occurrences, 6 users, 2026-08-04 to 2026-08-07
**PHP-LARAVEL-RR** — `Failed to decompress data: Invalid prefix`, 9 occurrences, 4 users, 2026-08-05 to 2026-08-08 (still recurring)
**PHP-LARAVEL-RK** — `Failed to decode base64: Invalid byte at offset 5`, 9 occurrences, 6 users, 2026-08-04 to 2026-08-07

All hit the same endpoint/controller (`MDTImportController`, via `MDTImportStringService::setEncodedString()`). The controller already has a broad catch for `MDTStringParseException`/`InvalidMDTStringException`/`CliWeakaurasParserNotFoundException` that map to a 400 response — these particular decode/decompress failures apparently throw a plain `\Error`/`\Exception` that isn't one of those caught types and falls through to the generic `\Throwable` re-throw, producing a 500 for what is really a 400-shaped "bad input" case.

Wotuu:
Closes #3906

:robot: `MDTBaseService::decode()` logged every codec decode failure at `error` level, which the `sentry` log channel forwards to Sentry regardless of whether the exception is genuinely a bug. `MDTStringFormat::detect()` only picks `MDT2` for a strict `!~MDT2~` prefix match (so a decode failure there means our own CBOR codec choked on a string that genuinely claimed to be an MDT export - worth reporting), but defaults to `Legacy` for *everything else*, including garbage that doesn't even look like a legacy export string.

`LegacyMDTCodec` shells out to `cli_weakauras_parser` for that path, and its stderr on malformed input becomes the exception message. I confirmed this directly by decoding one of the actual Sentry event's payloads through the binary:
```
$ printf "%s" "C8PAAAAA...WMA" > /tmp/x && sudo /usr/bin/cli_weakauras_parser decode /tmp/x
Error: Failed to decompress data: Invalid prefix
```
— an exact match for `PHP-LARAVEL-RR`'s message. All three reported clusters (`PHP-LARAVEL-RC`/`RR`/`RK`) are this same Legacy-path noise; the endpoint already returns a clean 400 for it via `MDTImportController`'s `MDTStringParseException` catch (confirmed the controller has zero `Log::error()` calls on that path) - the Sentry issue's "returns uncaught 500s" framing doesn't hold, this is log noise for an already-handled validation failure, not a crash.

Legacy-path decode failures now log at `warning` instead of `error` (still visible in local/daily logs, just not forwarded to Sentry); MDT2-path failures are untouched (`error`), since those are much more likely to be a real bug in our own codec rather than garbage input.

## Test plan
- [x] Added `MDTImportStringServiceDecodeTest::getDecoded_givenGarbageStringRoutedToLegacyCodec_logsAtWarningNotError` - uses `Log::spy()` to assert `warning` (not `error`) is called for garbage input. Confirmed it fails on the pre-fix code (`git stash` the service fix, rerun - `warning` was never called).

### DIFF
--- a/app/Service/MDT/MDTBaseService.php
+++ b/app/Service/MDT/MDTBaseService.php
@@ -45,26 +45,25 @@ abstract class MDTBaseService
      */
     protected function decode(string $string): ?array
     {
-        $format = MDTStringFormat::detect($string);
-
         try {
-            return $format->codec()->decode($string);
+            return MDTStringFormat::detect($string)->codec()->decode($string);
         } catch (CliWeakaurasParserNotFoundException $cliWeakaurasParserNotFoundException) {
             throw $cliWeakaurasParserNotFoundException;
         } catch (Throwable $throwable) {
             // Truncated: the input is unvalidated user input of arbitrary size.
             $context = ['string' => substr($string, 0, 2048)];
 
-            if ($format === MDTStringFormat::MDT2) {
-                // detect() only picks MDT2 for a strict `!~MDT2~` prefix match, so a decode
-                // failure here means our own CBOR codec choked on a string that genuinely claimed
-                // to be an MDT export - always worth reporting.
+            if (MDTStringFormat::isValid($string)) {
+                // The string at least plausibly claims to be an MDT export (one of the codecs'
+                // appliesTo() matched), so a decode failure here is worth reporting - it's either a
+                // genuine bug in our own codec, or (for the Legacy format) a real but corrupted/
+                // truncated export that a user has a right to expect us to import.
                 logger()->error($throwable->getMessage(), $context);
             } else {
-                // detect() defaults to Legacy for everything else, including garbage that doesn't
-                // even look like a legacy MDT export string (LegacyMDTCodec shells out to
-                // cli_weakauras_parser, whose stderr becomes the exception message) - most of these
-                // are just malformed user input, not application bugs (#3906).
+                // detect() unconditionally falls back to Legacy for anything that isn't MDT2, even
+                // garbage that doesn't look like a legacy MDT export at all - LegacyMDTCodec shells
+                // out to cli_weakauras_parser, whose stderr becomes the exception message, so this
+                // covers the common case of a user pasting non-MDT-string input (#3906).
                 logger()->warning($throwable->getMessage(), $context);
             }
 
--- a/app/Service/MDT/MDTBaseService.php
+++ b/app/Service/MDT/MDTBaseService.php
@@ -45,16 +45,28 @@ abstract class MDTBaseService
      */
     protected function decode(string $string): ?array
     {
+        $format = MDTStringFormat::detect($string);
+
         try {
-            return MDTStringFormat::detect($string)->codec()->decode($string);
+            return $format->codec()->decode($string);
         } catch (CliWeakaurasParserNotFoundException $cliWeakaurasParserNotFoundException) {
             throw $cliWeakaurasParserNotFoundException;
         } catch (Throwable $throwable) {
-            // detect() matched a format, so this string genuinely claimed to be an MDT export -
-            // always worth reporting. Truncated: the input is unvalidated user input of arbitrary size.
-            logger()->error($throwable->getMessage(), [
-                'string' => substr($string, 0, 2048),
-            ]);
+            // Truncated: the input is unvalidated user input of arbitrary size.
+            $context = ['string' => substr($string, 0, 2048)];
+
+            if ($format === MDTStringFormat::MDT2) {
+                // detect() only picks MDT2 for a strict `!~MDT2~` prefix match, so a decode
+                // failure here means our own CBOR codec choked on a string that genuinely claimed
+                // to be an MDT export - always worth reporting.
+                logger()->error($throwable->getMessage(), $context);
+            } else {
+                // detect() defaults to Legacy for everything else, including garbage that doesn't
+                // even look like a legacy MDT export string (LegacyMDTCodec shells out to
+                // cli_weakauras_parser, whose stderr becomes the exception message) - most of these
+                // are just malformed user input, not application bugs (#3906).
+                logger()->warning($throwable->getMessage(), $context);
+            }
 
             return null;
         }
--- a/app/Service/MDT/MDTBaseService.php
+++ b/app/Service/MDT/MDTBaseService.php
@@ -4,11 +4,16 @@ namespace App\Service\MDT;
 
 use App\Logic\MDT\Exception\CliWeakaurasParserNotFoundException;
 use App\Logic\MDT\IO\MDTStringFormat;
+use App\Service\MDT\Logging\MDTBaseServiceLoggingInterface;
 use Lua;
 use Throwable;
 
 abstract class MDTBaseService
 {
+    public function __construct(private readonly MDTBaseServiceLoggingInterface $log)
+    {
+    }
+
     /**
      * Gets a Lua instance and load all the required files in it.
      */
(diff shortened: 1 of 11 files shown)


app/Service/MDT/MDTBaseService.php
```diff
@@ -45,16 +45,28 @@ protected function encode(array $contents, MDTStringFormat $format = MDTStringFo
...
+                // detect() only picks MDT2 for a strict `!~MDT2~` prefix match, so a decode
+                // failure here means our own CBOR codec choked on a string that genuinely claimed
+                // to be an MDT export - always worth reporting.
+                logger()->error($throwable->getMessage(), $context);
+            } else {
+                // detect() defaults to Legacy for everything else, including garbage that doesn't
+                // even look like a legacy MDT export string (LegacyMDTCodec shells out to
+                // cli_weakauras_parser, whose stderr becomes the exception message) - most of these
+                // are just malformed user input, not application bugs (#3906).
+                logger()->warning($throwable->getMessage(), $context);
```
Wotuu: :robot: The discriminator here is "not MDT2 → warning", but the Legacy path is the *only* path that depends on an external binary — so this silences exactly the failures most worth paging on. `CliWeakaurasParserNotFoundException` is only rethrown when stderr contains the literal `cli_weakauras_parser: command not found`; everything else becomes a plain `LegacyMDTDecodeException` and now logs at `warning`:

- `sudo: no tty present…` / a sudoers change, permission denied, `/usr/bin/sudo` missing (Symfony `Process` RuntimeException)
- binary crash / non-zero exit with any other stderr
- `tempnam()` failure
- `cli_weakauras_parser output was not valid JSON: …` — that one is our own pipeline breaking, not user input

If the parser breaks in production, every legacy MDT import silently 400s and nothing reaches Sentry. Note `81e3d551f` (#2800, "Added additional check for whether cli_weakauras_parser is installed") — this failure mode has bitten before. And the compensating path is unreachable: `MDTImportController::import()`'s `report($ex)` (gated on `isValid()`) never fires for decode failures, because `decode()` swallows the Throwable and returns null before `MDTStringParseException` is raised. Also worth noting `AdminToolsMdtController::mdtviewasdungeonroute` (line ~135) round-trips *our own* export string back through `getDecoded()` — a failure there is by construction our bug, not garbage input, and it's now warning-only too.

The repo already has the exact "is this worth reporting" lever: `MDTStringFormat::isValid()`, whose docblock says it exists so "its only caller decides whether an unexpected decode failure elsewhere is worth a `report()` call", and which `MDTImportController:105` already uses for this decision. `isValid($string) ? error : warning` still fixes all three reported clusters — the PR's own evidence payload `C8PAAAAA…WMA` has no `!` prefix, so `isValid()` is false → warning — while keeping plausible-legacy and infra failures visible.

Tradeoff, honestly: `isValid()` would let a truncated-but-still-`!`-prefixed base64 paste (a common user error) alert again, so it's not a free win. Your call between adopting `isValid()` here and keeping the coarse check plus a narrower guard that keeps `error` for non-`LegacyMDTDecodeException` failures — but as it stands the infra blindness is a real regression in observability. (If you do adopt `isValid()`, the "its only caller" phrasing in `MDTStringFormat.php` goes stale.)

tests/Feature/App/Service/MDT/MDTImportStringServiceDecodeTest.php
```diff
@@ -47,4 +48,30 @@ public function getDecoded_givenInvalidString_returnsNull(): void
...
+
+        // Act
+        $decoded = app()->make(MDTImportStringServiceInterface::class)
+            ->setEncodedString('this_is_not_a_valid_mdt_string')
+            ->getDecoded();
+
+        // Assert
+        $this->assertNull($decoded);
+        $logSpy->shouldHaveReceived('warning');
+        $logSpy->shouldNotHaveReceived('error');
```
Wotuu: :robot: The whole semantic content of this change is a two-arm branch, but only the Legacy arm is tested — a future refactor that collapses both arms to `warning` (exactly the regression this test exists to prevent) still passes here. A companion test feeding `'!~MDT2~garbage'` and asserting `error` received / `warning` not received closes that gap, and as a bonus it needs no external binary since `MDT2Codec` decodes natively in PHP.

(Confirmed the existing assertion does fail pre-fix — pre-fix only ever calls `error` — so the claim in the PR body holds.)

Wotuu:

Wotuu:
:robot: Addressed both findings:
1. Replaced the `MDT2 vs Legacy` discriminator with `MDTStringFormat::isValid($string)` — this preserves `error`-level logging for anything that plausibly claims to be an MDT export (either codec's `appliesTo()` matches) but still fails to decode, including the CLI-crash/malformed-output cases you called out. Only input that fails `isValid()` entirely (which is exactly what all three reported Sentry payloads do — none start with `!`) now logs at `warning`.
2. Added `getDecoded_givenLegacyShapedStringThatFailsToDecode_logsAtError` — a `!`-prefixed-but-corrupt legacy string, proving the `error` arm still fires.

app/Service/MDT/MDTBaseService.php
Wotuu: :robot: Fixed as suggested — replaced the "which codec did detect() default to" discriminator with `MDTStringFormat::isValid($string)`. Anything that plausibly claims to be an MDT export (either codec's `appliesTo()` matches) but still fails to decode now logs at `error` regardless of which codec handled it, so the `cli_weakauras_parser` crash/sudo/tty/malformed-JSON infra failures you listed stay visible. Only input that fails `isValid()` entirely — which is what all three reported Sentry payloads do — logs at `warning`. Commit `96a676c8b`.

tests/Feature/App/Service/MDT/MDTImportStringServiceDecodeTest.php
Wotuu: :robot: Added `getDecoded_givenLegacyShapedStringThatFailsToDecode_logsAtError` — a `!`-prefixed-but-corrupt legacy string, proving the `error` arm still fires for the case a coarser two-arm collapse would silently break. Commit `96a676c8b`.

app/Service/MDT/MDTBaseService.php
```diff
@@ -50,11 +50,22 @@ protected function decode(string $string): ?array
...
-            ]);
+            // Truncated: the input is unvalidated user input of arbitrary size.
+            $context = ['string' => substr($string, 0, 2048)];
+
+            if (MDTStringFormat::isValid($string)) {
+                // The string at least plausibly claims to be an MDT export (one of the codecs'
+                // appliesTo() matched), so a decode failure here is worth reporting - it's either a
+                // genuine bug in our own codec, or (for the Legacy format) a real but corrupted/
+                // truncated export that a user has a right to expect us to import.
+                logger()->error($throwable->getMessage(), $context);
```
Wotuu: Structured logging should be used for these two lines. The way that you can solve this is that, in the implementation, you have the implementing class, except it has its own structured logging files. These structured logging files should extend a base class which mimics the log lines required in this base class. There should also be a base interface which is then extended by the implementing classes' structured logging implementations. The structured logging implementation will mirror the actual surface implementation with the base class and then one or more implementing classes.

Wotuu: :robot: Done in 03324bf4b — the two `logger()` calls now go through a structured logging hierarchy that mirrors the service hierarchy.

I mirrored the existing precedent at `app/Service/CombatLog/Splitters/Logging/CombatLogSplitterLogging.php` (base `CombatLogSplitter` + `CombatLogSplitterLoggingInterface`/`CombatLogSplitterLogging`, extended by `ZoneChangeSplitterLogging` and `ChallengeModeSplitterLogging`, with each subclass passing its own narrower logging interface up through `parent::__construct()`).

Added:
- `app/Service/MDT/Logging/MDTBaseServiceLoggingInterface.php` — extends `StructuredLoggingInterface`, declares the two decode events.
- `app/Service/MDT/Logging/MDTBaseServiceLogging.php` — `decodeFailed()` at `error` (Discord + Sentry), `decodeInvalidStringFailed()` at `warning`; the #3906 behaviour is unchanged, just moved.
- `app/Service/MDT/Logging/MDTExportStringServiceLogging.php` + `...Interface.php` — the second implementing pair, for the other `MDTBaseService` subclass.

Changed:
- `MDTImportStringServiceLoggingInterface`/`MDTImportStringServiceLogging` now extend the base pair instead of `StructuredLoggingInterface`/`StructuredLogging` directly.
- `MDTBaseService` takes an `MDTBaseServiceLoggingInterface`; `MDTImportStringService` and `MDTExportStringService` both call `parent::__construct($log)` with their own interface.
- No `LoggingServiceProvider` edit needed — it auto-binds every `*LoggingInterface` to its `*Logging` implementation by naming convention, and `app/Service/MDT/Logging/` is already at a covered depth.

Two things worth flagging:
1. **Sentry grouping changes for the better.** The error event's message is now the constant `MDTBaseServiceLogging::decodeFailed` instead of the varying `$throwable->getMessage()`, so it becomes one issue per log site with `exceptionClass`/`message` in the context — which is exactly what `FingerprintsStructuredErrorsHandler` fingerprints and tags on. The 2048-char truncation of the user input is kept.
2. **Test seam moved.** `Log::spy()` no longer observes these lines (StructuredLogging caches its logger at construction), so the two decode tests now bind a Mockery spy on `MDTImportStringServiceLoggingInterface` and assert the *event* (`decodeInvalidStringFailed` vs `decodeFailed`). Since that no longer asserts the level, I added `tests/Unit/App/Service/MDT/Logging/MDTBaseServiceLoggingTest.php` pinning the two events to `warning`/`error` — otherwise the actual #3906 guarantee could regress silently.

MDT feature suite (54 tests) + the new unit test pass, `composer run fix` and `composer run analyse` are clean.

PR #3933, opened 2026-08-09
#3915 Warn instead of 500 when an MDT rift-offset enemy isn't in the mapping version

### LINKED ISSUE #3915: New MDT-import flow 500s with 'No query results for model [Enemy]' on POST /new/mdtimport
:robot: `POST /new/mdtimport` raised an uncaught `ModelNotFoundException` for `App\Models\Enemy` — an `Enemy::findOrFail()` (or similar) call resolved no rows for whatever enemy id/reference the import string encoded.

- Occurrences: 1, Users impacted: 1
- First/last seen: 2026-08-08

```
No query results for model [App\Models\Enemy].
```
Single occurrence so far; flagging in case it recurs. Likely an MDT import string referencing an NPC/enemy clone that doesn't exist in the current mapping version, and the lookup should be tolerant of that (skip/warn) rather than 500ing. See `PHP-LARAVEL-S3` in Sentry for the full stack once more events accumulate — the `get_sentry_resource` full-event fetch didn't return a stacktrace for this one (defaulted to a bare message event).

Wotuu:
:robot: Closes #3915

`RiftOffsetImporter::parseRiftOffsets()` looked up an Awakened Obelisk skip's target enemy via `Enemy::where(...)->firstOrFail()`, inside a per-rift try/catch that only caught `ImportWarning` — so a `ModelNotFoundException` for an NPC id the current mapping version doesn't have an `Enemy` row for propagated uncaught out of `POST /new/mdtimport`.

Reproduced with real seeded data (no fixture hacking needed): The Underrot's current mapping version (id 158) has all 4 Awakened Obelisk map icons placed, but no matching `Enemy` rows for any of the 4 hardcoded obelisk NPC ids — exactly the "MDT import string referencing an NPC/enemy clone that doesn't exist in the current mapping version" the issue describes.

Fix: catch `ModelNotFoundException` alongside `ImportWarning` and record a warning for that one skip instead of letting it 500 the whole import — consistent with the surrounding "can't find X, skip it" pattern already used a few lines above for the obelisk map icons themselves.

Added a regression test (`parseRiftOffsets_givenRiftEnemyNotInCurrentMappingVersion_addsWarningInsteadOfThrowing`) using The Underrot's real current mapping version. Confirmed it reproduces the exact reported `ModelNotFoundException: No query results for model [App\Models\Enemy]` against the pre-fix code, and passes after.

### DIFF
--- a/app/Service/MDT/Import/RiftOffsetImporter.php
+++ b/app/Service/MDT/Import/RiftOffsetImporter.php
@@ -14,6 +14,7 @@ use App\Models\Polyline;
 use App\Service\MDT\Models\ImportStringRiftOffsets;
 use Exception;
 use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\ModelNotFoundException;
 use Illuminate\Support\Facades\Auth;
 
 class RiftOffsetImporter
@@ -125,6 +126,14 @@ class RiftOffsetImporter
                 $importStringRiftOffsets->getPaths()->push($pathAttributes);
             } catch (ImportWarning $warning) {
                 $importStringRiftOffsets->getWarnings()->add($warning);
+            } catch (ModelNotFoundException) {
+                // The import string references an NPC/enemy clone that isn't part of the current
+                // mapping version (#3915) - skip this one obelisk skip rather than 500ing the whole
+                // import.
+                $importStringRiftOffsets->getWarnings()->add(new ImportWarning(
+                    __('services.mdt.io.import_string.category.awakened_obelisks'),
+                    __('services.mdt.io.import_string.unable_to_find_awakened_obelisk_enemy'),
+                ));
             }
         }
 
--- a/app/Service/MDT/Import/RiftOffsetImporter.php
+++ b/app/Service/MDT/Import/RiftOffsetImporter.php
@@ -65,6 +65,17 @@ class RiftOffsetImporter
 
         // From the built array, construct our map icons / paths
         foreach ($rifts as $npcId => $mdtXy) {
+            // $npcId comes straight from the user-supplied import string and isn't constrained to
+            // the 4 hardcoded obelisk ids above - an arbitrary id would otherwise hit an undefined
+            // array key below and throw an uncaught Error (the same failure mode as #3915, just a
+            // different exception class than the catch further down handles).
+            if (!isset($npcIdToMapIconMapping[$npcId])) {
+                continue;
+            }
+
+            /** @var MapIcon $obeliskMapIcon */
+            $obeliskMapIcon = $npcIdToMapIconMapping[$npcId];
+
             try {
                 // Find out the floor where the NPC is standing on
                 /** @var Enemy $enemy */
@@ -73,9 +84,6 @@ class RiftOffsetImporter
                     ->whereNotNull('enemy_pack_id')
                     ->whereIn('floor_id', $floorIds)
                     ->firstOrFail();
-
-                /** @var MapIcon $obeliskMapIcon */
-                $obeliskMapIcon = $npcIdToMapIconMapping[$npcId];
 
                 if (isset($mdtXy['sublevel'])) {
                     throw new ImportWarning(
@@ -127,12 +135,16 @@ class RiftOffsetImporter
             } catch (ImportWarning $warning) {
                 $importStringRiftOffsets->getWarnings()->add($warning);
             } catch (ModelNotFoundException) {
-                // The import string references an NPC/enemy clone that isn't part of the current
-                // mapping version (#3915) - skip this one obelisk skip rather than 500ing the whole
-                // import.
+                // No enemy matching this NPC (packed, on the mapping version's floors) resolved -
+                // whether because the clone genuinely isn't in this mapping version, or (the common
+                // case seen while triaging #3915) it exists but with no enemy_pack_id on the current
+                // mapping version. Skip this one obelisk skip rather than 500ing the whole import.
                 $importStringRiftOffsets->getWarnings()->add(new ImportWarning(
                     __('services.mdt.io.import_string.category.awakened_obelisks'),
-                    __('services.mdt.io.import_string.unable_to_find_awakened_obelisk_enemy'),
+                    __(
+                        'services.mdt.io.import_string.unable_to_find_awakened_obelisk_enemy',
+                        ['name' => __($obeliskMapIcon->mapIconType->name)],
+                    ),
                 ));
             }
         }
--- a/lang/en_US/services.php
+++ b/lang/en_US/services.php
@@ -49,6 +49,7 @@ return [
                 'unable_to_find_enemies_pull_skipped_details'          => 'This may indicate MDT recently had an update that is not integrated in Keystone.guru yet.',
                 'unable_to_find_awakened_obelisks'                     => 'Cannot find Awakened Obelisks for your dungeon/week combination. Your Awakened Obelisk skips will not be imported.',
                 'unable_to_find_awakened_obelisk_different_floor'      => 'Unable to import Awakened Obelisk :name, it is on a different floor than the Obelisk itself. Keystone.guru does not support this at this time.',
+                'unable_to_find_awakened_obelisk_enemy'                => 'Unable to find the enemy for an Awakened Obelisk skip in your dungeon/week combination. This particular skip will not be imported.',
                 'unable_to_decode_mdt_import_string'                   => 'Unable to decode MDT import string',
                 'unable_to_validate_mdt_import_string'                 => 'Unable to validate MDT import string',
             ],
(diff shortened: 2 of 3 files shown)


app/Service/MDT/Import/RiftOffsetImporter.php
```diff
@@ -125,6 +126,14 @@ public function parseRiftOffsets(ImportStringRiftOffsets $importStringRiftOffset
                 $importStringRiftOffsets->getPaths()->push($pathAttributes);
             } catch (ImportWarning $warning) {
                 $importStringRiftOffsets->getWarnings()->add($warning);
+            } catch (ModelNotFoundException) {
+                // The import string references an NPC/enemy clone that isn't part of the current
```
Wotuu: :robot: **The premise of this fix looks wrong — the enemy *does* exist; it's `whereNotNull('enemy_pack_id')` (line 72) that excludes it.**

Verified against the seeded DB on this checkout:

```
Enemy::where('npc_id', 161124)->where('mapping_version_id', 158)->get(['id','floor_id','enemy_pack_id'])
=> [{"id":27269,"floor_id":51,"enemy_pack_id":null}]
```

(and `theunderrot`'s `getCurrentMappingVersion()->id === 158`). So the row for the Brutal obelisk NPC *is* part of the current mapping version — it just has `enemy_pack_id = NULL` (`seasonal_type: awakened`). The PR body's "genuinely zero matching `Enemy` rows", this comment ("isn't part of the current mapping version"), the test docblock and the new translation string all repeat that inaccurate framing.

And it isn't an Underrot anomaly — scanning `database/seeders/dungeondata/bfa/*/*/enemies.json` for the 4 obelisk npc ids, *packed* obelisk rows only exist on the older mapping versions of each dungeon:
- theunderrot: packed only at mv21/106 (floor 2); mv121→158 have unpacked rows only
- freehold: packed only at mv15/105; mv120/125/135/141/149/157/186 unpacked only

i.e. for essentially every *current* mapping version the `whereNotNull('enemy_pack_id')` filter matches nothing. That means this PR doesn't just stop a 500 — it turns "Awakened obelisk skips 500 the import" into "Awakened obelisk skips silently never import, for every dungeon", which is a functional behaviour change worth making deliberately rather than as a side effect.

Question for the maintainer: is the pack filter load-bearing here at all? The loop only consumes `$enemy->floor_id` and `$enemy->floor` — pack membership is irrelevant to what's actually needed, so `first()` with a fallback to an unpacked row would make these skips import correctly instead of being dropped. (A legitimate answer is "riftOffsets are legacy — `MDTExportStringService` literally comments `// Legacy - we don't do anything with it` — so warn-and-skip is fine"; but that's a decision, and the PR as written hides it.) Keeping this `catch` as defence-in-depth is right either way.

app/Service/MDT/Import/RiftOffsetImporter.php
```diff
@@ -125,6 +126,14 @@ public function parseRiftOffsets(ImportStringRiftOffsets $importStringRiftOffset
...
+                // mapping version (#3915) - skip this one obelisk skip rather than 500ing the whole
```
Wotuu: :robot: Same bug class, one line above the guard you just added (line 77): `$obeliskMapIcon = $npcIdToMapIconMapping[$npcId];`.

`$rifts` comes straight from the user-supplied MDT import string, so `$npcId` is arbitrary — it is not constrained to the 4 hardcoded obelisk ids. If a crafted string keys `riftOffsets[week]` with any npc id that happens to have a packed `Enemy` row on one of the dungeon's floors, the `firstOrFail()` succeeds, the array access hits an undefined key (PHP 8 warning → `null`), and `$obeliskMapIcon->mapIconType` on line 84 throws an uncaught `Error` — the same uncaught-500-out-of-`POST /new/mdtimport` failure mode #3915 is about, just a different exception class that your new `catch (ModelNotFoundException)` won't catch.

Cheap to close while you're here, e.g. before the lookup:

```php
if (!isset($npcIdToMapIconMapping[$npcId])) {
    continue;
}
```

lang/en_US/services.php
```diff
@@ -49,6 +49,7 @@
                 'unable_to_find_enemies_pull_skipped_details'          => 'This may indicate MDT recently had an update that is not integrated in Keystone.guru yet.',
                 'unable_to_find_awakened_obelisks'                     => 'Cannot find Awakened Obelisks for your dungeon/week combination. Your Awakened Obelisk skips will not be imported.',
                 'unable_to_find_awakened_obelisk_different_floor'      => 'Unable to import Awakened Obelisk :name, it is on a different floor than the Obelisk itself. Keystone.guru does not support this at this time.',
+                'unable_to_find_awakened_obelisk_enemy'                => 'Unable to find the enemy for an Awakened Obelisk skip in your dungeon/week combination. This particular skip will not be imported.',
```
Wotuu: :robot: This message can name the obelisk like its sibling key does. `unable_to_find_awakened_obelisk_different_floor` takes `:name` and fills it with `__($obeliskMapIcon->mapIconType->name)`; the same value is available in the new catch via `$npcIdToMapIconMapping[$npcId]` (guarding for the unknown-npc-id case — see my other comment), so the user can be told *which* skip was dropped rather than "an Awakened Obelisk skip".

Also, "Unable to find the enemy ... in your dungeon/week combination" isn't what actually happened: the enemy row exists on the current mapping version, it just has no `enemy_pack_id` (see my comment on `RiftOffsetImporter.php`). Worth rewording once the pack-filter question above is settled.

tests/Feature/App/Service/MDT/Import/RiftOffsetImporterTest.php
```diff
@@ -0,0 +1,60 @@
...
+     * matching Enemy row for one of those NPC ids (real seeded case: The Underrot's current mapping
+     * version, id 158), the underlying Enemy::where(...)->firstOrFail() must not surface an uncaught
+     * ModelNotFoundException - it should be recorded as a warning and that one skip omitted instead.
+     */
+    #[Test]
+    public function parseRiftOffsets_givenRiftEnemyNotInCurrentMappingVersion_addsWarningInsteadOfThrowing(): void
+    {
+        // Arrange
+        $dungeon        = Dungeon::where('key', 'theunderrot')->firstOrFail();
+        $mappingVersion = $dungeon->getCurrentMappingVersion();
```
Wotuu, line 36: :robot: This test asserts that the *current* seeded data produces a warning, which makes it a test of the data as much as of the code. Two consequences worth handling:

1. If the mapping data is ever fixed (obelisk enemies get an `enemy_pack_id` again), or if the `whereNotNull('enemy_pack_id')` filter is relaxed — which I'd argue it should be, see my comment on `RiftOffsetImporter.php` — this test fails with "Failed asserting that actual size 0 matches expected size 1". That reads as a regression when it is in fact the desired outcome, and the next person has no hint why.
2. It also depends on `getCurrentMappingVersion()` still having all 4 obelisk `MapIcon`s placed; if a future Underrot mapping version drops those, `parseRiftOffsets()` throws `ImportWarning` out of the method and this test errors for a reason unrelated to what it guards.

Cheap fix: assert the premise explicitly in Arrange with a message, e.g.

```php
self::assertSame(0, Enemy::where('npc_id', 161124)
    ->where('mapping_version_id', $mappingVersion->id)
    ->whereNotNull('enemy_pack_id')
    ->count(), 'Test premise no longer holds: this NPC now has a packed Enemy row on the current mapping version');
```

so a data change fails loudly with the reason rather than as a confusing count mismatch.

Wotuu:
:robot: Cold review (opus): 4 findings posted. Verified separately and OK: `Illuminate\Database\Eloquent\ModelNotFoundException` is the right class (the Sentry text `No query results for model [App\Models\Enemy]` is Eloquent MNFE specifically); no other unguarded `Enemy::...OrFail()` is reachable from the mdtimport chain (the only other one is `Console/Commands/Mapping/Restore.php`); no sibling MDT importer has the same narrow-catch-around-`firstOrFail()` pattern; translation key placement/alignment matches its neighbours.

app/Service/MDT/Import/RiftOffsetImporter.php
Wotuu: :robot: Good catch, and confirmed — you're right that the enemy row exists, just unpacked. I've corrected the framing in the code comment, warning message, and test docblock to say that accurately rather than claiming the enemy is absent from the mapping version.

On whether `whereNotNull('enemy_pack_id')` should be relaxed: leaving that as-is rather than changing it in this PR. This issue was specifically about the uncaught `ModelNotFoundException`, and whether packed-only is the correct semantics for Awakened Obelisk skip placement is a real question but a different one — it changes import *behavior* (obelisk skips would start actually importing rather than being warned-and-skipped) rather than just fixing the crash, and given `MDTExportStringService`'s own "Legacy" comment on this feature, that's a call I don't have enough context to make unilaterally. Filing it as a follow-up rather than bundling a behavior change into a crash fix.

Wotuu: :robot: Filed the follow-up: #3935

app/Service/MDT/Import/RiftOffsetImporter.php
Wotuu: :robot: Fixed as suggested — moved the map-icon lookup before the try block with an `isset()` guard, so an unrecognized rift npc id is silently skipped instead of hitting the undefined array key. Added a regression test.

lang/en_US/services.php
Wotuu: :robot: Fixed — the warning now interpolates `:name` from the resolved `$obeliskMapIcon`, and reworded to reflect the actual cause (per the other thread: enemy exists but couldn't be resolved, not necessarily absent).

tests/Feature/App/Service/MDT/Import/RiftOffsetImporterTest.php
Wotuu: :robot: Fixed as suggested — added the explicit precondition assertion with a message.

PR #3934, opened 2026-08-09
#3918 Stop paging Sentry when a run's combat-log segments aren't uploaded yet

### LINKED ISSUE #3918: Combat-log segment downloads failing on staging (segments not available / invalid Raider.IO API response)
:robot: Two structured-logging clusters on staging, both from the same combat-log-segment processing pipeline and correlated by shared `trace_id`s across events (e.g. run 42665055 hits both in the same job run):

**PHP-LARAVEL-RQ** — `ProcessCombatLogSegmentsLogging::handleSegmentsNotAvailable`, 69 occurrences, escalating substatus, 2026-08-05 to 2026-08-11 (still recurring)
**PHP-LARAVEL-RP** — `RaiderIOApiServiceLogging::getCombatLogSegmentsForRunInvalidResponse`, 150 occurrences, 2026-08-05 to 2026-08-17 (still recurring)

Representative event context (run 42665055, combat log version 22012000007):
```json
{"statusCode":404,"error":"Not Found","message":"No combat log segments found for this run"}
```
Requested URL path: `/api/v1/live-tracking/mythic-plus/combatlog-segments` (query params omitted per this triage's redaction policy — season and run id are already shown above).

The Raider.IO live-tracking API is returning 404 for these runs' combat-log segments, and `ProcessCombatLogSegments` logs it as an error-level event rather than treating "segments not yet available" as an expected, retryable state. Worth checking whether this should be a lower log level with a backoff/retry (segments may simply not have been uploaded yet) rather than an immediate error.

Two separate download-failure events for other runs are also part of this pipeline — see `PHP-LARAVEL-RM`/`PHP-LARAVEL-RX` (filed separately, lower volume).

## Update 2026-08-17: still recurring, both up substantially

RQ: 43 → 69 occurrences. RP: 43 → 150 occurrences. Both still actively recurring (RP as recently as 6 hours before this update). No sign this has resolved on its own — worth prioritizing over the newer, lower-volume issues filed in the same pass.

Wotuu:
:robot: Closes #3918

A run's combat-log segments simply not being available on Raider.IO yet (a 404 with body `{"statusCode":404,"error":"Not Found","message":"No combat log segments found for this run"}`) is an expected, recurring state — segments may just not have been uploaded yet — not a broken API integration. But `RaiderIOApiService::getCombatLogSegmentsForRun()` logged it via the same error-level `getCombatLogSegmentsForRunInvalidResponse` call used for a genuinely malformed/unexpected response, and `ProcessCombatLogSegments::handle()` logged its own echo of the same condition (`handleSegmentsNotAvailable`) at error too. Since the `sentry` log channel alerts on error-level logs (`config/logging.php`), both fired — as two correlated Sentry issues — on every occurrence (43 times in this triage pass).

Fix: distinguish the specific `{"statusCode": 404, ...}` response shape and log it at `info` via a new `getCombatLogSegmentsForRunNotYetAvailable` method, keeping `getCombatLogSegmentsForRunInvalidResponse` at `error` for anything that doesn't match either the success or the not-found shape (a genuine integration break still pages). Downgraded the job's own `handleSegmentsNotAvailable` echo to `info` too, since the service now logs the specific reason distinctly.

Added a regression test (`getCombatLogSegmentsForRun_givenSegmentsNotYetAvailable_logsDistinctlyFromInvalidResponse`) asserting the new method is called (and the error-level one isn't) for the 404 shape, and strengthened the existing invalid-response test to assert the reverse. `composer run fix`/`analyse` clean; existing `ProcessCombatLogSegmentsTest` suite (11 tests) still passes unaffected.

### DIFF
--- a/app/Jobs/Logging/ProcessCombatLogSegmentsLogging.php
+++ b/app/Jobs/Logging/ProcessCombatLogSegmentsLogging.php
@@ -13,10 +13,11 @@ class ProcessCombatLogSegmentsLogging extends StructuredLogging implements Proce
 
     public function handleSegmentsNotAvailable(int $runId): void
     {
-        // A run's segments not being available yet is an expected, recurring state (#3918) - the
-        // underlying RaiderIOApiServiceLogging call already logs the specific reason at the
-        // appropriate level (error for a genuinely malformed response, info for "not yet uploaded"),
-        // so this is just an echo and shouldn't page Sentry on its own.
+        // Fires for two distinct callers (#3918), both an expected, recurring state rather than a
+        // broken integration: getCombatLogSegmentsForRun() returned null (in which case
+        // RaiderIOApiServiceLogging already logged the specific reason, at error or info depending
+        // on cause), or it returned a well-formed response whose segments array is simply empty (no
+        // upstream log at all for that case - this is the only signal). Neither should page Sentry.
         $this->info(__METHOD__, get_defined_vars());
     }
 
--- a/app/Jobs/Logging/ProcessCombatLogSegmentsLogging.php
+++ b/app/Jobs/Logging/ProcessCombatLogSegmentsLogging.php
@@ -13,7 +13,11 @@ class ProcessCombatLogSegmentsLogging extends StructuredLogging implements Proce
 
     public function handleSegmentsNotAvailable(int $runId): void
     {
-        $this->error(__METHOD__, get_defined_vars());
+        // A run's segments not being available yet is an expected, recurring state (#3918) - the
+        // underlying RaiderIOApiServiceLogging call already logs the specific reason at the
+        // appropriate level (error for a genuinely malformed response, info for "not yet uploaded"),
+        // so this is just an echo and shouldn't page Sentry on its own.
+        $this->info(__METHOD__, get_defined_vars());
     }
 
     public function handleDownloadingSegment(int $runId, int $segmentId, string $downloadUrl, string $tempPath): void
--- a/app/Service/RaiderIO/RaiderIOApiService.php
+++ b/app/Service/RaiderIO/RaiderIOApiService.php
@@ -178,7 +178,14 @@ class RaiderIOApiService implements RaiderIOApiServiceInterface
             $json     = json_decode($response, true);
 
             if (!is_array($json) || !isset($json['sourceUserId'], $json['segments']) || !is_array($json['segments'])) {
-                $this->log->getCombatLogSegmentsForRunInvalidResponse($runId, $url, $response);
+                // A 404 with this shape means the run's segments simply haven't been uploaded to
+                // Raider.IO yet (#3918) - an expected, recurring state to log distinctly from a
+                // genuinely malformed/unexpected response, which stays error-level below.
+                if (($json['statusCode'] ?? null) === 404) {
+                    $this->log->getCombatLogSegmentsForRunNotYetAvailable($runId, $url);
+                } else {
+                    $this->log->getCombatLogSegmentsForRunInvalidResponse($runId, $url, $response);
+                }
 
                 return null;
             }
--- a/app/Service/RaiderIO/RaiderIOApiService.php
+++ b/app/Service/RaiderIO/RaiderIOApiService.php
@@ -178,11 +178,16 @@ class RaiderIOApiService implements RaiderIOApiServiceInterface
             $json     = json_decode($response, true);
 
             if (!is_array($json) || !isset($json['sourceUserId'], $json['segments']) || !is_array($json['segments'])) {
-                // A 404 with this shape means the run's segments simply haven't been uploaded to
-                // Raider.IO yet (#3918) - an expected, recurring state to log distinctly from a
-                // genuinely malformed/unexpected response, which stays error-level below.
-                if (($json['statusCode'] ?? null) === 404) {
-                    $this->log->getCombatLogSegmentsForRunNotYetAvailable($runId, $url);
+                // A 404 with this specific message means the run's segments simply haven't been
+                // uploaded to Raider.IO yet (#3918) - an expected, recurring state to log distinctly
+                // from a genuinely malformed/unexpected response, which stays error-level below.
+                // Matching on 'message' too (not just 'statusCode') matters: the upstream API is
+                // hapi-style, whose default route-not-found body is also {"statusCode":404,"error":
+                // "Not Found","message":"Not Found"} - identical statusCode, but a real integration
+                // break (wrong path, unrecognized season) that must still page.
+                if (($json['statusCode'] ?? null) === 404
+                    && str_contains((string)($json['message'] ?? ''), 'combat log segments')) {
+                    $this->log->getCombatLogSegmentsForRunNotYetAvailable($runId, $url, $response);
                 } else {
                     $this->log->getCombatLogSegmentsForRunInvalidResponse($runId, $url, $response);
                 }
--- a/app/Service/RaiderIO/RaiderIOApiService.php
+++ b/app/Service/RaiderIO/RaiderIOApiService.php
@@ -19,6 +19,7 @@ use App\Service\Season\SeasonAffixGroupServiceInterface;
 use App\Service\Season\SeasonServiceInterface;
 use App\Service\Traits\Curl;
 use Str;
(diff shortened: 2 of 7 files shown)


app/Service/RaiderIO/RaiderIOApiService.php
```diff
@@ -178,7 +178,14 @@ public function getCombatLogSegmentsForRun(Season $season, int $runId): ?CombatL
             $json     = json_decode($response, true);
 
             if (!is_array($json) || !isset($json['sourceUserId'], $json['segments']) || !is_array($json['segments'])) {
-                $this->log->getCombatLogSegmentsForRunInvalidResponse($runId, $url, $response);
+                // A 404 with this shape means the run's segments simply haven't been uploaded to
+                // Raider.IO yet (#3918) - an expected, recurring state to log distinctly from a
+                // genuinely malformed/unexpected response, which stays error-level below.
+                if (($json['statusCode'] ?? null) === 404) {
+                    $this->log->getCombatLogSegmentsForRunNotYetAvailable($runId, $url);
```
Wotuu: :robot: The not-yet-available branch keys only on `statusCode` and throws the body away — two related concerns:

1. **A 404 from a different cause has the identical shape.** The upstream API is hapi-style, and its default route-not-found body is `{"statusCode":404,"error":"Not Found","message":"Not Found"}` — indistinguishable here from "no segments for this run". So an endpoint path change, or a season string the API stops recognizing, would now be silently `info`-logged forever instead of paging, which is exactly the "genuine integration break still pages" property the PR body promises. Consider also matching on the `message` (or at least on the absence of `sourceUserId` *plus* a `message` that mentions segments).

2. **`getCombatLogSegmentsForRunNotYetAvailable(int $runId, string $url)` omits `$response`.** The sibling `getCombatLogSegmentsForRunInvalidResponse` takes it for precisely this reason. If the branch ever does misfire per (1), the log has nothing to diagnose from. Passing `$response` through is cheap and makes the loose discriminator survivable either way.

(For the record, the success path can't misfire into this branch — `isset($json['sourceUserId'], $json['segments'])` gates it — so this is only about *other* 404s.)

tests/Feature/App/Service/RaiderIO/RaiderIOApiServiceTest.php
```diff
@@ -92,6 +93,32 @@ public function getCombatLogSegmentsForRun_givenInvalidResponse_returnsNull(): v
...
+     * returns a 404 with this specific shape) is an expected, recurring state - it must be logged
+     * distinctly from a genuinely malformed response instead of paging Sentry as an error.
+     *
+     * @throws Exception
+     */
+    #[Test]
+    public function getCombatLogSegmentsForRun_givenSegmentsNotYetAvailable_logsDistinctlyFromInvalidResponse(): void
+    {
+        // Arrange
+        $this->log->expects($this->once())->method('getCombatLogSegmentsForRunNotYetAvailable');
```
Wotuu, line 108: :robot: The one behaviour this PR exists to change — the log *level* — isn't asserted anywhere.

This test mocks `RaiderIOApiServiceLoggingInterface`, so the level is unobservable by construction; it can only assert which method was dispatched. And the `ProcessCombatLogSegmentsLogging::handleSegmentsNotAvailable` `error` -> `info` downgrade has no coverage at all — revert that line to `$this->error(...)` and the whole suite still passes, while the Sentry noise the issue is about comes straight back.

A level assertion has to go against the *concrete* logging classes rather than the interface; `tests/Unit/App/Logging/StructuredLoggingTest.php` already shows the mechanism (`LoggingFixtures::createLogManager($this)` + expecting `log()` and inspecting the `$level` argument). Given the repo's test-enforcement rule, a small unit test pinning both methods below `error` seems warranted.

app/Jobs/Logging/ProcessCombatLogSegmentsLogging.php
```diff
@@ -13,7 +13,11 @@ public function handleStart(int $runId, int $combatLogVersion): void
 
     public function handleSegmentsNotAvailable(int $runId): void
     {
-        $this->error(__METHOD__, get_defined_vars());
+        // A run's segments not being available yet is an expected, recurring state (#3918) - the
+        // underlying RaiderIOApiServiceLogging call already logs the specific reason at the
+        // appropriate level (error for a genuinely malformed response, info for "not yet uploaded"),
+        // so this is just an echo and shouldn't page Sentry on its own.
+        $this->info(__METHOD__, get_defined_vars());
```
Wotuu, line 21: :robot: This comment's justification is factually wrong for one of the two callers.

`ProcessCombatLogSegments::handle()` calls `handleSegmentsNotAvailable()` for `$segmentsResponse === null` **or** `empty($segmentsResponse->segments)`. In the second case the API returned a well-formed, successful response that just lists zero segments — the *success* path of `getCombatLogSegmentsForRun()`, which logs only start/end. So `RaiderIOApiServiceLogging` logs no reason at all there, and this `info` line is the only signal for that case, not an "echo".

`info` is defensible for both branches (zero segments is arguably the same "not uploaded yet" state), so no level change is needed — but the durable comment will mislead the next reader into thinking a distinct upstream log always exists. Worth rewording to cover both branches.

Wotuu:

app/Service/RaiderIO/RaiderIOApiService.php
Wotuu: :robot: Fixed both — now also matches on `message` containing "combat log segments" so a same-shape-but-different-cause 404 (e.g. a hapi default route-not-found) still pages, and $response is now passed through to the new log method.

tests/Feature/App/Service/RaiderIO/RaiderIOApiServiceTest.php
Wotuu: :robot: Fixed — added unit tests asserting the actual emitted level against the concrete logging classes (via a mocked LogManager bound into the container), for both the new method and the downgraded handleSegmentsNotAvailable. Confirmed the latter fails against the pre-fix error level.

app/Jobs/Logging/ProcessCombatLogSegmentsLogging.php
Wotuu: :robot: Fixed — reworded to cover both callers accurately.

app/Service/RaiderIO/RaiderIOApiService.php
```diff
@@ -178,7 +178,19 @@ public function getCombatLogSegmentsForRun(Season $season, int $runId): ?CombatL
...
             if (!is_array($json) || !isset($json['sourceUserId'], $json['segments']) || !is_array($json['segments'])) {
-                $this->log->getCombatLogSegmentsForRunInvalidResponse($runId, $url, $response);
+                // A 404 with this specific message means the run's segments simply haven't been
+                // uploaded to Raider.IO yet (#3918) - an expected, recurring state to log distinctly
+                // from a genuinely malformed/unexpected response, which stays error-level below.
+                // Matching on 'message' too (not just 'statusCode') matters: the upstream API is
+                // hapi-style, whose default route-not-found body is also {"statusCode":404,"error":
+                // "Not Found","message":"Not Found"} - identical statusCode, but a real integration
+                // break (wrong path, unrecognized season) that must still page.
+                if (($json['statusCode'] ?? null) === 404
```
Wotuu: Project convention is to use the teapot status code library. Don't hardcode to 404, but use the not found static variable instead.

Wotuu: :robot: Fixed in 76b56c49b — now `use Teapot\StatusCode\Http;` + `Http::NOT_FOUND`, matching the existing style in the Ajax controllers.

That 404 was the only hardcoded HTTP status integer this PR introduced in app code; I also swapped the two `statusCode` values in the test fixtures (`RaiderIOApiServiceTest`) over to `Http::NOT_FOUND` for consistency, and added a test for the hapi generic route-not-found body (same status code, generic `message`) to pin that it still logs at error level. The literal `{"statusCode":404}` string in `RaiderIOApiServiceLoggingTest` is left as-is — it is an opaque raw-response payload passed straight to the logger, not a status comparison.

`composer run fix` and `composer run analyse` are clean; affected tests pass (7 tests, 21 assertions).

Issue #3198, opened 2026-05-22
Fix BroadcastException: Pusher payload too large when broadcasting KillZone events


Wotuu:
## Problem

`KillZoneChangedEvent` and `KillZoneDeletedEvent` were including the full set of kill zone path segments (all pathfinding waypoints for every pull in the route) inside the Pusher broadcast payload. For large routes this exceeds Pusher's 10KB message limit, crashing the queue worker with:

```
Illuminate\Broadcasting\BroadcastException: Pusher error: Payload too large..
```

Sentry issue: PHP-LARAVEL-6G (660 occurrences since January 2026)

## Root Cause

`KillZoneChangedEvent::broadcastWith()` called `KillZonePathService::calculateForRoute()` and embedded the result (all path segments for the entire route) in the broadcast. Payload size scales linearly with route complexity — a 20+ kill zone route with pathfinding waypoints easily exceeds 10KB.

## Fix

- Removed `killzone_paths` from `KillZoneChangedEvent` and `KillZoneDeletedEvent` broadcast payloads entirely.
- Added a new `GET /ajax/{dungeonRoute}/killzone/paths` endpoint.
- Collaborative editing clients now call `fetchAndRefresh(routeKey)` via AJAX after receiving a `killzone-changed` or `killzone-deleted` Echo event, instead of reading paths from the (now tiny) broadcast payload.
- The saving user's own client is unaffected — it still receives paths synchronously in the HTTP response from the existing `store()` endpoint.

Wotuu:
TODO: the test for this story is very weak. Make it better.

PR #3936, opened 2026-08-09
#3909 Drop redundant vertices_json from the Path, Arrow and Brushline broadcast payloads

### LINKED ISSUE #3909: Pusher broadcast fails with 'Payload too large' for some killzone path updates
:robot: `queue:work` occasionally fails to broadcast a `BroadcastEvent` via Pusher because the payload exceeds Pusher's size limit.

- Occurrences: 22, ongoing
- First/last seen: 2026-06-25 to 2026-07-27

```
Illuminate\Broadcasting\BroadcastException: Pusher error: Payload too large..
```
One representative event's `structured:debuginfocontextloggerlogging::handle` context shows the originating request was `PUT /ajax/{route}/path/{id}` — a killzone-path update — suggesting large route/path payloads (many nodes/edges) can exceed Pusher's message size cap when broadcast to collaborators. Worth checking whether the broadcast payload can be trimmed (e.g. broadcasting an id + "refetch" signal instead of the full path data) rather than the raw path graph.

## Update 2026-08-25: reopened — recurred after the fix shipped

The fix (`0d2b7313e`, "#3909 Drop redundant vertices_json from the Path, Arrow and Brushline broadcast payloads", PR #3936) landed 2026-08-10 and first shipped in **v15.10.2**. `PHP-LARAVEL-MM` has kept firing since: 22 → **45** lifetime occurrences, most recent event 2026-08-24 (~23 hours before this update), on releases well after v15.10.2.

So dropping `vertices_json` from those three payloads reduced the size but did not bring every broadcast under Pusher's cap — either some other payload on the same channel is oversized, or the path/killzone payloads are still large enough without `vertices_json`. Needs a fresh look at what the actual over-limit payload contains now, rather than a repeat of the same trim.

Wotuu:
:robot: Closes #3909

Pusher/Reverb rejects broadcasts over ~10KB, and large killzone paths were tripping that cap ("Payload too large", Sentry `PHP-LARAVEL-MM`, 22 occurrences).

## Root cause

`PathChangedEvent::broadcastWith()` sent the vertex data **three times**:
- the raw `polyline.vertices_json` (via `ContextModelEvent`'s full model serialization, included under `model`)
- two computed representations (`split_floors` and `facade`) under `model_data`

The receiving client (`PathChangedHandler::onReceive()`) always overwrites `e.model.polyline.vertices_json` with the `model_data` coordinates before use (see `ModelChangedHandler::_getCorrectLatLngFromEvent()`), so the raw copy is dead weight on every receiving client — it's never read before being replaced.

## Fix

`PathChangedEvent::broadcastWith()` now strips `polyline.vertices_json` from the serialized `model` before merging in `model_data`, cutting the payload by roughly a third. The rest of the polyline (color, weight, etc.) is untouched.

`ArrowChangedEvent` and `BrushlineChangedEvent` got the identical trim in `05d8b2f03` — they have the same shape (`broadcastWith()` unconditionally adds `model_data`, and `arrow/changed.js` / `brushline/changed.js` unconditionally overwrite `e.model.polyline.vertices_json` from it), so the raw copy is equally redundant there.

## Audit of every polyline-carrying broadcast event

The discriminator is whether a payload carries the raw `polyline.vertices_json` **and** a computed `model_data.coordinates` derived from the same vertices — only then is one of them redundant. Checked every `polyline()` relation in `app/Models`, every `HasVertices`/`SavesPolylines` user, and every `ModelChangedEvent`/`ModelDeletedEvent` subclass.

| Event | Carries polyline? | Also sends `model_data`? | Verdict |
|---|---|---|---|
| `PathChangedEvent` | yes | yes | **Trimmed** |
| `BrushlineChangedEvent` | yes | yes | **Trimmed** |
| `ArrowChangedEvent` | yes | yes | **Trimmed** |
| `EnemyPatrolChangedEvent` | yes (`$with = ['polyline']`) | **no** — no `broadcastWith()` override | Doesn't apply |
| `EnemyPackChangedEvent` | `vertices_json` via `HasVertices` | no | Doesn't apply |
| `MountableAreaChangedEvent` | `vertices_json` via `HasVertices` | no | Doesn't apply |
| `FloorUnionAreaChangedEvent` | `vertices_json` via `HasVertices` | no | Doesn't apply |
| `KillZoneChangedEvent` | no polyline; used to send `killzone_paths` | n/a | Already fixed in `902f16d8e` (#3198) |
| all `*DeletedEvent` | no | n/a | Nothing to trim |

- **EnemyPatrol** — `polyline.vertices_json` is the *only* representation in its payload, and `echohandler.js` registers no `EnemyPatrolChangedHandler`, so nothing client-side would recompute it. Trimming would delete data with no substitute.
- **EnemyPack / MountableArea / FloorUnionArea** — polygons carrying `vertices_json` directly on the model via `HasVertices`; their changed events add no `model_data`, so only one copy is sent.
- **KillZone** — the bloat there was `killzone_paths`, removed from the broadcast in `902f16d8e` (#3198) with the client re-fetching over a dedicated route. Verified still absent.

**Honesty note:** this reduces the payload, it doesn't hard-bound it. A polyline with enough vertices can still exceed the cap even with the one remaining representation; a hard cap (id + refetch signal, as the Sentry issue also suggested) would be a larger, separate change.

## Verification

- `AjaxPathControllerTest::store_givenNewValidPath_broadcastsPayloadWithoutRedundantVerticesJson`, plus the matching `AjaxArrowControllerTest` / `AjaxBrushlineControllerTest` tests, use a scoped `Event::fake([...Event::class])` around the real controller POST and assert on the payload of the event the controller actually dispatched — so deleting the `broadcast(...)` call fails them. All three verified non-vacuous by reverting the event changes and watching them go red.
- Scoped rather than bare `Event::fake()` deliberately: these tests depend on Eloquent model events staying live.
- `composer run analyse` and `composer run fix` clean; all 9 CI checks green.

### DIFF
--- a/app/Events/Models/Path/PathChangedEvent.php
+++ b/app/Events/Models/Path/PathChangedEvent.php
@@ -44,8 +44,18 @@ class PathChangedEvent extends ModelChangedEvent
         /** @var Path $model */
         $model = $this->model;
 
+        $broadcast = parent::broadcastWith();
+
+        // The receiving client always overwrites model.polyline.vertices_json with the
+        // model_data coordinates below (see ModelChangedHandler::_getCorrectLatLngFromEvent()
+        // and Path's changed.js), so broadcasting the raw vertices here too is dead weight that
+        // can push large paths over Reverb's message size cap (#3909).
+        $modelArray = $model->toArray();
+        unset($modelArray['polyline']['vertices_json']);
+        $broadcast['model'] = $modelArray;
+
         return array_merge(
-            parent::broadcastWith(),
+            $broadcast,
             [
                 'model_data' => $model->polyline->getCoordinatesData(
                     $this->coordinatesService,
--- a/tests/Feature/Controller/Ajax/AjaxPathControllerTest.php
+++ b/tests/Feature/Controller/Ajax/AjaxPathControllerTest.php
@@ -2,9 +2,12 @@
 
 namespace Tests\Feature\Controller\Ajax;
 
+use App\Events\Models\Path\PathChangedEvent;
 use App\Models\Floor\Floor;
 use App\Models\Path;
 use App\Models\Polyline;
+use App\Models\User;
+use App\Service\Coordinates\CoordinatesServiceInterface;
 use Illuminate\Support\Facades\Event;
 use PHPUnit\Framework\Attributes\Group;
 use PHPUnit\Framework\Attributes\Test;
@@ -42,6 +45,47 @@ final class AjaxPathControllerTest extends DungeonRouteTestBase
         $this->assertEquals($polyline['color_animated'], $responseArr['polyline']['color_animated']);
         $this->assertEquals($polyline['weight'], $responseArr['polyline']['weight']);
         $this->assertEquals($polyline['vertices_json'], $responseArr['polyline']['vertices_json']);
+    }
+
+    #[Test]
+    #[Group('Controller')]
+    public function store_givenNewValidPath_broadcastsPayloadWithoutRedundantVerticesJson(): void
+    {
+        // Arrange
+        /** @var Floor $randomFloor */
+        $randomFloor = $this->dungeonRoute->dungeon->floors()
+            ->where('facade', false)
+            ->get()
+            ->random();
+
+        $polyline = PolylineFixtures::createPolyline($randomFloor);
+
+        $response = $this->post(route('ajax.dungeonroute.path.create', ['dungeonRoute' => $this->dungeonRoute]), [
+            'floor_id' => $randomFloor->id,
+            'polyline' => $polyline,
+        ]);
+        $response->assertCreated();
+
+        /** @var Path $path */
+        $path = Path::findOrFail(json_decode($response->content(), true)['id']);
+
+        $event = new PathChangedEvent(
+            app(CoordinatesServiceInterface::class),
+            $this->dungeonRoute,
+            User::findOrFail(1),
+            $path,
+        );
+
+        // Act
+        $broadcastPayload = $event->broadcastWith();
+
+        // Assert - model.polyline.vertices_json is dead weight: PathChangedHandler always
+        // overwrites it client-side with model_data.coordinates before use, so broadcasting both
+        // just tripled the vertex payload and could exceed Reverb's message size cap on large
+        // paths (#3909). The rest of the polyline (used for e.g. color) must still be present.
+        $this->assertArrayNotHasKey('vertices_json', $broadcastPayload['model']['polyline']);
+        $this->assertEquals($polyline['color'], $broadcastPayload['model']['polyline']['color']);
+        $this->assertArrayHasKey('coordinates', $broadcastPayload['model_data']);
     }
 
     #[Test]
--- a/tests/Feature/Controller/Ajax/AjaxPathControllerTest.php
+++ b/tests/Feature/Controller/Ajax/AjaxPathControllerTest.php
@@ -6,8 +6,6 @@ use App\Events\Models\Path\PathChangedEvent;
 use App\Models\Floor\Floor;
 use App\Models\Path;
 use App\Models\Polyline;
-use App\Models\User;
-use App\Service\Coordinates\CoordinatesServiceInterface;
 use Illuminate\Support\Facades\Event;
 use PHPUnit\Framework\Attributes\Group;
 use PHPUnit\Framework\Attributes\Test;
@@ -52,6 +50,8 @@ final class AjaxPathControllerTest extends DungeonRouteTestBase
     public function store_givenNewValidPath_broadcastsPayloadWithoutRedundantVerticesJson(): void
     {
         // Arrange
+        Event::fake([PathChangedEvent::class]);
+
         /** @var Floor $randomFloor */
         $randomFloor = $this->dungeonRoute->dungeon->floors()
             ->where('facade', false)
@@ -60,32 +60,26 @@ final class AjaxPathControllerTest extends DungeonRouteTestBase
 
         $polyline = PolylineFixtures::createPolyline($randomFloor);
 
+        // Act
         $response = $this->post(route('ajax.dungeonroute.path.create', ['dungeonRoute' => $this->dungeonRoute]), [
             'floor_id' => $randomFloor->id,
             'polyline' => $polyline,
         ]);
         $response->assertCreated();
 
-        /** @var Path $path */
-        $path = Path::findOrFail(json_decode($response->content(), true)['id']);
-
-        $event = new PathChangedEvent(
-            app(CoordinatesServiceInterface::class),
-            $this->dungeonRoute,
-            User::findOrFail(1),
-            $path,
-        );
-
-        // Act
-        $broadcastPayload = $event->broadcastWith();
-
         // Assert - model.polyline.vertices_json is dead weight: PathChangedHandler always
         // overwrites it client-side with model_data.coordinates before use, so broadcasting both
         // just tripled the vertex payload and could exceed Reverb's message size cap on large
         // paths (#3909). The rest of the polyline (used for e.g. color) must still be present.
-        $this->assertArrayNotHasKey('vertices_json', $broadcastPayload['model']['polyline']);
-        $this->assertEquals($polyline['color'], $broadcastPayload['model']['polyline']['color']);
-        $this->assertArrayHasKey('coordinates', $broadcastPayload['model_data']);
+        Event::assertDispatched(PathChangedEvent::class, function (PathChangedEvent $event) use ($polyline) {
+            $broadcastPayload = $event->broadcastWith();
+
+            $this->assertArrayNotHasKey('vertices_json', $broadcastPayload['model']['polyline']);
+            $this->assertEquals($polyline['color'], $broadcastPayload['model']['polyline']['color']);
+            $this->assertArrayHasKey('coordinates', $broadcastPayload['model_data']);
+
+            return true;
+        });
     }
 
     #[Test]
--- a/app/Events/Models/Arrow/ArrowChangedEvent.php
+++ b/app/Events/Models/Arrow/ArrowChangedEvent.php
@@ -37,8 +37,18 @@ class ArrowChangedEvent extends ModelChangedEvent
         /** @var Arrow $model */
         $model = $this->model;
 
+        $broadcast = parent::broadcastWith();
+
+        // The receiving client always overwrites model.polyline.vertices_json with the
+        // model_data coordinates below (see ModelChangedHandler::_getCorrectLatLngFromEvent()
+        // and Arrow's changed.js), so broadcasting the raw vertices here too is dead weight
+        // that can push large arrows over Reverb's message size cap (#3909).
+        $modelArray = $model->toArray();
+        unset($modelArray['polyline']['vertices_json']);
+        $broadcast['model'] = $modelArray;
+
         return array_merge(
-            parent::broadcastWith(),
+            $broadcast,
             [
                 'model_data' => $model->polyline->getCoordinatesData(
                     $this->coordinatesService,
--- a/app/Events/Models/Brushline/BrushlineChangedEvent.php
+++ b/app/Events/Models/Brushline/BrushlineChangedEvent.php
@@ -44,8 +44,18 @@ class BrushlineChangedEvent extends ModelChangedEvent
         /** @var Brushline $model */
         $model = $this->model;
 
+        $broadcast = parent::broadcastWith();
+
+        // The receiving client always overwrites model.polyline.vertices_json with the
+        // model_data coordinates below (see ModelChangedHandler::_getCorrectLatLngFromEvent()
+        // and Brushline's changed.js), so broadcasting the raw vertices here too is dead weight
+        // that can push large brushlines over Reverb's message size cap (#3909).
+        $modelArray = $model->toArray();
+        unset($modelArray['polyline']['vertices_json']);
+        $broadcast['model'] = $modelArray;
+
         return array_merge(
-            parent::broadcastWith(),
+            $broadcast,
             [
                 'model_data' => $model->polyline->getCoordinatesData(
                     $this->coordinatesService,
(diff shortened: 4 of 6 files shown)


tests/Feature/Controller/Ajax/AjaxPathControllerTest.php
```diff
@@ -44,6 +47,47 @@ public function store_givenNewValidPath_shouldReturnPath(): void
...
+        // Act
+        $broadcastPayload = $event->broadcastWith();
+
+        // Assert - model.polyline.vertices_json is dead weight: PathChangedHandler always
+        // overwrites it client-side with model_data.coordinates before use, so broadcasting both
+        // just tripled the vertex payload and could exceed Reverb's message size cap on large
+        // paths (#3909). The rest of the polyline (used for e.g. color) must still be present.
+        $this->assertArrayNotHasKey('vertices_json', $broadcastPayload['model']['polyline']);
+        $this->assertEquals($polyline['color'], $broadcastPayload['model']['polyline']['color']);
+        $this->assertArrayHasKey('coordinates', $broadcastPayload['model_data']);
```
Wotuu: :robot: The test doesn't cover the seam its name implies. The POST is setup; the assertion target is `PathChangedEvent::broadcastWith()` invoked by hand on a re-fetched `Path`. Consequence: delete `broadcast(new PathChangedEvent(...))` from `AjaxPathController::store()` and this test still passes green — the controller's actual broadcast payload is never observed. As written it's a unit test of the event wearing a `store_...` name in `AjaxPathControllerTest`.

Faking the event scoped covers the real path *and* removes the local-run caveat noted in the PR body (no broadcast driver is ever touched, so it behaves identically with or without Reverb):

```php
Event::fake([PathChangedEvent::class]);

// ... existing POST ...

Event::assertDispatched(
    PathChangedEvent::class,
    fn (PathChangedEvent $event) => ! array_key_exists('vertices_json', $event->broadcastWith()['model']['polyline']),
);
```

Scoped `Event::fake([PathChangedEvent::class])` rather than a bare `Event::fake()` matters here — this file already works around global event faking at line 202 (`Event::forget('eloquent.updating: ' . Path::class)`), and the scoped form keeps the Eloquent model events that `Path`/`Polyline` saving depends on live.

Non-blocking.

Wotuu:

**Finding (non-blocking)** — the new test asserts against a hand-constructed `PathChangedEvent`, so the controller's actual broadcast is untested; scoped `Event::fake([PathChangedEvent::class])` + `assertDispatched` would cover the real seam and also dissolve the "no Reverb locally" caveat in the PR body.

**Verified, no defect found:**
- `'model' => $model->toArray()` is JSON-equivalent to the previous `'model' => $this->model` (both go through `attributesToArray()` + `relationsToArray()` honouring `Path::$visible`/`$appends`).
- The relation-load-ordering worry (`toArray()` now runs *before* `$model->polyline` is dereferenced) and the stale-relation worry (`Polyline::updateOrCreate()` in `SavesPolylines` returns a separate instance from `$path->polyline`) are both moot: `ContextEvent implements ShouldBroadcast` with `SerializesModels`, so `broadcastWith()` executes in the worker on a model restored via `newQueryForRestoration()` with `Path::$with = ['polyline', 'linkedawakenedobelisks']` applied. Same mechanism that made the #3198 killzone_paths fix work.
- `unset()` on the local `$modelArray` copy cannot reach `$model->polyline->vertices_json`, so `model_data` stays complete.
- Client-side premise holds: `PathChangedHandler::onReceive()` overwrites `e.model.polyline.vertices_json` whenever `_getCorrectLatLngFromEvent()` returns non-`false`, and `HasVertices::getCoordinatesData()` unconditionally returns a `coordinates` key, so the overwrite is not actually conditional in practice.

**Observation, not a finding** — the PR body already discloses this: the payload drops from ~3 vertex representations to ~2 (roughly a third), which does not hard-bound it. `Closes #3909` will auto-close a Sentry-derived issue that can plausibly recur on a sufficiently large path. Might be worth downgrading to a plain reference so the issue stays open until the Sentry issue is observably quiet.

CI checked at review time: 8/9 green, `php-tests (feature-other)` still pending; `php-tests (feature-controller)` — the shard that owns `AjaxPathControllerTest` — passed, which resolves the PR body's "couldn't run this locally" caveat.

Wotuu:
:robot: Addressed the cold-review finding in f3e1b3dd0: the test now uses `Event::fake([PathChangedEvent::class])` + `Event::assertDispatched()` against the controller's actual `broadcast()` call, instead of hand-constructing the event. Verified it fails if the `broadcast()` call is removed, fails on pre-fix `PathChangedEvent`, and passes with both present. This also drops the "no Reverb locally" caveat from the PR body — `Event::fake()` intercepts before the real broadcast pipeline, so the test passes locally without a driver override now.

app/Events/Models/Path/PathChangedEvent.php
```diff
@@ -44,8 +44,18 @@ public function broadcastWith(): array
         /** @var Path $model */
         $model = $this->model;
 
+        $broadcast = parent::broadcastWith();
```
Wotuu, line 47: This fix looks good, but don't you need to add it in more places, like the brush line, for example, or maybe enemy patrols? There are more models that use a polyline that then also send the vertices JSON along with the two other ones. Can you double-check all of those and see if you need to add this fix to more places?

Wotuu: :robot: Good catch — Brushline and Arrow did need the same fix. Done in 05d8b2f03, with a broadcast-payload test for each mirroring the Path one.

I audited this by mechanism rather than by eye: every `polyline()` relation in `app/Models`, every `HasVertices`/`SavesPolylines` user, and every `ModelChangedEvent`/`ModelDeletedEvent` subclass under `app/Events/`. The discriminator is whether the payload carries the raw `polyline.vertices_json` **and** a computed `model_data.coordinates` derived from the same vertices — only then is one of them redundant.

| Event | Carries polyline? | Also sends `model_data`? | Verdict |
|---|---|---|---|
| `PathChangedEvent` | yes | yes | **Trimmed** (original commit) |
| `BrushlineChangedEvent` | yes | yes | **Trimmed** in 05d8b2f03 |
| `ArrowChangedEvent` | yes | yes | **Trimmed** in 05d8b2f03 |
| `EnemyPatrolChangedEvent` | yes (`$with = ['polyline']`) | **no** — no `broadcastWith()` override | Doesn't apply |
| `EnemyPackChangedEvent` | `vertices_json` via `HasVertices` | no | Doesn't apply |
| `MountableAreaChangedEvent` | `vertices_json` via `HasVertices` | no | Doesn't apply |
| `FloorUnionAreaChangedEvent` | `vertices_json` via `HasVertices` | no | Doesn't apply |
| `KillZoneChangedEvent` | no polyline; used to send `killzone_paths` | n/a | Already fixed in 902f16d8e (#3198) |
| all `*DeletedEvent` | no | n/a | Nothing to trim |

Details on the three "doesn't apply" groups:

- **EnemyPatrol** — `EnemyPatrolChangedEvent` has no `broadcastWith()` override at all, so `model.polyline.vertices_json` is the *only* representation of the patrol's vertices in the payload. On top of that, `echohandler.js` registers no `EnemyPatrolChangedHandler` (only Brushline/Arrow/KillZone/MapIcon/Npc/OverpulledEnemy/Path have listeners), so nothing client-side would recompute them. Trimming here would delete data with no substitute.
- **EnemyPack / MountableArea / FloorUnionArea** — these carry `vertices_json` directly on the model via the `HasVertices` trait (polygons, not a `Polyline` relation), and their changed events add no `model_data`, so again only one copy goes over the wire.
- **KillZone** — the payload bloat there was `killzone_paths`, removed from the broadcast in 902f16d8e (#3198); the client now re-fetches them over the dedicated route added in that commit. Verified still absent, so not a miss.

The trim is only valid for Path/Brushline/Arrow because all three share the same client contract: `broadcastWith()` adds `model_data` **unconditionally**, and their `changed.js` handlers unconditionally do `e.model.polyline.vertices_json = JSON.stringify(coordinates)` from `ModelChangedHandler::_getCorrectLatLngFromEvent()` before handing the model to `loadMapObject()`. The rest of the polyline (colour, weight, animated colour) is still broadcast.

Also worth noting for the "is the payload now minimal" question: `ContextEvent::broadcastWith()` is already lean (just `__name`, the context route key/class and three user fields), so after this the vertices are genuinely the dominant term. As before, this reduces the payload rather than hard-bounding it — a sufficiently long polyline can still exceed the cap with the one remaining representation.

tests/Feature/Controller/Ajax/AjaxPathControllerTest.php
Wotuu: :robot: Fixed in 9cfc4f8dc, "Address cold review: assert against the controller's actual broadcast" (the commit was `f3e1b3dd0` when the fix went in; it was rebased onto master since, so cite the new SHA). Apologies — the "addressed" note went out as a top-level PR comment at the time instead of on this thread.

`store_givenNewValidPath_broadcastsPayloadWithoutRedundantVerticesJson` now does exactly what you suggested: scoped `Event::fake([PathChangedEvent::class])` up front, the real POST to `ajax.dungeonroute.path.create`, then `Event::assertDispatched(PathChangedEvent::class, ...)` inspecting `$event->broadcastWith()` on the instance the controller actually constructed — no hand-built event, no re-fetched `Path`. Deleting the `broadcast(...)` call from `AjaxPathController::store()` now fails the test.

The scoped form was kept for the reason you gave (the `Event::forget('eloquent.updating: ' . Path::class)` workaround further down the file shows Eloquent model events must stay live), and the local-run caveat is gone from the PR body.

The same pattern is now applied to the two new tests added in 05d8b2f03 for `ArrowChangedEvent` and `BrushlineChangedEvent`, and I verified all three are non-vacuous by reverting the event changes and watching them go red.

Wotuu:
:robot: **Review feedback addressed** — pushed as `05d8b2f03`.

> _Note: GitHub put my two in-thread replies into the still-unsubmitted draft review that holds the `PathChangedEvent.php:47` comment (the API refuses a normal reply while a pending review exists on the same account), so they will only become visible once that review is submitted. Reproducing the substance here so it is not lost — I did not submit or discard the draft review myself._

Good catch — Brushline and Arrow did need the same fix. Done in 05d8b2f03, with a broadcast-payload test for each mirroring the Path one.

I audited this by mechanism rather than by eye: every `polyline()` relation in `app/Models`, every `HasVertices`/`SavesPolylines` user, and every `ModelChangedEvent`/`ModelDeletedEvent` subclass under `app/Events/`. The discriminator is whether the payload carries the raw `polyline.vertices_json` **and** a computed `model_data.coordinates` derived from the same vertices — only then is one of them redundant.

| Event | Carries polyline? | Also sends `model_data`? | Verdict |
|---|---|---|---|
| `PathChangedEvent` | yes | yes | **Trimmed** (original commit) |
| `BrushlineChangedEvent` | yes | yes | **Trimmed** in 05d8b2f03 |
| `ArrowChangedEvent` | yes | yes | **Trimmed** in 05d8b2f03 |
| `EnemyPatrolChangedEvent` | yes (`$with = ['polyline']`) | **no** — no `broadcastWith()` override | Doesn't apply |
| `EnemyPackChangedEvent` | `vertices_json` via `HasVertices` | no | Doesn't apply |
| `MountableAreaChangedEvent` | `vertices_json` via `HasVertices` | no | Doesn't apply |
| `FloorUnionAreaChangedEvent` | `vertices_json` via `HasVertices` | no | Doesn't apply |
| `KillZoneChangedEvent` | no polyline; used to send `killzone_paths` | n/a | Already fixed in 902f16d8e (#3198) |
| all `*DeletedEvent` | no | n/a | Nothing to trim |

Details on the three "doesn't apply" groups:

- **EnemyPatrol** — `EnemyPatrolChangedEvent` has no `broadcastWith()` override at all, so `model.polyline.vertices_json` is the *only* representation of the patrol's vertices in the payload. On top of that, `echohandler.js` registers no `EnemyPatrolChangedHandler` (only Brushline/Arrow/KillZone/MapIcon/Npc/OverpulledEnemy/Path have listeners), so nothing client-side would recompute them. Trimming here would delete data with no substitute.
- **EnemyPack / MountableArea / FloorUnionArea** — these carry `vertices_json` directly on the model via the `HasVertices` trait (polygons, not a `Polyline` relation), and their changed events add no `model_data`, so again only one copy goes over the wire.
- **KillZone** — the payload bloat there was `killzone_paths`, removed from the broadcast in 902f16d8e (#3198); the client now re-fetches them over the dedicated route added in that commit. Verified still absent, so not a miss.

The trim is only valid for Path/Brushline/Arrow because all three share the same client contract: `broadcastWith()` adds `model_data` **unconditionally**, and their `changed.js` handlers unconditionally do `e.model.polyline.vertices_json = JSON.stringify(coordinates)` from `ModelChangedHandler::_getCorrectLatLngFromEvent()` before handing the model to `loadMapObject()`. The rest of the polyline (colour, weight, animated colour) is still broadcast.

Also worth noting for the "is the payload now minimal" question: `ContextEvent::broadcastWith()` is already lean (just `__name`, the context route key/class and three user fields), so after this the vertices are genuinely the dominant term. As before, this reduces the payload rather than hard-bounding it — a sufficiently long polyline can still exceed the cap with the one remaining representation.

---

**On the `AjaxPathControllerTest` cold-review thread** (`Event::fake` scoping):

Fixed in 9cfc4f8dc, "Address cold review: assert against the controller's actual broadcast" (the commit was `f3e1b3dd0` when the fix went in; it was rebased onto master since, so cite the new SHA). Apologies — the "addressed" note went out as a top-level PR comment at the time instead of on this thread.

`store_givenNewValidPath_broadcastsPayloadWithoutRedundantVerticesJson` now does exactly what you suggested: scoped `Event::fake([PathChangedEvent::class])` up front, the real POST to `ajax.dungeonroute.path.create`, then `Event::assertDispatched(PathChangedEvent::class, ...)` inspecting `$event->broadcastWith()` on the instance the controller actually constructed — no hand-built event, no re-fetched `Path`. Deleting the `broadcast(...)` call from `AjaxPathController::store()` now fails the test.

The scoped form was kept for the reason you gave (the `Event::forget('eloquent.updating: ' . Path::class)` workaround further down the file shows Eloquent model events must stay live), and the local-run caveat is gone from the PR body.

The same pattern is now applied to the two new tests added in 05d8b2f03 for `ArrowChangedEvent` and `BrushlineChangedEvent`, and I verified all three are non-vacuous by reverting the event changes and watching them go red.

PR #3938, opened 2026-08-09
#3914 Degrade gracefully on a dropped Redis connection in CacheService

### LINKED ISSUE #3914: Intermittent RedisException: Connection timed out (CacheService / MetricService)
:robot: Two related but separately-clustered Redis connection timeouts, both looking environmental (transient Redis connectivity) rather than application bugs:

**PHP-LARAVEL-6Q** — production, `CacheService::get()`, 96 occurrences, 7 users, 2026-01-30 to 2026-08-06 (long-running, low-frequency)
```
RedisException: Connection timed out
```
```php
211 │ public function get(string $key): mixed
212 │ {
213 │     return Cache::get($key);
```
Triggered via `TrustProxies` → `CloudflareService::getIpRanges()` → cache lookup of `cloudflare:ip-ranges-v4`, hit on `GET /status`.

**PHP-LARAVEL-RZ** — staging, `MetricService::aggregateMetrics()` → `modelCache:clear` → Redis tag flush, 1 occurrence, 2026-08-07

Both are the same failure mode (Redis connection timeout during a `PhpRedisConnector::connect()`) surfacing in different call paths. Given the low, sporadic frequency this looks like transient Redis network blips rather than a code defect — filing to track recurrence; no code fix planned unless frequency increases.

Wotuu:
:robot: Closes #3914

## Root cause

`CacheService::get()`/`remember()` are called from `TrustProxies::handle()` on **every production request** (via `CloudflareService::getIpRanges()`, cached under `cloudflare:ip-ranges-v4`/`-v6`). An uncaught `RedisException` from a transient Redis connection blip there isn't a failed cache lookup — Laravel's exception handler turns it into a 500 for the whole request, because `TrustProxies` runs before the router dispatches to the actual controller.

This is `PHP-LARAVEL-6Q`: 96 occurrences over 6+ months (2026-01-30 to 2026-08-06), 7 distinct users, production, `RedisException: Connection timed out` inside `PhpRedisConnector::connect()`.

## Fix

- `CacheService::get()` now catches `RedisException` and returns `null` (a cache miss) instead of letting it propagate. `remember()` already treats a `null` `get()` result as "compute fresh", so this just means the request proceeds instead of crashing.
- `remember()`'s existing "failed to write cache" catch around `set()` (previously only `InvalidArgumentException`) is widened to also catch `RedisException`, so a write-side blip degrades the same way a read-side one now does — the caller still gets its freshly-computed value, it just doesn't get cached for that request.
- Both paths log via the existing `CacheServiceLoggingInterface` (`error` level, matching the existing `rememberFailedToSetCache` precedent) so the failure stays visible/trackable, it just no longer takes the site down.

## Scope

The Sentry issue also cites a second, much lower-volume cluster: `MetricService::aggregateMetrics()` → `Artisan::call('modelCache:clear')`, 1 staging occurrence (`PHP-LARAVEL-RZ`). That's a scheduled console command failing, not a page-crashing path — a failed run just retries next schedule. Left alone; fixing it would be scope creep beyond what the evidence calls for.

## Verification

- Added `CacheServiceRedisResilienceTest` (3 cases: read-side failure, write-side failure, and the happy path implicitly covered by existing `CacheServiceHashTest`/`CacheServiceClearIdleKeysTest`).
- Verified all 3 new tests fail against pre-fix `CacheService.php` (uncaught `RedisException` propagates) and pass post-fix.
- Full `Cache` test group (8 tests) still green.

### DIFF
--- a/app/Service/Cache/CacheService.php
+++ b/app/Service/Cache/CacheService.php
@@ -211,10 +211,8 @@ class CacheService implements CacheServiceInterface
 
     public function get(string $key): mixed
     {
-        // A dropped Redis connection here must not bubble up as a 500 - this is called from
-        // TrustProxies on every production request (via CloudflareService::getIpRanges()), so an
-        // uncaught RedisException here takes the whole site down on a transient Redis blip (#3914).
-        // Degrading to a cache miss lets the caller recompute the value instead.
+        // Called from TrustProxies on every production request (via CloudflareService::getIpRanges()),
+        // so an uncaught RedisException here would take the whole site down on a Redis blip (#3914).
         try {
             return Cache::get($key);
         } catch (RedisException $e) {
--- a/app/Service/Cache/CacheService.php
+++ b/app/Service/Cache/CacheService.php
@@ -12,6 +12,7 @@ use Illuminate\Redis\Connections\Connection;
 use Illuminate\Support\Facades\Cache;
 use Illuminate\Support\Facades\Redis;
 use Psr\SimpleCache\InvalidArgumentException;
+use RedisException;
 
 class CacheService implements CacheServiceInterface
 {
@@ -137,7 +138,7 @@ class CacheService implements CacheServiceInterface
                             if ($this->set($key, $value, $ttl ?? $this->getTtl($key))) {
                                 $result = $value;
                             }
-                        } catch (InvalidArgumentException $e) {
+                        } catch (InvalidArgumentException|RedisException $e) {
                             $this->log->rememberFailedToSetCache($key, $e);
 
                             $result = $value;
@@ -210,7 +211,17 @@ class CacheService implements CacheServiceInterface
 
     public function get(string $key): mixed
     {
-        return Cache::get($key);
+        // A dropped Redis connection here must not bubble up as a 500 - this is called from
+        // TrustProxies on every production request (via CloudflareService::getIpRanges()), so an
+        // uncaught RedisException here takes the whole site down on a transient Redis blip (#3914).
+        // Degrading to a cache miss lets the caller recompute the value instead.
+        try {
+            return Cache::get($key);
+        } catch (RedisException $e) {
+            $this->log->getFailedRedisConnection($key, $e);
+
+            return null;
+        }
     }
 
     /**
--- a/app/Service/ReadOnlyMode/ReadOnlyModeService.php
+++ b/app/Service/ReadOnlyMode/ReadOnlyModeService.php
@@ -25,10 +25,6 @@ class ReadOnlyModeService implements ReadOnlyModeServiceInterface
      */
     public function isReadOnly(): bool
     {
-        // has() was redundant (a missing key already makes get() return null, which fails the
-        // cast/comparison below the same way) and, unlike get() since #3914, still throws
-        // uncaught on a dropped Redis connection - AppLayoutComposer's isReadOnly() check runs on
-        // every rendered page, so that uncaught exception 500'd the entire site on a Redis blip.
         return (bool)$this->cacheService->get('read_only_mode') === true;
     }
 
--- a/app/Service/ReadOnlyMode/ReadOnlyModeService.php
+++ b/app/Service/ReadOnlyMode/ReadOnlyModeService.php
@@ -25,7 +25,11 @@ class ReadOnlyModeService implements ReadOnlyModeServiceInterface
      */
     public function isReadOnly(): bool
     {
-        return $this->cacheService->has('read_only_mode') && (bool)$this->cacheService->get('read_only_mode') === true;
+        // has() was redundant (a missing key already makes get() return null, which fails the
+        // cast/comparison below the same way) and, unlike get() since #3914, still throws
+        // uncaught on a dropped Redis connection - AppLayoutComposer's isReadOnly() check runs on
+        // every rendered page, so that uncaught exception 500'd the entire site on a Redis blip.
+        return (bool)$this->cacheService->get('read_only_mode') === true;
     }
 
     /**
--- a/app/Service/Traits/Curl.php
+++ b/app/Service/Traits/Curl.php
@@ -5,7 +5,7 @@ namespace App\Service\Traits;
 trait Curl
 {
     /**
-     * @param array<int, mixed> $options CURLOPT_* constants as keys - they are ints, not strings
+     * @param array<int, mixed> $options
      */
     public function curlGet(string $url, array $options = []): string
     {
@@ -71,7 +71,7 @@ trait Curl
      * Note `httpCode` is 0 for non-HTTP protocols, and when the request never got a response at all, so callers
      * should judge failure on `errorNumber` first.
      *
-     * @param  array<int, mixed>                                         $options CURLOPT_* constants as keys - they are ints, not strings
+     * @param  array<int, mixed>                                         $options
      * @return array{body: string|bool, httpCode: int, errorNumber: int}
      */
     protected function curlGetResponse(string $url, array $options = []): array
--- a/app/Service/Traits/Curl.php
+++ b/app/Service/Traits/Curl.php
@@ -5,7 +5,7 @@ namespace App\Service\Traits;
 trait Curl
 {
     /**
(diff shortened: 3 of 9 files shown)


app/Service/Cache/CacheService.php
```diff
@@ -210,7 +211,17 @@ public function dropHashCache(string $hashKey): void
...
     {
-        return Cache::get($key);
+        // A dropped Redis connection here must not bubble up as a 500 - this is called from
+        // TrustProxies on every production request (via CloudflareService::getIpRanges()), so an
+        // uncaught RedisException here takes the whole site down on a transient Redis blip (#3914).
+        // Degrading to a cache miss lets the caller recompute the value instead.
+        try {
+            return Cache::get($key);
+        } catch (RedisException $e) {
+            $this->log->getFailedRedisConnection($key, $e);
```
Wotuu, line 219: :robot: **The new `get()` protection can't fire on the hottest page-render path, because an unprotected sibling call runs first.**

`ReadOnlyModeService::isReadOnly()` (`app/Service/ReadOnlyMode/ReadOnlyModeService.php:27`) is:

```php
return $this->cacheService->has('read_only_mode') && (bool)$this->cacheService->get('read_only_mode') === true;
```

`&&` short-circuits left-to-right, so `has()` — which this PR leaves as a bare `Cache::has($key)` — evaluates *first*. On the exact Redis blip this PR targets, `has()` throws the `RedisException` before `get()` is ever reached, and the request still 500s.

That path isn't obscure: `AppLayoutComposer::compose()` is registered for `layouts.app`, `layouts.sitepage` and `layouts.map` (`app/Providers/KeystoneGuruServiceProvider.php:332-336`), i.e. effectively every rendered page. It calls `messageBannerService->getMessage()` (now protected by this change) on one line and `readOnlyModeService->isReadOnly()` (still unprotected) on the next. `ReadOnlyMode` middleware is also registered in two middleware groups in `bootstrap/app.php:72,84`, giving a second per-request hit.

Cheapest fix is at the call site rather than wrapping `has()`: drop the `has()` call entirely — `(bool)$this->cacheService->get('read_only_mode') === true` is equivalent (a stored `false` and a missing key both yield `false`), removes a redundant Redis round-trip, and inherits the protection added here.

Secondary, same class of gap and lower priority: `CacheService::lock()` (`Cache::lock(...)->block()`) is unprotected and is hit on every API request via `ApiMetrics` middleware -> `MetricService::storeMetricAsync()`. A lock you can't acquire being fatal is more defensible, so this one is fine to consciously decline — just worth stating in the PR body if so.

app/Service/Cache/CacheService.php
```diff
@@ -210,7 +211,17 @@ public function dropHashCache(string $hashKey): void
 
     public function get(string $key): mixed
     {
-        return Cache::get($key);
+        // A dropped Redis connection here must not bubble up as a 500 - this is called from
+        // TrustProxies on every production request (via CloudflareService::getIpRanges()), so an
+        // uncaught RedisException here takes the whole site down on a transient Redis blip (#3914).
+        // Degrading to a cache miss lets the caller recompute the value instead.
+        try {
+            return Cache::get($key);
```
Wotuu, line 217: :robot: **Failing open here routes a Redis outage into the unbounded Cloudflare fetch — and, if that fetch also fails, into an empty trusted-proxy list.**

Two consequences worth bounding before this ships, both in `TrustProxies` -> `CloudflareService::getIpRangesV4/V6()`:

1. Each degraded request now performs two synchronous outbound cURL calls, and `Curl::curlGetResponse()` (`app/Service/Traits/Curl.php`) defaults to `CURLOPT_CONNECTTIMEOUT => 120` / `CURLOPT_TIMEOUT => 120` — up to 240s per request, inside middleware that runs before routing. During a sustained Redis outage that converts fast 500s into pinned FPM workers. `curlGet()` already accepts an `$options` array, so capping both timeouts (a few seconds) at the two `CloudflareService` call sites is a one-liner.

2. `curlGet()` has no transport-failure check: `curl_exec` returning `false` is coerced to `''` by the `: string` return type, `validateIpAddressRanges('')` yields `[]`, and `TrustProxies::handle()` then sets `$this->proxies = []`. Symfony trusts nothing, `$request->ip()` resolves to the Cloudflare edge IP, and rate limiting buckets swaths of visitors under a handful of IPs — exactly what `useCloudflareConnectingIp()`'s own docblock says it exists to prevent. Before this PR that state required cURL to fail *while Redis was up*; now a Redis outage reaches it too. Note `curlSaveToFile()` in the same trait explicitly guards `body === false || errorNumber !== CURLE_OK` (#3789) — the codebase already knows about this failure mode, `curlGet()` callers just don't get it.

The change is clearly net-positive over a hard 500; this is about bounding what it degrades into.

Wotuu:

Wotuu:
:robot: Addressed both cold-review findings in fde083b1a:

1. **`ReadOnlyModeService::isReadOnly()`'s unprotected `has()` call** — dropped it (redundant: `get()` already returns `null` for a missing key, which fails the same `=== true` comparison). Added `ReadOnlyModeServiceTest` covering both the "must not call `has()`" and "null degrades to false" cases; verified it fails on the pre-fix `&&`-short-circuit version.

2. **Unbounded Cloudflare fetch on a degraded read** — bounded `CURLOPT_CONNECTTIMEOUT`/`CURLOPT_TIMEOUT` to 5s on both `getIpRangesV4()`/`getIpRangesV6()` calls, so a Redis outage can no longer cascade into requests hanging for up to 240s inline in `TrustProxies`. Added a test asserting the options are passed, verified it fails pre-fix. (Also fixed `Curl::curlGet()`'s `$options` PHPDoc while touching it — `CURLOPT_*` constants are int keys, not string, which is what PHPStan caught.)

**Intentionally not addressed**: a Cloudflare fetch that itself fails silently still yields an empty proxy list, which defeats `CF-Connecting-IP` trust and buckets visitors under the Cloudflare edge IP for rate limiting. This is a pre-existing gap reachable via any cold-cache + Cloudflare-API-down combination, independent of Redis — out of scope for a Redis-resilience PR. Noted in the commit message as worth its own follow-up.

app/Service/ReadOnlyMode/ReadOnlyModeService.php
```diff
@@ -25,7 +25,11 @@ public function setReadOnly(bool $readOnly): bool
      */
     public function isReadOnly(): bool
     {
-        return $this->cacheService->has('read_only_mode') && (bool)$this->cacheService->get('read_only_mode') === true;
+        // has() was redundant (a missing key already makes get() return null, which fails the
```
Wotuu: Okay, you removed some code, but no need to add a comment for it. It's gone, so no need to reference it anymore.

app/Service/Traits/Curl.php
```diff
@@ -5,7 +5,7 @@
 trait Curl
 {
     /**
-     * @param array<string, mixed> $options
+     * @param array<int, mixed> $options CURLOPT_* constants as keys - they are ints, not strings
```
Wotuu: "they are ints not strings." Okay, good for you, but remove that. I can read that from the line itself.

app/Service/ReadOnlyMode/ReadOnlyModeService.php
Wotuu: :robot: Removed — the four-line comment explaining the removed `has()` call is gone entirely, leaving just the `return`. Also swept the rest of the diff for the same thing: dropped the duplicated "why" comment from the `CloudflareServiceTest` arrange block (it just restated the constant's docblock), the narration above the assert in `CacheServiceRedisResilienceTest`, and the `has()`-referencing comment in `ReadOnlyModeServiceTest`; trimmed the `CacheService::get()` and `FETCH_TIMEOUT_SECONDS` comments down to the part that is actually non-obvious.

app/Service/Traits/Curl.php
Wotuu: :robot: Removed. Both `@param` lines are back to a bare `@param array<int, mixed> $options` — the corrected `int` key type stays, all the prose describing it is gone.

PR #3944, opened 2026-08-10
#3939 Update MDT to v6.2.0-alpha5

### LINKED ISSUE #3939: Update MDT to v6.2.0-alpha5
Tracking the update of the Mythic Dungeon Tools (MDT) composer package and reimport of the
current retail season's dungeon mappings (Season 2 for Midnight).

- [ ] Bump composer package reference (version + commit SHA) and `composer update`
- [ ] Reimport each retail-season dungeon mapping
- [ ] Handle any new POI types/templates and missing map-icon translations
- [ ] Run per-dungeon tests (rewrite stale CorrectEvents fixtures)
- [ ] Bump `config/keystoneguru.php` MDT version
- [ ] Re-export NPC translations and re-seed
- [ ] Run the full test suite

Wotuu:
:robot: Closes #3939

## Summary
- Bumps `nnoggie/mythicdungeontools` composer package from 6.2.0-alpha3 to 6.2.0-alpha5.
- Reimports all 8 Midnight Season 2 dungeon mappings (`kingsrest`, `templeofsethraliss`, `rubylifepools`, `altar_of_fangs`, `den_of_nalorakk`, `murder_row`, `the_blinding_vale`, `voidscar_arena`). Two of them (`murder_row`, `the_blinding_vale`) have actual mapping data changes; the rest re-export as effectively unchanged.
- Fixes `app/Logic/MDT/Data/MDTDungeon.php::getLua()` to handle upstream's new vararg-unpacking header (`local _, MDT = ...` replacing `local addonName = ...` / `local MDT = MDT`), which upstream commit `ebe8a6e0` ("split UI into load-on-demand addon") introduced. Without this, every Season 2 dungeon import failed with `attempt to index a nil value (local 'MDT')`.
- Adds a Step 0 (baseline full test suite run) to the `update-mdt-package` skill for future bumps.
- Cold review fixes: stamped the two reimported mapping versions with the correct `mdt_addon_version` (6205, was 6203), aligned `MDTMappingExportService`'s Lua header with the new alpha5 format, simplified the `getLua()` replaceStrings entry, and dropped three unrelated whitespace-only file diffs.
- Fixed a pre-existing, unrelated flaky test (`AjaxKillZoneControllerTest::store_givenAFacadeLocationWhileTheMapFacadeStyleSaysSplitFloors_persistsTheLocationOnItsRealFloor`) that intermittently picked an enemy outside floor union area coverage in `lowerkarazhan` — reproduced at the same rate on unmodified master, confirming it predates this PR.

## Test plan
- [x] Full MDT-filtered test suite passes (301 passed, 653 assertions) after rebasing onto latest master.
- [x] Full `feature-controller` CI group (387 tests) green, including 30 repeated runs of the previously-flaky test to confirm the fix.
- [x] `composer run analyse` (PHPStan) and `composer run fix` (PHP CS Fixer) clean.
- [x] Visual verification: `murder_row` and `the_blinding_vale` explore pages screenshotted — see PR comment below. Both render cleanly with sane enemy/pack placement.


database/seeders/dungeondata/mapping_versions.json
```diff
@@ -9022,5 +9006,53 @@
...
+        "game_version_id": 1,
+        "dungeon_id": 143,
+        "version": 3,
+        "enemy_forces_required": 690,
+        "enemy_forces_required_teeming": 0,
+        "enemy_forces_shrouded": null,
+        "enemy_forces_shrouded_zul_gamux": null,
+        "timer_max_seconds": 0,
+        "mdt_mapping_hash": "9e40d6afe469a1cb4bade0e001032491",
+        "mdt_addon_version": 6203,
```
Wotuu: :robot: **The two new mapping versions are stamped `mdt_addon_version: 6203` (alpha3), not `6205` (alpha5)** — the import evidently ran before `config/keystoneguru.php` was bumped. `MDTAddonVersionService::getCurrentAddonVersion()` strips non-digits from `keystoneguru.mdt.version`, so `v6.2.0-alpha5` → `6205` (which does exist in the refreshed `addon_versions.json`, dated `2026-08-04T12:40:34Z`).

This is not just cosmetic. The predecessors are stamped `6203` too (murder_row v2 = id 812, the_blinding_vale v3 = id 814), so in `MappingService::getMappingVersionForMdtAddonVersion()` a string exported by the *real* MDT alpha3 (`addonVersion = 6203`, release date `2026-07-21T19:20:06Z`) now:

- matches 812 (candidate date == string date, so not `lessThan`), then
- hits 832 with the **same** date → the `equalTo($matchDate)` branch fires and the higher `version` wins → **832**.

So an alpha3 route silently binds to the alpha5 mapping and loses the outdated/upgrade prompt #3380 added. With `6205` the dates differ, the loop `break`s after 812, and 812 correctly wins.

Fix is a two-line hand-edit here (`6203` → `6205` on ids 832 and 833).

app/Logic/MDT/Data/MDTDungeon.php
```diff
@@ -384,6 +385,7 @@ function GetDungeonIndex()
 
             $replaceStrings = [
                 'local addonName = ...' => 'local addonName = "MythicDungeonTools"',
+                'local _, MDT = ...'    => 'local _, MDT = "MythicDungeonTools", MDT',
```
Wotuu: :robot: The import side is fixed here, but **the export side still emits the pre-alpha5 header**. `MDTMappingExportService::getHeader()` (`app/Service/MDT/MDTMappingExportService.php:73`) hardcodes:

```
local addonName = ...
local MDT = MDT
local L = MDT.L
```

while every mainline dungeon file at `94473b39` now starts with (verified for all 16 files under `Midnight/`):

```
local _, MDT = ...
local addonName = MDT.AddonName
local L = MDT.L
```

`MDTMappingExportService` exists precisely to generate those files (`mdt:exportmapping` + `AdminToolsMdtController`), so its output is now in a stale format relative to the addon version this PR pins.

No test catches this because keystone's own `getLua()` prelude declares `local MDT = {}` in the same chunk, so `local MDT = MDT` still resolves through that outer local on the internal round-trip — the mismatch only shows up in the real addon. Switching the emitted header to the new form is safe both ways, since the `local _, MDT = ...` replacement added in this PR handles it on re-import.

Minor, on this line: the `"MythicDungeonTools"` value lands in `_` and is discarded (`addonName` now comes from `MDT.AddonName`, which is why the prelude line above is needed) — `'local _, MDT = ...' => 'local MDT = MDT'` would be equivalent and clearer about intent.

database/seeders/seasondata/affix_groups.json
```diff
@@ -4192,8 +4192,7 @@
                 "key_level": 12
             }
         ]
-    }
-,
+    },
```
Wotuu: :robot: Three files in this PR contain **zero content changes** — only reformatting artifacts from the seeder rewrite:

- `database/seeders/seasondata/affix_groups.json` — this hunk is just `}\n,` → `},`
- `database/seeders/seasondata/seasons.json` — only the trailing newline removed
- `database/seeders/dungeondata/spells.json` — only the trailing newline removed

None of them are MDT mapping data, so `git checkout -- <file>` on all three is zero-risk and matches the repo's "stage only the files you actually intended to touch" convention. Keeps the diff (already 15k lines) free of unrelated churn.

Wotuu:

database/seeders/dungeondata/mapping_versions.json
Wotuu: :robot: Fixed — stamped both new mapping versions (murder_row id 832, the_blinding_vale id 833) with `mdt_addon_version: 6205` instead of 6203. Verified `6205` exists in the refreshed `addon_versions.json` (2026-08-04T12:40:34Z), later than 6203's date, so the equalTo-match regression you described no longer happens.

app/Logic/MDT/Data/MDTDungeon.php
Wotuu: :robot: Fixed — updated `MDTMappingExportService::getHeader()` to emit `local _, MDT = ...` / `local addonName = MDT.AddonName` / `local L = MDT.L`, matching the real Midnight/*.lua header format. Also took your suggestion for MDTDungeon.php's replaceStrings entry — simplified to `'local _, MDT = ...' => 'local MDT = MDT'` since the addon-name value was landing in a discarded `_` anyway.

database/seeders/seasondata/affix_groups.json
Wotuu: :robot: Fixed — reverted all three files (`affix_groups.json`, `seasons.json`, `spells.json`) to their pre-PR content; confirmed no diff remains against master.

Wotuu:
:robot: Visual verification of the two dungeons with actual mapping-data changes — both explore pages render cleanly with sane enemy/pack placement, no console/page errors beyond the known benign cookie-parse warning.

**murder_row**
murder_row

**the_blinding_vale**
the_blinding_vale

PR #3948, opened 2026-08-10
#3945 Revamp the compendium pages into the Field Record design

### LINKED ISSUE #3945: Revamp the /compendium page designs for production readiness
:robot: The information architecture and factual content of the Compendium pages (`/compendium`, `/compendium/npc[/{npc}]`, `/compendium/spell[/{spell}]`, `/compendium/activity[/{date}]`, `/compendium/class[/{class}]`) is OK, but the visual design is not production-ready — the pages currently style entirely with stock Bootstrap classes (see the placeholder note in `resources/assets/css/sections/compendium.css`) and the design does not meet the bar to go live.

## Scope

Redesign the Compendium surface within the incumbent visual system (see `DESIGN.md`: "The War Room" — dark tonal layering, Keystone Green single accent, sturdy/legible components, theme-variable driven so darkly/lux/vapor all hold):

- `resources/views/compendium/**` (index, npc index/show, spell index/show, class index/show, activity index/day + section partials)
- `resources/assets/css/sections/compendium.css` (the designated home for compendium CSS)

**Content stays factual and unchanged** — this is a design pass, not a copy/data change.

## Definition of done

- All compendium pages restyled, verified in headless Chrome with before/after screenshots posted on the MR (desktop + mobile).
- Works in all three themes (darkly flagship; lux and vapor sanity-checked).
- No hardcoded theme hexes; everything through `var(--theme-*)` / `--bs-*` per DESIGN.md.
- Tests green, `composer run analyse` + `composer run fix` clean.

Wotuu:
Closes #3945

:robot: Redesigns all `/compendium` pages within the site's incumbent visual system ("The War Room", per the new `DESIGN.md`). Every page now reads as one continuous **field record**: an identity band on top, then labeled register sections (label rail left, content right) separated by hairline rules, everything on the theme-variable tonal ladder — no hardcoded theme hexes, so darkly, lux and vapor all hold from one rule set. Content is factually unchanged; two lang keys were added (`event.count`, `event.more`, plus characteristics `empty`/`not_observed` labels).

Highlights:
- Landing: icon-card grid → **directory register** (icon plate / title+count / description / go), data-source panel, "how it works" register rows.
- NPC/spell detail: identity band with portrait + classification badges + stat chips; characteristics as **labeled chips** with a muted "not observed" strip; spell details table (1 row × 8 columns) → **key-value data plate**; event feeds → aligned **log rows** with a fixed time column and one colored glyph per event type.
- Indexes: DataTables restyled into the register language (Dungeon Black header row, hairline dividers); class picker tiles restyled on the ladder.
- Activity: day headers with event counts; the index caps each day at 15 rows with an "and :count more" link to the day page (day page still shows everything).

Local verification used fabricated combat-log-derived demo data (the isolated worktree DB has none); the seed script is kept out of the diff. Note: `spells/ability_mage_invisibility.jpg` is missing from the asset set (pre-existing, visible on class pages as a broken icon) — not addressed here.

| | Before | After |
|---|---|---|
| Landing | before | after |
| NPC record | before | after |
| Spell record | before | after |

Activity index (after): activity

Theme sanity — lux: lux · vapor: vapor · mobile: mobile

Status: design-iteration draft — Wotuu is reviewing the direction first; cold review + full checklist follow once the direction is approved.

https://claude.ai/code/session_0111gac1BcpqYNzAMjvvqFA6

### DIFF
--- a/resources/views/common/spell/link.blade.php
+++ b/resources/views/common/spell/link.blade.php
@@ -8,6 +8,6 @@ use App\Models\Spell\Spell;
  */
 $size ??= 20;
 ?>
-<a href="{{ route('spell.compendium.show', $spell) }}"><img src="{{ $spell->icon_url }}"
+<a href="{{ route('spell.compendium.show', $spell) }}" data-wowhead="spell={{ $spell->id }}"><img src="{{ $spell->icon_url }}"
          width="{{ $size }}" height="{{ $size }}"
          class="me-1" loading="lazy" alt=""/>{{ __($spell->name) }}</a>
--- a/resources/views/common/spell/link.blade.php
+++ b/resources/views/common/spell/link.blade.php
@@ -8,6 +8,6 @@ use App\Models\Spell\Spell;
  */
 $size ??= 20;
 ?>
-<a href="{{ route('spell.compendium.show', $spell) }}" data-wowhead="spell={{ $spell->id }}" data-wh-iconize-link="false"><img src="{{ $spell->icon_url }}"
+<a href="{{ route('spell.compendium.show', $spell) }}" data-wowhead="{{ $spell->wowhead_tooltip_data }}" data-wh-iconize-link="false"><img src="{{ $spell->icon_url }}"
          width="{{ $size }}" height="{{ $size }}"
          class="me-1" loading="lazy" alt=""/>{{ __($spell->name) }}</a>
--- a/resources/views/common/spell/link.blade.php
+++ b/resources/views/common/spell/link.blade.php
@@ -8,6 +8,6 @@ use App\Models\Spell\Spell;
  */
 $size ??= 20;
 ?>
-<a href="{{ route('spell.compendium.show', $spell) }}" data-wowhead="spell={{ $spell->id }}"><img src="{{ $spell->icon_url }}"
+<a href="{{ route('spell.compendium.show', $spell) }}" data-wowhead="spell={{ $spell->id }}" data-wh-iconize-link="false"><img src="{{ $spell->icon_url }}"
          width="{{ $size }}" height="{{ $size }}"
          class="me-1" loading="lazy" alt=""/>{{ __($spell->name) }}</a>
--- a/resources/views/compendium/index.blade.php
+++ b/resources/views/compendium/index.blade.php
@@ -52,19 +52,6 @@ $howItWorks = [
 @endsection
 
 @section('content')
-    {{--
-    Direction contract (#3945, "The Field Record"):
-    THESIS: the Compendium is a field record - one continuous, dense record sheet per subject;
-      it refuses the icon-card dashboard arrangement.
-    OWN-WORLD: DESIGN.md's War Room (tonal ladder via theme variables, Keystone Green accent,
-      hairline registers, sturdy controls); no new palette or faces.
-    STORY: a player looks up a thing, reads its record, trusts the combat-log-backed data.
-    FIRST VIEWPORT: intro line, then a four-row directory register (icon plate / title+count /
-      description / go), data-source panel below.
-    FORM: field-record register, candidate 5 of the grounded list, seed key f2671375.
-    FINISH: unreviewed and undocumented is unfinished; this build ends with the finish review,
-      the verdict, and DESIGN.md.
-    --}}
     <div class="row justify-content-center">
         <div class="col-12 col-lg-8 text-center">
             <p class="lead">{{ __('view_compendium.index.intro') }}</p>
--- a/resources/views/compendium/index.blade.php
+++ b/resources/views/compendium/index.blade.php
@@ -4,7 +4,7 @@
  * @var array{npc: int, spell: int, class: int} $stats
  */
 
-$cards = [
+$sections = [
     [
         'icon'     => 'fa-dragon',
         'title'    => __('view_compendium.index.cards.npc.title'),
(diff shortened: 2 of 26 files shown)


Wotuu:
:robot: Fixed the double icon on spell links: power.js has \`iconizeLinks\` on site-wide, so every \`data-wowhead\` anchor got Wowhead's icon injected next to our own. Spell links now carry \`data-wh-iconize-link="false"\` — one icon again, tooltip verified still working on hover.

after fix

resources/views/common/spell/link.blade.php
```diff
@@ -8,6 +8,6 @@
  */
 $size ??= 20;
 ?>
-<a href="{{ route('spell.compendium.show', $spell) }}"><img src="{{ $spell->icon_url }}"
+<a href="{{ route('spell.compendium.show', $spell) }}" data-wowhead="spell={{ $spell->id }}" data-wh-iconize-link="false"><img src="{{ $spell->icon_url }}"
```
Wotuu: :robot: The `data-wowhead` payload drops the game version, so non-retail spells get the wrong tooltip.

`Spell::getWowheadLink()` (app/Models/Spell/Spell.php:228) deliberately branches on `game_version_id` and prefixes `/wrath`, `/classic` or `/mop-classic` — and the seed data really does contain non-retail spells (`database/seeders/dungeondata/spells.json`: 341 rows with `game_version_id = 2` / Classic Era). For those spells this page will show a *retail* tooltip on hover, while the link's own `href` and the "Wowhead" button rendered from `$spell->wowhead_url` in `compendium/spell/sections/header.blade.php` point at `/classic` — two different spells (or none) behind one row.

Suggest deriving the tooltip's domain from `game_version_id` the same way `getWowheadLink()` already does, ideally via a small accessor on `Spell` so the mapping lives in one place rather than being duplicated here and in the Handlebars template.

Same applies to `resources/assets/js/handlebars/spell_template.handlebars` — the DataTables payloads in `compendium/{npc,spell}/index.blade.php` would need to send the domain alongside `spell_id`.

resources/views/compendium/index.blade.php
```diff
@@ -52,73 +52,81 @@
...
+    Direction contract (#3945, "The Field Record"):
+    THESIS: the Compendium is a field record - one continuous, dense record sheet per subject;
+      it refuses the icon-card dashboard arrangement.
+    OWN-WORLD: DESIGN.md's War Room (tonal ladder via theme variables, Keystone Green accent,
+      hairline registers, sturdy controls); no new palette or faces.
+    STORY: a player looks up a thing, reads its record, trusts the combat-log-backed data.
+    FIRST VIEWPORT: intro line, then a four-row directory register (icon plate / title+count /
+      description / go), data-source panel below.
+    FORM: field-record register, candidate 5 of the grounded list, seed key f2671375.
+    FINISH: unreviewed and undocumented is unfinished; this build ends with the finish review,
```
Wotuu: :robot: This "Direction contract" block is session-process scaffolding, not view documentation, and it is now committed into a production Blade file — "candidate 5 of the grounded list, seed key f2671375", "FINISH: unreviewed and undocumented is unfinished; this build ends with the finish review, the verdict, and DESIGN.md".

None of that means anything to a future reader of `compendium/index.blade.php`, and the FINISH line is an instruction to the implementing session that will read as stale the moment this merges. The design rationale that *is* durable already lives in `DESIGN.md` and in the header comment of `resources/assets/css/sections/compendium.css`. Please drop this block (or keep at most a one-line pointer to DESIGN.md).

resources/views/compendium/sections/event_list.blade.php
```diff
@@ -154,45 +169,58 @@
...
-                        @endif
-                    </small>
-                </div>
-            </li>
+                    <span>{!! $eventDescription($event, $subjectHtml !== '') !!}</span>
+                </span>
+            </div>
         @endforeach
-    </ul>
+        @if($hiddenCount > 0 && $contextDungeon !== null && $date !== null)
```
Wotuu: :robot: Rows can be silently dropped: the truncation itself (`$displayedEvents = $events->take($limit)` above) is unconditional, but the "and :count more" escape hatch is gated on `$contextDungeon !== null && $date !== null`. A caller that passes `$limit` without a context dungeon/date therefore renders a short list with no indication that anything was cut.

No current caller hits it — `activity/index.blade.php` is the only one passing `limit` and it always supplies both — but this is a shared partial that now advertises `$limit` in its `@var` block, so the next caller (e.g. capping the NPC/spell detail event feeds, which pass neither) inherits a data-loss trap. Either fall back to a plain non-linked "and :count more" line, or only compute `$hiddenCount` when the link can actually be rendered so the list stays complete otherwise.

Related nit: `compendium/activity/sections/event_list.blade.php` now reads `$limit ?? null` but does not declare `@var int|null $limit` in its docblock, unlike this file.

resources/views/compendium/sections/event_list.blade.php
```diff
@@ -154,45 +169,58 @@
...
-                               class="text-muted">
-                                {{ $event->created_at->diffForHumans() }}
-                            </a>
-                        @else
-                            {{ $event->created_at->diffForHumans() }}
-                        @endif
-                    </small>
-                </div>
-            </li>
+                    <span>{!! $eventDescription($event, $subjectHtml !== '') !!}</span>
```
Wotuu, line 210: :robot: This line widens an XSS sink in a way worth a deliberate decision: the description is now emitted with `{!! !!}`, and while every interpolated *value* is correctly `e()`-escaped or a trusted rendered view, the **translation line itself** is now also unescaped output. Per CLAUDE.md only `lang/en_US` is edited in-repo — all other locales are managed externally — so a translated `view_compendium.event.*` string is now raw HTML on the page.

Low severity (translators are semi-trusted and the current strings are fine), and the design is defensible, but it is a real change in trust boundary that is not called out anywhere. A one-line note in the closure's docblock saying translation lines are trusted HTML here would at least make it a decision rather than an accident.

resources/views/compendium/sections/event_list.blade.php
```diff
@@ -82,9 +74,20 @@
...
 };
 
-$eventDescription = static function (CombatLogNpcEvent|CombatLogSpellEvent $event) use ($spellPropertyName): string {
+/**
+ * Returns HTML: spell names inside descriptions render as links to their compendium page, so
+ * every interpolated plain string is escaped here and the caller prints with {!! !!}.
+ * When the row already leads with the spell as its subject link ($subjectShown), the
+ * description switches to the subject-less wording instead of repeating the spell name.
+ */
+$eventDescription = static function (CombatLogNpcEvent|CombatLogSpellEvent $event, bool $subjectShown) use ($spellPropertyName): string {
```
Wotuu, line 85: :robot: No test coverage for the new logic in this file. Per CLAUDE.md ("Every change must be programmatically tested"), two behaviours added here deserve a feature test before this leaves draft:

1. The `$limit` path — that the activity index renders exactly 15 rows for a day with more events and emits the "and :count more" link to the day page, while the day page still renders all of them.
2. The `$subjectShown` switch — that a spell event on the activity page uses the `_no_subject` wording (e.g. `Added to database`) rather than repeating the spell name, and that a spell name inside an NPC `SpellAssigned` description renders as a link to `spell.compendium.show`.

`tests/Feature/Controller/Compendium/NpcCompendiumControllerActivityTest.php` is the natural home; it currently asserts nothing about rendered event text, so neither change would be caught if it regressed.

Wotuu:

resources/views/common/spell/link.blade.php
Wotuu: :robot: Fixed in 3a724455f: added `Spell::getWowheadTooltipData()` which appends `domain=wrath`/`classic`/`mop-classic` for non-retail game versions (matching `getWowheadLink()`'s URL prefixes and the vendored power.js domain resolver). Exposed as the appended `wowhead_tooltip_data` attribute so the blade partial and both index DataTables render the same value. Covered by a new test asserting `domain=classic` renders for a classic-era spell.

resources/views/compendium/index.blade.php
Wotuu: :robot: Fixed in 3a724455f: removed the direction-contract comment from the view. The design documentation lands in DESIGN.md as part of the finish pass once the direction is approved.

resources/views/compendium/sections/event_list.blade.php
Wotuu: :robot: Fixed in 3a724455f: the "and :count more" row now renders whenever the list is truncated — linked when the dungeon/date context is available, plain text otherwise — so a future caller passing `limit` alone can no longer silently drop rows. Also declared `$limit` in the activity wrapper docblock.

resources/views/compendium/sections/event_list.blade.php
Wotuu: :robot: Documented in 3a724455f: the `$eventDescription` docblock now states explicitly that the translation lines are part of the HTML trust boundary (en_US in-repo, other locales via the managed translation pipeline). Keeping the sink: the descriptions must embed spell links, and every interpolated value is escaped at the interpolation sites.

resources/views/compendium/sections/event_list.blade.php
Wotuu: :robot: Fixed in 3a724455f: two new tests in `NpcCompendiumControllerActivityTest` — one drives the activity index with 17 same-day events and asserts the 15-row cap collapses into the "and 2 more" day-page link; the other inserts a counter event for a classic-era spell and asserts the subject link, the game-version-aware `data-wowhead` payload, and the `_no_subject` wording switch (including that the subject-repeating variant does NOT render).

PR #3950, opened 2026-08-10
#3949 Fix InfluxDB port and php-cs-fixer traversal breaking fresh setups

### LINKED ISSUE #3949: Fix two setup-blocking defects found bringing up a fresh dev machine
:robot: Filed by Claude while repairing a development machine using the `new-machine-setup` skill.

## Background

Bringing up a dev environment from a clean state surfaced two defects in tracked files. Both are
invisible on an already-working machine, because a pre-existing `.env` and a warm toolchain mask
them — but both bite anyone setting up from scratch, and the second one breaks a workflow step
that `.claude/CLAUDE.md` mandates after *every* task.

## Defect 1 — `.env.docker.example` has the wrong InfluxDB port

`.env.docker.example` ships `INFLUXDB_PORT=3000`. The container listens on **8086** internally;
`docker-compose.yml` publishes it on the host as `8096` to avoid a conflict. Since
`INFLUXDB_HOST=influxdb` is the compose service name, the value must be the *internal* port —
neither the published `8096` nor `3000` (which looks like a Grafana port that crept in).

Consequence: **15 test failures** on a fresh setup, all identical:

```
FAILED Tests\Feature\Console\Commands\Scheduler\... Exception
cURL error 7: Failed to connect to influxdb port 3000 after 0 ms: Could not connect to server
  at vendor/influxdb/influxdb-php/src/InfluxDB/Database.php:177
  app/Console/Commands/Traits/SavesToInfluxDB.php:25
  app/Console/Commands/Scheduler/SchedulerCommand.php:36
```

Every `SchedulerCommand` subclass writes a metric through `SavesToInfluxDB`, so the whole
`Tests\Feature\Console\Commands\Scheduler` tree goes red. An older `.env` restored from a backup
has `8086` and is therefore *more* correct than the example it is supposed to be derived from,
which is why this has stayed hidden.

After the fix: `php artisan test --filter=Scheduler` -> 20 passed.

## Defect 2 — `composer run fix` is broken outright

```
In RecursiveDirectoryIterator.php line 111:
  RecursiveDirectoryIterator::__construct(/var/www/docker-compose/mysql-combatlog/#innodb_temp):
  Failed to open directory: Permission denied

Script php-cs-fixer fix handling the fix event returned with error code 1
```

The finder in `.php-cs-fixer.php` scans `__DIR__` and excludes only
`vendor`, `node_modules`, `storage`, `bootstrap/cache`, `resources/views`. It therefore descends
into `docker-compose/`, which holds the bind-mounted MySQL data dirs (`docker-compose/mysql`,
`docker-compose/mysql-combatlog`). Directories MySQL creates inside them — e.g. `#innodb_temp` —
are owned by the container's mysql uid with mode `0750`, so neither the host user nor the
container's `ksg` user can read them, and the traversal aborts the entire run.

This matters because `.claude/CLAUDE.md` requires `composer run fix` after every task, so on a
fresh machine that instruction cannot be followed at all.

**Non-obvious detail worth recording:** the intuitive fix, `->ignoreVCSIgnored(true)`, does
**not** work, even though the data dirs *are* gitignored (`.gitignore:41-45`). Symfony Finder
applies that as a *filter over results*, after `RecursiveDirectoryIterator` has already tried to
descend — so it still throws. Only `exclude()` prevents the descent. Verified empirically.

There is no PHP under `docker-compose/` (`git ls-files -- docker-compose | grep -c '\.php$'` -> 0),
so excluding the whole directory is safe.

After the fix: `Fixed 0 of 3886 files` — i.e. the repo had no style drift, the tool simply could
never get far enough to say so.

## Remaining work

1. `.env.docker.example`: `INFLUXDB_PORT=3000` -> `8086`, with a comment explaining internal vs
   published port so it does not regress.
2. `.php-cs-fixer.php`: add `docker-compose` to the finder's `exclude()`, with a comment recording
   the `ignoreVCSIgnored` dead end.
3. `.claude/skills/new-machine-setup/SKILL.md`: fold in both, plus the stale probes found while
   following it (see below).

## Also: stale probes in the `new-machine-setup` skill

The skill invites correcting it in-session when reality disagrees. Found while following it:

- Phase 6 probe checks `public/js/app.js` / `public/css/app.css`; the build emits
  version-suffixed names (`app-<sha>.js`), so the probe can never pass.
- Phase 7 probe uses `Dungeon::count()`; the model is not root-namespaced, so tinker errors with
  `Class "Dungeon" not found`. Needs `\App\Models\Dungeon`.
- Phase 9 tells you to use the `headless-browser-verify` skill, whose `chrome` service is defined
  in `docker-compose.worktree.yml` — not loaded by the main checkout's `docker compose`, so
  `--profile chrome up -d chrome` fails there with "no such service".
- `npm ci` dies in puppeteer's postinstall when the host lacks `unzip` (a phase-1 package), with
  an error that reads like a network problem.
- Nothing documented the case where `~/Git` / `~/Git/private` are root-owned: the assets clone
  fails with `Permission denied`, and because such clones are routinely piped (`git clone ... | tail`)
  the pipe's exit code makes it look like it succeeded. Docker then creates the missing sibling as
  a root-owned empty dir, which is the silently-broken-images failure the skill already warns about.
- ~10 `Unable to find enemy patrol <id> that is coupled to enemy <id>` console errors appear on
  some dungeon maps. These are stale cross-mapping-version references in the tracked seeder data
  (53 of 7319 patrol-carrying enemies; zero patrols missing outright) and are **not** a bad seed —
  worth documenting so the next person does not chase them.

## Definition of done

- `.env.docker.example` and `.php-cs-fixer.php` fixed, skill updated.
- `composer run fix` completes (expect `Fixed 0 of N files`).
- `composer run analyse` clean.
- Full suite green except `Tests\Unit\MapTiles\MapTilesExistenceTest`, which is the documented
  known-good failure when `keystone.guru.assets/tiles/` (~36G, gitignored, in no repo) has not been
  transferred to the machine.

Wotuu:
:robot: Closes #3949

Two defects in tracked files, found while bringing up a development machine from a clean state.
Both are invisible on an already-working machine — a pre-existing `.env` and a warm toolchain mask
them — and both bite anyone setting up from scratch.

## 1. `.env.docker.example` had the wrong InfluxDB port

`INFLUXDB_PORT=3000` -> `8086`.

The container listens on **8086** internally; `docker-compose.yml:376` publishes it on the host as
`8096` to avoid a conflict. `INFLUXDB_HOST=influxdb` is the compose service name, so the value must
be the *internal* port — neither the published `8096` nor `3000` (which looks like a Grafana port
that crept in).

Every `SchedulerCommand` subclass writes a metric via `SavesToInfluxDB`, so the wrong port took out
the whole `Tests\Feature\Console\Commands\Scheduler` tree — **15 failures**, all
`cURL error 7: Failed to connect to influxdb port 3000`.

A `.env` restored from an old backup has `8086` and is therefore *more* correct than the example it
is supposed to be derived from, which is why this stayed hidden.

## 2. `composer run fix` was broken outright

The finder in `.php-cs-fixer.php` scanned `__DIR__` while excluding only `vendor`, `node_modules`,
`storage`, `bootstrap/cache`, `resources/views` — so it descended into `docker-compose/`, which
holds the bind-mounted MySQL data dirs. Directories MySQL creates there (e.g.
`mysql-combatlog/#innodb_temp`) are owned by the container's mysql uid with mode `0750`, so the
traversal aborted the entire run:

```
RecursiveDirectoryIterator::__construct(/var/www/docker-compose/mysql-combatlog/#innodb_temp):
Failed to open directory: Permission denied
Script php-cs-fixer fix handling the fix event returned with error code 1
```

`.claude/CLAUDE.md` requires `composer run fix` after every task, so on a fresh machine that step
could not be performed at all.

**Worth knowing:** the intuitive fix, `->ignoreVCSIgnored(true)`, does **not** work — even though
the data dirs *are* gitignored (`.gitignore:41-45`). Symfony Finder applies it as a filter over
*results*, after `RecursiveDirectoryIterator` has already tried to descend, so it still throws. Only
`exclude()` prevents the descent. I tried `ignoreVCSIgnored` first and it failed; the comment in the
diff records that so the next person does not repeat it.

Excluding the whole directory is safe — there is no PHP under it:
`git ls-files -- docker-compose | grep -c '\.php$'` -> `0`.

## 3. `new-machine-setup` skill corrections

The skill asks to be corrected in-session when reality disagrees. Folded in both fixes above as
troubleshooting rows, plus probes that could not pass as written:

| Correction | Why it mattered |
|---|---|
| Assets are version-suffixed (`app-<sha>.js`), not `app.js` | Phase 6 probe could never pass |
| `\App\Models\Dungeon`, not `Dungeon` | Phase 7 tinker probe errored `Class "Dungeon" not found` |
| `chrome` service lives in `docker-compose.worktree.yml` | Not loaded by the main checkout, so `--profile chrome` fails there |
| `npm ci` needs host `unzip` | Puppeteer postinstall failure reads like a network problem |
| Root-owned `~/Git` breaks the assets clone | Piped clones (`\| tail`) hide the failure; Docker then creates a root-owned empty dir — the silently-broken-images trap the skill already warns about |
| `Unable to find enemy patrol` console errors are benign | Stale cross-mapping-version refs in tracked seeder data (53 of 7319); not a bad seed |

## Verification

Run in this worktree:

- `composer run fix` -> `Fixed 0 of 3886 files` (i.e. the repo had no drift; the tool simply could
  never get far enough to say so)
- `composer run analyse` -> `[OK] No errors`
- `php artisan test --filter=Scheduler` -> 19 passed, 1 skipped (was 15 failures)
- Full suite on the repaired machine -> **1968 passed**, 1 failed

The single remaining failure is `Tests\Unit\MapTiles\MapTilesExistenceTest`, the documented
known-good one: it asserts `keystone.guru.assets/tiles/` (~36G, gitignored, in no repo) exists on
disk. Unrelated to this change.

## Note on tests

No new automated test accompanies this. Both changes are to build/config files that the suite does
not load — `.env.docker.example` is a template consumed only by humans setting up a machine, and
`.php-cs-fixer.php` configures the formatter itself. The fixes are verified by running the actual
tools, as above. Happy to add a guard test if you would prefer one.

### DIFF
--- a/.claude/skills/new-machine-setup/SKILL.md
+++ b/.claude/skills/new-machine-setup/SKILL.md
@@ -153,6 +153,21 @@ git clone git@github.com:RaiderIO/keystone.guru.assets.git   # ~342M tracked (im
 `sh/worktree.sh` also expects `~/Git/private/keystone.guru-worktrees/` as its worktree root; it
 creates that itself on first use.
 
+**Check `~/Git` and `~/Git/private` are owned by you, not root.** If an earlier attempt cloned the
+repo under `sudo`, the parent dirs end up `root:root` while `keystone.guru` itself looks correctly
+owned — so nothing is obviously wrong until the second clone fails with
+`could not create work tree dir 'keystone.guru.assets': Permission denied`. That failure is easy to
+miss because it is routinely piped (`git clone … | tail`), and a pipe reports the *tail* exit code,
+so the clone reads as success. Worse, the missing sibling does not stay missing: the Docker daemon
+runs as root and will happily create `../keystone.guru.assets` as a root-owned empty dir for the
+bind mount, which is exactly the silent-broken-images failure above. Fix ownership before cloning:
+
+```bash
+sudo chown -R "$(id -u):$(id -g)" ~/Git
+# no sudo available? the daemon runs as root, so borrow it:
+docker run --rm -v ~/Git:/git alpine chown -R "$(id -u):$(id -g)" /git
+```
+
 The clone above only brings `images/` and `webfonts/`. The checkout also has a `tiles/` directory
 (~36G, gitignored per `keystone.guru.assets/.gitignore:1`) that is **not** part of the git history —
 map tiles have to be transferred separately (rsync/external drive/etc. from the old machine). See
@@ -309,10 +324,18 @@ nvm use && npm ci && npm run production
 `npm run production` can be memory-hungry; if WSL OOMs, `npm run development` produces a working
 (unminified) build and is fine for dev.
 
+**`npm ci` dies in puppeteer's postinstall if the host has no `unzip`** — the error is
+`Failed to set up chrome-headless-shell … no zip archiver is available`, which reads like a
+puppeteer/network problem rather than the missing phase-1 package it is. Install `unzip` (phase 1)
+and re-run. To unblock without sudo, `PUPPETEER_SKIP_DOWNLOAD=1 npm ci` installs everything else —
+asset builds do not need puppeteer's browser, and the in-container thumbnail path uses the worker
+image's own Chrome. Run `npx puppeteer browsers install` once `unzip` is available.
+
 **Probe:**
 ```bash
 [ -s version ] && echo "version: $(cat version)"
-ls public/js/app.js public/css/app.css        # both exist
+# Assets are version-suffixed from the git rev — there is no unsuffixed app.js/app.css
+ls public/js/app-$(cat version).js public/css/app-$(cat version).css
 docker compose exec -T app php artisan --version
 ```
 
@@ -387,7 +410,7 @@ provision, and tests run against this seeded data (see the `writing-tests` skill
 
 **Probe:**
 ```bash
-docker compose exec -T app php artisan tinker --execute 'echo User::count()." users, ".Dungeon::count()." dungeons";'
+docker compose exec -T app php artisan tinker --execute 'echo \App\Models\User::count()." users, ".\App\Models\Dungeon::count()." dungeons";'
 ```
 Expect 3 users and a few hundred dungeons. Zero dungeons = the seed did not finish.
 
@@ -416,9 +439,25 @@ MySQL 34006 / 34007.
 Do not report the machine as ready until all six pass. Fix failures here rather than noting them.
 
 1. **Site renders**: `curl -s http://localhost:8008 | grep -q "Keystone.guru"`.
-2. **A map renders**: use the `headless-browser-verify` skill against a dungeon explore page and
-   look at the screenshot with `Read`. Tiles must be visible (they come from the CDN), enemies must
-   be plotted. A blank grey map means `ASSETS_BASE_URL` is wrong.
+2. **A map renders**: use the `headless-browser-verify` skill against a dungeon explore page (e.g.
+   `/explore/retail/court-of-stars`) and look at the screenshot with `Read`. Tiles must be visible
+   (they come from the CDN), enemies must be plotted. A blank grey map means `ASSETS_BASE_URL` is
+   wrong.
+
+   That skill's `chrome` service is defined in `docker-compose.worktree.yml`, which the **main**
+   checkout's `docker compose` does not load — `--profile chrome up -d chrome` there fails with
+   "no such service". Run it standalone on the stack network instead (with
+   `ASSETS_BASE_URL` on the CDN the socat asset forward is unnecessary):
+
+   ```bash
+   docker run -d --name ksg-chrome --network keystoneguru_keystone.guru \
+       --network-alias chrome --shm-size 1gb chromedp/headless-shell:latest
+   ```
+
+   Expect ~10 `Unable to find enemy patrol <id> that is coupled to enemy <id>` console errors on
+   some dungeons. Those are stale cross-mapping-version references in the tracked seeder data
+   (~53 of 7319 patrol-carrying enemies; zero patrols missing outright), not a bad seed — the map
+   still plots every enemy. Do not chase them.
 3. **Login works**: `admin@app.com` / `password`.
 4. **Tests are green**:
    ```bash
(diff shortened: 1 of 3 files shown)


.claude/skills/new-machine-setup/SKILL.md
```diff
@@ -416,9 +439,25 @@ MySQL 34006 / 34007.
...
+   (they come from the CDN), enemies must be plotted. A blank grey map means `ASSETS_BASE_URL` is
+   wrong.
+
+   That skill's `chrome` service is defined in `docker-compose.worktree.yml`, which the **main**
+   checkout's `docker compose` does not load — `--profile chrome up -d chrome` there fails with
+   "no such service". Run it standalone on the stack network instead (with
+   `ASSETS_BASE_URL` on the CDN the socat asset forward is unnecessary):
+
+   ```bash
+   docker run -d --name ksg-chrome --network keystoneguru_keystone.guru \
```
Wotuu: :robot: Nit (instruction-file usability): this `docker run` creates a long-lived, fixed-name container with no teardown step, so the second time an agent walks phase 9 (or retries after a failed screenshot) it dies with `Conflict. The container name "/ksg-chrome" is already in use`. Suggest `docker run -d --rm ...` plus an explicit `docker rm -f ksg-chrome` line once the check passes, so the acceptance step is idempotent and does not leave a stray container on the stack network.

Also worth a half-sentence: `keystoneguru_keystone.guru` is the Compose-derived project network name for a checkout at `~/Git/private/keystone.guru` (verified on this machine). It is correct given phase 1 mandates that path, but a `docker network ls | grep keystone` hint would save the next person if the checkout ever lives elsewhere.

Wotuu:
:robot: Cold review (opus): 1 finding posted (a nit on the standalone chrome container in the setup skill).

Verified, all held up: `8086` is the container-internal influx port (`docker-compose.yml:376` publishes `8096:8086`, `INFLUXDB_HOST=influxdb` is the service name, `config/influxdb.php` defaults to 8086); no tracked PHP under `docker-compose/` and Symfony `Finder` really does apply `exclude()` before the `RecursiveIteratorIterator` (`vendor/symfony/finder/Finder.php:788/791`) while `VcsIgnoredFilterIterator` is added afterwards at line 826, so the `ignoreVCSIgnored` note is accurate; assets are version-suffixed for both `development` and `production` builds (`scripts/build/version.mjs`); `App\\Models\\Dungeon` correct; `chrome` exists only in `docker-compose.worktree.yml` and the main checkout sets no `COMPOSE_FILE`; `.gitignore:41-45` matches; `/explore/retail/court-of-stars` resolves (302 -> `/1`). The patrol claim checks out exactly: 7319 enemies with a patrol, 0 dangling, 53 cross-mapping-version.

.claude/skills/new-machine-setup/SKILL.md
Wotuu: :robot: Fixed in a55b6a24b.

Good catch — the snippet was written from a single run and never re-run, so the name conflict never showed up.

- `--rm` plus a preceding `docker rm -f ksg-chrome 2>/dev/null` makes it re-runnable.
- Added an explicit teardown line noting `docker compose down` will *not* clean it up, since the container is not part of the compose stack.
- The network name is now derived (`docker network ls | grep keystone`) rather than hardcoded, per your secondary note about it being Compose-project-derived.

Verified: two consecutive runs both succeed, and `docker ps -a --filter name=ksg-chrome` is empty after teardown.

Wotuu:
:robot: Ready for review.

- **Cold review**: 1 finding (nit on the standalone chrome container), fixed in `a55b6a24b`, thread resolved. Reviewer independently re-verified the InfluxDB port reasoning, the php-cs-fixer exclusion, the `ignoreVCSIgnored` claim at the Symfony Finder source level, and every factual claim in the skill edits.
- **Visual verification**: not applicable — config and documentation only, no rendered output changed.
- **CI**: 9/9 green on `a55b6a24b`.

Worktree left in place for teardown on merge.

PR #3953, opened 2026-08-10
#3935 Anchor Awakened Obelisk rift-skip enemy lookup to the icon's floor, not enemy_pack_id

### LINKED ISSUE #3935: Awakened Obelisk rift skips likely never import: enemy_pack_id filter excludes current mapping versions
:robot: Follow-up from #3915 / PR #3933's cold review.

`RiftOffsetImporter::parseRiftOffsets()` resolves an Awakened Obelisk skip's target enemy via:

```php
Enemy::where('npc_id', $npcId)
    ->where('mapping_version_id', $importStringRiftOffsets->getMappingVersion()->id)
    ->whereNotNull('enemy_pack_id')
    ->whereIn('floor_id', $floorIds)
    ->firstOrFail();
```

A scan of `database/seeders/dungeondata/bfa/*/*/enemies.json` for the 4 hardcoded obelisk NPC ids (161124/161241/161244/161243) shows packed (`enemy_pack_id` not null) rows only exist on **older** mapping versions of each dungeon — e.g. The Underrot: packed only at mv21/106, current mapping version (158) has unpacked rows only. Freehold: packed only at mv15/105, several newer mapping versions unpacked only.

This means `whereNotNull('enemy_pack_id')` matches nothing for essentially every dungeon's *current* mapping version, so Awakened Obelisk rift-offset skips likely never import successfully today — PR #3933 turns the resulting crash into a silent skip-with-warning, which is strictly better, but doesn't address the underlying "does this feature ever actually work" question.

Worth checking whether `enemy_pack_id` is even needed here: the code only reads `$enemy->floor_id`/`$enemy->floor` afterward, so pack membership may not be load-bearing at all. Note `MDTExportStringService` comments this feature as "Legacy" for at least the export direction, so it's possible this is a deliberately deprioritized/dying feature rather than a bug worth fixing — needs a maintainer call on whether to invest here.

Wotuu:
Closes #3935

:robot: `RiftOffsetImporter::parseRiftOffsets()` resolved an Awakened Obelisk skip's target Enemy via `whereNotNull('enemy_pack_id')`. Per #3935's investigation, packed rows for these 4 hardcoded obelisk NPC ids only exist on **older** mapping versions of most BFA dungeons - so this lookup silently failed (as a warning, since #3933) for essentially every dungeon's current mapping version.

## Investigated: is `enemy_pack_id` even load-bearing?

No - only `$enemy->floor_id`/`$enemy->floor` are read afterward. But naively dropping the filter isn't safe either: some dungeons (Tol Dagor, Shrine of the Storm) place the same obelisk NPC on *multiple floors*, so an unqualified lookup can land on the wrong one.

**Fix**: anchor the lookup to the obelisk `MapIcon`'s own floor (`$obeliskMapIcon->floor_id`) instead of filtering by pack membership. This is the correct disambiguator regardless of pack data: the NPC being looked up *is* the obelisk the icon represents, and the icon's floor is exactly where the gateway line needs to land.

## Blast radius (swept all `expansion_id=2` dungeons with an obelisk MapIcon seeded, all 4 NPC ids, current mapping version, resolving the icon exactly as the importer does including the `seasonal_index` predicate)

Cold review caught that my first pass at this table was wrong - the packed-row disambiguation in Tol Dagor and Shrine of the Storm was **not** correct pre-fix, it just happened to not throw:

| Dungeon | Before (packed-row filter) | After (icon-floor anchor) |
|---|---|---|
| freehold | FAIL (no packed row) | resolves on floor 43 (fixed) |
| theunderrot | FAIL (no packed row) | resolves on floor 51 (fixed) |
| shrineofthestorm | resolved, but on floor 46 (wrong - icon is on floor 45) | resolves on floor 45 (icon's floor - **also fixed**, not just preserved) |
| toldagor | resolved, but on floor 59 for all 4 obelisks (wrong - icons are on floors 53/58/55/57) | resolves each obelisk on its own icon's floor (**also fixed**) |
| ataldazar, kingsrest, siegeofboralus, templeofsethraliss, themotherlode, waycrestmanor, mechagonworkshop | FAIL (NPC not seeded on this mapping version at all) | still FAIL - unchanged, out of scope (needs seeder data, not a query fix) |

So this fix repairs the gateway floor for **4 dungeons outright** (freehold, theunderrot newly resolve; shrineofthestorm, toldagor now land on the correct floor instead of a wrong one that happened not to throw) and leaves ~7 dungeons unchanged where the NPC simply isn't seeded on the current mapping version at all - flagging per the issue's "needs a maintainer call" note, since fixing those needs seeder/mapping data, not a code change. Given the fix here is small and strictly improves correctness, it seemed worth landing regardless of that remaining gap.

One dependency worth noting: the obelisk `MapIcon` lookup above isn't scoped to `mapping_version_id`, so it resolves the *oldest* mapping version's icon (e.g. Underrot picks the icon from mapping version 21 while the current version is 158). This was harmless before since pack membership drove enemy resolution; now the icon's floor does. It holds on today's seeded data (verified: no dungeon regresses from resolving to failing), but rests on obelisk icons never having moved floors across a dungeon's mapping versions.

Also per the issue: `MDTExportStringService` comments this feature as "Legacy" (export direction only) - this fix doesn't touch export.

## Test plan
- [x] `RiftOffsetImporterTest`: added coverage for (a) unpacked-only resolution (The Underrot), (b) multi-floor disambiguation via the obelisk icon's floor, mirroring the importer's `seasonal_index` predicate and pinning that the floor-unfiltered lookup would otherwise disagree (Tol Dagor), (c) genuine absence still warns instead of throwing (Atal'Dazar, replacing the old pinned-negative-premise dungeon since it now resolves)
- [x] `php artisan test --group=MDT` - 232 passed

### DIFF
--- a/app/Service/MDT/Import/RiftOffsetImporter.php
+++ b/app/Service/MDT/Import/RiftOffsetImporter.php
@@ -77,12 +77,16 @@ class RiftOffsetImporter
             $obeliskMapIcon = $npcIdToMapIconMapping[$npcId];
 
             try {
-                // Find out the floor where the NPC is standing on
+                // Find out the floor where the NPC is standing on. Anchored to the obelisk map
+                // icon's own floor rather than filtered by enemy_pack_id: some dungeons only carry
+                // an unpacked Enemy row for this NPC on their current mapping version (packed rows
+                // exist only on older ones - #3935), and the same npc_id can additionally be placed
+                // on multiple floors of the same dungeon, so the obelisk icon's floor is the only
+                // reliable disambiguator between those duplicates.
                 /** @var Enemy $enemy */
                 $enemy = Enemy::where('npc_id', $npcId)
                     ->where('mapping_version_id', $importStringRiftOffsets->getMappingVersion()->id)
-                    ->whereNotNull('enemy_pack_id')
-                    ->whereIn('floor_id', $floorIds)
+                    ->where('floor_id', $obeliskMapIcon->floor_id)
                     ->firstOrFail();
 
                 if (isset($mdtXy['sublevel'])) {
@@ -135,10 +139,10 @@ class RiftOffsetImporter
             } catch (ImportWarning $warning) {
                 $importStringRiftOffsets->getWarnings()->add($warning);
             } catch (ModelNotFoundException) {
-                // No enemy matching this NPC (packed, on the mapping version's floors) resolved -
-                // whether because the clone genuinely isn't in this mapping version, or (the common
-                // case seen while triaging #3915) it exists but with no enemy_pack_id on the current
-                // mapping version. Skip this one obelisk skip rather than 500ing the whole import.
+                // No enemy matching this NPC on the obelisk icon's floor resolved for this mapping
+                // version - the clone genuinely isn't seeded there (e.g. #3935: several BFA
+                // dungeons never carry this NPC at all). Skip this one obelisk skip rather than
+                // 500ing the whole import.
                 $importStringRiftOffsets->getWarnings()->add(new ImportWarning(
                     __('services.mdt.io.import_string.category.awakened_obelisks'),
                     __(
--- a/tests/Feature/App/Service/MDT/Import/RiftOffsetImporterTest.php
+++ b/tests/Feature/App/Service/MDT/Import/RiftOffsetImporterTest.php
@@ -4,6 +4,8 @@ namespace Tests\Feature\App\Service\MDT\Import;
 
 use App\Models\Dungeon;
 use App\Models\Enemy;
+use App\Models\MapIcon;
+use App\Models\MapIconType;
 use App\Service\MDT\Import\RiftOffsetImporter;
 use App\Service\MDT\Models\ImportStringRiftOffsets;
 use Illuminate\Support\Collection;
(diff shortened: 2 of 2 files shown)


app/Service/MDT/Import/RiftOffsetImporter.php
```diff
@@ -77,12 +77,16 @@ public function parseRiftOffsets(ImportStringRiftOffsets $importStringRiftOffset
...
+                // an unpacked Enemy row for this NPC on their current mapping version (packed rows
+                // exist only on older ones - #3935), and the same npc_id can additionally be placed
+                // on multiple floors of the same dungeon, so the obelisk icon's floor is the only
+                // reliable disambiguator between those duplicates.
                 /** @var Enemy $enemy */
                 $enemy = Enemy::where('npc_id', $npcId)
                     ->where('mapping_version_id', $importStringRiftOffsets->getMappingVersion()->id)
-                    ->whereNotNull('enemy_pack_id')
-                    ->whereIn('floor_id', $floorIds)
+                    ->where('floor_id', $obeliskMapIcon->floor_id)
```
Wotuu, line 89: :robot: The code change looks right to me, but the PR body's blast-radius table is empirically wrong about the two dungeons it lists as unchanged. I swept all `expansion_id=2` dungeons on the seeded DB, resolving the obelisk `MapIcon` exactly as the importer does (including the `seasonal_index` predicate) and comparing the old `whereNotNull('enemy_pack_id')` result against the new `where('floor_id', $obeliskMapIcon->floor_id)`:

```
freehold          npc161124/161241/161244/161243  icon f43  before=FAIL after=43
theunderrot       npc161124/161241/161244/161243  icon f51  before=FAIL after=51
shrineofthestorm  npc161124/161241/161244/161243  icon f45  before=46   after=45
toldagor          npc161124 icon f53  before=59 after=53
toldagor          npc161241 icon f58  before=59 after=58
toldagor          npc161244 icon f55  before=59 after=55
toldagor          npc161243 icon f57  before=59 after=57
```

So Tol Dagor and Shrine of the Storm are *not* preserved: before this change all four Tol Dagor obelisks resolved to floor 59 (the packed clones) while their icons sit on floors 53/58/55/57, and Shrine went 46 -> 45. The body's "the packed row happened to be the one that disambiguates correctly there ... this fix preserves that correct disambiguation" doesn't hold — the packed rows were the *wrong* floor in both dungeons.

That doesn't change my read that the new behaviour is correct: the NPC being looked up (161124 = Urg'roth, Brutal Spire) *is* the obelisk the `MapIcon` represents, `$enemy->floor` is only used to convert the MDT destination coordinate, and MDT gives that coordinate on the obelisk's own sublevel unless `sublevel` is set (already warned about above). Pre-fix, the gateway icon and `Path` landed on floor 59/46 while the polyline's first vertex was the icon's lat/lng from a different floor — visibly broken. So this is a fix for 4 dungeons, not 2.

Please correct the table and the "preserves that correct disambiguation" sentence — the body is the record of what shipped, and "unchanged" vs "gateway now lands on a different floor" is a meaningful difference for anyone bisecting a later report.

Context worth one line: the `MapIcon` lookup above is not scoped to `mapping_version_id`, so `firstOrFail()` returns the *oldest* mapping version's icon (Underrot picks the mv21 icon while the current mapping version is 158; Tol Dagor picks mv22 icons). That was harmless before, but the icon's floor now drives enemy resolution, so this fix's correctness rests on obelisk icons never having moved floors between mapping versions. It holds on today's seeded data (my sweep found no before-success -> after-FAIL), and scoping the lookup would regress dungeons whose icon only exists on an old version (Atal'Dazar: mv14 only -> whole-import warning), so I'm not asking for a change — just noting the new dependency.

tests/Feature/App/Service/MDT/Import/RiftOffsetImporterTest.php
```diff
@@ -15,24 +17,139 @@
...
+    {
+        // Arrange
+        $dungeon        = Dungeon::where('key', self::DUNGEON_KEY_MULTI_FLOOR)->firstOrFail();
+        $mappingVersion = $dungeon->getCurrentMappingVersion();
+        $floorIds       = $dungeon->floors->pluck('id');
+
+        $obeliskMapIcon = MapIcon::where(
+            'map_icon_type_id',
+            MapIconType::ALL[MapIconType::MAP_ICON_TYPE_AWAKENED_OBELISK_BRUTAL],
+        )->whereIn('floor_id', $floorIds)->firstOrFail();
```
Wotuu: :robot: This lookup doesn't replicate the importer's `seasonal_index` predicate, and has no deterministic ordering — so it agrees with what the importer actually resolves only by row order.

Tol Dagor has two Brutal obelisk icons on the current data: `id=1997299 floor=53 seasonal_index=1` and `id=1997309 floor=58 seasonal_index=0`. The importer builds its mapping with `->where(fn($q) => $q->whereNull('seasonal_index')->orWhere('seasonal_index', $seasonalIndex ?? 1))`, so with `seasonalIndex: null` it deterministically picks the `si=1` icon on floor 53. The test's unfiltered `->first()` today returns 1997299 as well — but only because that's the lower id. If the seeder emits them in the other order, `$obeliskMapIcon->floor_id` becomes 58 while the importer resolves 53, and the test fails spuriously on a change that isn't a regression. Suggest mirroring the same `seasonal_index` closure here (or at least an explicit `orderBy('id')` plus a comment saying why floor 53 is the expected one).

Second, smaller point on the premise pin below: `assertGreaterThan(1, $distinctFloors->count())` pins that the NPC is on multiple floors, but not the property that actually makes this test non-vacuous — that the row a *floor-unfiltered* query would return first (`id=4195`, floor 58) is on a different floor than the icon (53). As written, a seeder id shift that made the floor-53 row the first one would let this test pass even with the floor anchor removed from the importer. Asserting that explicitly would keep the guard honest.

Wotuu:

app/Service/MDT/Import/RiftOffsetImporter.php
Wotuu: :robot: Fixed — corrected the blast-radius table and the "preserves that correct disambiguation" sentence to reflect that Tol Dagor and Shrine of the Storm were also broken pre-fix (wrong floor, not a throw), so this fix repairs 4 dungeons, not 2. Also kept the note about the un-scoped MapIcon lookup as informational context, per your suggestion.

tests/Feature/App/Service/MDT/Import/RiftOffsetImporterTest.php
Wotuu: :robot: Fixed — the lookup now mirrors the importer's seasonal_index predicate exactly (whereNull or =1), so it deterministically picks the same icon (floor 53) regardless of row order. Also added an explicit assertion that a floor-unfiltered lookup disagrees with the icon's floor, pinning the property that makes this test non-vacuous.

PR #3954, opened 2026-08-10
Fix multi-select dungeon picker truncating preselected values to one

### LINKED ISSUE #3952: Multi-select dungeon picker only keeps the first selected option
## Background

Found while investigating why `/admin/season/18` only showed Kings' Rest as enabled, even though
`season_dungeons` correctly links all 8 dungeons for that season (verified via tinker — all rows
present, all dungeons and their expansions active).

The DB is correct. The bug is in how `resources/views/common/dungeon/select.blade.php` builds the
`<select multiple>` element.

## Root cause

```php
html()->select($name, $dungeonsSelect, $selected)->attributes(array_merge(['id' => $id],
    $multiple ? ['multiple' => 'multiple'] : [], [...]))
```

`Html::select($name, $options, $value)` (spatie/laravel-html) internally calls
`->options($options)->value($value)` *before* this blade ever chains `->attributes(...)`. Calling
`->value()` triggers `Select::applyValueToOptions()`:

```php
protected function applyValueToOptions()
{
    $value = Collection::make($this->value);

    if (! $this->hasAttribute('multiple')) {
        $value = $value->take(1);
    }

    $this->children = $this->applyValueToElements($value, $this->children);
}
```

At that point the `multiple` HTML attribute hasn't been attached yet (it's added afterward via
`->attributes([...])`, a plain attribute setter — not the builder's dedicated `->multiple()`
method, which also re-triggers `applyValueToOptions()`). So `hasAttribute('multiple')` is `false`,
and the value collection is truncated to `->take(1)` — only the *first* selected id survives, no
matter how many were passed in.

Confirmed directly in tinker:

```php
$html->select("test[]", [1=>"one",2=>"two",3=>"three"], [1,2,3])->attributes(["multiple"=>"multiple"])->render();
// => only option 1 gets `selected`, even though multiple="multiple" is present in the markup

$html->select("test[]", [1=>"one",2=>"two",3=>"three"], [1,2,3])->multiple()->render();
// => all three options correctly get `selected`
```

## Fix

In `resources/views/common/dungeon/select.blade.php`, stop passing `multiple` as a plain attribute
and instead call the builder's `->multiple()` method (conditionally on the existing `$multiple`
blade variable) so `applyValueToOptions()` re-runs once the `multiple` attribute is actually set.
`->multiple()` already guards against double-appending `[]` to a name that ends with `[]`, so it's
safe to use with this partial's existing `dungeon_ids[]` naming.

## Blast radius

Any admin (or other) page using `common.dungeon.select` with `multiple => true` is affected —
season edit (`/admin/season/{id}`) is the one that surfaced it. Search for other
`multiple => true` usages of this partial as part of the fix.

## Definition of done

- `/admin/season/{id}` edit page shows every dungeon actually linked to the season as selected in
  the multi-select.
- Regression test covering the fix (a feature test hitting the season edit page, or a focused test
  around the select partial/builder usage).

Wotuu:
:robot: Closes #3952

## Summary
- The season/NPC admin edit pages' dungeon multi-select only ever showed the first linked
  dungeon as selected — `season_dungeons`/`npc_dungeons` data was correct, the rendering was not.
- `spatie/laravel-html`'s `Select::applyValueToOptions()` truncates the preselected values to
  `->take(1)` when the `multiple` HTML attribute isn't present yet. `resources/views/common/dungeon/select.blade.php`
  attached `multiple` via a plain `->attributes([...])` call *after* `html()->select()` had already
  applied the value internally, so the attribute wasn't there yet and every id after the first was
  dropped.
- Fix: use the builder's own `->multiple()` method (conditionally, via `->when($multiple, ...)`),
  which re-runs `applyValueToOptions()` once the attribute is actually set.

## Test plan
- [x] New regression test `edit_givenSeasonWithMultipleDungeons_preselectsAllOfThem` — confirmed it
      fails against the pre-fix blade (reverted the file, saw the assertion fail on the second
      dungeon id) and passes with the fix.
- [x] `AdminSeasonControllerTest` + `AdminToolsNpcControllerTest` green.
- [x] `composer run fix` clean, `composer run analyse` clean.

### DIFF
--- a/tests/Feature/Controller/Admin/AdminSeasonControllerTest.php
+++ b/tests/Feature/Controller/Admin/AdminSeasonControllerTest.php
@@ -168,6 +168,41 @@ final class AdminSeasonControllerTest extends PublicTestCase
     }
 
     #[Test]
+    public function edit_givenSeasonWithMultipleDungeons_preselectsAllOfThem(): void
+    {
+        // Arrange - a regression test for a real bug: the multi-select builder truncated the
+        // preselected values down to the first one whenever `multiple` was applied as a plain
+        // HTML attribute instead of via the builder's dedicated `multiple()` method.
+        $expansion = Expansion::query()->firstOrFail();
+        $dungeons  = Dungeon::query()->limit(3)->get();
+        $season    = Season::create([
+            'expansion_id'            => $expansion->id,
+            'index'                   => 987,
+            'start'                   => '2026-01-01 00:00:00',
+            'presets'                 => 0,
+            'affix_group_count'       => 4,
+            'start_affix_group_index' => 0,
+            'key_level_min'           => 2,
+            'key_level_max'           => 30,
+        ]);
+        $season->syncDungeons($dungeons->pluck('id')->toArray());
+
+        try {
+            // Act
+            $response = $this->get(route('admin.season.edit', $season));
+
+            // Assert
+            $response->assertOk();
+            foreach ($dungeons as $dungeon) {
+                $response->assertSee(sprintf('value="%d" selected="selected"', $dungeon->id), false);
+            }
+        } finally {
+            $season->syncDungeons([]);
+            $season->delete();
+        }
+    }
+
+    #[Test]
     public function update_givenValidDataWithDungeons_syncsDungeonsAndDefaultsPresetsAndSeasonalAffixId(): void
     {
         // Arrange
--- a/tests/Feature/Controller/Admin/AdminSeasonControllerTest.php
+++ b/tests/Feature/Controller/Admin/AdminSeasonControllerTest.php
@@ -174,7 +174,7 @@ final class AdminSeasonControllerTest extends PublicTestCase
         // preselected values down to the first one whenever `multiple` was applied as a plain
         // HTML attribute instead of via the builder's dedicated `multiple()` method.
         $expansion = Expansion::query()->firstOrFail();
-        $dungeons  = Dungeon::query()->limit(3)->get();
+        $dungeons  = Dungeon::query()->orderBy('id')->limit(3)->get();
         $season    = Season::create([
             'expansion_id'            => $expansion->id,
             'index'                   => 987,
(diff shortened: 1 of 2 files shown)


tests/Feature/Controller/Admin/AdminSeasonControllerTest.php
```diff
@@ -167,6 +167,41 @@ public function update_givenRealSeasonalAffixId_persistsAndPreselectsIt(): void
...
+        $season->syncDungeons($dungeons->pluck('id')->toArray());
+
+        try {
+            // Act
+            $response = $this->get(route('admin.season.edit', $season));
+
+            // Assert
+            $response->assertOk();
+            foreach ($dungeons as $dungeon) {
+                $response->assertSee(sprintf('value="%d" selected="selected"', $dungeon->id), false);
```
Wotuu: :robot: The first iteration of this loop is vacuous. `Expansion::query()->firstOrFail()` returns id `1` and `Dungeon::query()->limit(3)` returns ids `[1, 2, 3]` on the seeded DB (verified via tinker), and the season edit page also renders an `expansion_id` select where the season's expansion is marked selected — so `value="1" selected="selected"` is already present in the response regardless of what the dungeon multi-select does. The test still catches the regression via ids 2 and 3, so it isn't broken, but the assertion isn't actually scoped to the element under test.

Cheap fix — include the option text so it can't collide with any other select on the page:

```php
$response->assertSee(sprintf('value="%d" selected="selected">%s<', $dungeon->id, __($dungeon->name)), false);
```

Secondary, same theme: `Dungeon::query()->limit(3)->get()` has no `orderBy`, so which three dungeons the test picks is up to MySQL. Not a practical problem today (and the sibling `update_givenValidDataWithDungeons_...` does the same), but `->orderBy('id')` would make it deterministic.

Wotuu:

tests/Feature/Controller/Admin/AdminSeasonControllerTest.php
Wotuu: :robot: Fixed in 179af884e — assertion now checks `value="%d" selected="selected">%s` (value + option text together) and the dungeon query got `->orderBy('id')` for determinism.

Wotuu:
:robot: Cold review: 1 finding

Issue #3940, opened 2026-08-09
6 pre-existing Ajax test failures (Arrow/EnemyForcesCheckpoint/Path) fail locally on master


Wotuu:
:robot: Found while baselining the full test suite for #3939 (MDT alpha5 update) — these 6
failures are **not** related to that work. Confirmed they fail identically on a clean `master`
checkout with no code changes, so this is pre-existing breakage in the current local/Docker test
environment, not a regression from any recent change.

Filter to reproduce:
```
docker compose exec -T app php artisan test --compact --filter="AjaxArrowControllerTest|AjaxEnemyForcesCheckpointControllerTest|AjaxPathControllerTest"
```
Result: `6 failed, 10 passed`.

## 1. `AjaxEnemyForcesCheckpointControllerTest` (4 failures)

All fail with:
```
Illuminate\Broadcasting\BroadcastException: Pusher error: cURL error 7: Failed to connect to
127.0.0.1 port 6001 ... for http://127.0.0.1:6001/apps/.../events
```

`phpunit.xml` forces `REVERB_HOST=http://127.0.0.1` (`REVERB_HOST` env, `force="true"`, around
line 41) with a comment explaining `ReverbHttpApiService` builds its Guzzle client in its
constructor from this value. That works when Reverb runs in the same network namespace as the
app process, but in the local Docker Compose setup `reverb` is a separate container
(`keystone.guru-reverb`), not reachable at `127.0.0.1` from inside the `app` container — only via
the `reverb` service hostname. This may be fine in CI (if the runner uses host networking or a
sidecar) but breaks every local `php artisan test` run that exercises broadcasting through
`AjaxEnemyForcesCheckpointController`.

## 2. `AjaxArrowControllerTest > delete` and `AjaxPathControllerTest > store` (2 failures)

Both fail with a 404 where a 204/201 is expected:
```
AjaxArrowControllerTest: Expected response status code [204] but received 404.
  at tests/Feature/Controller/Ajax/AjaxArrowControllerTest.php:132

AjaxPathControllerTest: Expected response status code [201] but received 404.
  at tests/Feature/Controller/Ajax/AjaxPathControllerTest.php:37
```
No exception/broadcast error in these two — looks like a genuine routing/binding issue
independent of the Reverb problem above. Needs separate investigation (route registration,
model binding, or a stale route cache are the likely suspects).

## Suggested next steps
[...]

PR #3955, opened 2026-08-10
#3941 Wrap every controller broadcast() failure in its own narrow catch

### LINKED ISSUE #3941: Ajax broadcast() failures are inconsistently swallowed across ~20 controllers (follow-up to #3940)
:robot: Found while fixing #3940 (6 pre-existing Ajax test failures caused by an unreachable
Reverb host in local Docker test runs).

## Background

Fixing #3940 required forcing `BROADCAST_CONNECTION=log` in `phpunit.xml` so `broadcast()` calls
in tests stop making a real HTTP call to the (intentionally unreachable) `REVERB_HOST` pinned
there. While diagnosing the two "distinct root cause" 404s the issue asked to investigate
separately (`AjaxArrowControllerTest > delete`, `AjaxPathControllerTest > store`), it turned out
they were **not** a distinct root cause at all - they were the *same* Reverb-connectivity
`BroadcastException`, just caught by a broader `catch (\Exception)` block that turns any failure
(including a successful mutation whose broadcast afterward failed) into a generic 404.

This exposed an inconsistent, and in places incorrect, pattern for handling broadcast failures
across the Ajax controllers:

1. **Correct pattern** - `AjaxArrowController::store()` (`app/Http/Controllers/Ajax/AjaxArrowController.php:92-94`)
   wraps only the `broadcast()` call in `catch (BroadcastException)` and ignores it, so a
   broadcast hiccup doesn't affect the response.
2. **Broad-catch pattern (bug)** - `AjaxArrowController::delete()` (line 127) and
   `AjaxPathController::store()`/`delete()` (`app/Http/Controllers/Ajax/AjaxPathController.php:98,130`)
   call `broadcast()` inside a `try` block whose `catch (Exception)` covers the whole method body.
   A transient broadcast failure here doesn't just get ignored - the delete/update has *already
   succeeded* (in `delete()`'s case, the row really is gone; in `store()`'s case the DB
   transaction hasn't committed yet, so at least that one rolls back), but the client is told
   `404 not found`, which is misleading and, for `delete()`, describes a state that isn't true.
3. **No catch at all** - `AjaxMappingModelBaseController::storeModel()` (`app/Http/Controllers/Ajax/AjaxMappingModelBaseController.php:96`)
   and `AjaxEnemyController::delete()` (`app/Http/Controllers/Ajax/AjaxEnemyController.php:195`)
   call `broadcast()` with no surrounding catch for it specifically - a `BroadcastException` here
   propagates as an unhandled exception (500, or in one of #3940's 4 failing tests, straight
   through to PHPUnit as a `BroadcastException`).

There are ~20+ controllers with this `broadcast(new ...)` shape (see
`grep -rn "broadcast(new" app/Http/Controllers/`), so this wasn't fixed as part of #3940 to keep
that MR scoped to the test-infra problem it was filed for.

## Remaining work

Audit every `broadcast(new ...Event(...))` call site under `app/Http/Controllers/` and make sure
each one follows pattern 1 above: wrap only the `broadcast()` call itself in
`catch (\Illuminate\Broadcasting\BroadcastException)` (ignore it - the mutation already
succeeded/is about to commit and shouldn't be reported as failed because a websocket push
couldn't be delivered), never inside a `catch (\Exception)` that also covers unrelated failure
modes, and never left uncaught.

- This is a real, if narrow, behavioral bug: a **successful** delete/update currently gets
  reported as `404`/`500` to the client if Reverb is briefly unreachable, which could cause the
  map editor's frontend to retry or show a stale/incorrect state even though the backend mutation
  went through.
- Some of the "broad catch" blocks are wrapping actual DB work too (e.g. `AjaxPathController::store()`
  runs the whole thing inside `DB::transaction()`) - narrowing the catch must not accidentally stop
  catching the *real* failure modes those blocks exist for (failed validation cascades, DB errors,
  etc.). Read each method's `try` block fully before narrowing it.
- Touches ~20 files - do this as a careful, mechanical sweep with a test added per representative
  case (force a `BroadcastException` the way `AjaxPathControllerTest::store_givenUpdateThatFailsToSave_shouldReturnNotFoundAndRollBackTransaction`
  already forces `Path::update()` to fail, and assert the success response still comes back).

## How to measure / verify

- `docker compose exec -T app php artisan test --compact tests/Feature/Controller/Ajax` after the
  fix, plus new regression test(s) for at least one `delete()` and one `store()` broadcast-failure
  case.

## Key files

- `app/Http/Controllers/Ajax/AjaxArrowController.php`
- `app/Http/Controllers/Ajax/AjaxPathController.php`
- `app/Http/Controllers/Ajax/AjaxMappingModelBaseController.php`
- `app/Http/Controllers/Ajax/AjaxEnemyController.php`
- every other controller under `app/Http/Controllers/` matching `broadcast(new `

## Definition of done

- Every `broadcast()` call site in a controller is wrapped in its own narrow
  `catch (BroadcastException)`, not a broader `catch (Exception)`.
- At least one regression test per pattern (broad-catch and no-catch) proving a broadcast failure
  no longer turns a successful mutation into an error response.
- Full test suite green (aside from the known, unrelated `MapTilesExistenceTest` worktree-env
  failure).

Wotuu:
:robot: Closes #3941

## Summary

Follow-up to #3940/#3942, which surfaced that `broadcast(new ...Event(...))` failure handling was
inconsistent across ~20 controllers:

1. **Broad-catch (bug)** — `broadcast()` called inside a `try` whose `catch (\Exception)` also
   covers the surrounding mutation. A transient Reverb blip turned an already-successful
   delete/update into a misleading 404 (or, for `store()` inside a `DB::transaction()`, actually
   rolled back the just-saved row).
2. **No catch at all (bug)** — `broadcast()` had no surrounding catch, so a `BroadcastException`
   propagated as an unhandled 500.

Every site is now narrowed to wrap only the `broadcast()` call itself in
`catch (\Illuminate\Broadcasting\BroadcastException)`, matching the one controller
(`AjaxArrowController::store()`) that already did this correctly — including the shared
`AjaxMappingModelBaseController::storeModel()` used by 6 of the `store()` methods.

Note on production impact (raised in cold review): `BroadcastManager::queue()` only calls the
broadcaster inline for `ShouldBroadcastNow` events; every event here is a plain `ShouldBroadcast`,
so in production (`QUEUE_CONNECTION=redis`) `broadcast()` just pushes a job and the actual Reverb
call/`BroadcastException` happens in the Horizon worker, not in the web request. The bug patterns
above are real and this fix is what #3941 asked for, but they can only bite a live request under
`QUEUE_CONNECTION=sync` — today that's `phpunit.xml`. The immediate payoff is correctness in tests
(and #3940's flakiness) plus future-proofing; it doesn't currently prevent a misleading 404/500 in
a production request.

## Files touched

- `app/Http/Controllers/Ajax/AjaxMappingModelBaseController.php` (shared `storeModel()`)
- `app/Http/Controllers/Ajax/AjaxArrowController.php`, `AjaxBrushlineController.php`,
  `AjaxDungeonFloorSwitchMarkerController.php`, `AjaxEnemyController.php`,
  `AjaxEnemyForcesCheckpointController.php`, `AjaxEnemyPackController.php`,
  `AjaxEnemyPatrolController.php`, `AjaxKillZoneController.php` (4 sites),
  `AjaxLiveSessionController.php`, `AjaxMapIconController.php`, `AjaxMountableAreaController.php`,
  `AjaxNpcController.php`, `AjaxOverpulledEnemyController.php` (2 sites),
  `AjaxPathController.php`, `AjaxPridefulEnemyController.php` (2 sites),
  `Ajax/Floor/AjaxFloorUnionAreaController.php`, `Ajax/Floor/AjaxFloorUnionController.php`
- `app/Http/Controllers/LiveSessionController.php`, `NpcController.php`, `ProfileController.php`

## Tests

Added `store_givenBroadcastFails_stillCreatesCheckpoint` and
`delete_givenBroadcastFails_stillDeletesCheckpoint` to `AjaxEnemyForcesCheckpointControllerTest`
(representative of the no-catch `store()` and broad-catch `delete()` patterns respectively). Both
register a fake `failing` broadcast connection whose `broadcast()` always throws
`BroadcastException`, so the test forces a real synchronous broadcast failure without a network
call, and assert the mutation still succeeds.

## Test plan

- [x] `php artisan test --compact tests/Feature/Controller/Ajax` — 110 passed
- [x] `php artisan test --compact` (full suite) — 1969 passed, 1 pre-existing unrelated
      `MapTilesExistenceTest` worktree-env failure (excluded per issue), 1 skipped
- [x] `composer run fix` — clean

### DIFF
--- a/app/Http/Controllers/Ajax/AjaxMappingModelBaseController.php
+++ b/app/Http/Controllers/Ajax/AjaxMappingModelBaseController.php
@@ -13,6 +13,7 @@ use App\Service\Coordinates\CoordinatesServiceInterface;
 use Closure;
 use DB;
 use Exception;
+use Illuminate\Broadcasting\BroadcastException;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Facades\Auth;
 use Throwable;
@@ -93,7 +94,12 @@ abstract class AjaxMappingModelBaseController extends Controller
                     /** @var Floor|null $floor */
                     $floor = $model->getAttribute('floor');
                     $echoContext ??= $floor?->dungeon;
-                    broadcast($this->getModelChangedEvent($coordinatesService, $echoContext, Auth::user(), $model));
+
+                    try {
+                        broadcast($this->getModelChangedEvent($coordinatesService, $echoContext, Auth::user(), $model));
+                    } catch (BroadcastException) {
+                        // Ignore broadcast failures
+                    }
                 }
 
                 return $model;
--- a/app/Http/Controllers/LiveSessionController.php
+++ b/app/Http/Controllers/LiveSessionController.php
@@ -14,6 +14,7 @@ use App\Service\Reverb\ReverbHttpApiServiceInterface;
 use Auth;
 use Exception;
 use Illuminate\Auth\Access\AuthorizationException;
+use Illuminate\Broadcasting\BroadcastException;
 use Illuminate\Contracts\Foundation\Application;
 use Illuminate\Contracts\View\Factory;
 use Illuminate\Contracts\View\View;
@@ -65,8 +66,12 @@ class LiveSessionController extends Controller
                 }
 
                 if ($invitees->isNotEmpty()) {
-                    // Broadcast that channel that a team member has started a live session and that we're invited!
-                    broadcast(new InviteEvent($liveSession, $user, $invitees));
+                    try {
+                        // Broadcast that channel that a team member has started a live session and that we're invited!
+                        broadcast(new InviteEvent($liveSession, $user, $invitees));
+                    } catch (BroadcastException) {
+                        // Ignore broadcast failures
+                    }
                 }
             } catch (Exception $exception) {
                 report($exception);
--- a/app/Http/Controllers/LiveSessionController.php
+++ b/app/Http/Controllers/LiveSessionController.php
@@ -69,8 +69,9 @@ class LiveSessionController extends Controller
                     try {
                         // Broadcast that channel that a team member has started a live session and that we're invited!
                         broadcast(new InviteEvent($liveSession, $user, $invitees));
-                    } catch (BroadcastException) {
-                        // Ignore broadcast failures
+                    } catch (BroadcastException $exception) {
+                        // Ignore broadcast failures, but keep the visibility the outer catch below used to give this
+                        report($exception);
                     }
                 }
             } catch (Exception $exception) {
--- a/app/Http/Controllers/Ajax/AjaxArrowController.php
+++ b/app/Http/Controllers/Ajax/AjaxArrowController.php
@@ -124,7 +124,11 @@ class AjaxArrowController extends Controller
                     /** @var \App\Models\User $user */
                     $user = Auth::getUser();
 
-                    broadcast(new ArrowDeletedEvent($dungeonRoute, $user, $arrow));
+                    try {
+                        broadcast(new ArrowDeletedEvent($dungeonRoute, $user, $arrow));
+                    } catch (BroadcastException) {
+                        // Ignore broadcast failures
+                    }
                 }
 
                 $this->dungeonRouteChanged($dungeonRoute, $arrow, null);
--- a/app/Http/Controllers/Ajax/AjaxBrushlineController.php
+++ b/app/Http/Controllers/Ajax/AjaxBrushlineController.php
@@ -128,7 +128,11 @@ class AjaxBrushlineController extends Controller
                     /** @var \App\Models\User $user */
                     $user = Auth::getUser();
 
-                    broadcast(new BrushlineDeletedEvent($dungeonRoute, $user, $brushline));
+                    try {
+                        broadcast(new BrushlineDeletedEvent($dungeonRoute, $user, $brushline));
+                    } catch (BroadcastException) {
+                        // We don't really care if the broadcast fails, so just catch the exception and move on
+                    }
                 }
 
                 $this->dungeonRouteChanged($dungeonRoute, $brushline, null);
--- a/app/Http/Controllers/Ajax/AjaxDungeonFloorSwitchMarkerController.php
+++ b/app/Http/Controllers/Ajax/AjaxDungeonFloorSwitchMarkerController.php
@@ -11,6 +11,7 @@ use App\Models\Mapping\MappingVersion;
 use App\Models\User;
 use App\Service\Coordinates\CoordinatesServiceInterface;
 use Exception;
+use Illuminate\Broadcasting\BroadcastException;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Http\Request;
 use Illuminate\Http\Response;
@@ -43,7 +44,12 @@ class AjaxDungeonFloorSwitchMarkerController extends AjaxMappingModelBaseControl
                 if (Auth::check()) {
                     /** @var User $user */
                     $user = Auth::getUser();
-                    broadcast(new DungeonFloorSwitchMarkerDeletedEvent($dungeon, $user, $dungeonFloorSwitchMarker));
+
+                    try {
+                        broadcast(new DungeonFloorSwitchMarkerDeletedEvent($dungeon, $user, $dungeonFloorSwitchMarker));
+                    } catch (BroadcastException) {
+                        // Ignore broadcast failures
+                    }
                 }
 
                 // Trigger mapping changed event so the mapping gets saved across all environments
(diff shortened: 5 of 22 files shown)


app/Http/Controllers/LiveSessionController.php
```diff
@@ -65,8 +66,12 @@ public function create(
                 }
 
                 if ($invitees->isNotEmpty()) {
-                    // Broadcast that channel that a team member has started a live session and that we're invited!
-                    broadcast(new InviteEvent($liveSession, $user, $invitees));
+                    try {
```
Wotuu, line 69: :robot: **Narrowing here silently drops the deliberate "Reverb is down" signal.**

Before this change, a `BroadcastException` from this `broadcast()` fell into the outer `catch (Exception $exception)` on line 76, which does `report($exception)` (→ Sentry) plus `Log::error('Reverb server is probably not running!')`. That catch was clearly written *for* this failure mode. The new inner `catch (BroadcastException) { }` swallows it first, so a broadcast-only failure now leaves no trace anywhere.

Same shape in `ProfileController::update()` (the inner `catch (Exception)` there logs `Unable to find users for channel %s`, and the outer one reports + warns "Reverb server is probably not running!").

Mitigating: in both methods the preceding `$reverbHttpApiService->getChannelUsers()` call would itself throw on a full Reverb outage, so the outer report/log still fires in that case — only a broadcast-specific failure goes dark. So this may well be intentional; if it is, worth a one-line note here, otherwise consider `report()`/`Log::debug()` inside the catch on these two sites rather than a fully empty body.

app/Http/Controllers/Ajax/AjaxMappingModelBaseController.php
```diff
@@ -93,7 +94,12 @@ protected function storeModel(
                     /** @var Floor|null $floor */
                     $floor = $model->getAttribute('floor');
                     $echoContext ??= $floor?->dungeon;
-                    broadcast($this->getModelChangedEvent($coordinatesService, $echoContext, Auth::user(), $model));
+
+                    try {
+                        broadcast($this->getModelChangedEvent($coordinatesService, $echoContext, Auth::user(), $model));
```
Wotuu, line 99: :robot: **These catches only fire when the event is dispatched *synchronously* — worth checking the PR body's impact claim.**

`BroadcastManager::queue()` only performs the broadcast inline for `ShouldBroadcastNow` (or an event with `shouldBroadcastNow()`); everything else is pushed as a `BroadcastEvent` job onto `queue.default`. `ContextEvent` is a plain `ShouldBroadcast` and no event in `app/Events` defines `shouldBroadcastNow()`, so the only place in this repo where `broadcast()` can throw a `BroadcastException` at the call site is where `queue.default === 'sync'` — i.e. `phpunit.xml` (`QUEUE_CONNECTION=sync`). Under any queued connection, `broadcast()` here just pushes a job; the Reverb HTTP call and its `BroadcastException` happen in the worker, and what could fail *in the request* is the queue push (a Redis/connection exception, which these catches deliberately don't cover).

The change itself is correct and is exactly what #3941 asked for — this is only about the claimed impact: if prod's `QUEUE_CONNECTION` isn't `sync`, "a transient Reverb blip turned an already-successful delete into a misleading 404" can't happen in a web request, and the real value of this PR is test-environment correctness plus future-proofing. Worth confirming the prod value and softening that line in the PR/issue if so.

(Side note, deliberately *not* a suggestion to change this PR: Laravel 12 has `Illuminate\Contracts\Broadcasting\ShouldRescue`, which would centralise this on `ContextEvent`. I'd leave it — `rescue()` swallows *every* Throwable, so a bug in `broadcastOn()`/`broadcastWith()` would be silently rescued too, which is the opposite of what #3941 wanted, and for queued events it wraps the push rather than the broadcast.)

Wotuu:

app/Http/Controllers/LiveSessionController.php
Wotuu: :robot: Fixed — added `report($exception)` inside the narrow catch on both this site and the equivalent one in `ProfileController::update()`, so a broadcast-only failure still reaches Sentry while leaving the response/redirect unaffected. See commit 81b99da65.

app/Http/Controllers/Ajax/AjaxMappingModelBaseController.php
Wotuu: :robot: Confirmed and agreed — checked `.env.example`, prod's `QUEUE_CONNECTION` is `redis`, not `sync`, so this is accurate: the fix's real teeth today are in tests (matches #3940's flakiness) plus future-proofing, not a currently-reachable production 404/500. Softened the PR body's impact claim to say so explicitly rather than leave it overstated. Not changing the code — the fix itself is still exactly what #3941 asked for, and I agree `ShouldRescue` would be the wrong tool here for the reason you gave.

PR #3960, opened 2026-08-11
#3957 Open the NPC Compendium in a new tab on enemy right-click when the flag is enabled

### LINKED ISSUE #3957: Right-clicking an enemy should open its NPC Compendium page when the flag is enabled (follow-up to #3945)
:robot: Follow-up to #3945.

## Background

Right-clicking an enemy on a route/explore map currently always opens the enemy details modal (`#enemy_details_modal`), rendered from the `map_sidebar_enemy_info_template` Handlebars template. Now that the NPC Compendium exists (#3945, feature-flagged via `App\Features\NpcCompendium`), that modal is the inferior view of the same NPC.

## Requested behaviour

When the `NpcCompendium` feature flag is **enabled** (for the current user), right-clicking an enemy should open the corresponding NPC Compendium entry (`compendium.npc.show`, `/compendium/npc/{id}-{slug}`) **in a new browser tab** instead of showing the modal.

When the flag is **disabled**, behaviour stays exactly as it is today (modal).

Shift+right-click is reserved for the raid marker circle menu and must remain untouched in both cases.

## Implementation notes

- The handler is `_onEnemyContextMenu()` in `resources/assets/js/custom/inline/common/maps/map.js` (~line 576); it receives the `enemy:contextmenu` signal from `resources/assets/js/custom/models/enemy.js` (~line 1103, layer-level `contextmenu` event; the shift+right-click raid-marker branch lives just above it).
- There is currently **no plumbing that exposes Pennant feature flags to the map JS** — nothing in the map blades or `map.js` reads `Feature::active()`. The flag state needs to reach the client, e.g. via the map context payload or a `@feature`-conditional data attribute/JS bool on the map blade, along with the compendium NPC base URL (same pattern as `spellShowBaseUrl` in `resources/views/compendium/npc/index.blade.php`).
- The slug part of the route key comes from the NPC name (`Npc::getRouteKey()` style `{id}-{slug}`); the bare `/compendium/npc/{id}` also resolves, so the JS can link without slugifying if that is simpler.
- Popup blockers: the tab must be opened synchronously from the user gesture (`window.open` directly in the contextmenu handler), not after an async hop.
- Note the admin mapping pages: the modal is already a no-op there (`enemy_details_modal` is not rendered), and admins right-click enemies for mapping work — decide explicitly whether the new-tab behaviour should apply on admin maps or only on route/explore maps.

## Definition of done

- Flag on: right-click opens the NPC Compendium page in a new tab; no modal. Flag off: modal, unchanged.
- Shift+right-click raid marker menu unaffected in both states.
- JS tests / feature tests as applicable, `composer run fix` + `composer run analyse` clean.

Wotuu:
:robot: Closes #3957

## Summary
- Right-clicking an enemy on a route/explore map now opens that NPC's Compendium page (`/compendium/npc/{id}`) in a new tab when the `NpcCompendium` Pennant flag is active for the current user, instead of the `#enemy_details_modal`.
- Flag off: unchanged modal behaviour.
- Admin mapping pages: unchanged (still a no-op) — the modal/compendium-tab logic is gated on `#enemy_details_modal` existing in the DOM, which only route/explore maps render (`MapContextDungeonRoute`/`MapContextDungeonExplore`). Decided against opening the compendium tab there since right-click currently does nothing on admin maps and admins are there for mapping work, not to view NPC details.
- Shift+right-click raid marker menu is untouched — it's handled earlier in `enemy.js` and returns before the `enemy:contextmenu` signal fires.
- The flag state and compendium base URL reach the client via new `npcCompendiumEnabled`/`npcCompendiumBaseUrl` inline options on `common/maps/map.blade.php`, following the same pattern as `spellShowBaseUrl` in `compendium/npc/index.blade.php`.
- The tab is opened synchronously from the contextmenu handler (`window.open` directly, no async hop) so popup blockers don't intervene.
- No slugification needed client-side — the bare `/compendium/npc/{id}` route resolves and 301-redirects to the canonical `{id}-{slug}` URL.

## Test plan
- [x] New JS test: `resources/assets/js/custom/inline/common/maps/map.enemycontextmenu.test.js` — flag enabled opens the tab and skips the modal, flag disabled falls through to the modal, and no-op when `#enemy_details_modal` isn't in the DOM (admin pages).
- [x] New feature test: `tests/Feature/Controller/DungeonRoute/DungeonRouteViewNpcCompendiumFlagTest.php` — asserts the real Blade render exposes the correct `npcCompendiumEnabled`/`npcCompendiumBaseUrl` options for both flag states.
- [x] Full JS suite green (`npx vitest run resources/assets/js/custom`).
- [x] Backend-only/JS-behaviour change, no rendered-output/visual change — screenshot verification skipped.

### DIFF
--- a/resources/assets/js/custom/inline/common/maps/map.enemycontextmenu.test.js
+++ b/resources/assets/js/custom/inline/common/maps/map.enemycontextmenu.test.js
@@ -1,0 +1,94 @@
+// #3957: right-clicking an enemy should open its NPC Compendium page in a new tab when the
+// NpcCompendium feature flag is enabled, instead of the `#enemy_details_modal` bootstrap modal.
+//
+// Follows the global-script recipe from map.favorite.test.js: stub the collaborators the class
+// body touches at load time, then require the source.
+
+globalThis.$ = globalThis.jQuery = require('jquery');
+
+const {InlineCode} = require('../../inlinecode');
+globalThis.InlineCode = InlineCode;
+
+globalThis.SettingsTabMap = class SettingsTabMap {
+};
+globalThis.SettingsTabPull = class SettingsTabPull {
+};
+
+const {CommonMapsMap} = require('./map');
+
+describe('CommonMapsMap._onEnemyContextMenu', () => {
+    let visualData;
+    let enemy;
+    let windowOpenSpy;
+
+    beforeEach(() => {
+        document.body.innerHTML = '';
+
+        visualData = {};
+        enemy = {
+            npc: {id: 123, name: 'Some Npc'},
+            getVisualData: () => visualData,
+        };
+
+        globalThis.getState = vi.fn(() => ({
+            isMapAdmin: () => false,
+        }));
+        globalThis.lang = {get: (key) => key};
+
+        windowOpenSpy = vi.spyOn(window, 'open').mockImplementation(() => {
+        });
+    });
+
+    afterEach(() => {
+        vi.clearAllMocks();
+    });
+
+    /**
+     * @param {Object} options
+     * @returns {CommonMapsMap}
+     */
+    function buildMap(options = {}) {
+        return new CommonMapsMap('map', 'common/maps/map', options);
+    }
+
+    it('_onEnemyContextMenu_givenNoEnemyDetailsModalInDom_doesNothing', () => {
+        // Arrange
+        let map = buildMap({npcCompendiumEnabled: true, npcCompendiumBaseUrl: '/compendium/npc'});
+
+        // Act
+        map._onEnemyContextMenu({context: enemy});
+
+        // Assert
+        expect(windowOpenSpy).not.toHaveBeenCalled();
+    });
+
+    it('_onEnemyContextMenu_givenFlagEnabled_opensCompendiumPageInNewTabAndSkipsModal', () => {
+        // Arrange
+        document.body.innerHTML = '<div id="enemy_details_modal"></div>';
+        let map = buildMap({npcCompendiumEnabled: true, npcCompendiumBaseUrl: '/compendium/npc'});
+
+        // Act
+        map._onEnemyContextMenu({context: enemy});
+
+        // Assert
+        expect(windowOpenSpy).toHaveBeenCalledWith('/compendium/npc/123');
+    });
+
+    it('_onEnemyContextMenu_givenFlagDisabled_doesNotOpenNewTab', () => {
+        // Arrange
+        document.body.innerHTML = '<div id="enemy_details_modal"></div><div id="enemy_details_modal_title_text"></div><div id="enemy_details_modal_body"></div>';
+        globalThis.Handlebars = {templates: {map_sidebar_enemy_info_template: () => ''}};
+        globalThis.refreshTooltips = vi.fn();
+        globalThis.bootstrap = {
+            Collapse: {getOrCreateInstance: () => ({hide: vi.fn()})},
+            Modal: {getOrCreateInstance: () => ({show: vi.fn()})},
+        };
+        let map = buildMap({npcCompendiumEnabled: false, npcCompendiumBaseUrl: '/compendium/npc'});
+
+        // Act
+        map._onEnemyContextMenu({context: enemy});
+
+        // Assert
+        expect(windowOpenSpy).not.toHaveBeenCalled();
+    });
+});
--- a/resources/assets/js/custom/inline/common/maps/map.enemycontextmenu.test.js
+++ b/resources/assets/js/custom/inline/common/maps/map.enemycontextmenu.test.js
@@ -20,6 +20,7 @@ describe('CommonMapsMap._onEnemyContextMenu', () => {
     let visualData;
     let enemy;
     let windowOpenSpy;
+    let modalShowSpy;
 
     beforeEach(() => {
         document.body.innerHTML = '';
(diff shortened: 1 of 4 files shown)


resources/assets/js/custom/inline/common/maps/map.js
```diff
@@ -577,10 +579,17 @@ class CommonMapsMap extends InlineCode {
...
 
-        // The modal is only rendered on route/explore maps - on e.g. the admin mapping pages this is a no-op
+        // The modal (and the NPC Compendium tab below) is only relevant on route/explore maps -
+        // on e.g. the admin mapping pages this is a no-op
         let enemyDetailsModal = document.getElementById('enemy_details_modal');
 
         if (visualData !== null && enemyDetailsModal !== null) {
+            if (this.options.npcCompendiumEnabled) {
+                // Opened synchronously from the user gesture so popup blockers don't intervene
+                window.open(`${this.options.npcCompendiumBaseUrl}/${enemy.npc.id}`);
```
Wotuu: :robot: `window.open()` returns `null` when the popup is blocked, but the return value is discarded and the `return` fires unconditionally — so in any environment where the popup is refused, right-click becomes a silent dead action with no fallback (the modal branch below is never reached).

The concrete case is embeds, not desktop popup blockers: `resources/views/dungeonroute/embed.blade.php` includes `common.maps.map` with a `MapContextDungeonRoute`, so `#enemy_details_modal` *is* rendered on embedded route maps and this new branch applies inside third-party `<iframe>`s. An iframe with a `sandbox` attribute that omits `allow-popups` blocks `window.open` outright, and the embedder controls that attribute — we don't.

Cheap fix that keeps the happy path identical:

```js
if (this.options.npcCompendiumEnabled && window.open(`${this.options.npcCompendiumBaseUrl}/${enemy.npc.id}`) !== null) {
    return;
}
```

so a blocked popup degrades to the existing modal instead of to nothing.

resources/assets/js/custom/inline/common/maps/map.js
```diff
@@ -577,10 +579,17 @@ class CommonMapsMap extends InlineCode {
...
+                return;
```
Wotuu, line 592: :robot: This early `return` also removes the only entry point to the **enemy user report** form, which has no replacement on the compendium page.

`enemy_report_category` / `enemy_report_enemy_id` / `enemy_report_username` / `enemy_report_message` exist only in `resources/views/common/modal/enemydetails.blade.php`, and the only code that populates them and opens the collapse is the modal branch immediately below this line (feeding `POST /ajax/userreport/enemy/{id}`). `resources/views/compendium/npc/show.blade.php` has no report affordance, so with the flag on, "report a mapping issue with this enemy" becomes unreachable from the map.

Not a live regression — `App\Features\NpcCompendium::resolve()` gates on `ROLE_ADMIN`/`ROLE_INTERNAL_TEAM`, so today only staff hit this path, and #3957 did consciously decide the modal is the inferior NPC view. But the report form isn't an NPC view, and the issue doesn't mention it. Worth a deliberate decision (surface it on the compendium NPC page, or keep a secondary entry point) before this flag goes public rather than discovering the reports dried up afterwards.

resources/assets/js/custom/inline/common/maps/map.enemycontextmenu.test.js
```diff
@@ -0,0 +1,94 @@
...
+    });
+
+    it('_onEnemyContextMenu_givenFlagDisabled_doesNotOpenNewTab', () => {
+        // Arrange
+        document.body.innerHTML = '<div id="enemy_details_modal"></div><div id="enemy_details_modal_title_text"></div><div id="enemy_details_modal_body"></div>';
+        globalThis.Handlebars = {templates: {map_sidebar_enemy_info_template: () => ''}};
+        globalThis.refreshTooltips = vi.fn();
+        globalThis.bootstrap = {
+            Collapse: {getOrCreateInstance: () => ({hide: vi.fn()})},
+            Modal: {getOrCreateInstance: () => ({show: vi.fn()})},
```
Wotuu: :robot: This stub mints a throwaway `vi.fn()` on every call, so the flag-disabled test can only assert the *negative* (`window.open` not called) — it can't prove the modal actually opened. That leaves the "flag off falls through to the modal" claim untested: any future change that returns early for a different reason would still pass this test.

Hoisting the spy makes it a real assertion:

```js
const modalShowSpy = vi.fn();
globalThis.bootstrap = {
    Collapse: {getOrCreateInstance: () => ({hide: vi.fn()})},
    Modal: {getOrCreateInstance: () => ({show: modalShowSpy})},
};
// ...
expect(modalShowSpy).toHaveBeenCalled();
```

Wotuu:

resources/assets/js/custom/inline/common/maps/map.js
Wotuu: :robot: Fixed in bdcc3fcb — `window.open()`'s return value now gates the early return, so a blocked popup (e.g. an embed iframe without `allow-popups`) falls back to the modal instead of doing nothing. Added a JS test (`_onEnemyContextMenu_givenFlagEnabledButPopupBlocked_fallsBackToModal`) covering the fallback.

resources/assets/js/custom/inline/common/maps/map.js
Wotuu: :robot: Agreed this is a real gap, but treating it as intentional scope for this PR: #3957 explicitly asks for the modal to be fully replaced by the Compendium page when the flag is on ("that modal is the inferior view of the same NPC"), and the flag is currently staff-only (`ROLE_ADMIN`/`ROLE_INTERNAL_TEAM`), so no regression for real users today. Surfacing a report affordance on the Compendium NPC page is a legitimate follow-up, but it's a product decision (where/how to report from that page) that belongs in its own issue rather than being decided inline here — leaving this thread open for Wotuu's call on whether to file one before the flag goes public.

resources/assets/js/custom/inline/common/maps/map.enemycontextmenu.test.js
Wotuu: :robot: Fixed in bdcc3fcb — hoisted the `Modal.show()` spy and now assert it was actually called in both the flag-disabled test and the new popup-blocked-fallback test.

PR #3962, opened 2026-08-11
#3951 Own spell tooltips from wago.tools DB2 data

### LINKED ISSUE #3951: Own spell tooltips from wago.tools DB2 data (follow-up to #3945)
:robot: #3945 shipped spell hover tooltips by putting `data-wowhead="spell={id}"` on our spell links: the vendored Wowhead power.js attaches its tooltip without touching the href, so links keep pointing at keystone.guru. That works well but depends on Wowhead's CDN and goodwill, and spells Wowhead doesn't know silently get no tooltip. This issue tracks replacing it with a self-hosted tooltip pipeline.

## Data source

https://wago.tools/db2 exposes the game client's DB2 tables as per-build CSVs (e.g. `https://wago.tools/db2/Spell/csv` → `Spell.12.1.0.69189.csv`). Verified contents: `Spell.Description_lang` holds the raw description templates.

## Why this is a real subsystem, not a column import

Descriptions are game-engine templates, e.g.:

```
Stab the target, causing ${$s2*$<mult>} Physical damage. Damage increased by $s4% when you
are behind your target$?s319949[, and critical strikes apply Find Weakness for $319949s1 sec][].
```

Faithful rendering needs, at minimum:
- **Tables**: `Spell` (descriptions), `SpellName`, `SpellEffect` (`$s1..$sN` effect points), `SpellMisc` (links duration/range/cast-time indices), `SpellDuration` (`$d`), plus `SpellRadius`/`SpellCastTimes` for less common vars.
- **A template parser**: `$sN`/`$wN` effect substitution, `$d` durations, inline math `${...}`, cross-spell refs (`$319949s1`), conditionals `$?s<id>[...][...]` / `$?a<id>[...][...]` (player-knowledge-dependent — render the base branch like Wowhead does for anonymous visitors), plural/gender macros (`$lpoint:points;`).
- **Per-patch reimport** (artisan command keyed on build, like the existing DB2-derived data flows).
- A tooltip UI component replacing the power.js dependency on compendium pages.

## Definition of done

- Spell tooltips on compendium pages render from our own DB with no wow.zamimg.com / whTooltips dependency, including spells Wowhead doesn't know.
- Descriptions resolve at least `$sN`, `$d`, `${math}`, cross-spell refs and conditionals' base branch; unresolvable tokens degrade gracefully (omitted, never rendered raw).
- Import command documented + repeatable per game build.

Wotuu:
:robot: Closes #3951

Spell tooltips now render from our own data instead of Wowhead's tooltip script. Spell descriptions
are imported from the game client's DB2 tables (published per build on https://wago.tools), rendered
up front, and shown in a tooltip of our own on compendium spell links.

## What was decided up front

- **Descriptions are rendered at import time**, not per request. Nothing about a rendered description
  varies per visitor — conditionals render the anonymous base branch — so the DB2 tables are streamed
  once and thrown away, leaving two new nullable columns on `spells`: the rendered text and the
  template it came from. No new tables; `SpellEffect` alone is ~1M rows.
- **Only spells already in `spells`** get a description (~7.2k of 7.9k have one). Cross-spell
  references (`$319949s1`, `$@spelldesc2823`) still resolve — they are read from the same stream.
- **`power.js` stays loaded.** A spell link renders our tooltip when we have a description, and keeps
  its `data-wowhead` attribute when we do not, so Wowhead covers the gaps. Its `iconizeLinks` also
  decorates the bare wowhead hrefs in the admin tools and the compendium's "View on Wowhead" button,
  which removing the script would silently change. Note this leaves the definition of done's "no
  wow.zamimg.com dependency" partly open by design — a follow-up can finish it once every
  `data-wh-*` behaviour has a replacement.

## The parser

Descriptions are a small template language of the game's own:

```
Stab the target, causing ${$s2*$<mult>} Physical damage. Damage increased by $s4% when you are
behind your target$?s319949[, and critical strikes apply Find Weakness for $319949s1 sec][].
```

Supported: effect values (`$s1`, `$m1`, `$o1`), periods, radii, durations, the same with a spell id
prefix, inline arithmetic, named variables (`$<mult>`), cross-spell name/description inserts,
conditionals (including the `$?a[x]?b[y][z]` if/else-if chain and comparisons on real spell values),
plural and gender macros, and the UI escape codes. Unresolvable tokens are omitted rather than left
raw, and arithmetic containing one is dropped whole. `${...}` never goes near `eval()` — a small
recursive descent evaluator accepts its grammar and nothing else.

Boss spells — what the compendium is for — render in full. Player abilities carry no base points in
DB2 and scale off the casting character, so those render without a number; there is no character to
scale against up front.

## Safety

The import refuses to run when a build describes none of our spells, so a wrong `--build` or a bad
download cannot clear every description in one pass. (Found the hard way while testing.)

## Repeatable per patch

```sh
php artisan wagotools:importspelldescriptions   # --product / --build to pin
php artisan mapping:save                        # descriptions ship in spells.json, like icon_name
```

Documented in the new `spell-descriptions` skill.

## Verification

- 61 new tests: the parser table-driven over real template shapes, the evaluator (including that it
  rejects anything that is not arithmetic), the import end-to-end against CSVs placed in the download
  cache (no network), the CSV reader, and the link partial (including that a description is escaped).
- Rendered all 7,203 descriptions locally and checked no raw token, bracket or pipe code survives.

## Cold review

`cold-reviewer-fable` raised 4 findings, all fixed in `ebed65e90` and resolved on their threads:
a cross-game-version wipe (an import is now scoped to the game version its product is the client
of, and only clears a description when the build actually holds a row for that spell), two-digit
effect indices rendering a wrong number, an unvalidated build string reaching a filesystem path,
and one-hop reference collection. The seeder was re-exported: 55 classic spells had picked up
retail text in the first pass and now carry none.

## Deliberately not in scope

- **The spell's own compendium page still does not print its description as body text** — only
  links carry it, which is what the issue asked for ("spell tooltips on compendium pages"). Happy
  to add it if you want it, but it is a design change to that page rather than part of this.
- **Descriptions are English only.** DB2 serves per-locale `_lang` columns, so this could be
  extended, but every other locale would need its own column or table.


Wotuu:
:robot: Verification: hovering a spell link on an NPC compendium page (demo `npc_spells` rows, local dev). The description, the 25 damage and the 4 yard radius all come from our own `spells.description` — no Wowhead tooltip involved.

verification

app/Service/Spell/Description/SpellDescriptionImportService.php
```diff
@@ -0,0 +1,342 @@
...
+        $updatedCount = 0;
+        $handled      = 0;
+        $total        = $spells->count();
+
+        foreach ($spells as $spell) {
+            $handled++;
+
+            $template = $templates[$spell->id] ?? null;
+
+            // A spell that lost its description in this build must lose ours as well
```
Wotuu, line 121: :robot: **Cross-game-version wipe**: the import loads *every* row of `spells` regardless of `game_version_id`, then nulls the description of any spell the chosen product's build does not describe. The spells table holds 341 classic-era spells (`game_version_id = 2`) alongside retail ones, and the command explicitly documents `--product=wow_classic` / `--product=wowt`. Running with `wow_classic` would describe the classic spells and **null out ~7,100 retail descriptions in the same pass** — and the "nothing was imported" guard will not trip, because the classic build does describe our classic spells (plus 55 overlapping ids that already picked up retail text in this PR's shipped run). Suggest scoping the spell query (and the null-out) to spells whose game version matches the product, or at minimum never clearing spells belonging to a different game version than the one being imported.

app/Service/Spell/Description/SpellDescriptionParser.php
```diff
@@ -0,0 +1,711 @@
...
+     * an optional effect index.
+     */
+    private function renderValueToken(
+        SpellDescriptionContextInterface $context,
+        int                              $spellId,
+        string                           $template,
+        int                              &                              $offset,
+        ?float                           &                           $lastNumber,
+    ): string {
+        $pattern = sprintf('/\G\$(\d*)(%s|[a-zA-Z])(\d?)/', implode('|', array_keys(self::ALIAS_TOKENS)));
```
Wotuu: :robot: **Two-digit effect indices are misparsed.** The `(\d?)` here (and in `substituteValueToken`) only takes one digit, so `$s10`..`$s32` parse as effect 1 followed by literal text — e.g. `$s11%` renders effect 1's value with a stray `1%` appended, i.e. a *wrong number* rather than an omission. Spells carry up to 32 effects and the shipped seeder data already contains 10 such tokens (spells 31884, 196840, 319454 — currently all inside dropped conditional branches, so the bug is latent, but any boss spell using `$s10+` in its main text will render a wrong value). `(\d{0,2})` plus the existing `SPELL_WIDE_TOKENS` trailing-digit handling would cover it.

app/Service/WagoTools/WagoToolsService.php
```diff
@@ -0,0 +1,197 @@
...
+                yield array_combine($header, $row);
+            }
+        } finally {
+            fclose($handle);
+        }
+    }
+
+    private function getTableCsvTargetPath(string $table, string $build): string
+    {
+        return storage_path(sprintf('app/db2/%s/%s.csv', $build, $table));
```
Wotuu, line 141: :robot: **Unvalidated `$build` flows into a filesystem path.** `$build` comes either from the CLI or from wago.tools' JSON (`getLatestBuild` returns whatever string the API hands back) and is interpolated into `storage_path('app/db2/%s/%s.csv')` — a compromised/misbehaving response like `../../foo` writes and reads outside `storage/app/db2` (and mangles the download URL). Cheap hardening: validate the build against `/^\d+(\.\d+){3}$/` in `getLatestBuild`/`getTableCsvPath` before using it.

app/Service/Spell/Description/SpellDescriptionImportService.php
```diff
@@ -0,0 +1,342 @@
...
+            // import - carrying on would clear every description we have on the strength of a bad file.
+            if ($templates === [] && $spells->isNotEmpty()) {
+                $this->log->importDescriptionsNoDescriptionsFound($product, $build);
+
+                return null;
+            }
+
+            // Descriptions reference other spells for their values and even for their entire text, so
+            // everything they point at must be looked up too
+            $referencedIds = $this->collectReferencedSpellIds($templates);
```
Wotuu: :robot: Minor: reference collection is one hop deep, while the parser's `MAX_RECURSION_DEPTH = 5` suggests deeper chains render. `collectReferencedSpellIds` only scans *our* spells' templates; the templates fetched for those referenced ids are never scanned themselves, so a referenced description that itself contains `$@spelldescY`, `$@spellnameY` or `$Ys1` (Y outside `$wantedIds`) silently renders those tokens as omissions — dropped text/values with no signal. With 2,143 `$@spelldesc` references in the shipped data, some nested chains are likely. Either iterate collection to a fixpoint over the fetched templates, or note the one-hop limit in the class docblock.

Wotuu:
:robot: Cold review (fable): 4 findings posted.

app/Service/Spell/Description/SpellDescriptionImportService.php
Wotuu: :robot: Fixed in ebed65e90. The import now takes a `--gameVersion` (default `retail`) and only loads spells of that game version, so a classic build cannot reach retail spells at all. On top of that a description is only cleared when the build actually holds a row for the spell — one it has never heard of is left alone rather than treated as "description dropped". Two tests cover both.

Good catch on the 55 classic spells that had picked up retail text in the shipped seeder: those are cleared and the seeder re-exported, so `spells.json` now carries descriptions for retail spells only.

app/Service/Spell/Description/SpellDescriptionParser.php
Wotuu: :robot: Fixed in ebed65e90 — `(\d?)` is now `(\d{0,2})` in both places, with a parser test for `$s11`. Agreed this was the worse failure mode of the two: a wrong number rather than an omission.

app/Service/WagoTools/WagoToolsService.php
Wotuu: :robot: Fixed in ebed65e90. `getTableCsvPath()` now refuses a build that is not four dot-separated numbers, and a table name that is not alphabetic, before either reaches a path; `getLatestBuild()` validates what the API hands back against the same pattern.

app/Service/Spell/Description/SpellDescriptionImportService.php
Wotuu: :robot: Fixed in ebed65e90. The Spell table is now re-read until no new referenced spell is pulled in (bounded at 3 passes, matching how deep the parser recurses). It also surfaced a second bug: referenced spells without a description of their own must still be looked up for their values and names, or a cross-spell duration stops resolving.

Wotuu:
:robot: Back to draft — Wotuu reviewed the rendered output directly and the damage numbers are wrong by orders of magnitude (a boss ability reads as "10 damage"). Creature spell damage in DB2 is a coefficient, not a number; working on the scaling model plus a place to keep the inputs so it can be scaled by key level later.

Wotuu:
When you make these pull request bodies, can you ensure that you don't insert line breaks on your own? Just insert one long line, and GitHub will automatically word wrap it for you. Right now, you pre-formatted several lines, so there are multiple line breaks there, which reads annoyingly.

app/Service/WagoTools/WagoToolsService.php
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    /**
+     * Streams the response straight to disk - these CSVs run to tens of megabytes, and the shared Curl
```
Wotuu, line 145: Is it an idea to change the shared curl handler to do the same streaming thing, or is that not wise to do for a global function?

PR #3964, opened 2026-08-11
#3963 Run the combat_log_parse_failures drop migration on combatlog_migrate

### LINKED ISSUE #3963: combat_log_parse_failures drop migration uses the unprivileged combatlog connection, not combatlog_migrate
:robot: The drop migration `database/migrations_combatlog/2026_08_06_000001_drop_combat_log_parse_failures_table.php` hardcodes `protected $connection = 'combatlog';`, the least-privilege runtime connection (`DB_COMBATLOG_USERNAME`), instead of the elevated `combatlog_migrate` connection (`DB_MIGRATION_USERNAME`) that config/database.php defines specifically for privileged schema changes.

Result: the runtime combatlog DB user lacks DROP grant, so the migration fails on staging with:

```
SQLSTATE[42000]: Syntax error or access violation: 1142 DROP command denied to user
'keystoneguru_combatlog_staging_u'@'...' for table 'combat_log_parse_failures'
(Connection: combatlog, ...)
```

Sentry: PHP-LARAVEL-SJ (unresolved).

This has been failing on every staging migrate attempt since the migration was merged in #3822 (present since v15.10.0). It doesn't affect the app's runtime behavior (the table is just never dropped), but it means combat_log_parse_failures never actually gets cleaned up, and any future destructive combatlog migration will hit the same wall — every combatlog migration in the repo hardcodes `combatlog` instead of relying on the CLI `--database` flag the way the main-DB migrations do (they declare no $connection at all and rely on `artisan migrate --database=migrate`).

Fix: change this migration's $connection to `combatlog_migrate`.

A hotfix for v15.10.2 has already been applied (S3 hotfix overlay, same tag, redeployed to staging) so the current release cleans this up. This issue tracks landing the same one-line fix on master through the normal flow so the next real release contains it too.

Wotuu:
:robot: Closes #3963

## Summary
- The `combat_log_parse_failures` drop migration hardcoded `protected $connection = 'combatlog'` — the least-privilege runtime connection — instead of the elevated `combatlog_migrate` connection that `config/database.php` defines specifically for privileged schema changes.
- The runtime combatlog DB user lacks DROP grant, so this migration has been failing on every staging migrate attempt since #3822 merged it (present since v15.10.0). Sentry: PHP-LARAVEL-SJ.
- Already hotfixed onto the deployed v15.10.2 tag (S3 hotfix overlay + staging redeploy); this PR lands the same one-line fix on master so the next real release contains it.

## Test plan
- [x] Added `DropCombatLogParseFailuresTableTest` asserting the migration's connection, proved RED before the fix / GREEN after
- [x] CI green

### DIFF
--- a/database/migrations_combatlog/2026_08_06_000001_drop_combat_log_parse_failures_table.php
+++ b/database/migrations_combatlog/2026_08_06_000001_drop_combat_log_parse_failures_table.php
@@ -5,7 +5,7 @@ use Illuminate\Database\Schema\Blueprint;
 use Illuminate\Support\Facades\Schema;
 
 return new class extends Migration {
-    protected $connection = 'combatlog';
+    protected $connection = 'combatlog_migrate';
 
     public function up(): void
     {
--- a/tests/Feature/Database/Migrations/DropCombatLogParseFailuresTableTest.php
+++ b/tests/Feature/Database/Migrations/DropCombatLogParseFailuresTableTest.php
@@ -1,0 +1,28 @@
+<?php
+
+namespace Tests\Feature\Database\Migrations;
+
+use PHPUnit\Framework\Attributes\Group;
+use PHPUnit\Framework\Attributes\Test;
+use Tests\TestCases\PublicTestCase;
+
+/**
+ * The runtime `combatlog` connection is least-privilege and lacks DROP - only the elevated
+ * `combatlog_migrate` connection may run this migration, see #3963.
+ */
+#[Group('CombatLog')]
+final class DropCombatLogParseFailuresTableTest extends PublicTestCase
+{
+    private const MIGRATION = 'migrations_combatlog/2026_08_06_000001_drop_combat_log_parse_failures_table.php';
+
+    #[Test]
+    public function migration_usesTheElevatedCombatlogMigrateConnection(): void
+    {
+        // Arrange
+        // Act
+        $migration = require database_path(self::MIGRATION);
+
+        // Assert
+        $this->assertSame('combatlog_migrate', $migration->getConnection());
+    }
+}
--- a/tests/Feature/Database/Migrations/DropCombatLogParseFailuresTableTest.php
+++ b/tests/Feature/Database/Migrations/DropCombatLogParseFailuresTableTest.php
@@ -16,7 +16,7 @@ final class DropCombatLogParseFailuresTableTest extends PublicTestCase
     private const MIGRATION = 'migrations_combatlog/2026_08_06_000001_drop_combat_log_parse_failures_table.php';
 
     #[Test]
-    public function migration_usesTheElevatedCombatlogMigrateConnection(): void
+    public function getConnection_givenTheMigration_returnsAConfiguredElevatedConnection(): void
     {
         // Arrange
         // Act
@@ -24,5 +24,10 @@ final class DropCombatLogParseFailuresTableTest extends PublicTestCase
 
         // Assert
         $this->assertSame('combatlog_migrate', $migration->getConnection());
+        $this->assertArrayHasKey(
+            $migration->getConnection(),
+            config('database.connections'),
+            'The migration points at a connection that no longer exists in config/database.php.',
+        );
     }
 }


tests/Feature/Database/Migrations/DropCombatLogParseFailuresTableTest.php
```diff
@@ -0,0 +1,28 @@
...
+
+    #[Test]
+    public function migration_usesTheElevatedCombatlogMigrateConnection(): void
+    {
+        // Arrange
+        // Act
+        $migration = require database_path(self::MIGRATION);
+
+        // Assert
+        $this->assertSame('combatlog_migrate', $migration->getConnection());
```
Wotuu, line 26: :robot: This assertion compares a literal to the literal in the file it just read, so it only catches someone editing this one migration — it still passes if `combatlog_migrate` is renamed or removed from `config/database.php`, which is the other half of the failure mode described in the docblock. Worth one extra line:

```php
$this->assertArrayHasKey($migration->getConnection(), config("database.connections"));
```

That matters more than usual here: `database/schema/combatlog-schema.dump` stops well before `2026_05_*`, so this migration *does* run in CI/fresh worktrees — but via `migrate --database=combatlog`, where the CI user is `root` (`.env.ci.example`). CI therefore stays green either way, and this test is the only guard the fix has.

tests/Feature/Database/Migrations/DropCombatLogParseFailuresTableTest.php
```diff
@@ -0,0 +1,28 @@
...
+ * The runtime `combatlog` connection is least-privilege and lacks DROP - only the elevated
+ * `combatlog_migrate` connection may run this migration, see #3963.
+ */
+#[Group('CombatLog')]
+final class DropCombatLogParseFailuresTableTest extends PublicTestCase
+{
+    private const MIGRATION = 'migrations_combatlog/2026_08_06_000001_drop_combat_log_parse_failures_table.php';
+
+    #[Test]
+    public function migration_usesTheElevatedCombatlogMigrateConnection(): void
```
Wotuu: :robot: Test name does not follow the `[functionname]_given[Condition]_returns[ExpectedResult]` convention from CLAUDE.md — `getConnection_returnsTheElevatedCombatlogMigrateConnection` would. The sibling in this same directory (`RepointDungeonRoutesOffRemovedRetailMappingVersionsTest::up_givenRouteOnARemovedMappingVersion_repointsItToTheCurrentLegionRemixMappingVersion`) complies.

database/migrations_combatlog/2026_08_06_000001_drop_combat_log_parse_failures_table.php
```diff
@@ -5,7 +5,7 @@
 use Illuminate\Support\Facades\Schema;
 
 return new class extends Migration {
-    protected $connection = 'combatlog';
+    protected $connection = 'combatlog_migrate';
```
Wotuu, line 8: :robot: Non-blocking, scope question for Wotuu: this is a point fix for one migration, but the underlying asymmetry stays. `app/Console/Commands/Database/Migrate.php` and `app/Console/Commands/Environment/Update.php` run the main DB with `--database=migrate` (elevated) yet the combatlog DB with `--database=combatlog` (least-privilege) — so *every* combatlog schema migration defaults to the user without DROP, and the next DROP/ALTER there repeats this exact failure unless whoever writes it remembers the override. Both connections read the same `DB_MIGRATION_USERNAME`/`DB_MIGRATION_PASSWORD` pair, so `--database=combatlog_migrate` in those two commands (plus `.github/actions/php-ci-setup/action.yml` and `sh/worktree.sh`) would fix the class rather than the instance, and would let this property be dropped entirely — matching the sibling migrations here that set no `$connection` at all.

Concretely relevant now: `database/migrations_combatlog/2026_06_09_000001_drop_enemy_positions_table.php` still hardcodes `combatlog` with a `dropIfExists`. I cannot check whether it applied on staging, but it is the same pattern and worth confirming it is not stuck the same way.

Wotuu:
:robot: Cold review (opus): 3 findings posted — none blocking. The one-line connection fix itself is correct (`combatlog_migrate` matches `config/database.php`, and `Migrator::runMethod()` swaps the default connection so `Schema::` inside `up()`/`down()` uses it; `down()` needs no separate treatment). Findings are: a test-quality gap, a test-name convention nit, and a scope question about the point fix vs. the command-level asymmetry.

tests/Feature/Database/Migrations/DropCombatLogParseFailuresTableTest.php
Wotuu: :robot: Fixed — added `assertArrayHasKey($migration->getConnection(), config('database.connections'))` in cd186fd73.

tests/Feature/Database/Migrations/DropCombatLogParseFailuresTableTest.php
Wotuu: :robot: Fixed — renamed to `getConnection_givenTheMigration_returnsAConfiguredElevatedConnection` in cd186fd73.

database/migrations_combatlog/2026_08_06_000001_drop_combat_log_parse_failures_table.php
Wotuu: :robot: Agreed this is a real gap, keeping it out of scope for this point fix as you note. Checked Sentry for `enemy_positions` DROP failures over the last 90 days — none found, so that migration already succeeded before the combatlog runtime user's DROP grant was tightened; it is not currently stuck. Leaving this thread open for your call on whether to file the class-level fix (switch Migrate/Update/CI/worktree.sh to `combatlog_migrate` and drop the per-migration $connection override entirely) as a follow-up issue.

Wotuu:

database/migrations_combatlog/2026_08_06_000001_drop_combat_log_parse_failures_table.php
Wotuu: All combat log migrations should use `combatlog_migrate` connection. I think eventually it should not be defined per-migration since it's easy to forget. 

I checked the infra repository and it says:

```
      this.createBootstrapTaskDefinition(`${stage}-migrate`, stage, tag, [
        `php artisan migrate --database=migrate --force`,
        `php artisan migrate --database=combatlog_migrate --path=database/migrations_combatlog --force`,
        `php artisan db:seed --database=migrate --force`,
      ]);
```

So the fact that this migration added a `$connection` at all is what broke it.

Wotuu: :robot: Filed the class-level fix as #3965 (switch `Migrate`/`Update`/CI/`worktree.sh` to `combatlog_migrate` and drop the per-migration `$connection` override) — resolving this thread.

PR #3972, opened 2026-08-11
#3971 Scale spell damage from DB2 coefficients to real numbers

### LINKED ISSUE #3971: Scale spell damage from DB2 coefficients to real numbers
:robot: Follow-up to #3951, split out of #3962 so the two can be reviewed and merged separately.

#3962 renders spell descriptions from the game client's DB2 data. What it cannot do is print a
correct damage number: the game stores creature damage and healing as a **coefficient**, not an
amount. "10" in the data is 50,845 in the game.

## The model

Verified exactly against the game's own rendered numbers:

```
amount = coefficient / 10
       × the multiplier of the content the caster belongs to
       × the M+ season factor            (ExpectedStatMod, gated by
                                          Min/MaxMythicPlusSeasonID)
```

The season factor being a separate term is what makes key-level scaling a multiplication rather
than a re-render.

Nothing in the client data links a spell to the content it is cast in — `SpellMisc.ContentTuningID`
is 0, `SpellLevels` is empty, and only 65 of our 6,149 NPCs carry a `ContentTuningID`. The link we
do have is our own: `spell → npc_spells → npcs → npc_dungeons → dungeons`, which resolves a single
dungeon for 97% of NPCs.

## Scope

- `spell_effects` — the per-effect coefficients, in the database rather than baked into a sentence.
- Descriptions stored as a sprintf format plus one tagged value per number, so damage and healing
  can be recalculated later while durations and radii stay literal.
- `dungeons.spell_damage_coefficient` — the per-dungeon multiplier, calibrated once per patch and
  shipped as seeded data (never fetched at runtime).
- The spell detail page showing its description, and a denser tooltip (cast time, range, duration,
  stacks, school, dispel type, mechanic).

## Definition of done

- A boss ability's tooltip shows the damage the game shows, not the raw coefficient.
- The coefficient and the season factor stay separately available, so a later key-level selector is
  a multiplication rather than a re-import.

Wotuu:
:robot: Closes #3971. Stacks on #3962 — split out of it so the two can be reviewed and merged
separately. Targets `master` per the repo convention (checks only run against `master`); the diff
collapses to just these commits once #3962 merges.

#3962 renders spell descriptions from the game client's DB2 data, but prints the raw coefficient
where a damage number belongs — "10" where the game shows 50,845. This PR is the machinery that
turns one into the other, and keeps the inputs around so it can be redone per key level.

## Descriptions are a format, not a sentence

```
FORMAT  Leaps at an enemy, inflicting %1$s Physical damage and causing
        a %2$s Physical damage every %3$s sec for %4$s.
VALUES  [{"kind":"damage","coefficient":10,"spellId":357827,"effectIndex":0,"text":"10"},
         {"kind":"damage","coefficient":2, "spellId":357827,"effectIndex":1,"text":"2"},
         {"kind":"period","text":"0.5"},
         {"kind":"duration","text":"5 sec"}]
```

A later key-level selector rewrites the `damage` entries and re-runs `sprintf` — no re-import, no
re-parse. **Which numbers are damage comes from the effect's own type in the game data**
(`Effect` / `EffectAura`), not from reading the sentence, so "increases damage taken by 20%" is
correctly left alone while real damage scales.

Literal text is `%%`-escaped into the format — descriptions are full of real percent signs, and
without that the stored format would corrupt the first time a spell said "50%".

## The scaling model

Verified exactly against the game's own rendered numbers:

```
amount = coefficient / 10 × dungeon multiplier × M+ season factor
```

- `spell_effects` holds the per-effect coefficients — the durable input.
- `dungeons.spell_damage_coefficient` holds the multiplier, already threaded through the parser.
- The season factor (`ExpectedStatMod`, gated by `Min`/`MaxMythicPlusSeasonID`) stays a separate
  term, which is what makes key-level scaling a multiplication.

Nothing in the client data links a spell to the content it is cast in — `SpellMisc.ContentTuningID`
is 0, `SpellLevels` is empty, and only 65 of our 6,149 NPCs carry a `ContentTuningID`. The link that
does work is ours: `spell → npc_spells → npcs → npc_dungeons → dungeons`, single-valued for 97% of
NPCs.

## Not finished yet

- `spell_damage_coefficient` is **null everywhere**, so damage still renders as the coefficient.
  Calibrating it needs real spell-to-dungeon pairings, which are being collected from live combat
  logs (`combatlog:pollruns`). The multiplier is then committed as seeded data — never fetched at
  runtime.
- The spell detail page does not show its description yet.
- The tooltip does not show cast time, range, duration, stacks, school, dispel type or mechanic yet.

Marked draft until those land and it has been cold-reviewed.


app/Service/Spell/Description/SpellDamageCalibrationService.php
```diff
@@ -0,0 +1,158 @@
...
+
+        // The description is the last block of the tooltip, after the name and the range and cast line
+        $paired = array_slice($theirNumbers, -count($values));
+        $agreed = [];
+
+        foreach ($values as $index => $value) {
+            // Measured against the coefficient rather than against what we display, because what we
+            // display already has a multiplier in it once one has been measured - comparing that would
+            // report a multiplier of one and quietly undo the previous run
+            if ($value->coefficient === null || $value->coefficient <= 0.0) {
```
Wotuu: :robot: **Calibration can never bootstrap a spell whose multiplier is still null — which is exactly its default target set.** `measure()` reads coefficients out of `description_values`, but `SpellDescriptionParser::appendNumber()` *omits* scalable values entirely (no placeholder, no value, no coefficient) when `damageMultiplier <= 0.0`. So after `wagotools:importspelldescriptions` runs for a spell with `damage_multiplier = null`, its `description_values` contains only non-scalable entries — every one is skipped by this `coefficient === null` check, `$agreed` stays empty, and the spell is counted `unreadable`. Meanwhile the default query in `calibrate()` selects precisely the `whereNull(\x27damage_multiplier\x27)` spells. Net effect: the documented SKILL.md sequence (import → calibrate → import) measures 0 new spells with the shipped code — the 3,647 multipliers in the seeder must have been produced by an earlier code state. Any brand-new boss spell after the next patch will never get a damage number. Fix options: have calibration derive the coefficients from `spell_effects`/the template rather than from `description_values`, or have the import store the coefficient on the value even when the multiplier is unknown (with the placeholder omitted from the format).

app/Service/Spell/Description/SpellDamageCalibrationService.php
```diff
@@ -0,0 +1,158 @@
...
+        }
+
+        if ($agreed === []) {
+            return ['multiplier' => null, 'disagreed' => false];
+        }
+
+        // Every scaled number in one description is scaled by the same thing; if ours are not, we paired
+        // them up wrongly and the measurement is worthless
+        foreach ($agreed as $candidate) {
+            if (abs($candidate - $agreed[0]) > $agreed[0] * self::AGREEMENT_TOLERANCE) {
```
Wotuu, line 153: :robot: **The agreement check protects almost none of the measured spells, and non-scalable pairs are never verified — a misaligned pairing stores a wrong multiplier silently.** In the shipped seeder, 3,134 of the 3,647 spells with a multiplier carry exactly *one* scalable value, so this every-pair-must-agree loop is vacuous for 86% of measurements. With single-pair spells, any count mismatch inside the description (Wowhead rendering a conditional branch we omit, a token we drop, "stacks up to 3 times") shifts the tail slice by one, and as long as `wrong_number / coefficient > 100` the bogus multiplier is stored with no log. The class docblock even states the safeguard — "A pair that is equal is a literal" — but the code never checks it: literal pairs (`coefficient === null` values like durations and radii) are skipped instead of being compared for equality against `$paired[$index]`. Verifying that every non-scalable pair matches (within formatting tolerance) is a cheap alignment check that would catch most of these silent misparings; treat a mismatch as `unreadable`.

database/migrations/2026_08_11_000000_add_description_to_spells_table.php
```diff
@@ -0,0 +1,36 @@
...
+        Schema::table('spells', function (Blueprint $table) {
+            // The raw DB2 template as shipped by the game client. Kept alongside the rendered text as its
+            // provenance: it is what makes a changed description reviewable in a seeder diff, and what
+            // tells you whether a description reads oddly because of the parser or because of the game.
+            $table->text('description_template')->nullable()->after('name');
+            // The description as a sprintf format with positional placeholders, and one entry per
+            // placeholder saying what that number is. Damage and healing are coefficients of the content
+            // the caster belongs to, so they have to stay separable from the sentence around them to be
+            // recalculated per key level rather than fixed at import time.
+            $table->text('description_format')->nullable()->after('description_template');
```
Wotuu: :robot: **This edits #3962's migration in place — safe only if both PRs land in the same release.** If #3962 merges and a release (or staging deploy) cuts before #3972 merges, the migrate cron records `2026_08_11_000000` as run with the old shape (`description` text column). When #3972 then merges, this edited file never re-runs: those environments end up with `description`, but without `description_format`/`description_values`, while the new code selects and writes both (`SpellDescriptionImportService`, `Spell` accessors, the seeder JSON) — 500s and a broken reseed, and no migration left that can repair it. Either (a) guarantee the two PRs merge into the same release and say so in the PR body, or (b) leave #3962's migration untouched and put `description_format`/`description_values` (+ dropping `description`) in a new migration in this PR, which is safe in every ordering.

app/Service/Spell/Description/SpellDescriptionImportService.php
```diff
@@ -0,0 +1,428 @@
...
+                    'period_ms'     => $effect->periodMs,
+                    'chain_targets' => $effect->chainTargets,
+                    'radius'        => $effect->radius,
+                    'max_radius'    => $effect->maxRadius,
+                ];
+            }
+        }
+
+        foreach (array_chunk($rows, self::PERSIST_CHUNK_SIZE) as $chunk) {
+            SpellEffect::query()->upsert($chunk, ['spell_id', 'effect_index']);
```
Wotuu, line 123: :robot: **Stale `spell_effects` rows survive forever.** `upsert()` inserts/updates but never deletes, so when a patch removes an effect from a spell (or shrinks its effect list), the old row stays. Because the export nests whatever is in the table under each spell, the stale row then round-trips through `spells.json` into every environment permanently. The description render itself is unaffected (it reads the in-memory DB2 stream), but the table exists precisely to recalculate amounts per key level later — a stale index will feed that with a coefficient the spell no longer has. Cheap fix here: after the upserts, delete rows for the imported spell ids whose `(spell_id, effect_index)` is not in the new set.

Wotuu:
:robot: Cold review (fable): 4 findings posted.

app/Service/Spell/Description/SpellDamageCalibrationService.php
Wotuu: :robot: Fixed in 305aeb484 — and you were right that the seeded multipliers came from an earlier code state. My own test appeared to pass only because I nulled `damage_multiplier` without re-importing, so the values still held coefficients from a previous run. Re-importing first reproduces it exactly:

```
before re-import: [{"kind":"damage","coefficient":10,...},{"kind":"damage","coefficient":2,...},...]
after  re-import: [{"kind":"period",...},{"kind":"duration",...}]
```

`appendNumber()` now always records the value and its coefficient, and leaves the *text* empty when there is no multiplier — so nothing is shown, exactly as before, but the coefficient survives. The gap that an empty value leaves is closed at render time instead of when the format is built (`RenderedSpellDescription::closeGaps()`, mirrored in the JS).

Proven from cold: wiped every multiplier, re-imported, ran the command — **3,671 spells measured**.

app/Service/Spell/Description/SpellDamageCalibrationService.php
Wotuu: :robot: Fixed in 305aeb484. The values that are *not* amounts — durations, radii, periods — are now compared against the game's numbers, and any mismatch rejects the spell outright. That is the alignment check the agreement check could not be: with a single damage value it had nothing to disagree with.

It bites, which is the point: **782 spells are now rejected** as misaligned that would previously have stored a plausible wrong multiplier.

database/migrations/2026_08_11_000000_add_description_to_spells_table.php
Wotuu: :robot: Fixed in 305aeb484. #3962's migration is restored to exactly what that branch carries, and the two new columns moved to `2026_08_11_000003_add_description_format_to_spells_table`.

The superseded `description` column is deliberately left in place rather than dropped here — dropping it in the same release that stops writing it is the #3497 failure mode. Follow-up filed as #3977 to drop it once this has shipped in a release.

app/Service/Spell/Description/SpellDescriptionImportService.php
Wotuu: :robot: Fixed in 305aeb484. After the upsert, effects whose index is no longer present for that spell are deleted, so an effect a patch removes does not linger and ride into every environment through the seeder.

Wotuu:
Generic question - when will this all get run? Manually? Do we wire it up into the `create a release` skill (regenerating all the spell tooltips)? Is it already? I don't want to remember when to call this exactly.

Wotuu:
`spells.json` is growing very big with this update. It's 14MB big now. Do we think this is a problem? If so, should we split out the new data to a separate `spell_descriptions.json` for example? Just so it's easier for version control? If that is not a concern then I'm fine with keeping it this way.

We can also do the same thing with `spell_effects` and put those in their own JSON? They already have their own table, too.

app/Logic/MapContext/MapContextDungeonData.php
```diff
@@ -105,8 +105,10 @@ function () {
                         'fetched_data_at',
                         // Nothing on the map renders a spell description, and one per spell of the dungeon
                         // would grow the map context by a fair amount of text
-                        'description',
                         'description_template',
+                        'description_format',
+                        'description_values',
+                        'tooltip_data',
```
Wotuu, line 113: Should `spell_effects` also be a part of this?

app/Service/Spell/Description/Dtos/SpellEffectData.php
```diff
@@ -7,7 +7,27 @@
...
+    private const int EFFECT_WEAPON_DAMAGE_NOSCHOOL = 121;
+
+    /** SpellEffect.EffectAura values that do the same over time. */
+    private const int AURA_PERIODIC_DAMAGE = 3;
+
+    private const int AURA_PERIODIC_HEAL = 8;
+
+    private const int AURA_PERIODIC_LEECH = 53;
+
+    private const int AURA_SCHOOL_ABSORB = 69;
```
Wotuu: Should this not be an enum then?

app/Service/Spell/Description/SpellDescriptionImportService.php
```diff
@@ -93,6 +102,49 @@ public function importDescriptions(
...
+                    'spell_id'      => $spellId,
+                    'effect_index'  => $effectIndex,
+                    'effect_type'   => $effect->effectType,
+                    'aura_type'     => $effect->auraType,
+                    'base_points'   => $effect->basePoints,
+                    'variance'      => $effect->variance,
+                    'period_ms'     => $effect->periodMs,
+                    'chain_targets' => $effect->chainTargets,
+                    'radius'        => $effect->radius,
+                    'max_radius'    => $effect->maxRadius,
```
Wotuu: I'd prefer if you used an `$effect->toArray()` for these values. Easy to add something and forget to change it here without it.

Wotuu:
> What would actually remove "I don't want to remember" — happy to file this as a follow-up if you
want it: a small scheduled command that asks wago.tools for the latest retail build, compares it
against the last build we imported (which we'd need to start recording — right now nothing persists
it), and on a change files a GitHub issue / pings Discord saying "patch 12.x is out, re-run the spell
description import". That turns it from something you remember into something that tells you, without
pretending a human-reviewed seeder diff can be generated unattended. Say the word and I'll open it.

Yes, file this follow-up. I'd like that command as a reminder. Make it file a Github issue. 

Re: splitting the spells.json, all good then, we'll leave it as-is.

PR #3973, opened 2026-08-11
#3968 Scope the Compendium by dungeon: /compendium/dungeon/{dungeon}/{section}

### LINKED ISSUE #3968: NPC Compendium class pages don't reflect the selected dungeon in the URL (SEO/shareability)
:robot: Found while manually testing staging for v15.10.2.

`/compendium/class/{class}` (`ClassCompendiumController::show()`) renders content that depends
on the currently selected dungeon (`Dungeon::getUserOrDefaultDungeon()` -
`app/Http/Controllers/Compendium/ClassCompendiumController.php:35`), but that selection lives in
session/user state rather than the URL. So `https://staging.keystone.guru/compendium/class/rogue`
can show entirely different content (different dungeon's NPCs/spell data) depending on who's
looking and what they last had selected, with no way to tell from the URL alone, share a link to
a specific dungeon's view, or have search engines index the per-dungeon variants separately.

This likely affects the other Compendium controllers too (`CompendiumController`,
`NpcCompendiumController`, `SpellCompendiumController`) - worth checking whether they have the
same pattern.

## Needs investigating
- Best way to make the dungeon part of the URL - e.g. `/compendium/class/rogue/{dungeon}` (or a
  query param), while keeping a sensible default/redirect when no dungeon is specified (current
  `getUserOrDefaultDungeon()` behavior).
- How this interacts with the existing route/URL structure for `/compendium/**`, and any
  front-end (dungeon picker) wiring that would need updating to push the new URL on selection
  change (browser history / `pushState`) rather than just re-rendering in place.
- Whether canonical/meta tags need adjusting once the dungeon is URL-addressable.

No fix proposed yet - filing to track the investigation and pick the right approach before
implementing.

Wotuu:
:robot: Closes #3968

`/compendium/class/{class}`, `/compendium/npc` and `/compendium/spell` all rendered content scoped to the visitor's *context dungeon* (`users.dungeon_id` for members, the `dungeon_context` cookie for guests). The dungeon appeared nowhere in the URL, so the same URL showed different content per visitor, a link could not be shared for a specific dungeon, and the per-dungeon variants could not be indexed separately.

## URL shape

| Page | Bare (302 →) | Canonical |
|---|---|---|
| Class | `/compendium/class/rogue` | `/compendium/dungeon/the-dawnbreaker/class/rogue` |
| NPC index | `/compendium/npc` | `/compendium/dungeon/the-dawnbreaker/npc` |
| Spell index | `/compendium/spell` | `/compendium/dungeon/the-dawnbreaker/spell` |
| Activity | `/compendium/activity` | `/compendium/dungeon/the-dawnbreaker/activity` |
| Activity day | — | `/compendium/dungeon/the-dawnbreaker/activity/{date}` |

Every dungeon-scoped page hangs off the dungeon it is about, so all four share one shape. The literal `dungeon/` segment keeps the slug unambiguous, so no section can ever be mistaken for a dungeon (or vice versa) whatever a dungeon is called — and it removes the need for any route-ordering trick. Pages that are not dungeon-scoped stay where they are: `/compendium`, `/compendium/npc/{npc}`, `/compendium/spell/{spell}`, `/compendium/class`.

The pre-move `/compendium/activity/{dungeon}[/{date}]` URLs are **removed rather than redirected**: the Compendium is gated to admins/internal team and has never been enabled on production, so there is no audience to keep them working for.

Otherwise this follows the `{gameVersion}/{dungeon}` pattern the rest of the site already uses (explore / heatmap / discover): the bare URL redirects to the canonical dungeon-ful one, the page writes the URL dungeon back into the dungeon context, and the header dungeon picker gets a `dungeonContextLinks` override so its cards link straight to the same page for another dungeon.

Decisions worth calling out:

- **302, not 301**, for the bare → canonical redirect: the target is derived from the visitor's own cookie / `users.dungeon_id`, so a permanent redirect would get browser- and proxy-cached and pin one dungeon into every later session. (The neighbouring slug guards in `NpcCompendiumController::show()` *are* 301, correctly — those targets are static.)
- **A separate bare route + redirect, not an optional `{dungeon?}` param** — an optional param would serve identical content at two URLs, which is the duplicate-content problem this issue is about.
- **Two Laravel details worth knowing**, both now commented in place: a route registered with first-class callable syntax binds its parameters **positionally**, so `showDungeon()` takes `(Dungeon, CharacterClass)` to match the URL; and a nested binding with a custom key (`{characterClass:key}` under `{dungeon}`) is **scoped to its parent** by default, which had Laravel resolving the class through `$dungeon->characterClasses()` — hence `->withoutScopedBindings()`.
- **No season forcing.** Unlike `activity()`, the class/index pages accept any dungeon; an out-of-season dungeon renders with no data rather than silently redirecting a shared link elsewhere. Consequence, intended: for such a dungeon the header picker (current season only) has no matching card.
- **The in-page dungeon filter on the NPC/spell index now navigates** instead of reloading the DataTable, which keeps URL, header selection and persisted context in one round-trip. If the select ever holds a non-dungeon-id value (it can emit season/expansion options when configured to), it falls back to the old `table.ajax.reload()`.
- **The index tables take their dungeon from the URL, not from that filter** (cold review). The filter only lists dungeons mapped for the visitor's game version, so reading the table's dungeon off it would list a different dungeon whenever the URL named one the select does not offer. A guard rejecting such URLs was deliberately not added: it would make the same URL 404/redirect for one visitor and render for another, which is the per-visitor variability this issue exists to remove. Residual, cosmetic: for such a dungeon the dropdown's visible label falls back to its first option while URL, breadcrumb, header identity and table all name the right dungeon.
- **Only pages that validate their dungeon write the site-wide context** (cold review). `activity()` validates against the current season and redirects, so it writes; `activityDay()` does not validate, so it does not write — a direct link to a day page therefore leaves the header dungeon highlight stale, by design.
- Also fixed: the activity pages never persisted their URL dungeon as the context, letting the header's dungeon selection drift away from the page being viewed.

## Not doing: canonical / meta tags (issue bullet 3)

Checked, deferred. There is no `<link rel="canonical">` anywhere in `resources/views` or `app` today, and no `<meta name="description">` either — only Open Graph via `common/general/linkpreview.blade.php`. Adding canonical tags would mean new shared-layout infrastructure, and it cannot pay off yet: `NpcCompendium::resolve()` gates on `ROLE_ADMIN`/`ROLE_INTERNAL_TEAM`, so `feature_active` 404s Googlebot on every Compendium URL — nothing is indexable until the flag goes public.

The issue's `pushState` bullet is answered by the design rather than implemented: with the dungeon in the path and `dungeonContextLinks` overridden, the picker is plain link navigation — real URLs, working back button, no history API.

## Verification

- `php artisan test --filter=Compendium` — 71 passed. New per page: bare URL redirects to the context dungeon (asserting 302), a cross-dungeon case that sets `users.dungeon_id` to A, requests B's URL and asserts B renders *and* becomes the context, and an unknown-slug 404. Activity gained context-persistence tests (that `activity()` writes and `activityDay()` does not), and each index page has a test that a dungeon outside the visitor's game version still drives the table from the URL.
- Real headless Chrome, logged in as an admin with the flag active: all four bare URLs 302 to their canonical form; the class page renders the URL dungeon (The Dawnbreaker) while the stored context was Black Rook Hold; header cards link to `/compendium/class/rogue/<other-dungeon>`; picking another dungeon in the index filter navigates to `/compendium/dungeon/pit-of-saron/npc`; the back button returns to the previous dungeon's page; no page errors.

No visual change beyond the URLs, so no before/after screenshots.

### DIFF
--- a/app/Http/Controllers/Compendium/ClassCompendiumController.php
+++ b/app/Http/Controllers/Compendium/ClassCompendiumController.php
@@ -50,7 +50,11 @@ class ClassCompendiumController extends Controller
         ]);
     }
 
-    public function showDungeon(CharacterClass $characterClass, Dungeon $dungeon, DungeonServiceInterface $dungeonService): View
+    /**
+     * Parameters follow the URL's order (/compendium/dungeon/{dungeon}/class/{characterClass}) - a
+     * route registered with first-class callable syntax binds them positionally, not by name.
+     */
+    public function showDungeon(Dungeon $dungeon, CharacterClass $characterClass, DungeonServiceInterface $dungeonService): View
     {
         // The URL is the source of truth for which dungeon is being viewed - make it the context
         // dungeon as well, so the header's dungeon selection follows along (as on explore/heatmap)
--- a/app/Http/Controllers/Compendium/ClassCompendiumController.php
+++ b/app/Http/Controllers/Compendium/ClassCompendiumController.php
@@ -10,7 +10,10 @@ use App\Models\Npc\Npc;
 use App\Models\Spell\Spell;
 use App\Service\CombatLog\DataExtractors\SpellCounters\SpellCounterDefinitionInterface;
 use App\Service\CombatLog\DataExtractors\SpellCounters\SpellCounterDefinitions;
+use App\Service\Dungeon\DungeonServiceInterface;
+use Illuminate\Http\RedirectResponse;
 use Illuminate\Support\Collection;
+use Illuminate\Support\Facades\Auth;
 use Illuminate\View\View;
 
 class ClassCompendiumController extends Controller
@@ -32,9 +35,27 @@ class ClassCompendiumController extends Controller
         ]);
     }
 
-    public function show(CharacterClass $characterClass): View
+    /**
+     * The class page without a dungeon in the URL - bounces to the canonical URL of the visitor's
+     * context dungeon.
+     *
+     * Deliberately a 302: the target depends on the visitor's own context dungeon, so a permanent
+     * redirect would get cached and pin one dungeon into every later visit.
+     */
+    public function show(CharacterClass $characterClass): RedirectResponse
     {
-        $dungeon        = Dungeon::getUserOrDefaultDungeon();
+        return redirect()->route('compendium.class.show.dungeon', [
+            'characterClass' => $characterClass,
+            'dungeon'        => Dungeon::getUserOrDefaultDungeon(),
+        ]);
+    }
+
+    public function showDungeon(CharacterClass $characterClass, Dungeon $dungeon, DungeonServiceInterface $dungeonService): View
+    {
+        // The URL is the source of truth for which dungeon is being viewed - make it the context
+        // dungeon as well, so the header's dungeon selection follows along (as on explore/heatmap)
+        $dungeonService->setDungeonContext($dungeon, Auth::user());
+
         $mappingVersion = $dungeon->getCurrentMappingVersion();
 
         $spells = Spell::query()
@@ -71,6 +92,9 @@ class ClassCompendiumController extends Controller
             'npcsByCharacteristicId' => $npcsByCharacteristicId,
             'counterSections'        => $this->getCounterSections($characterClass, $dungeon, $mappingVersion),
             'reflectSection'         => $this->getReflectSection($characterClass, $dungeon, $mappingVersion),
+            // HeaderComposer only injects this into the header view itself - the dungeon context
+            // links this page overrides are built in the view, so it needs its own copy
+            'gameVersionDungeons' => $dungeonService->getDungeonsForGameVersion(),
         ]);
     }
 
--- a/app/Http/Controllers/Compendium/NpcCompendiumController.php
+++ b/app/Http/Controllers/Compendium/NpcCompendiumController.php
@@ -89,21 +89,6 @@ class NpcCompendiumController extends Controller
         return redirect()->route('compendium.activity', ['dungeon' => $dungeon]);
     }
 
-    /**
-     * The activity pages used to hang off /compendium/activity/{dungeon}, before the dungeon moved to
-     * the front of the path. A 301 is right here where it is wrong for the context-dungeon
-     * redirects: this target is derived purely from the URL, so it is the same for every visitor.
-     */
-    public function activityLegacy(Dungeon $dungeon): RedirectResponse
-    {
-        return redirect(route('compendium.activity', ['dungeon' => $dungeon]), 301);
-    }
-
-    public function activityDayLegacy(Dungeon $dungeon, string $date): RedirectResponse
-    {
-        return redirect(route('compendium.activity.day', ['dungeon' => $dungeon, 'date' => $date]), 301);
-    }
-
     public function activity(
         Dungeon                       $dungeon,
         SeasonServiceInterface        $seasonService,
--- a/app/Http/Controllers/Compendium/NpcCompendiumController.php
+++ b/app/Http/Controllers/Compendium/NpcCompendiumController.php
@@ -137,9 +137,10 @@ class NpcCompendiumController extends Controller
             abort(404);
         }
 
-        // The URL is the source of truth for which dungeon is being viewed - make it the context
-        // dungeon as well, so the header's dungeon selection follows along (as on explore/heatmap)
-        $dungeonService->setDungeonContext($dungeon, Auth::user());
+        // Deliberately no setDungeonContext() here, unlike every other dungeon-in-the-URL page: only
+        // pages that validate their dungeon write the site-wide context, and this one does not.
+        // activity() one route up rejects any dungeon outside the current season, so persisting one
+        // from here would leave the visitor with a context its own overview refuses to show.
 
         return view('compendium.activity.day', [
             'contextDungeon'      => $dungeon,
--- a/app/Http/Controllers/Compendium/NpcCompendiumController.php
+++ b/app/Http/Controllers/Compendium/NpcCompendiumController.php
@@ -13,20 +13,46 @@ use App\Models\GameVersion\GameVersion;
 use App\Models\Npc\Npc;
 use App\Models\Npc\NpcHealth;
 use App\Service\Compendium\NpcCompendiumServiceInterface;
+use App\Service\Dungeon\DungeonServiceInterface;
 use App\Service\Season\SeasonServiceInterface;
 use Exception;
 use Illuminate\Database\Query\JoinClause;
 use Illuminate\Http\RedirectResponse;
 use Illuminate\Http\Request;
 use Illuminate\Support\Carbon;
+use Illuminate\Support\Facades\Auth;
 use Illuminate\View\View;
 
 class NpcCompendiumController extends Controller
 {
-    public function index(): View
+    /**
+     * The NPC index without a dungeon in the URL - bounces to the canonical URL of the visitor's
+     * context dungeon.
+     *
+     * Deliberately a 302: the target depends on the visitor's own context dungeon, so a permanent
+     * redirect would get cached and pin one dungeon into every later visit.
+     */
+    public function index(): RedirectResponse
     {
+        return redirect()->route('npc.compendium.index.dungeon', [
+            'dungeon' => Dungeon::getUserOrDefaultDungeon(),
+        ]);
+    }
+
+    public function indexDungeon(Dungeon $dungeon, DungeonServiceInterface $dungeonService): View
+    {
+        // The URL is the source of truth for which dungeon is being viewed - make it the context
+        // dungeon as well, so the header's dungeon selection follows along (as on explore/heatmap)
+        $dungeonService->setDungeonContext($dungeon, Auth::user());
+
         return view('compendium.npc.index', [
-            'contextDungeon' => Dungeon::getUserOrDefaultDungeon(),
+            'contextDungeon' => $dungeon,
+            // HeaderComposer only injects this into the header view itself - the dungeon context
+            // links this page overrides are built in the view, so it needs its own copy
+            'gameVersionDungeons' => $dungeonService->getDungeonsForGameVersion(),
+            // The dungeon filter's options are dungeon ids; navigating to another dungeon's page
+            // needs their slugs
+            'dungeonSlugsById' => Dungeon::query()->pluck('slug', 'id'),
         ]);
     }
 
@@ -67,6 +93,7 @@ class NpcCompendiumController extends Controller
         Dungeon                       $dungeon,
         SeasonServiceInterface        $seasonService,
         NpcCompendiumServiceInterface $npcCompendiumService,
+        DungeonServiceInterface       $dungeonService,
     ): View|RedirectResponse {
         $contextDungeon = $this->getContextDungeonOrDefault($seasonService, $dungeon);
         if ($contextDungeon === null) {
@@ -74,6 +101,10 @@ class NpcCompendiumController extends Controller
         } elseif ($contextDungeon->id !== $dungeon->id) {
             return redirect()->route('compendium.activity', ['dungeon' => $contextDungeon]);
         }
+
+        // The URL is the source of truth for which dungeon is being viewed - make it the context
+        // dungeon as well, so the header's dungeon selection follows along (as on explore/heatmap)
+        $dungeonService->setDungeonContext($dungeon, Auth::user());
 
         $dates       = $npcCompendiumService->getActivityDates(10, $dungeon);
         $eventsByDay = [];
@@ -83,14 +114,19 @@ class NpcCompendiumController extends Controller
         }
 
         return view('compendium.activity.index', [
-            'contextDungeon' => $dungeon,
-            'dates'          => $dates,
-            'eventsByDay'    => $eventsByDay,
+            'contextDungeon'      => $dungeon,
+            'dates'               => $dates,
+            'eventsByDay'         => $eventsByDay,
+            'gameVersionDungeons' => $dungeonService->getDungeonsForGameVersion(),
         ]);
     }
 
-    public function activityDay(Dungeon $dungeon, string $date, NpcCompendiumServiceInterface $npcCompendiumService): View
-    {
+    public function activityDay(
+        Dungeon                       $dungeon,
+        string                        $date,
+        NpcCompendiumServiceInterface $npcCompendiumService,
+        DungeonServiceInterface       $dungeonService,
+    ): View {
         try {
             $carbon = Carbon::createFromFormat('Y-m-d', $date);
         } catch (Exception) {
@@ -101,10 +137,15 @@ class NpcCompendiumController extends Controller
             abort(404);
         }
 
+        // The URL is the source of truth for which dungeon is being viewed - make it the context
+        // dungeon as well, so the header's dungeon selection follows along (as on explore/heatmap)
+        $dungeonService->setDungeonContext($dungeon, Auth::user());
+
         return view('compendium.activity.day', [
-            'contextDungeon' => $dungeon,
-            'date'           => $carbon,
-            'events'         => $npcCompendiumService->getEventsForDate($carbon, $dungeon),
+            'contextDungeon'      => $dungeon,
+            'date'                => $carbon,
+            'events'              => $npcCompendiumService->getEventsForDate($carbon, $dungeon),
+            'gameVersionDungeons' => $dungeonService->getDungeonsForGameVersion(),
         ]);
     }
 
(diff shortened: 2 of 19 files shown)


app/Http/Controllers/Compendium/NpcCompendiumController.php
```diff
@@ -13,20 +13,46 @@
...
+        return redirect()->route('npc.compendium.index.dungeon', [
+            'dungeon' => Dungeon::getUserOrDefaultDungeon(),
+        ]);
+    }
+
+    public function indexDungeon(Dungeon $dungeon, DungeonServiceInterface $dungeonService): View
     {
+        // The URL is the source of truth for which dungeon is being viewed - make it the context
+        // dungeon as well, so the header's dungeon selection follows along (as on explore/heatmap)
+        $dungeonService->setDungeonContext($dungeon, Auth::user());
```
Wotuu, line 46: :robot: **The "renders with no data" rationale doesn't hold for the two index pages — and the context write happens before anything validates the dungeon.**

The PR body justifies accepting any dungeon with "an out-of-season dungeon renders with no data rather than silently redirecting a shared link elsewhere". That's true for `ClassCompendiumController::showDungeon()`, which filters queries by `$dungeon->id` directly. It is *not* true here (nor in `SpellCompendiumController::indexDungeon()`): the table's entire content comes from `d.dungeon_id = $('#compendium_filter_dungeon').val()`, and that select only contains dungeons that are active *and* have a mapping version for the visitor's current game version. Verified on this worktree's seed: 77 of 155 dungeons make it into the retail select — e.g. `cathedral-of-eternal-night` is absent. So `/compendium/npc/dungeon/cathedral-of-eternal-night` under a retail context renders a page whose URL, header and breadcrumb say one dungeon while the table does *not* show that dungeon's NPCs (whatever the select falls back to). Silently wrong data, not empty data.

Second, `setDungeonContext($dungeon, Auth::user())` runs before anything checks the dungeon is usable, and `users.dungeon_id` / the `dungeon_context` cookie are site-wide. A shared Compendium link therefore repoints the visitor's dungeon context to a dungeon their game version doesn't have, and that persists onto /explore, /discover and route-creation defaults — more than the cosmetic "header picker has no matching card" the PR body notes.

The in-repo precedent is `DungeonExploreController::viewDungeonFloor()`, which guards before doing anything else:

```php
if (!$dungeon->active || $currentMappingVersion === null) {
    return redirect()->route('dungeon.explore.gameversion.select', ['gameVersion' => $gameVersion]);
}
```

An equivalent guard (404, or redirect to the context dungeon) on `indexDungeon()` for both controllers would close both halves.

Note the new tests can't catch this: they all pick `Dungeon::active()->firstOrFail()`, which is always in the current game version. A case with a dungeon that has no mapping version for the visitor's game version would.

app/Http/Controllers/Compendium/NpcCompendiumController.php
```diff
@@ -101,10 +137,15 @@ public function activityDay(Dungeon $dungeon, string $date, NpcCompendiumService
             abort(404);
         }
 
+        // The URL is the source of truth for which dungeon is being viewed - make it the context
+        // dungeon as well, so the header's dungeon selection follows along (as on explore/heatmap)
+        $dungeonService->setDungeonContext($dungeon, Auth::user());
```
Wotuu: :robot: `activityDay()` now persists an arbitrary URL dungeon as the visitor's site-wide context, but — unlike its own parent route — never validates it. `activity()` above runs `getContextDungeonOrDefault($seasonService, $dungeon)` and redirects when the URL dungeon isn't acceptable; `activityDay()` calls `setDungeonContext()` unconditionally. So `/compendium/activity/{dungeon}/{date}` will happily write a dungeon into `users.dungeon_id` that `/compendium/activity/{dungeon}` would have refused one segment earlier.

Wotuu:

app/Http/Controllers/Compendium/NpcCompendiumController.php
Wotuu: :robot: Fixed in 54f386fd4, though not with a guard — a guard on "is this dungeon offered for the visitor’s game version" would make the same URL 404/redirect for one visitor and render for another, which is the per-visitor variability this issue exists to remove.

Instead the table now takes its dungeon straight from the URL (`const contextDungeonId = {{ $contextDungeon->id }}`, sent as `d.dungeon_id`) rather than from the select. The select value was only ever read for the table, and since picking a dungeon now navigates, it is not needed for that at all. So the table provably shows the URL dungeon in every case, with no visitor-state dependence. New test per index page pins it, using a dungeon deliberately outside the visitor’s game version (`indexDungeon_givenDungeonNotOfferedByTheFilter_stillDrivesTheTableFromTheUrlDungeon`).

Residual, stated rather than engineered around: for such a dungeon the dropdown’s *visible label* still falls back to its first option, while URL, breadcrumb, header identity and table all name the right dungeon. Making the select include it needs `ignoreGameVersion`, which would not fully close it anyway (the `$expansion->active` gate and the season-dungeon reject still run) and would make the dropdown’s contents vary per URL.

On `setDungeonContext()` repointing the site-wide context: that is the approved design of this PR — the URL wins and the context follows — and it is what explore, heatmap and discover already do. Not changed.

app/Http/Controllers/Compendium/NpcCompendiumController.php
Wotuu: :robot: Fixed in 54f386fd4 by removing the `setDungeonContext()` call from `activityDay()` entirely — it was mine, added beyond what the issue needed.

That gives a rule the two routes can both follow: only pages that validate their dungeon write the site-wide context. `activity()` validates via `getContextDungeonOrDefault()` and redirects, so it writes; `activityDay()` does not validate, so it does not write. Preferable to gating the write on the same check, which would have papered over the inconsistency rather than removing it.

Cost is cosmetic: a direct link to a day page leaves the header dungeon highlight stale. The day view's `dungeonContextLinks` override still points its cards at the right per-dungeon URLs. New test: `activityDay_givenDungeonInUrl_leavesTheContextDungeonAlone`.

Wotuu:

Focused on the unreviewed dungeon-first restructure (7bff28684). Verified:

- **Legacy URL coverage is complete.** Every previously-*released* shape still resolves: `/compendium/activity/{dungeon}` and `/compendium/activity/{dungeon}/{date}` 301 to the new form; `/compendium/npc`, `/compendium/spell`, `/compendium/class/{class}` are still registered and 302 to the canonical dungeon URL. The `/compendium/npc/dungeon/{dungeon}`, `/compendium/spell/dungeon/{dungeon}` and `/compendium/class/{class}/{dungeon}` shapes only ever existed on this branch (fda50f88a), so they correctly need no redirect.
- **The positional-binding fix on `showDungeon(Dungeon, CharacterClass)` was genuinely required** by the reorder, and the order now matches the URL. Checked the same trap on `activityDay(Dungeon, string $date)` and `activityDayLegacy(Dungeon, string $date)` — both already in URL order.
- **`withoutScopedBindings()` is newly needed and correctly placed.** `characterClass` went from first parameter to non-first-with-a-binding-field, which is exactly the condition that turns on implicit parent scoping. No other new route has a second model binding with a custom key, so none of the others need it.
- **No stale references.** Swept `app/`, `resources/views/`, `routes/`, `lang/`, `config/` and the JS for `compendium.class.show` / `npc.compendium.index` / `spell.compendium.index` / `compendium.activity` route names and for hardcoded `/compendium/...` paths. The only remaining bare-route links (header nav, `/compendium` cards, class index tiles) are the intended entry points that 302 to canonical. No sitemap generator exists, so nothing else emits these URLs.
- **`route:cache` survives the restructure.** Per the repo's reserved-word/first-class-callable caveat this is a production-only failure mode that the normal (uncached) test run cannot see, so I re-ran the suite with `bootstrap/cache/routes-v7.php` present: 73 passed. The `scope_bindings => false` action flag serializes fine. Cache cleared afterwards.
- `php artisan test --compact --filter=Compendium` — 73 passed, 174 assertions.

The two earlier findings were not reintroduced: the index tables still read `contextDungeonId` from the URL dungeon rather than the filter, and `activityDay()` still deliberately does not write the dungeon context.

app/Http/Controllers/Compendium/ClassCompendiumController.php
```diff
@@ -32,9 +35,31 @@ public function index(): View
...
 
-    public function show(CharacterClass $characterClass): View
+    /**
+     * The class page without a dungeon in the URL - bounces to the canonical URL of the visitor's
+     * context dungeon.
+     *
+     * Deliberately a 302: the target depends on the visitor's own context dungeon, so a permanent
+     * redirect would get cached and pin one dungeon into every later visit.
+     */
+    public function show(CharacterClass $characterClass): RedirectResponse
```
Wotuu, line 45: Is this even used anymore? Can you maybe get rid of it or not?

app/Http/Controllers/Compendium/NpcCompendiumController.php
```diff
@@ -13,20 +13,46 @@
...
 {
-    public function index(): View
+    /**
+     * The NPC index without a dungeon in the URL - bounces to the canonical URL of the visitor's
+     * context dungeon.
+     *
+     * Deliberately a 302: the target depends on the visitor's own context dungeon, so a permanent
+     * redirect would get cached and pin one dungeon into every later visit.
+     */
+    public function index(): RedirectResponse
```
Wotuu, line 35: Same here. Is this being used? If it's being used, I prefer it not be used. We can maybe keep the endpoints, but don't use this in the code if possible.

app/Http/Controllers/Compendium/SpellCompendiumController.php
```diff
@@ -10,17 +10,43 @@
...
+    /**
+     * The spell index without a dungeon in the URL - bounces to the canonical URL of the visitor's
+     * context dungeon.
+     *
+     * Deliberately a 302: the target depends on the visitor's own context dungeon, so a permanent
+     * redirect would get cached and pin one dungeon into every later visit.
+     */
+    public function index(): RedirectResponse
     {
+        return redirect()->route('spell.compendium.index.dungeon', [
```
Wotuu, line 31: Same here as for the previous two comments.

Wotuu:
Verify this visually before merging This is a note to self.

app/Http/Controllers/Compendium/ClassCompendiumController.php
Wotuu: :robot: Checked — this is used. `resources/views/compendium/class/index.blade.php:21` links every class tile to `route('compendium.class.show', $characterClass)` (the dungeon-less form), which 302s here to the visitor's context-dungeon canonical URL. That's the whole point of the redirect (see the docblock added in this diff): a class tile link can't hardcode a dungeon, so it always needs a context-dependent bounce. Left as-is — no code change.

app/Http/Controllers/Compendium/NpcCompendiumController.php
Wotuu: :robot: Checked — this is used. The header nav dropdown (`resources/views/common/layout/header.blade.php:203`) and `resources/views/compendium/index.blade.php:13` both link to `route('npc.compendium.index')`, the dungeon-less form, which 302s here to the visitor's context-dungeon canonical URL. Same pattern as the Class controller — kept because nav links can't hardcode a dungeon. Left as-is — no code change.

app/Http/Controllers/Compendium/SpellCompendiumController.php
Wotuu: :robot: Checked — same as the previous two: `resources/views/common/layout/header.blade.php:204` and `resources/views/compendium/index.blade.php:21` both link to `route('spell.compendium.index')`, the dungeon-less form, which 302s here to the visitor's context-dungeon canonical URL. Genuinely used as the nav entry point, not dead code. Left as-is — no code change.

app/Http/Controllers/Compendium/ClassCompendiumController.php
Wotuu: :robot: Was only used from `compendium/class/index.blade.php`'s tile links, which routed through this 302 to reach the visitor's context-dungeon URL. Changed that link to build `compendium.class.show.dungeon` directly (context dungeon resolved once in the blade), so this redirect method is no longer used internally. Kept the `compendium.class.show` route + this handler registered, since `/compendium/class/{class}` is a previously-released URL shape that external bookmarks may still hit (confirmed by the cold review's legacy-URL sweep) - removing it would 404 those. The redirect target is still covered by `ClassCompendiumControllerTest` (asserts the 302 destination and the invalid-class 404).

app/Http/Controllers/Compendium/NpcCompendiumController.php
Wotuu: :robot: Was still used internally: the header nav dropdown, the `/compendium` landing page card, and the `compendium.npc.index` breadcrumb link all pointed at this bare route. Changed all three to build `npc.compendium.index.dungeon` directly with the visitor's context dungeon, so this handler is no longer reached from anywhere in the app. Kept `npc.compendium.index` registered - `/compendium/npc` is a previously-released URL shape (external bookmarks may still hit it), so removing the route would 404 them; only the 302 handler itself is now dead code path-wise. `compendium.activity.index` next to it in the header dropdown is a different case and intentionally left alone: it's not a pure context-dungeon bounce, it validates the target dungeon against the current season and can redirect to `home`, so there's no single canonical dungeon-scoped URL to link to directly from a static nav item.

app/Http/Controllers/Compendium/SpellCompendiumController.php
Wotuu: :robot: Same fix as the NPC controller: the header nav dropdown, the `/compendium` landing page card, and the `compendium.spell.index` breadcrumb link all pointed at this bare route. Changed all three to build `spell.compendium.index.dungeon` directly with the visitor's context dungeon, so this handler is no longer reached internally. Kept `spell.compendium.index` registered for the same backward-compat reason (`/compendium/spell` is a previously-released URL shape).

Wotuu:
:robot: Update on the three "is this used" threads above: my earlier replies said these redirect routes were used and left as-is, but a better fix landed afterward in fffdb6882 — internal navigation (header dropdown, `/compendium` cards, class-index tiles, breadcrumbs) now links straight to the dungeon-scoped canonical routes instead of the bare redirect endpoints, so in-app clicks skip the 302 hop entirely. The bare routes stay registered for backward-compatible bookmarks, matching what was asked ("keep the endpoints, but don't use this in the code if possible"). CI is running on this commit now.

routes/breadcrumbs.php
```diff
@@ -517,7 +520,10 @@
 
 Breadcrumbs::for('compendium.spell.index', static function (Generator $trail) {
     $trail->parent('home');
-    $trail->push(__('breadcrumbs.home.compendium.spell'), route('spell.compendium.index'));
+    // Link straight to the dungeon-scoped canonical route rather than the bare (no dungeon) endpoint -
```
Wotuu: Remove this comment, please.

routes/breadcrumbs.php
```diff
@@ -507,7 +507,10 @@
  */
 Breadcrumbs::for('compendium.npc.index', static function (Generator $trail) {
     $trail->parent('home');
-    $trail->push(__('breadcrumbs.home.compendium.npc'), route('npc.compendium.index'));
+    // Link straight to the dungeon-scoped canonical route rather than the bare (no dungeon) endpoint -
```
Wotuu: Remove this comment, please.

resources/views/compendium/index.blade.php
```diff
@@ -1,24 +1,31 @@
 <?php
 
+use App\Models\Dungeon;
+
 /**
  * @var array{npc: int, spell: int, class: int} $stats
  */
 
+// Link straight to the dungeon-scoped canonical routes rather than the bare (no dungeon) endpoints -
```
Wotuu: Remove this comment and any similar comments that are in this same pull request.

routes/breadcrumbs.php
Wotuu: :robot: Removed in 2789a40f5.

routes/breadcrumbs.php
Wotuu: :robot: Removed in 2789a40f5.

resources/views/compendium/index.blade.php
Wotuu: :robot: Removed all 5 instances of this comment (routes/breadcrumbs.php x2, resources/views/compendium/class/index.blade.php, resources/views/compendium/index.blade.php, resources/views/common/layout/header.blade.php) plus the matching one in CompendiumControllerTest.php — pushed in 2789a40f5.

PR #3984, opened 2026-08-12
#3980 Update MDT to v6.2.1

### LINKED ISSUE #3980: Update MDT to v6.2.1
:robot: A new Mythic Dungeon Tools release is available.

- Upstream: v6.2.1
- Currently pinned: v6.2.0-alpha5

Run the update via the update-mdt-package skill.

Wotuu:
:robot: Closes #3980

Bumps the `nnoggie/mythicdungeontools` composer package from **6.2.0-alpha5 → 6.2.1** (upstream commit `33a9cfb`) and reimports the mappings of all 8 dungeons in the current retail season. All 8 changed upstream, so each produced a new mapping version.

| Dungeon | New mapping version | `mdt_addon_version` |
| --- | --- | --- |
| kingsrest | v4 | 621 |
| templeofsethraliss | v3 | 621 |
| rubylifepools | v9 | 621 |
| altar_of_fangs | v4 | 621 |
| den_of_nalorakk | v4 | 621 |
| murder_row | v4 | 621 |
| the_blinding_vale | v5 | 621 |
| voidscar_arena | v4 | 621 |

No new MDT POI types or templates, and no missing map-icon translations. One NPC renamed upstream: `263181` "Egg Marker" → "Egg". No new floors or dungeons, so no map tiles are needed in the assets repo.

### The config bump happens *before* the reimport, on purpose

`MappingService` stamps `mapping_versions.mdt_addon_version` from `config('keystoneguru.mdt.version')`. The skill used to bump that config *after* the imports, which is exactly how #3944 ended up stamping its new mapping versions with the previous release's addon version — and that silently mis-binds imported MDT strings through the release-date tie-break in `getMappingVersionForMdtAddonVersion()` (#3380). Cold-review finding #1 on #3944. All 8 new rows verified as `621` against the DB rather than assumed.

`621` looks "lower" than alpha5's `6205`, which is correct and not a bug: the integer is `version:gsub("%.", "")` exactly as MDT computes it, it has never been monotonic across MDT's version schemes, and every comparison in `MappingService` happens on release dates for that reason. `mdt:syncaddonversions --refresh` maps `621 → 2026-08-11`.

### Incidental fixes

Three things broke on the way that are not MDT-version-specific; fixed here rather than deferred.

1. **`MappingExportService::serializeSpells()` leaked a computed accessor.** #3951 added `wowhead_tooltip_data` to `Spell::$appends` but not to the `makeHidden()` list next to the other two appended accessors, so every `mapping:save` rewrote all ~7900 entries of `spells.json` with a value the seeder cannot load back. New `MappingExportServiceTest` guards *every* entry of `Spell::$appends`, not just the three that exist today, so the next added accessor fails the test instead of the diff.
2. **`spell:assignmissingdungeons` crashed** on `$npcSpell->spell` being `null`. The importer deliberately records spells the `spells` table does not have yet ("Save spells that we don't know of yet" in `MDTMappingImportService`) — this import added 206 such rows. The command now skips them.
3. **`ClassCompendiumControllerTest::show_givenReflectedSpellInDungeon_displaysSpellAndCastingNpc`** picked a dungeon first and then assumed it held an NPC with neither a characteristic nor a spell. MDT 6.2.1 gave every NPC in `altar_of_fangs` (the deterministic pick) a spell, so there was no candidate left. The NPC requirement is now part of the dungeon selection via `findDungeon()`'s `resolve` closure — the same fix shape as the `AjaxKillZoneControllerTest` flake in #3944.

### Midnight Season 2: start date, and the manual mapping edits

Folded in at Wotuu's request. Season 18 (Midnight S2) `start` moves **2026-10-19 → 2026-08-17** (a Monday, like every other season start) and `active` **false → true**, and its `season_dungeons` are re-exported in the admin-panel order. Applied to the database and exported with `mapping:save` rather than hand-edited, per the `SeasonsSeeder` docblock.

Also carries manual admin-panel mapping edits to `altar_of_fangs`: enemy `134904` moved from pack `26108` to `26125`, three enemy pack polygons reshaped, and a new admin map icon on floor 1.

> ⚠️ **@Wotuu — two things to check before merging.**
> 1. `active: true` is included because it was part of the state you'd built in the admin panel, not because the date change required it. It is the flag that advertises S2 in the header nav and the Create Route dungeon list. Say the word and I'll set it back to `false`.
> 2. Edits you made *after* the `mapping:save` that captured the above were lost to a full `db:seed` I ran and could not be recovered (no binlog locally). Only `altar_of_fangs` was in that export, so the other seven dungeons are worth a look.

**Moving the date broke eight tests, and they were genuinely wrong rather than brittle.** Four classes (`HeaderComposerTest`, `CreateRouteFormComposerTest`, `DungeonExploreControllerTest`, `DungeonRouteControllerCreateSeasonGateTest`) each create their own upcoming season and then assert on a question about *global* state — which season is soonest, whether any is advertised at all. That only worked while no seeded season happened to be upcoming nearby; with S2 five days out the seeded season answers instead, and `compose_givenAnInactiveUpcomingSeason_omitsTheNextSeasonEverywhere` asserts a global *absence* that no date the test picks could satisfy. Verified the date alone breaks them independent of `active`.

The new `IsolatesSeededUpcomingSeasons` trait parks every seeded upcoming season 50 years out and deactivates it for the test, restoring in `tearDown` — the test DB persists between runs, so a leaked park would corrupt every later run, not just the one test.

### Skill update

`.claude/skills/update-mdt-package` gains: the config-bump ordering above; the fact that the tracking issue is now bot-filed; the missing `docker compose exec` prefix on Step 7; the worktree/draft-MR flow (it still said "do not commit"); the expected-and-recoverable import log lines; and the fork-sync failure below.

The fork's `master` is now protected by a repository ruleset, so `gh repo sync Wotuu/MythicDungeonTools` fails with `422 Repository rule violations found`. **This is not a blocker** — forks share an object store, so `33a9cfb` is already reachable in the fork by SHA and the dist archive returns 200, which is the path CI and production use (`config.preferred-install` is `dist`). What it does break is a *locally* source-installed `vendor/nnoggie/mythicdungeontools`: composer keeps the existing install type on update and a clone of the fork genuinely lacks the commit. Removing the directory and reinstalling from the lock fixes it. Worth knowing that the fork can no longer be fast-forwarded, though — @Wotuu, if that ruleset is unintentional it's worth relaxing.

### Verification

- Full test suite before the bump: **1 failed, 1976 passed** — `MapTilesExistenceTest`, the known worktree asset-mount gap.
- Full test suite after everything above, season change included: **1 failed, 1979 passed** — same single failure, no regressions. Season 18's row verified unchanged after the run, confirming the test isolation restores what it parks.
- `composer run fix` and `composer run analyse` (PHPStan) clean.
- `mapping:save` confirmed idempotent: a second no-op run re-dirties only the three files that always churn on trailing newlines (`seasondata/affix_groups.json`, `seasondata/seasons.json`, `dungeondata/spells.json`), which are excluded from this diff.

Screenshots of the reimported dungeons follow.


Wotuu:
:robot: **Visual verification** — the two dungeons with the largest upstream Lua diff, explore page, master (before) vs this branch (after).

Both pages return 200 with no page errors; the single `SyntaxError: "undefined" is not valid JSON` console line is the documented benign `hidden_map_object_groups` cookie read that every map page emits.

### The Blinding Vale — the big one

| Before (master, MDT 6.2.0-alpha5) | After (this branch, MDT 6.2.1) |
| --- | --- |
| before | after |

This one changes a lot, and it is the change worth your eyes. Upstream went from **276 enemies / 0 enemy packs** to **168 enemies / 63 enemy packs** — the previous import had no packs at all, so this looks like MDT replacing a work-in-progress mapping with a real one. Enemy forces required is unchanged at 655.

### Den of Nalorakk

| Before (master) | After (this branch) |
| --- | --- |
| before | after |

116 → 170 enemies, still 0 enemy packs (unchanged from before — upstream has not packed this one yet).

### Enemy / pack counts for all 8

| Dungeon | Enemies | Packs | Enemy forces required |
| --- | --- | --- | --- |
| the_blinding_vale | 276 → **168** | 0 → **63** | 655 → 655 |
| den_of_nalorakk | 116 → **170** | 0 → 0 | 729 → 729 |
| kingsrest | 101 → **98** | 29 → **25** | 654 → **608** |
| templeofsethraliss | 128 → 128 | 43 → **42** | 661 → **649** |
| rubylifepools | 166 → **150** | 34 → **35** | 531 → **553** |
| altar_of_fangs | 160 → **157** | 46 → 46 | 836 → **817** |
| murder_row | 221 → 221 | 56 → 56 | 690 → 690 |
| voidscar_arena | 218 → **214** | 63 → **62** | 738 → 738 |

### Note on the baseline

The main stack's header shows `Season 2`, `Heatmaps`, `Next season` and per-dungeon affix letters that the worktree's freshly-seeded DB does not. That is a worktree-environment difference, **not** something this PR causes: `skyreach` (not reimported) renders pixel-identical enemies on both sides while showing the same header difference.

.claude/skills/update-mdt-package/SKILL.md
```diff
@@ -276,7 +343,15 @@ formatting of the rewritten `npcs.php` files, per the project's finishing-up con
...
-4. Leave all changes uncommitted for the user to review, and leave the tracking issue **open**
-   (don't close it yourself unless asked).
+3. **Verify visually.** Enemy/pack positions are rendered output, so this is *not* a backend-only
+   change — screenshot the explore page of the dungeons whose mapping actually changed with the
+   `headless-browser-verify` skill (A/B against the main stack as baseline) and post them on the
+   MR. Also remind the user to browse the explore pages themselves.
+4. **Also check what the suite can't see:** `MapTilesExistenceTest` fails on its *first* dungeon
+   in a worktree (the assets dir isn't mounted), so it never reaches the newly imported ones and
+   cannot tell you a new floor is missing map tiles. Confirm directly that
+   `git status database/seeders/dungeondata/**` lists no changed `floors.json` / `dungeon.json` —
```
Wotuu: :robot: This map-tile safety check can never fire: there is no `floors.json` or `dungeon.json` anywhere under `database/seeders/dungeondata/` — verified with `git ls-files database/seeders/dungeondata | grep -iE "floors|dungeon\.json"` (only `dungeons.json`, `npcs.json`, `spells.json`, `mapping_versions.json`, `mapping_commit_logs.json` at the root, and per-floor `enemies/enemy_packs/map_icons/floor_unions/...` files below it). Floors are stored **nested inside `database/seeders/dungeondata/dungeons.json`** (each dungeon object has a `floors` array — e.g. `altar_of_fangs` has 4).

So a future run following this step will always see "no changed floors.json / dungeon.json" and conclude no map tiles are needed, which is exactly the failure mode the step was added to prevent. Suggest rewording to check `database/seeders/dungeondata/dungeons.json` for changes (and specifically for added entries in a dungeon's `floors` array).

(For this PR the conclusion still holds — `dungeons.json` is genuinely unchanged — but the instruction itself is wrong and would mislead every future run.)

.claude/skills/update-mdt-package/SKILL.md
```diff
@@ -276,7 +343,15 @@ formatting of the rewritten `npcs.php` files, per the project's finishing-up con
...
+   if it does, the assets repo needs the tiles and the update is not complete.
+5. Commit and push the worktree branch and open the **draft** MR (`Closes #<issue>` in the body),
```
Wotuu, line 363: :robot: The rewrite left two contradictions with the old text that survive elsewhere in this file — a future run will get conflicting instructions.

1. **Step 1 still says do not commit** (unchanged lines ~88–91):
> **Record the issue number** it prints — reference it in the final summary [...] and (**if they later ask you to commit**) use it as the `#<number>` prefix [...] Do **not** commit or close the issue yourself unless asked.

That directly contradicts the new Conventions bullet ("commit the whole run [...] as you go and open a draft MR at the end") and this Step 8 item 5. "Do not close the issue yourself" is still correct; the "do not commit / if they later ask you to commit" half is stale and should go.

2. **The tracking-issue body checklist (line ~81) still lists the config bump after the reimport steps:**
```
- [ ] Reimport each retail-season dungeon mapping
- [ ] Handle any new POI types/templates and missing map-icon translations
- [ ] Run per-dungeon tests (rewrite stale CorrectEvents fixtures)
- [ ] Bump `config/keystoneguru.php` MDT version
```
That is precisely the ordering the new Step 2.5 exists to prevent (#3944 stamping `mdt_addon_version` from the previous release). Since the checklist is what a reader ticks through, it should list the config bump directly after the composer bump.

Wotuu:

Both are in `.claude/skills/update-mdt-package/SKILL.md`. The hand-written code changes verified clean:

- **(a) `wowhead_tooltip_data` makeHidden + test** — correct and complete. `makeHidden()` does suppress `$appends` entries in `toArray()`; `Spell::$appends` is exactly the three hidden names; the committed `spells.json` on master is still clean (no leaked keys), so this is preventive rather than a data repair; `serializeNpcs()` is unaffected (`Npc::$appends` is only `enemy_portrait_url`, already hidden, and `npcs.json` contains none of the accessor keys). The generic `getAppends()`-driven test does guard future additions.
- **(b) skipping null spells in `AssignMissingSpellDungeons`** — not masking a bug. `MDTMappingImportService` (lines 251–308) deliberately `NpcSpell::insert()`s rows for spell IDs with no `spells` row, and `npc_spells` is not exported to the seeders at all, so nothing dangling ships in this diff. A bare `return` inside `each()` is the right skip.
- **(c) `ClassCompendiumControllerTest`** — still tests what it claims. The `resolve` closure runs the identical NPC query (no characteristic, no spell) against the same `MappingVersion` that is returned, so the "passes for the wrong reason" guard is intact; the only change is that the dungeon scan continues past a dungeon with no candidate instead of failing.
- **(e) `mdt_addon_version` 621** — verified, not taken on faith. `MDTAddonVersionService::versionStringToAddonVersion()` is `preg_replace('/\\D/', '', $version)`, and the only consumers of `config('keystoneguru.mdt.version')` are that stamp and the footer string in `ViewService::formatAppNameAndVersion()`. Every comparison in `MappingService::getMappingVersionForMdtAddonVersion()` resolves through `getReleaseDate()`; `621` is new in `addon_versions.json` (no key collision, dates strictly increasing), and the export string service never emits an `addonVersion`.
- **Generated mapping data** — spot-checked programmatically: all 8 new `mapping_versions` rows carry `mdt_addon_version: 621`, every new enemy references an existing `npc_id` and a pack belonging to the same mapping version (0 dangling), no NPCs added/removed (so no missing `lang/en_US/npcs.php` keys), and the only `npcs.json` deletions are health value updates.

.claude/skills/update-mdt-package/SKILL.md
Wotuu: :robot: Fixed in 7be16a5 — good catch, the check was vacuous. Confirmed there is no `floors.json` anywhere under `database/seeders/dungeondata/`; each dungeon object in `dungeondata/dungeons.json` carries a nested `floors` array (e.g. `the_blinding_vale` has 2). The step now checks that one file, with a grep for added `"index"` keys.

The PR's own conclusion is unaffected: `dungeons.json` is genuinely unchanged in this diff, so no new floors and no missing map tiles.

.claude/skills/update-mdt-package/SKILL.md
Wotuu: :robot: Fixed in 7be16a5. Both contradictions removed:

- Step 1's trailing note no longer says "do not commit unless asked" / "if they later ask you to commit" — it now just records what the issue number is used for (commit subject prefix, MR title, `Closes #<issue>`), leaving the worktree/draft-MR flow in the Conventions block and Step 8 as the single source of truth.
- The tracking-issue checklist now lists the config bump directly after the composer bump, with the reason inline, and gained the addon-version-map refresh step it was missing.

Also corrected while in there: the composer `require` constraint is `^6.0@alpha`, not `^6.0` — the step said the latter.

tests/Feature/Traits/IsolatesSeededUpcomingSeasons.php
```diff
@@ -0,0 +1,80 @@
...
+     * Parks every currently-upcoming season. Call from setUp(), before the test creates its own season -
+     * seasons created afterwards are deliberately untouched, they are the test's subject.
+     */
+    protected function hideSeededUpcomingSeasons(): void
+    {
+        $this->parkedUpcomingSeasons = [];
+
+        $upcomingSeasons = DB::table('seasons')
+            ->select(['id', 'start', 'active'])
+            ->where('start', '>', now()->toDateTimeString())
```
Wotuu: :robot: **This predicate doesn't match the app's definition of "upcoming", and the gap opens on 2026-08-17.**

The trait parks `start > now()`. But the resolution these tests assert on is `Expansion::nextSeason()` (`app/Models/Expansion.php:191-199`), which selects on

```
DATE_ADD(DATE_ADD(`start`, INTERVAL ? day), INTERVAL ? hour) >= now()
```

with the region's `reset_day_offset` / `reset_hours_offset`. `GameServerRegionsSeeder` seeds those at **1d15h – 2d23h**. So for roughly 1–3 days *after* a season's `start`, that season is still the app's `nextSeason` while no longer matching `start > now()` — it is not parked.

Concretely for season 18 (`start` 2026-08-17, `active: true`): from 2026-08-17 until ~2026-08-20 it stays eligible, and since `nextSeason()` is `orderBy('start')->limit(1)`, its earlier start beats the `now()->addWeek()` season each test creates. All four classes regress into exactly the failure mode this trait was written to remove — and then heal by themselves a couple of days later, which is the worst shape of flake to debug.

Note that the "full suite green" verification in the PR body was necessarily run before 2026-08-17, so it can't cover this window.

I'm deliberately not prescribing the replacement predicate: simply widening to the adjusted window also parks a season that is *current* for other regions during those same days, which may break assertions that need a current season to exist. `Expansion::nextSeason()` is the predicate of record — worth deciding explicitly how the trait should mirror it.

tests/Feature/Traits/IsolatesSeededUpcomingSeasons.php
```diff
@@ -0,0 +1,80 @@
...
+            ->get();
+
+        foreach ($upcomingSeasons as $season) {
+            $this->parkedUpcomingSeasons[(int)$season->id] = [
```
Wotuu, line 59: :robot: **The park is not idempotent, so a single missed restore becomes permanent and silent.**

The capture takes whatever is in the row as the "original". If a restore is ever skipped, the row is left at `start = now()+50y`, which still satisfies `start > now()` — so the **next** run parks it again and captures `+50y / active=0` as the original, then faithfully restores that. From then on the seeded season is gone from the persistent test DB for good, with nothing anywhere reporting it. Given the DB persists between runs (no `DatabaseTransactions`/`RefreshDatabase` in this suite), that's a corruption that survives until someone re-seeds by hand.

There are two distinct holes and they need two distinct fixes:

1. **Restore never runs at all.** Each consuming class has to remember to write a `tearDown()` that calls `restoreSeededUpcomingSeasons()`; four have been added by hand here, and the fifth user of the trait will silently leak. `$this->beforeApplicationDestroyed(fn () => $this->restoreSeededUpcomingSeasons())` registered inside `hideSeededUpcomingSeasons()` closes that, and collapses the four hand-written `tearDown()` overrides into nothing.
2. **Restore *can't* run.** PHPUnit does not call `tearDown()` when `setUp()` throws — so an exception partway through the park loop (some rows already parked) leaks regardless, as does a fatal error or a killed run. The callback above doesn't help there either.

Hole 2 is what makes the non-idempotency matter, and only a guard fixes it: refuse to capture/park a row whose `start` is already ~50 years out (or otherwise recognise the parked sentinel and treat it as "already parked, not mine to capture"). That turns a leaked park into something a later run can detect instead of laundering into the recorded original.

Wotuu:

tests/Feature/Traits/IsolatesSeededUpcomingSeasons.php
Wotuu: :robot: Fixed in 75636cf — this was a real time bomb, thank you. The trait now mirrors `Expansion::nextSeason()` instead of approximating it: it selects `DATE_ADD(start, INTERVAL ? hour) >= now()` using the widest offset across all regions, read from `game_server_regions` (`MAX(reset_day_offset * 24 + reset_hours_offset)` = 71h, KR) rather than hardcoded, so seeding a new region cannot silently reopen the window.

Verified rather than assumed, by moving season 18's start two days into the past to simulate the window:

```
max region offset hours = 71
old predicate parks: [7]          <- misses 18
new predicate parks: [18,7]
app nextSeason(kr) for Midnight = 18
```

So the failure was exactly as you described, and the four test classes pass in that simulated window with the new predicate. Season 18 was restored to 2026-08-17 afterwards.

On the concern you flagged but did not prescribe: widening does mean a season that is concurrently *current* for an earlier region gets parked during those few days. That is the behaviour these tests want — they take ownership of the upcoming-season set — and the alternative (leaving it unparked) is the bug itself.

tests/Feature/Traits/IsolatesSeededUpcomingSeasons.php
Wotuu: :robot: Fixed in 75636cf, taking both halves of this.

**Restore now runs without a `tearDown()` override.** `beforeApplicationDestroyed()` is registered *before* anything is parked, so it covers a throw part-way through the park loop as well as a class that forgets the override. All four hand-written `tearDown()` overrides are gone as you suggested.

**Re-capture is now impossible rather than merely unlikely.** Parked rows carry a fixed sentinel start (`2999-01-01 00:00:00`) instead of a relative one, so a row left behind by a run that could not restore at all (fatal error, killed process) is recognisable. Encountering one throws with an explicit instruction to re-seed rather than absorbing the parked values as originals. Tested by planting the sentinel on season 7:

```
Season 7 is still parked by an earlier test run that could not restore it; its real start and
active flag are unrecoverable. Re-seed the test database (php artisan db:seed) before running
this suite again.
```

Season 7 was restored to 2050-12-28 afterwards, and the full suite is back to its 1 pre-existing failure with seasons 7 and 18 verified unchanged after the run.

Wotuu:
Can you create a follow-up issue where I have a to-do list of all eight dungeons in Season 2? For each dungeon, I should check:
- whether there is a dungeon start icon
- whether the floor unions are set properly
- whether the Blinding Veil needs to be checked properly, additionally, since I think that's wrong in this pull request


That's it.

app/Console/Commands/Spell/AssignMissingSpellDungeons.php
```diff
@@ -47,9 +47,18 @@ public function handle(): int
 
                     // For each spell this NPC can cast, ensure that that spell is also assigned to the same dungeon(s)
                     $npc->npcSpells->each(function (NpcSpell $npcSpell) use ($npc) {
-                        /** @var Spell $spell */
+                        /** @var Spell|null $spell */
                         $spell = $npcSpell->spell;
 
+                        // The MDT import deliberately records every spell an NPC casts, including ones the
```
Wotuu, line 53: MDT does, but we don't want to do anything with this information. We gathered this information ourselves with the parsing of combat log data. If they are still a part of the code that imports NPC spell data or couples spells to dungeons or NPCs, please ensure that it doesn't do that in a follow-up issue.

tests/Feature/View/CreateRouteFormComposerTest.php
```diff
@@ -31,6 +34,10 @@ protected function setUp(): void
         // the array store the tests run on - survives between test runs. Seasons created below would otherwise be
         // invisible to the composer, or worse, stay visible to whatever runs next.
         $this->flushSeasonCaches();
+
+        // These tests assert on which season is upcoming site-wide, so a seeded upcoming season would answer
+        // for them - and no assertion of "nothing is advertised" could hold at all.
+        $this->hideSeededUpcomingSeasons();
```
Wotuu, line 40: Is this enough to also clean or restore the parked seasons?

Wotuu:
I added the can merge label. If there are no changes required, you can merge it.

Wotuu:
:robot: Filed the follow-up checklist issue: #3988 — one item per Season 2 dungeon (start icon present, floor unions set correctly), with The Blinding Vale called out specifically for manual verification since you flagged it as possibly wrong in this PR.

app/Console/Commands/Spell/AssignMissingSpellDungeons.php
Wotuu: :robot: Confirmed - MDT-sourced coupling is still live and untouched by this PR: `MDTMappingImportService::importNpcsDataFromMDT()` still inserts `NpcSpell` rows straight from the MDT export (the `NpcSpell::insert($npcSpellsAttributes)` call), and `importSpellDataFromMDT()` still couples spells to dungeons via `SpellDungeon` (and creates placeholder `Spell` rows for unknown spell IDs). Neither is something this PR added - they're pre-existing behaviour that #3980 only had to work around (the null-spell guard in `AssignMissingSpellDungeons.php`, since this reimport left 206 `NpcSpell` rows pointing at spell IDs MDT knows about but the combat-log pipeline doesn't yet).

Given the blast radius (touches every future mapping reimport, and needs verifying nothing else currently depends on MDT seeding those placeholder `Spell`/`SpellDungeon` rows), I filed a follow-up rather than changing it in this already-large PR: #3989.

tests/Feature/View/CreateRouteFormComposerTest.php
Wotuu: :robot: Yes, it's enough. `hideSeededUpcomingSeasons()` (called at line 40) registers `beforeApplicationDestroyed(fn() => $this->restoreSeededUpcomingSeasons())` internally (tests/Feature/Traits/IsolatesSeededUpcomingSeasons.php:46), before parking anything - so restore runs automatically regardless of whether the test passes, fails, or throws partway through setUp(), with no tearDown() override needed in this class. That's exactly the fix from the earlier thread on this trait (75636cf): the four consuming test classes, including this one, deliberately have no explicit tearDown() calling restoreSeededUpcomingSeasons() anymore - I checked all four and confirmed none of them do. Re-ran this file to confirm: 2 passed.

PR #3986, opened 2026-08-12
#3983 Poll combat log runs across key level bands instead of a flat minimum of 10

### LINKED ISSUE #3983: Poll combat log runs across key level bands instead of a flat minimum of 10
:robot:

## Background

`combatlog:pollruns` currently polls Raider.IO with a single flat `mythicLevel >= 10`
(`COMBAT_LOG_POLLING_MYTHIC_LEVEL_MIN`, `config/keystoneguru.php:535`). Budgets are tracked per
criterion model (`Dungeon`, `CharacterClassSpecialization`) per day in
`combat_log_parsing_criteria`, thresholds currently 300 on production.

The result is a sampling bias: L10-L16 dominates the supply, so that is nearly all we parse. For
the NPC Compendium what matters is not the key level itself — a L2 observation and a L24
observation are equally valid facts — but the **behaviour of the players in the run**. Low keys
try silly things (good coverage of "did anyone even try X"), high keys only ever do the thing that
works. We want coverage of the whole spectrum, plus a guarantee that the very top keys are always
represented.

Explicitly **out of scope**: storing the key level on observations. The compendium does not need to
know which level a fact came from.

### Measured supply (yesterday, full UTC day, `hasAutoRoute=1`, timed, season-mn-1)

| band | runs/day | per dungeon (÷8) |
|---|---|---|
| 2-6 | 1,231 | 154 |
| 7-11 | 7,667 | 958 |
| 12-16 | 7,929 | 991 |
| 17-21 | 2,205 | 276 |
| 22+ | 542 | 68 |
| 23+ | 219 | 27 |

Every band can sustain a 60/day per-dungeon sub-budget.

### Verified API behaviour

- `mythicLevel[0][gte]` **and** `[lte]`, and `[eq]`, are all honoured by
  `https://raider.io/api/search-advanced`.
- `total.value` reflects the active filter and is capped at 10,000.
- An empty result set returns `"total": 0` as a **scalar**, not `{"value": n}` —
  `RaiderIOApiService::searchAdvancedRuns()` currently assumes the object shape and must handle both.

### 1. Key level bands as a criterion dimension

Add a band column to `combat_log_parsing_criteria` and make it part of the unique key, so budgets
become per `(model_class, model_id, band)` and one band cannot spend another's budget.

- Bands are 5 wide: 2-6, 7-11, 12-16, 17-21, ... up to the dynamic top band floor.
- Per-band threshold ~60 (config-seeded, not inherited). **Gotcha:**
  `CombatLogParsingCriteriaService::getDefaultThreshold()` picks the most recent threshold row per
  *model class*, ignoring the new band — left as-is, all 5 bands inherit the production 300 and
  nominal daily volume goes from ~2,400 to ~12,000 runs. Thresholds must become per-band and
  config-driven.
- Total budget stays roughly where it is today: 8 dungeons x 5 bands x 60 ~= 2,400 runs/day, and
  40 specs x 5 bands x 60 / 5 specs-per-run ~= 2,400.

### 2. One band per hour, rotating

Polling every band for every criterion each hour would take API calls from ~48/hour to ~240/hour.
Rotate: each hourly run polls a single band, so each band is polled 4-5 times a day and call
volume is unchanged.

### 3. Dynamic top band, always parsed

Runs at `max_key_level - 2` and above bypass the budget entirely and are always dispatched
(~540/day at floor 22).

- `max_key_level` = the highest level with **at least ~25 runs in the last 7 days** (noise floor).
  Highest-nonzero is wrong: over the last 7 days L25 had only 4 runs worldwide while L24 had 292,
  so highest-nonzero would silently shrink the always-parse set by 60%.
- Determined by probing `total` with `limit=1` walking down from the season cap; cached per
  season+week and recomputed on a schedule (affix week and season progression shift it).
- Top-band runs **bypass `shouldParse()`** and must **not** increment the dungeon/spec counters of
  the spread bands — otherwise ~540 runs/day eat ~22% of the dungeon budget and starve the low
  bands, which is the exact bias this issue exists to remove. Count them into their own criterion
  row for observability only.

## Key files

- `app/Console/Commands/CombatLog/PollCombatLogRunsCommand.php`
- `app/Service/CombatLog/CombatLogParsingCriteriaService.php` (+ interface)
- `app/Models/CombatLog/CombatLogParsingCriterion.php`
- `app/Service/RaiderIO/Dtos/SearchAdvancedRunsFilter.php`
- `app/Service/RaiderIO/RaiderIOApiService.php`
- `config/keystoneguru.php` (`raider_io.combat_log_polling`)
- new migration in `database/migrations_combatlog/`

## Definition of done

- Migration is additive and backward compatible with currently-running code.
- Feature tests cover: band rotation picks the expected band per hour, per-band budgets are
  independent, top-band runs bypass the budget and do not increment spread-band counters, the
  max-key-level probe handles the scalar `total: 0` response.

Wotuu:
:robot:

Closes #3983

## What this changes

`combatlog:pollruns` asked Raider.IO for every run at level 10 and up, so what got parsed was
whatever the 10-16 range happened to produce — by far the most populous part of the ladder. For the
NPC Compendium the key level of an observation carries no meaning of its own, but the behaviour of
the players in the run does: low keys reveal whether anything was even attempted, high keys only
ever show the thing that works. Coverage of the whole spectrum is what is worth having, so this
does **not** store the key level anywhere — it only changes which runs get sampled.

- **5-wide bands, one per hour on rotation.** `2-6`, `7-11`, `12-16`, `17-21` at the moment.
  Rotating keeps the API call volume exactly where it was (~48/hour) instead of multiplying it by
  the band count.
- **A budget per (criterion, band).** The band is part of the `combat_log_parsing_criteria` key, so
  no band can spend another's. Default 60 per band per dungeon/spec — the same ~2,400 runs/day the
  current 300-per-criterion setting produces.
- **The top band is always parsed.** Everything from `max_key_level - 2` upwards bypasses the
  budgets entirely, in a single all-dungeons query per hour. It counts into its own criterion rows
  so it cannot starve the bands below it.
- **The max key level is probed weekly** by walking `total` down from a ceiling, with a noise floor
  of 25 runs over 7 days so a handful of world-record attempts cannot drag it up. Cached per
  season + ISO week; no scheduled command needed.

## Measured supply this season (yesterday, full UTC day)

| band | runs/day | per dungeon |
|---|---|---|
| 2-6 | 1,231 | 154 |
| 7-11 | 7,667 | 958 |
| 12-16 | 7,929 | 991 |
| 17-21 | 2,205 | 276 |
| 22+ | 542 | 68 |

Every band sustains a 60/day per-dungeon budget. Probing live from the branch resolves to
`max=24`, top band `22+`, spread bands `2-6, 7-11, 12-16, 17-21` — about 540 always-parsed runs a
day on top of the budgeted ~2,400.

## Two traps this had to avoid

- **Threshold inheritance.** `getDefaultThreshold()` resolved the threshold per *model class*, so
  adding a band dimension would have handed all 5 bands the 300 currently configured on production
  — a nominal 12,000 runs/day. It now resolves per (class, band), which also keeps a hand-edited
  threshold sticky for the band it was edited on.
- **Sort stability.** The search sorted on `hasAutoRoute`, which the filter already pins to 1 — a
  no-op ordering, so repeated polls of the same filter got back the exact same page, every run on
  it already parsed. Harmless while the filter was wide, fatal once bands narrow the result set to
  ~68 runs/dungeon/day. Now sorted by `completedAt desc`, so each poll leads with what appeared
  since the last one.

## Notes for review

- `COMBAT_LOG_POLLING_MYTHIC_LEVEL_MIN` is **no longer read** — deliberately a new config key
  rather than a reused one, because a stale `=10` in the production env would have silently killed
  the low bands. Worth removing from the production env. It is not in `.env.example`, and the new
  `COMBAT_LOG_POLLING_BAND_*` / `_TOP_BAND_*` vars all have working defaults.
- The migration is additive (two columns, unique index widened by one column) and backward
  compatible: old code writes `mythic_level_min = 0` and stays unique exactly as before. Existing
  rows get `mythic_level_max` backfilled to 0, because a null there marks the always-parsed top
  band.
- `CombatLogPollingBandService` logs its one operational warning through the `Log` facade rather
  than a `{Service}Logging` companion — the sibling `CombatLogParsingCriteriaService` has none
  either, but flagging it in case a companion is preferred.
- `getModelsEligibleForPolling()`'s top-band branch is defensive: `pollTopBand()` issues one
  all-dungeons query and never calls it today.

## Verification

- `--group=CombatLog` (318) and `--group=RaiderIO` (13) green; `composer run fix` and
  `composer run analyse` clean.
- Migration applied, rolled back and re-applied against the worktree DB.
- Band service smoke-tested against the live Raider.IO API (see numbers above).

### DIFF
--- a/app/Console/Commands/CombatLog/PollCombatLogRunsCommand.php
+++ b/app/Console/Commands/CombatLog/PollCombatLogRunsCommand.php
@@ -11,8 +11,10 @@ use App\Models\Dungeon;
 use App\Models\Interfaces\CombatLogCriterionModelInterface;
 use App\Models\Season;
 use App\Service\CombatLog\CombatLogParsingCriteriaServiceInterface;
+use App\Service\CombatLog\CombatLogPollingBandServiceInterface;
 use App\Service\CombatLog\Dtos\CombatLogParsingCriterionCheck;
 use App\Service\CombatLog\Dtos\CombatLogRunContext;
+use App\Service\CombatLog\Dtos\KeyLevelBand;
 use App\Service\RaiderIO\Dtos\SearchAdvancedRun;
 use App\Service\RaiderIO\Dtos\SearchAdvancedRunsFilter;
 use App\Service\RaiderIO\RaiderIOApiServiceInterface;
@@ -29,6 +31,7 @@ class PollCombatLogRunsCommand extends Command
 
     public function __construct(
         private readonly CombatLogParsingCriteriaServiceInterface $criteriaService,
+        private readonly CombatLogPollingBandServiceInterface     $bandService,
         private readonly RaiderIOApiServiceInterface              $raiderIOApiService,
         private readonly SeasonServiceInterface                   $seasonService,
     ) {
@@ -48,26 +51,17 @@ class PollCombatLogRunsCommand extends Command
         $combatLogVersion = array_key_last(CombatLogVersion::RETAIL_ALL);
 
         $windowDays      = (int)config('keystoneguru.raider_io.combat_log_polling.completed_at_window_days');
-        $mythicLevelMin  = (int)config('keystoneguru.raider_io.combat_log_polling.mythic_level_min');
         $limit           = (int)config('keystoneguru.raider_io.combat_log_polling.limit');
         $completedAtFrom = Carbon::now()->subDays($windowDays);
 
-        $allModelsByClass      = [];
-        $eligibleModelsByClass = [];
-        $criteriaSummary       = [];
-
-        foreach (array_keys(CombatLogParsingCriterion::VALID_CRITERIA) as $modelClass) {
-            $allModelsByClass[$modelClass]      = $this->criteriaService->getAllModelsForCriteria($modelClass, $season);
-            $eligibleModelsByClass[$modelClass] = $this->criteriaService->getModelsEligibleForPolling($combatLogVersion, $modelClass, $season);
-
-            $shortClass        = class_basename($modelClass);
-            $allCount          = $allModelsByClass[$modelClass]->count();
-            $eligibleCount     = $eligibleModelsByClass[$modelClass]->count();
-            $criteriaSummary[] = sprintf('%s=%d/%d eligible', $shortClass, $eligibleCount, $allCount);
-        }
+        // Only one band is polled per hour: polling all of them every hour would multiply the
+        // number of Raider.IO calls by the band count for no extra coverage.
+        $spreadBand = $this->bandService->getSpreadBandForHour($season, Carbon::now()->hour);
+        $topBand    = $this->bandService->getTopBand($season);
 
         $force = (bool)$this->option('force');
 
+        /** @var array<int, true> $existingRunIds */
         $existingRunIds = $force ? [] : ParsedCombatLog::query()
             ->whereNotNull('run_id')
             ->pluck('run_id')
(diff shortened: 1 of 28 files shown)


Wotuu:
:robot: Admin criteria page, same seeded rows rendered by both views.

**Before** — every dungeon appears once per band with nothing saying which, and the always-parsed top band shows a nonsensical editable `27 / 0` threshold:

before

**After** — grouped per band, and the top band is a read-only count because it has no budget to configure:

after

app/Service/CombatLog/CombatLogParsingCriteriaService.php
```diff
@@ -88,28 +99,47 @@ public function getModelsEligibleForPolling(int $combatLogVersion, string $model
...
+     * volume that gets parsed each day by the number of bands.
+     */
+    private function getDefaultThreshold(string $modelClass, KeyLevelBand $band): int
     {
-        return CombatLogParsingCriterion::query()
+        if ($band->isTopBand()) {
+            return 0;
+        }
+
+        $lastConfiguredThreshold = CombatLogParsingCriterion::query()
```
Wotuu, line 138: :robot: **Top-band rows can poison a spread band's default threshold to 0.** This inheritance query matches on `mythic_level_min` only, but top-band rows (created by `recordParsed()` with `threshold = 0`) share that column. With today's live values (max=24 → top band min 22, which sits exactly on the `2 + 5k` spread-band grid) rows with `mythic_level_min = 22, threshold = 0` accumulate daily. The moment the max key level rises to ≥26 later in the season, `22-26` becomes a spread band, this query picks up yesterday's top-band row (`ORDER BY date DESC`), and the new band's row is created with `threshold = 0` — `shouldParse()` then sees `count (0) >= threshold (0)` and the band never polls. It self-perpetuates: each following day inherits the 0 from the now-dead spread row, until an admin hand-edits it. (Same root cause bites within a single day if a mid-day cache flush re-probes to a different max and `findOrCreate()` lands on an existing threshold-0 top row for the same date.) Suggested fix: exclude top-band rows here, e.g. `->whereNotNull('mythic_level_max')`.

database/migrations/2026_08_12_120000_add_key_level_band_to_combat_log_parsing_criteria_table.php
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    public function down(): void
+    {
+        Schema::table('combat_log_parsing_criteria', function (Blueprint $table) {
+            $table->index(['model_class', 'date'], 'clpc_model_class_date_index');
+            $table->dropIndex('clpc_model_class_band_date_index');
+
+            $table->dropUnique('clpc_version_class_id_date_band_unique');
+            $table->unique(
```
Wotuu, line 62: :robot: **`down()` is not runnable once banded rows exist.** After a single day of the new poller the table holds multiple rows per `(combat_log_version, model_class, model_id, date)` — one per band — so restoring this narrower unique index throws a duplicate-entry error mid-migration and leaves the rollback half-applied (old `model_class/date` index restored, band index dropped, no unique at all). If the rollback is meant to be usable, `down()` should collapse the band rows first — e.g. keep one row per `(version, class, id, date)` group (or simply delete rows where `mythic_level_min > 0`) before recreating the unique.

Wotuu:

app/Service/CombatLog/CombatLogParsingCriteriaService.php
Wotuu: :robot: Fixed in c7e8456cf — the inheritance query now also requires `mythic_level_max` to be non-null, so a top band row can never hand its threshold of 0 to the spread band that later starts on the same level. Added `shouldParse_givenBandThatUsedToBeTheTopBand_ignoresItsThresholdOfZero`, which arranges yesterday's `22+` top band row and asserts today's `22-26` spread band gets the configured default instead.

database/migrations/2026_08_12_120000_add_key_level_band_to_combat_log_parsing_criteria_table.php
Wotuu: :robot: Fixed in c7e8456cf — `down()` now collapses to the lowest-id row per (version, model_class, model_id, date) before restoring the narrower unique. These rows are daily counters the poller recreates on demand, so nothing of value is lost. Verified against the worktree DB with 10 banded rows across 4 keys present: rollback collapses them to 4 and re-running the migration succeeds.

app/Service/CombatLog/CombatLogPollingBandService.php
```diff
@@ -0,0 +1,183 @@
...
+        try {
+            $maxKeyLevel = $this->probeMaxKeyLevel($season);
+        } catch (Throwable $throwable) {
+            $lastKnown = Cache::get($this->getMaxKeyLevelLastKnownCacheKey($season));
+
+            Log::warning('combatlog:pollruns - probing the max key level failed', [
+                'season'     => $season->id,
+                'last_known' => $lastKnown,
+                'exception'  => $throwable->getMessage(),
+            ]);
```
Wotuu: What's this? Please use structured logging.

app/Console/Commands/CombatLog/PollCombatLogRunsCommand.php
```diff
@@ -48,128 +51,225 @@ public function handle(): int
...
+        $topBand    = $this->bandService->getTopBand($season);
 
         $force = (bool)$this->option('force');
 
+        /** @var array<int, true> $existingRunIds */
         $existingRunIds = $force ? [] : ParsedCombatLog::query()
             ->whereNotNull('run_id')
             ->pluck('run_id')
             ->flip()
             ->all();
```
Wotuu: Can we maybe figure out a way to not query all of the parsed combat logs? In this case, they can run into the thousands and thousands of rows. It may not be too bad, but maybe we can find a way to determine the season of the parsed combat log and write that along in the database so that we can filter on it. 

I think the run ID is known at some point, though I cannot find where it is written. Maybe we can also extract the season that this parsed combat log is from so that we can filter on it.  
  
I do notice that the parsed combat log model itself does not contain an `@property` for the run ID, which threw me for a loop here. I do see that you filter on it in this query, but it's not in the model. It's there in the database, so it should also go in the PHP doc on the model. If you can correct that as well, that would be great.

Wotuu (commented):
One thing I want to verify: when a new season starts, everyone's gonna start from a low level, I think level 2, and they will slowly creep their way up. It will happen quickly, but it can take a few hours for people to get to level 10. Especially that first day, the max kill level done for the season is going to rise really quickly.

Can you verify that this doesn't cause any issues? Not that when it reads the first runs coming in and it will read only 2s, it's then stuck parsing +2s for the rest of the day or something? Does it refresh that max level often enough that this won't be a problem, especially on the first day of the season?

If this is not a problem, then it's fine and we can merge. If this is a problem, then we need to fix it and then merge it.

Wotuu:
I've just taken a look at the admin pages, and the admin page can become really long when all the parsing criteria are now repeated multiple times across the different bands. Can you, for each band, introduce a bootstrap accordion so that they all start collapsed and I can expand them as necessary?

Maybe there's a different idea to make the page not so long and still be able to get an overview of sorts. If everything was collapsed in an accordion, I need to start expanding things in order to see something. I don't know, maybe we can color the accordion's collapsed state green if everything is fulfilled or orange if it's not. Let's spar on this, perhaps.

PR #3990, opened 2026-08-12
#3989 Stop MDT import from coupling NPC spells / spell-dungeon assignments

### LINKED ISSUE #3989: Stop MDT import from coupling NPC spells / spell-dungeon assignments (follow-up to #3980)
:robot: Follow-up to #3980 / #3984, requested by Wotuu in review
(https://github.com/RaiderIO/keystone.guru/pull/3984#discussion_r3767041888):

> MDT does, but we don't want to do anything with this information. We gathered this information
> ourselves with the parsing of combat log data. If they are still a part of the code that imports
> NPC spell data or couples spells to dungeons or NPCs, please ensure that it doesn't do that in a
> follow-up issue.

## What's still there

`app/Service/MDT/MDTMappingImportService.php` still couples spells to NPCs and dungeons purely
from the MDT Lua export, independently of the combat-log-derived data:

- `importNpcsDataFromMDT()` (~line 251-269, insert at line 308) - for every spell an MDT NPC
  export lists, inserts an `NpcSpell` row (`npc_id` + `spell_id`) if one doesn't already exist.
- `importSpellDataFromMDT()` (~line 324-385) - for every spell an MDT NPC casts, either couples an
  *existing* `Spell` to the dungeon via `SpellDungeon` (if the spell ID is already known), or
  inserts a brand-new placeholder `Spell` row (empty name/icon/school, `category.unknown`, etc.)
  and couples *that* to the dungeon.

Neither of these were touched by #3980/#3984 - they're pre-existing MDT import behaviour. #3980
did touch the consumer of this data, `app/Console/Commands/Spell/AssignMissingSpellDungeons.php`,
which had to be hardened to skip `NpcSpell` rows whose `spell` relation is `null` - a NpcSpell row
that points at a `spell_id` `importNpcsDataFromMDT()` created and `spells.php` doesn't have a
readable name for yet. This reimport alone produced 206 such rows.

## Why it matters

Per Wotuu, NPC-spell and spell-dungeon associations are meant to come exclusively from combat log
parsing now (see `app/Service/CombatLog/DataExtractors/SpellDataCollectors/NpcSpellAssignmentCollector.php`,
`SpellCounterDataExtractor.php`, `ImmunityBypassDataExtractor.php` - all of which create `NpcSpell`
rows from parsed combat log data). The MDT-sourced coupling is redundant with that pipeline and, per
Wotuu, actively unwanted: MDT's spell list per NPC is not curated data we want driving `NpcSpell`/
`SpellDungeon` rows.

## Suggested approach

- Remove (or no-op) the `NpcSpell::insert($npcSpellsAttributes)` path in
  `importNpcsDataFromMDT()` and the spell-coupling half of `importSpellDataFromMDT()` -
  scope needs deciding: does anything else currently depend on MDT populating placeholder `Spell`
  rows for spell IDs the combat-log pipeline hasn't seen yet? (`Spell::EXCLUDE_MDT_IMPORT_SPELLS`
  and the `$existingSpells` lookup suggest this has evolved organically and may have other
  consumers worth checking first.)
- If `AssignMissingSpellDungeons.php`'s null-spell handling (added in #3980) becomes unreachable
  once the MDT coupling is removed, consider whether it can be simplified back, or left as
  defensive code.
- Needs a full mapping reimport cycle to verify no regression (missing spell school data, broken
  Compendium spell listings, etc.) - see the `update-mdt-package` skill for the reimport runbook.

## Key files

- `app/Service/MDT/MDTMappingImportService.php`
- `app/Console/Commands/Spell/AssignMissingSpellDungeons.php`
- `app/Models/Spell/Spell.php` (`EXCLUDE_MDT_IMPORT_SPELLS`)

Wotuu:
:robot: Closes #3989

Follow-up to #3980 / #3984, requested by Wotuu in review: NPC-spell and spell-dungeon
associations come exclusively from our own combat log parsing, and the MDT import must not
write them.

## What was removed

| Where | What it did |
|---|---|
| `importNpcsDataFromMDT()` | Inserted an `NpcSpell` row for every spell an MDT NPC lists |
| `importSpellDataFromMDT()` | Coupled existing spells to the dungeon via `SpellDungeon`, and inserted empty placeholder `Spell` rows (no name/icon/school, `category.unknown`) for spell IDs `spells` didn't have yet |

`importSpellDataFromMDT()` is deleted outright rather than emptied, along with its only caller,
the `mdt:importspells` command (zero other references), its interface entry and its five logging
methods. `importNpcsDataFromMDTCharacteristicsAndSpellsUpdate()` loses its spell counters and is
renamed `importNpcsDataFromMDTNpcsUpdate()`; `importNpcsDataFromMDTSpellInExcludeList()` is gone.

The issue asked whether anything else depended on the MDT-created placeholder spells. Answer, by
grep: no. `Spell::EXCLUDE_MDT_IMPORT_SPELLS` had exactly the two call sites removed here, and
`Spell::isAssignedDungeon()` had exactly one — both are removed as dead code. Real `Spell` rows
come from `wowhead:fetchmissingspells`, the DB2 pipeline (#3951) and the combat log extractors.

## What was deliberately kept

- **`MDTNpc::toArray()` still includes `'spells'`.** It feeds `getMDTMappingHash()`; dropping it
  would change every dungeon's hash and trigger a spurious mapping-version bump on the next
  scheduled import. MDT's spell list no longer *produces* anything, but it may still legitimately
  signal "MDT's data for this dungeon changed".
- **The null-`spell` guard in `AssignMissingSpellDungeons` (#3980) stays.** Removing the writer
  doesn't make the reader safe: the ~206 orphan `NpcSpell` rows the last reimport created still
  exist in dev/staging/production, and `NpcSpellAssignmentCollector` independently records a
  (npc, spell) pair the moment it observes a cast — before the `spells` row exists. Its comment is
  updated to say so.
- **No cleanup command or migration to purge the existing MDT-sourced rows.** Not asked for, and
  a data-destructive sweep over `npc_spells` / `spell_dungeons` is Wotuu's call. Happy to file a
  follow-up if wanted.

## Testing

New `MDTMappingImportNoSpellCouplingTest` runs `importNpcsDataFromMDT()` against real MDT Lua and
asserts `npc_spells`, `spells` and `spell_dungeons` are all unchanged.

The precondition is the part that makes it worth having: it *selects* a dungeon whose MDT export
lists a spell that isn't already coupled to the NPC casting it, and `fail()`s if no such dungeon
exists — otherwise the pre-change code would have had nothing to insert either and every assertion
would pass regardless. Verified discriminating: with the `NpcSpell` insert temporarily put back,
the test fails.

Also ran a real single-dungeon reimport (`mdt:importmapping hallsofvalor retail --force`) in the
worktree: the mapping version builds end to end (117 enemies, hash stamped) and all three spell
tables gained zero rows.

- 484 related tests green (`--filter="MDT|Spell|Compendium"`), `composer run fix` and
  `composer run analyse` clean.
- `MDTNpcMappingCoverageTest` fails in my worktree with 320 unmatched MDT clones across
  `kingsrest`, `rubylifepools`, `den_of_nalorakk`, `templeofsethraliss`, `altar_of_fangs`,
  `the_blinding_vale`, `voidscar_arena` — **pre-existing**: it fails identically with my changes
  stashed, and none of the implicated dungeons is one this PR or my verification run touched.
- Zero `database/seeders/**` diff.
- No visual verification: backend-only, nothing rendered changed.

## Cold review

1 finding, fixed in ffe7a804f: the new test's `finally` reverted `NpcHealth`/`NpcDungeon` but not the
`npcs` / `npc_enemy_forces` rows the import writes for a genuinely *new* NPC — zero residue today only
because every MDT NPC is already seeded, an invariant the next MDT bump breaks. Thread resolved.


tests/Feature/App/Service/MDT/MDTMappingImportNoSpellCouplingTest.php
```diff
@@ -0,0 +1,160 @@
...
+            NpcHealth::query()
+                ->where('game_version_id', $retailGameVersion->id)
+                ->whereIn('npc_id', $mdtNpcIds)
+                ->whereNotIn('id', $preExistingNpcHealthIds)
+                ->delete();
+
+            NpcDungeon::query()
+                ->whereIn('npc_id', $mdtNpcIds)
+                ->whereNotIn('id', $preExistingNpcDungeonIds)
+                ->delete();
```
Wotuu, line 121: :robot: The cleanup here is narrower than the comment above promises ("snapshot what exists to clean up anything new"). `importNpcsDataFromMDT()` writes to two more unscoped tables that this `finally` never touches:

- **`npcs`** — for an MDT NPC the dungeon doesn't have, `$newlyCreated` is true and `$npc->save()` inserts a brand-new row.
- **`npc_enemy_forces`** — for each such new NPC with `getCount() > 0`, `createNpcEnemyForcesForExistingMappingVersions()` writes a row into *every* historical mapping version.

I re-ran the test twice and confirmed zero residue, but that only holds because on today's seeded DB every MDT NPC already exists, so `npcsInserted == 0`. That's an invariant of the current seeder data, not of the test — and an MDT reimport like #3980 is exactly the event that makes MDT's NPC list stop being a subset of what's seeded. When that happens the test silently leaks an `npcs` row plus N `npc_enemy_forces` rows into the shared seeded DB with nothing to signal it.

Worth noting that `MDTMappingImportCrashRecoveryTest` **deliberately stubs `importNpcsDataFromMDT()` out** for precisely this reason — its docblock says it "writes to shared Npc/NpcHealth/NpcSpell/NpcDungeon rows that are not scoped to the new mapping version, so a mapping-version delete cannot revert them… exactly what #3737 itself warns about". This test runs it for real against retail on a fully-seeded dungeon, so it's the higher-blast-radius variant of that same call.

Secondary, one line: the assertions compare row *counts*, not values. The import also overwrites pre-existing `npc_healths.health/percentage` and the `npcs` row's `name`/`level`/`display_id`/`mdt_scale`/`npc_type_id`/`truesight` for every MDT NPC in the selected dungeon and restores none of it — unlike the sibling `MDTMappingImportGameVersionScopingTest`, which targets `beta` where little pre-exists.

Latent rather than blocking (the sibling test has the same shape and the repo has evidently accepted it), but at minimum the two missing tables deserve the same snapshot-and-delete treatment, or the comment should say which mutations are knowingly left behind.

Wotuu:

Verified independently and found clean: no remaining references anywhere in the repo to `importSpellDataFromMDT`, `mdt:importspells`, `Spell::EXCLUDE_MDT_IMPORT_SPELLS`, `Spell::isAssignedDungeon()` or the old logging method names; `wowhead:fetchmissingspells` is driven off `npc_spells` (via `getMissingSpellIds()`), not off the deleted placeholder `spells` rows, so removing the placeholder writer doesn't starve the spell pipeline; the new test is picked up by CI (`--testsuite "Feature Tests" --exclude-group=Controller…` shard, `UsesLua`/`MDT` are not excluded anywhere) and passes in 0.4s with zero DB residue on two consecutive runs.

Two notes that didn't warrant their own thread:

- `MDTMappingImportCrashRecoveryTest`'s two docblocks (lines ~118 and ~185) still describe `importNpcsDataFromMDT()` as mass-inserting `NpcSpell`, which this PR makes untrue. Harmless, but it's doc drift this change introduces in the one file that documents that method's write set most explicitly.
- On keeping `'spells'` in `MDTNpc::toArray()` / the mapping hash: the "avoid one-time hash churn" half of the justification argues for a one-time cost, while the retained field imposes an ongoing one (a purely-spell-list MDT change now bumps every affected dungeon's mapping version for data nothing consumes). The second half of the justification — that it still signals "MDT's data changed" — stands on its own, so this is Wotuu's call, not a finding.

tests/Feature/App/Service/MDT/MDTMappingImportNoSpellCouplingTest.php
Wotuu: :robot: Fixed in ffe7a804f. The `finally` now also snapshots the `npcs` rows that already exist among the MDT NPC IDs, and afterwards deletes any newly-inserted NPC plus its `npc_enemy_forces` rows (query-builder deletes — both are `SeederModel`s, whose `->delete()` on a model instance is silently refused). `Npc` goes last since the other cleanups key off `npc_id`.

You're right that it's zero-residue today only by accident of every MDT NPC already being seeded, and that an MDT bump is exactly what breaks it — the comment now says so rather than leaving it as an unstated assumption.

Also took the docblock note from the summary comment: dropped `NpcSpell` from the two `MDTMappingImportCrashRecoveryTest` docblocks describing writes `importNpcsDataFromMDT()` no longer makes.

app/Service/MDT/MDTMappingImportService.php
```diff
@@ -200,21 +197,17 @@ public function importNpcsDataFromMDT(MDTDungeon $mdtDungeon, Dungeon $dungeon,
             // Get a list of NPCs and update/save them. npcHealths must be eager loaded here - it is read
             // per NPC below via getHealthByGameVersion(), which would otherwise be an N+1 that hard-fails
             // under preventLazyLoading (dev) for any dungeon that already has NPCs.
-            $existingNpcs = $dungeon->npcs()->with(['npcSpells', 'npcHealths'])->get()->keyBy('id');
+            $existingNpcs = $dungeon->npcs()->with(['npcHealths'])->get()->keyBy('id');
```
Wotuu, line 214: Can we create a follow-up issue to also remove the importing of health from MDT? I think we also gather that information from combat logs now. If we don't, can you create a follow-up ticket to fix that one to-do in the NPC health extractor somewhere and use the new DB2 parsing logic to extract the proper health? The incoming health comes in through the combat log and is scaled by key level, so it needs to be scaled down to a base health. I think you're going to need to file a follow-up issue for this.

Issue #2473, opened 2024-09-10
Add ability to add arrows


Wotuu:
Don't forget integration with MDT for this.

PR #3992, opened 2026-08-12
#3966 Restore adding vertices to lines during route edit

### LINKED ISSUE #3966: Route edit mode no longer offers adding new vertices to lines
:robot: Found while manually testing staging for v15.10.2.

Hitting the edit button on a route used to expose an option to add new vertices to existing
lines (paths/arrows/brushlines) while editing. That option no longer appears.

Likely related to `L.Draw.Polyline.prototype.addVertex` wiring in
`resources/assets/js/custom/models/arrow.js` (and whatever equivalent exists for paths/brushlines)
- needs investigation into whether this regressed from a recent Leaflet.draw/editing change, or
was removed intentionally.

## Repro
1. Open an existing route on staging.
2. Click the edit button to enter map edit mode.
3. Try to add a new vertex to an existing line (path/arrow/brushline).

## Expected
The vertex-adding affordance (usually a small marker/handle on the line, or a context menu
option) should be available while editing, same as before.

## Actual
No such option is shown.

Wotuu:
:robot: Closes #3966

## What was wrong

`8318433de` (#2473, shipped 2026-06-09) added a **global** monkeypatch of Leaflet.draw's
middle-marker factory so arrows could not gain extra vertices during edit. The guard was written
opt-**in**:

```js
if (!this._poly.options.allowVertexCreationDuringEdit) { return; }
```

but `allowVertexCreationDuringEdit` is only ever *written* in one place in the codebase —
`arrow.js`, where it is set to `false`. Nothing ever set it to `true`, so for every other polyline
the option was `undefined`, `!undefined === true`, and the patch returned early. Middle markers are
the only way to insert a vertex into an existing line, so every editable polyline silently lost that
ability for ~2 months:

- **route editor**: paths and brushlines
- **mapping-version editor**: admin enemy patrols, floor unions and floor union areas

(Killzone paths and non-admin enemy patrols both return `false` from `isEditable()`, so they never
reach `editableLayers` and were never affected either way.)

## The fix

1. Invert the guard to opt-**out** (`=== false`). Arrow's explicit `false` still wins; everything
   else behaves as Leaflet.draw ships it, i.e. exactly as it did before June.
2. Move Arrow's flag from the constructor into `onLayerInit()`. `MapObjectGroup#setLayerToMapObject()`
   replaces `mapObject.layer` wholesale on every rebuild (floor switch, live-session update), so a
   flag set once at construction is lost — and the inversion in (1) is what makes that loss
   *observable*: a rebuilt arrow would have regained its middle markers. Verified below.

## Tests

New `resources/assets/js/custom/mapcontrols/drawcontrols.test.js` drives the real `leaflet` +
`leaflet-draw` packages through Leaflet.draw's own editing handler:

- `createMiddleMarker_givenPolylineWithoutFlag_createsMiddleMarker` — 3-point polyline gets 2 middle markers
- `createMiddleMarker_givenArrowWithFlagFalse_skipsMiddleMarker` — the arrow restriction survives
- `onLayerInit_givenRebuiltLayer_reappliesVertexCreationFlag` — guards (2)

Both regression assertions are **red on master** (`expected +0 to be 2`) and green here. Full JS
suite: 634 passed.

## Browser verification

Real headless Chrome, route edit mode with a 3-point path and a 2-point arrow, edit handler active:

| | path middle markers | arrow middle markers | arrow flag after a real `setLayerToMapObject()` swap |
|---|---|---|---|
| before | **0** | 0 | **lost (`undefined`)** |
| after | **2** | 0 | `false` (kept) |

before
after

The faint extra handles along the red dashed path in the "after" shot are the restored middle
markers; the blue dashed arrow correctly keeps only its two endpoints in both.

The one page error logged (`Cannot read properties of undefined (reading 'length')`) is present
identically on the baseline — pre-existing, unrelated.

Note: JS change, so it cannot be hotfixed onto the current release — it ships with the next one.

### DIFF
--- a/resources/assets/js/custom/mapcontrols/drawcontrols.js
+++ b/resources/assets/js/custom/mapcontrols/drawcontrols.js
@@ -69,10 +69,12 @@ L.DrawToolbar.prototype.getModeHandlers = function (map) {
     ];
 };
 
-// Prevent the editing of
+// Prevent the creation of new vertices during edit for layers that opted out (arrows, which are
+// always exactly two vertices). Every other polyline keeps Leaflet.draw's middle markers - dragging
+// one is the only way to insert a vertex into an existing line (#3966).
 const _originalCreateMiddleMarker = L.Edit.PolyVerticesEdit.prototype._createMiddleMarker;
 L.Edit.PolyVerticesEdit.prototype._createMiddleMarker = function (marker1, marker2) {
-    if (!this._poly.options.allowVertexCreationDuringEdit) {
+    if (this._poly.options.allowVertexCreationDuringEdit === false) {
         return;
     }
     _originalCreateMiddleMarker.call(this, marker1, marker2);
--- a/resources/assets/js/custom/mapcontrols/drawcontrols.test.js
+++ b/resources/assets/js/custom/mapcontrols/drawcontrols.test.js
@@ -1,0 +1,124 @@
+// ---------------------------------------------------------------------------
+// Covers #3966: adding a new vertex to an existing line during edit.
+//
+// Leaflet.draw inserts a vertex when you drag the translucent "middle marker"
+// it renders halfway between two vertices. drawcontrols.js monkeypatches
+// L.Edit.PolyVerticesEdit#_createMiddleMarker so arrows - which are always
+// exactly two vertices - do not get them. That guard has to be opt-OUT: every
+// other polyline (path, brushline, killzone path, enemy patrol, floor union
+// area) must keep its middle markers.
+//
+// These tests drive the REAL leaflet + leaflet-draw packages so the patch is
+// exercised through Leaflet.draw's own editing handler, not a stand-in.
+// ---------------------------------------------------------------------------
+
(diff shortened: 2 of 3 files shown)


Wotuu:

Checked: the opt-in→opt-out inversion restores pre-`8318433de` behaviour for every other polyline (verified `FloorUnion` is a Marker and `FloorUnionArea` an arbitrary polygon, so no other layer type wants the arrow restriction); the flag move to `onLayerInit()` is safe on ordering — `MapObjectGroup#setLayerToMapObject()` adds the layer only to the plain `L.LayerGroup` before calling `onLayerInit()`, and the draw feature group (`DungeonMap#editableLayers`, which is what triggers leaflet-draw `_enableLayerEdit` → `_createMiddleMarker`) is only fed from the later `object:layerchanged` handler, so the flag is always set before leaflet-draw reads `options.allowVertexCreationDuringEdit`; `Arrow` is only ever constructed via `ArrowMapObjectGroup#_createMapObject()`, which is always immediately followed by `onLayerInit()`. Test file ran green locally and the guarded `module.exports` matches the established global-script pattern.

PR #3994, opened 2026-08-13
#3993 Import MDT genericItem POIs and report the ones we have no icon for

### LINKED ISSUE #3993: MDT genericItem POIs are silently dropped on mapping import - Murder Row and 5 other Midnight dungeons are missing 19 map icons
## Background

MDT draws item pickups on its dungeon maps as `mapPOIs` of type `genericItem`. Keystone.guru
imports **none of them** - Murder Row shows Fel Contraband, Overload Golem, Felwyrm Egg etc. in
MDT, and nothing on our map.

The cause is a silent fall-through in `MDTMappingImportService::importMapPOIs()`
(`app/Service/MDT/MDTMappingImportService.php`, ~line 1005). It builds a `$mapIconTypeMapping`
covering 9 MDT POI types, handles `mapLink` in an `elseif`, and **anything else does nothing at
all** - no icon, no warning, no log line.

`MDTMapPOIType::GenericItem = 'genericItem'` *is* a known enum case
(`app/Logic/MDT/Entity/MDTMapPOIType.php`, added in #3192), it is just not in the mapping. That
is also why the MDT 6.2.1 import (#3980) reported "no new POI types": the `update-mdt-package`
runbook detects new types by `MDTMapPOI` **throwing** on an unknown type string. A
known-but-unmapped type never throws, so this class of gap is invisible.

## Scope - 19 POIs, 13 distinct spells, 6 Midnight dungeons

| Dungeon | spellId | Name | texture FDID | icon file | count |
|---|---|---|---|---|---|
| Algeth'ar Academy | 389501 | Red Dragonflight Pledge Pin | 132532 | `inv_bijou_red` | 1 |
| Algeth'ar Academy | 389512 | Bronze Dragonflight Pledge Pin | 132530 | `inv_bijou_orange` | 1 |
| Algeth'ar Academy | 389516 | Black Dragonflight Pledge Pin | 132533 | `inv_bijou_silver` | 1 |
| Algeth'ar Academy | 389521 | Blue Dragonflight Pledge Pin | 132526 | `inv_bijou_blue` | 1 |
| Algeth'ar Academy | 389536 | Green Dragonflight Pledge Pin | 132529 | `inv_bijou_green` | 1 |
| Magisters' Terrace | 1254550 | Arcane Empowerment | 135740 | `spell_arcane_mindmastery` | 1 |
| Maisara Caverns | 1269056 | Hearty Vilebranch Stew | 4659336 | `inv_cooking_10_heartystew` | 1 |
| Murder Row | 1270638 | Fel Contraband | 1003586 | `inv_enchanting_wod_crystal` | 3 |
| Murder Row | 1223133 | Overload Golem | 4620673 | `ui_profession_engineering` | 3 |
| Murder Row | 1223570 | Felwyrm Egg | 236999 | `inv_egg_08` | 1 |
| Murder Row | 1223537 | Felstone | 1394959 | `inv_infernalbrimstone` | 1 |
| Murder Row | 1217960 | Loaded Pistol | 331438 | `inv_weapon_rifle_40` | 1 |
| Murder Row | 1223607 | Heartstop Poison | 1120423 | `ability_boss_kilrogg_heartseeker` | 1 |
| Seat of the Triumvirate | 244300 | Void Infusion | 6891020 | `inv_112_arcane_beam` | 1 |
| The Blinding Vale | 1265942 | Flourishing Stride | 348567 | `spell_lifegivingspeed` | 1 |

Map icon types for Maisara Caverns and Seat of the Triumvirate already exist (a18f9799e) but were
never wired into `$mapIconTypeMapping` either, so they too must have been placed by hand. That
leaves **13 new map icon types** to create.

## Remaining work

1. **Report unhandled POIs instead of dropping them.** Add an `else` branch that logs every
   known-but-unmapped POI type (at ERROR, like `importMapPOIsMissingTranslation`) and collect them
   so `mdt:importmapping` can render a summary table of `dungeon / spellId / texture / icon name /
   count`. This must **not** feed `$importFailures` - that throws
   `MDTMappingImportPartialFailureException` and deletes the whole mapping version.
2. **Key the mapping on `(type, spellId)` for `genericItem`.** `genericItem` is one MDT type
   covering many different items, so a flat `type => mapIconTypeId` map cannot express it. The
   other 8 types stay type-only.
3. **Create the 13 missing map icon types** - constant + id in `MapIconType::ALL`,
   `MapIconTypesSeeder` entry, `lang/en_US/mapicontypes.php` string, `GenerateItemIcons` entry, and
   the generated png in the `keystone.guru.assets` repo.

## Resolving an MDT texture to a downloadable icon, automatically

Wowhead is **not** usable for this: `https://www.wowhead.com/spell=<id>` returns **403** to the
app container (CloudFront blocks it), so `WowheadService::getSpellData()` returns an empty
`SpellDataResult` - worth a separate look, it means `wowhead:fetchspelldata` is also degraded.

A deterministic path exists instead. MDT's `info.texture` is a **FileDataID**:

1. `https://wago.tools/db2/ManifestInterfaceData/csv` maps FileDataID -> icon file name
   (`4620673 -> Interface\ICONS\UI_Profession_Engineering.blp`).
2. `https://wow.zamimg.com/images/wow/icons/large/<lowercased name>.jpg` downloads the icon.

Both are reachable from the container (200). This validates against existing data: Maisara
Caverns' FDID resolves to exactly the `inv_cooking_10_heartystew` Wotuu picked by hand.

Note Seat of the Triumvirate's texture now resolves to `inv_112_arcane_beam`, while the existing
icon uses `inv_enchant_voidsphere` - MDT changed the texture since #3070.

## Notes

- Wiring the importer does **not** backfill. `importMapPOIs` only runs inside
  `importMappingVersionFromMDT`, gated on `mdt_mapping_hash`, so existing mapping versions keep
  their missing icons until a `mdt:importmapping <key> retail --force` (which creates a new
  mapping version per dungeon). Deliberately out of scope here - the icons land at the next MDT
  bump.
- `mapicon:generateitemicons` re-renders **all** icons and produces a git diff on already
  existing ones; visually the output is equivalent. Stage only the new pngs.

## Key files

- `app/Service/MDT/MDTMappingImportService.php` (`importMapPOIs`)
- `app/Service/MDT/Logging/MDTMappingImportServiceLogging{,Interface}.php`
- `app/Console/Commands/MDT/ImportMapping.php`
- `app/Logic/MDT/Entity/MDTMapPOIType.php`
- `app/Models/MapIconType.php`, `database/seeders/MapIconTypesSeeder.php`
- `lang/en_US/mapicontypes.php`
- `app/Console/Commands/MapIcon/GenerateItemIcons.php`
- `keystone.guru.assets/images/mapicon_gen/`, `keystone.guru.assets/images/mapicon/`

## Definition of done

- All 19 `genericItem` POIs across the 6 dungeons map to a real map icon type on import.
- An unmapped POI type produces a visible ERROR log line and a summary table on
  `mdt:importmapping`, and never silently disappears.
- Tests green, `composer run analyse` and `composer run fix` clean.

Wotuu:
:robot: Closes #3993

## What was wrong

MDT draws item pickups as `mapPOIs` of type `genericItem`. `MDTMappingImportService::importMapPOIs()`
matched POI types against a flat `type => mapIconTypeId` array, handled `mapLink` in an `elseif`, and
did **nothing at all** for anything else - no icon, no warning, no log line. 19 POIs across 6 Midnight
dungeons disappeared that way, including Murder Row's Fel Contraband, Overload Golem and Felwyrm Egg.

`MDTMapPOIType::GenericItem` is a *known* enum case, so it never tripped the `Found new type ...`
exception the `update-mdt-package` runbook uses to spot gaps - which is why the MDT 6.2.1 import
(#3980) reported "no new POI types".

## What changed

- **The POI -> map icon type mapping moves to `Conversion`** and gains a second, spell-id keyed table.
  `genericItem` is a single MDT type covering every item it draws, distinguished only by the spell it
  shows the tooltip of, so a type-keyed map cannot express it.
- **Unmapped POIs are reported, not dropped.** `importMapPOIsUnhandledMapPOI` logs at ERROR, and
  `mdt:importmapping` renders a summary table. It reads MDT directly rather than the import, so it
  shows up even on `no change detected` - the gap exists either way. Deliberately **not** routed
  through `$importFailures`, which throws `MDTMappingImportPartialFailureException` and deletes the
  whole mapping version.
- **`zoom` and `textFrame` are marked as never-imported** (MDT's own zoom affordance and the user's
  text annotations), so they don't show up as false positives.
- **13 new map icon types** covering every `genericItem` in the current Midnight dungeons.
- **`mapicon:downloadmdtitemicons <dungeon|seasonId>`** fetches the source icons automatically.

## Getting an icon out of MDT, automatically

Wowhead is not usable from the app container - `https://www.wowhead.com/spell=<id>` returns **403**
(CloudFront), so `WowheadService::getSpellData()` returns an empty `SpellDataResult`. Worth a separate
issue: `wowhead:fetchspelldata` is degraded by the same thing.

A deterministic path exists instead. MDT's `info.texture` is a FileDataID:

1. `wago.tools/db2/ManifestInterfaceData/csv` maps FileDataID -> icon file name
   (`4620673` -> `UI_Profession_Engineering.blp`) - the new `WagoToolsService`.
2. `wow.zamimg.com/images/wow/icons/large/<name>.jpg` serves it - `WowheadService::downloadIcon()`,
   split out of `downloadSpellIcon()`.

It validates against existing hand-picked data: Maisara Caverns' FileDataID resolves to exactly the
`inv_cooking_10_heartystew` that was chosen by hand in a18f9799e.

## The 19 POIs

| Dungeon | spellId | Name | texture FDID | icon |
|---|---|---|---|---|
| Algeth'ar Academy | 389501 | Red Dragonflight Pledge Pin | 132532 | `inv_bijou_red` |
| Algeth'ar Academy | 389512 | Bronze Dragonflight Pledge Pin | 132530 | `inv_bijou_orange` |
| Algeth'ar Academy | 389516 | Black Dragonflight Pledge Pin | 132533 | `inv_bijou_silver` |
| Algeth'ar Academy | 389521 | Blue Dragonflight Pledge Pin | 132526 | `inv_bijou_blue` |
| Algeth'ar Academy | 389536 | Green Dragonflight Pledge Pin | 132529 | `inv_bijou_green` |
| Magisters' Terrace | 1254550 | Arcane Empowerment | 135740 | `spell_arcane_mindmastery` |
| Maisara Caverns | 1269056 | Hearty Vilebranch Stew | 4659336 | existing |
| Murder Row | 1270638 x3 | Fel Contraband | 1003586 | `inv_enchanting_wod_crystal` |
| Murder Row | 1223133 x3 | Overload Golem | 4620673 | `ui_profession_engineering` |
| Murder Row | 1223570 | Felwyrm Egg | 236999 | `inv_egg_08` |
| Murder Row | 1223537 | Felstone | 1394959 | `inv_infernalbrimstone` |
| Murder Row | 1217960 | Loaded Pistol | 331438 | `inv_weapon_rifle_40` |
| Murder Row | 1223607 | Heartstop Poison | 1120423 | `ability_boss_kilrogg_heartseeker` |
| Seat of the Triumvirate | 244300 | Void Infusion | 6891020 | existing |
| The Blinding Vale | 1265942 | Flourishing Stride | 348567 | `spell_lifegivingspeed` |

## Verification

Force-reimported Murder Row against an isolated worktree DB - the new mapping version carries exactly
what MDT draws:

```
dungeon_start x1
murder_row_felwyrm_egg x1
murder_row_fel_contraband x3
murder_row_loaded_pistol x1
murder_row_felstone x1
murder_row_heartstop_poison x1
murder_row_overload_golem x3
```

All 6 Midnight dungeons now report 0 unhandled POIs. The table was verified against `therookery`,
which still has an unmapped `rookItem`.

## Out of scope

- **No reimport.** `importMapPOIs` is gated on `mdt_mapping_hash`, so existing mapping versions keep
  their missing icons until the next MDT bump. Agreed with Wotuu - a `--force` across 6 dungeons means
  6 new mapping versions with enemy/checkpoint churn and seeder diffs.
- **The generalised reporting surfaced 62 more unhandled POIs in 12 older dungeons**
  (`nwItem`, `brackenhideCage`, `neltharusChain`, `mistsItem`, `cotItem`, `rookItem`, ...). None are in
  the current season, so they are left for whenever those dungeons come back around.
- **Assets:** the 13 generated `mapicon/*.png` and their 13 `mapicon_gen/*.jpg` sources are staged in
  `keystone.guru.assets` but **not committed** - that repo's `main` is not mine to push to. They must
  land before this merges, or the new icons 404.

## Cold review

4 findings, all fixed in a3ea1a1e9, all threads resolved. The blocking one: spell IDs `1269056`
(Maisara Caverns) and `244300` (Seat of the Triumvirate) already have hand-placed icons that
`copyMappingVersionContentsToDungeon()` clones into every new mapping version, sitting 5.3 and 13.9
lat/lng off MDT's position - outside `getMapIconNearLocation()`'s ±5 window, so the import would have
added a second copy of each. Fixed by having `importMapPOIs()` delete the cloned
`genericItem`-derived icons and re-create the full set from MDT: every map icon type in
`MAP_POI_GENERIC_ITEM_SPELL_ID_MAP_ICON_TYPE_MAPPING` exists for exactly one MDT POI, so MDT owning
them is the right semantic. Verified by force-reimporting both dungeons - one icon each, at MDT's
coordinates.

## Held in draft on purpose

**Resolved:** the `keystone.guru.assets` commit has been pushed, so the icons are in place.

### DIFF
--- a/app/Console/Commands/MDT/ImportMapping.php
+++ b/app/Console/Commands/MDT/ImportMapping.php
@@ -2,6 +2,7 @@
 
 namespace App\Console\Commands\MDT;
 
+use App\Logic\MDT\Entity\MDTMapPOI;
 use App\Models\Dungeon;
 use App\Models\GameVersion\GameVersion;
 use App\Models\Season;
@@ -9,6 +10,7 @@ use App\Service\Mapping\MappingServiceInterface;
 use App\Service\MDT\MDTMappingImportServiceInterface;
 use Exception;
 use Illuminate\Console\Command;
+use Illuminate\Support\Collection;
 
 class ImportMapping extends Command
 {
@@ -45,12 +47,17 @@ class ImportMapping extends Command
             throw new Exception(sprintf('Game version %s not found', $gameVersionKey));
         }
 
+        /** @var Collection<int, Dungeon> $dungeons */
+        $dungeons = new Collection();
+
         if (is_numeric($dungeonKey)) {
             // If it's an ID we should treat it as a season instead
             $season = Season::findOrFail($dungeonKey);
 
             // Cannot do ->with('npcs') here - it won't load the relationship properly due to orWhere(dungeon_id = -1)
             foreach ($season->dungeons as $dungeon) {
+                $dungeons->push($dungeon);
+
                 try {
                     $dungeon->setRelation('npcs', $dungeon->npcs()->get());
                     $mappingImportService->importMappingVersionFromMDT($mappingService, $dungeon, $gameVersion, $force);
@@ -62,10 +69,82 @@ class ImportMapping extends Command
             // Cannot do ->with('npcs') here - it won't load the relationship properly due to orWhere(dungeon_id = -1)
             /** @var Dungeon $dungeon */
             $dungeon = Dungeon::where('key', $dungeonKey)->firstOrFail();
+            $dungeons->push($dungeon);
 
             $dungeon->setRelation('npcs', $dungeon->npcs()->get());
 
             $mappingImportService->importMappingVersionFromMDT($mappingService, $dungeon, $gameVersion, $force);
         }
+
+        $this->renderUnhandledMapPOIs($mappingImportService, $dungeons);
+    }
+
+    /**
+     * Reports every POI MDT draws that we have no map icon type for. These are read straight from MDT rather
+     * than from the import, so they show up even when the import no-ops on an unchanged mapping hash - the
+     * gap exists either way, and used to be entirely invisible (#3993).
+     *
+     * @param Collection<int, Dungeon> $dungeons
+     */
+    private function renderUnhandledMapPOIs(
+        MDTMappingImportServiceInterface $mappingImportService,
+        Collection                       $dungeons,
+    ): void {
+        $rows                  = [];
+        $unhandledMapPOIsTotal = 0;
+
+        foreach ($dungeons as $dungeon) {
+            try {
+                $unhandledMapPOIs = $mappingImportService->getUnhandledMapPOIs($dungeon);
+            } catch (Exception $exception) {
+                $this->error($exception->getMessage());
+
+                continue;
+            }
+
+            $unhandledMapPOIsTotal += $unhandledMapPOIs->count();
+
+            // One row per distinct item rather than per POI - the same item is usually placed several times
+            $unhandledMapPOIsByItem = $unhandledMapPOIs->groupBy(
+                static fn(MDTMapPOI $mdtMapPOI): string => sprintf(
+                    '%s-%s-%s',
+                    $mdtMapPOI->getType()->value,
+                    $mdtMapPOI->getSpellId() ?? '',
+                    $mdtMapPOI->getTextureFileDataId() ?? '',
+                ),
+            );
+
+            foreach ($unhandledMapPOIsByItem as $unhandledMapPOIsForItem) {
+                /** @var MDTMapPOI $mdtMapPOI */
+                $mdtMapPOI = $unhandledMapPOIsForItem->first();
+
+                $rows[] = [
+                    $dungeon->key,
+                    $mdtMapPOI->getType()->value,
+                    $mdtMapPOI->getSpellId() ?? '-',
+                    $mdtMapPOI->getTextureFileDataId() ?? '-',
+                    $unhandledMapPOIsForItem->count(),
+                    $mdtMapPOI->getSpellId() === null
+                        ? '-'
+                        : sprintf('https://www.wowhead.com/spell=%d', $mdtMapPOI->getSpellId()),
+                ];
+            }
+        }
+
+        if ($rows === []) {
+            return;
+        }
+
+        $this->newLine();
+        $this->error(sprintf(
+            '%d MDT map POI(s) across %d distinct item(s) have no map icon type and were NOT imported:',
+            $unhandledMapPOIsTotal,
+            count($rows),
+        ));
+        $this->table(
+            ['Dungeon', 'MDT type', 'Spell ID', 'Texture ID', 'Count', 'Wowhead'],
+            $rows,
+        );
+        $this->line('Run <comment>mapicon:downloadmdtitemicons</comment> for these dungeons to fetch their icons, then add a map icon type for each.');
     }
 }
--- a/app/Console/Commands/MapIcon/DownloadMDTItemIcons.php
+++ b/app/Console/Commands/MapIcon/DownloadMDTItemIcons.php
@@ -1,0 +1,135 @@
+<?php
+
+namespace App\Console\Commands\MapIcon;
+
+use App\Logic\MDT\Entity\MDTMapPOI;
+use App\Models\Dungeon;
+use App\Models\Season;
+use App\Service\MDT\MDTMappingImportServiceInterface;
+use App\Service\Wago\WagoToolsServiceInterface;
+use App\Service\Wowhead\WowheadServiceInterface;
+use Exception;
+use Illuminate\Console\Command;
+use Illuminate\Support\Collection;
+
+class DownloadMDTItemIcons extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'mapicon:downloadmdtitemicons {dungeon}';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Downloads the source icons of all MDT map POIs that have no map icon type yet';
+
+    /**
+     * Execute the console command.
+     *
+     * @throws Exception
+     */
+    public function handle(
+        MDTMappingImportServiceInterface $mappingImportService,
+        WagoToolsServiceInterface        $wagoToolsService,
+        WowheadServiceInterface          $wowheadService,
+    ): int {
+        $dungeonArgument = $this->argument('dungeon');
+
+        if (is_numeric($dungeonArgument)) {
+            $dungeons = Season::findOrFail($dungeonArgument)->dungeons;
+        } else {
+            $dungeons = new Collection([Dungeon::where('key', $dungeonArgument)->firstOrFail()]);
+        }
+
+        /** @var Collection<int, MDTMapPOI> $unhandledMapPOIs */
+        $unhandledMapPOIs = new Collection();
+        $dungeonKeyByPOI  = [];
+
+        foreach ($dungeons as $dungeon) {
+            foreach ($mappingImportService->getUnhandledMapPOIs($dungeon) as $mdtMapPOI) {
+                $dungeonKeyByPOI[spl_object_id($mdtMapPOI)] = $dungeon->key;
+                $unhandledMapPOIs->push($mdtMapPOI);
+            }
+        }
+
+        if ($unhandledMapPOIs->isEmpty()) {
+            $this->info('No unhandled MDT map POIs - nothing to download.');
+
+            return 0;
+        }
+
+        // One texture may be shared by several POIs, and resolving them costs a ~9MB DB2 download each time
+        $fileDataIds = $unhandledMapPOIs
+            ->map(static fn(MDTMapPOI $mdtMapPOI): ?int => $mdtMapPOI->getTextureFileDataId())
+            ->filter()
+            ->unique()
+            ->values()
+            ->all();
+
+        $iconFileNames = $wagoToolsService->getIconFileNamesByFileDataIds($fileDataIds);
+
+        $targetFolder = realpath(base_path('../keystone.guru.assets/images/mapicon_gen'));
+        if ($targetFolder === false) {
+            $this->error('Unable to find the keystone.guru.assets repository next to this checkout - is it cloned, and is it mounted into this container?');
+
+            return 1;
+        }
+
+        $rows        = [];
+        $downloaded  = [];
+        $hasFailures = false;
+
+        foreach ($unhandledMapPOIs as $mdtMapPOI) {
+            $fileDataId   = $mdtMapPOI->getTextureFileDataId();
+            $iconFileName = $fileDataId === null ? null : ($iconFileNames[$fileDataId] ?? null);
+
+            $status = 'no texture';
+            if ($iconFileName !== null) {
+                if (isset($downloaded[$iconFileName])) {
+                    $status = $downloaded[$iconFileName];
+                } elseif (file_exists(sprintf('%s/%s.jpg', $targetFolder, $iconFileName))) {
+                    $status = 'already present';
+                } elseif ($wowheadService->downloadIcon($iconFileName, $targetFolder)) {
+                    $status = 'downloaded';
+                } else {
+                    $status      = 'DOWNLOAD FAILED';
+                    $hasFailures = true;
+                }
+
+                $downloaded[$iconFileName] = $status;
+            } elseif ($fileDataId !== null) {
+                $status      = 'NOT AN ICON IN ManifestInterfaceData';
+                $hasFailures = true;
+            }
+
+            $rows[] = [
+                $dungeonKeyByPOI[spl_object_id($mdtMapPOI)],
+                $mdtMapPOI->getType()->value,
+                $mdtMapPOI->getSpellId() ?? '-',
+                $fileDataId ?? '-',
+                $iconFileName ?? '-',
+                $status,
+                $mdtMapPOI->getSpellId() === null
+                    ? '-'
+                    : sprintf('https://www.wowhead.com/spell=%d', $mdtMapPOI->getSpellId()),
+            ];
+        }
+
+        $this->table(
+            ['Dungeon', 'MDT type', 'Spell ID', 'Texture ID', 'Icon', 'Status', 'Wowhead'],
+            $rows,
+        );
+
+        $this->newLine();
+        $this->line('Next: add a MapIconType for each item (constant + id in MapIconType::ALL, MapIconTypesSeeder,');
+        $this->line('lang/en_US/mapicontypes.php, GenerateItemIcons and Conversion::MAP_POI_GENERIC_ITEM_SPELL_ID_MAP_ICON_TYPE_MAPPING),');
+        $this->line('then run <comment>mapicon:generateitemicons</comment>.');
+
+        return $hasFailures ? 1 : 0;
+    }
+}
--- a/app/Console/Commands/MapIcon/DownloadMDTItemIcons.php
+++ b/app/Console/Commands/MapIcon/DownloadMDTItemIcons.php
@@ -94,6 +94,9 @@ class DownloadMDTItemIcons extends Command
                     $status = $downloaded[$iconFileName];
                 } elseif (file_exists(sprintf('%s/%s.jpg', $targetFolder, $iconFileName))) {
                     $status = 'already present';
+                    // Despite the service name this fetches from `wow.zamimg.com`, Wowhead's asset CDN, which
+                    // serves us fine. It is wowhead.com's *pages* that answer 403 - so the file name has to
+                    // come from wago.tools above rather than by scraping the icon page (#3993).
                 } elseif ($wowheadService->downloadIcon($iconFileName, $targetFolder)) {
                     $status = 'downloaded';
                 } else {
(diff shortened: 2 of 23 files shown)


app/Logic/MDT/Conversion.php
```diff
@@ -287,6 +290,69 @@ class Conversion
...
+        // Algeth'ar Academy
+        389501 => MapIconType::MAP_ICON_TYPE_ALGETHAR_ACADEMY_RED_DRAGONFLIGHT_PLEDGE_PIN,
+        389512 => MapIconType::MAP_ICON_TYPE_ALGETHAR_ACADEMY_BRONZE_DRAGONFLIGHT_PLEDGE_PIN,
+        389516 => MapIconType::MAP_ICON_TYPE_ALGETHAR_ACADEMY_BLACK_DRAGONFLIGHT_PLEDGE_PIN,
+        389521 => MapIconType::MAP_ICON_TYPE_ALGETHAR_ACADEMY_BLUE_DRAGONFLIGHT_PLEDGE_PIN,
+        389536 => MapIconType::MAP_ICON_TYPE_ALGETHAR_ACADEMY_GREEN_DRAGONFLIGHT_PLEDGE_PIN,
+        // Magisters' Terrace
+        1254550 => MapIconType::MAP_ICON_TYPE_MAGISTERS_TERRACE_ARCANE_EMPOWERMENT,
+        // Maisara Caverns
+        1269056 => MapIconType::MAP_ICON_TYPE_MAISARA_CAVERNS_HEARTY_VILEBRANCH_STEW,
```
Wotuu, line 330: :robot: **These two spell IDs already have hand-placed map icons — the next reimport of those dungeons will create a *second* one.**

`maisara_caverns` (109) and `theseatofthetriumvirate` (108) were mapped by hand before this PR (a18f9799e); those icons live in the current mapping version and `MappingService::copyMappingVersionContentsToDungeon()` clones every `mapIcon` into the new mapping version. Now that the genericItem POI also imports, `importMapPOIs()` will additionally create one from MDT — unless `MappingVersion::getMapIconNearLocation()` matches, and its dedupe window is only ±5 lat/lng.

Probed live against the worktree DB (mv 801 / current retail mapping):

```
maisara_caverns   existing icon: floor 403, lat -150.9799, lng 103.8161  (comment: mapping.map_icons.midnight.maisara_caverns.hearty_vilebranch_stew)
                  MDT genericItem: floor 403, lat -152.5680, lng 109.1280   -> dlng 5.31 > 5 -> getMapIconNearLocation() == NULL

theseatofthetriumvirate  existing icon: floor 37, lat -143.8043, lng 158.2035
                         MDT genericItem: floor 37, lat -141.8593, lng 172.1114 -> dlng 13.9 -> getMapIconNearLocation() == NULL
```

So on the next MDT bump for either dungeon the map gets two Hearty Vilebranch Stew / Void Infusion icons a few units apart, and the MDT-created one carries no `comment` (the import only sets one for `GenericAssignablePOI`), so the seeder diff will look like a legitimate new icon rather than a duplicate.

The PR body explicitly defers reimport as out of scope, which is exactly what makes this land unnoticed later. Worth either dropping these two spell IDs from the map (they're already covered by hand), or moving/deleting the hand-placed icons so the MDT position wins.

app/Console/Commands/MDT/ImportMapping.php
```diff
@@ -62,10 +69,75 @@ public function handle(
...
+    }
+
+    /**
+     * Reports every POI MDT draws that we have no map icon type for. These are read straight from MDT rather
+     * than from the import, so they show up even when the import no-ops on an unchanged mapping hash - the
+     * gap exists either way, and used to be entirely invisible (#3993).
+     *
+     * @param Collection<int, Dungeon> $dungeons
+     */
+    private function renderUnhandledMapPOIs(
```
Wotuu, line 89: :robot: **`.claude/skills/update-mdt-package/SKILL.md` still tells the next session that `exit=0` means no gaps.**

That runbook is the reason #3993 went unnoticed. It currently says (line ~268):

> A new POI type/template would make the command **exit non-zero**, so an `exit=0` with zero `Found new` matches confirms no enum changes were needed.

and its "Read the output for each dungeon and handle:" list has bullets for `Found new type`, `No change detected` and `importMapPOIsMissingTranslation` — but nothing for the new unhandled-POI table. `handle()` returns `void`, so the table never changes the exit code, and the runbook's own grep recipe (`grep -ciE 'Found new (template|type)'`) won't surface it either. A session following the runbook will do exactly what the MDT 6.2.1 import did and report "no new POI types".

Per `.claude/CLAUDE.md` ("record durable knowledge in-repo"), this PR should add a bullet to that list: the new "N MDT map POI(s) have no map icon type and were NOT imported" table, what to do with it (`mapicon:downloadmdtitemicons`, then the 5 places listed in `DownloadMDTItemIcons`' closing hint), and a correction to the `exit=0` claim.

app/Service/Wago/WagoToolsServiceInterface.php
```diff
@@ -0,0 +1,22 @@
...
+     *
+     * @param  array<int>         $fileDataIds
+     * @return array<int, string> FileDataID => icon file name, without extension, lowercased.
+     */
+    public function getIconFileNamesByFileDataIds(array $fileDataIds): array;
+
+    /**
+     * @see WagoToolsServiceInterface::getIconFileNamesByFileDataIds()
+     */
+    public function getIconFileNameByFileDataId(int $fileDataId): ?string;
```
Wotuu: :robot: `getIconFileNameByFileDataId()` has no callers anywhere in `app/` or `tests/` — `DownloadMDTItemIcons` (the only consumer) deliberately uses the batch method. It's also a footgun on a brand-new interface: each call downloads the full ~9MB ManifestInterfaceData CSV, so a loop over it is a 9MB-per-iteration download — precisely what the batch method's docblock exists to prevent. Suggest dropping it until something needs it.

app/Console/Commands/MDT/ImportMapping.php
```diff
@@ -62,10 +69,75 @@ public function handle(
...
+                ];
+            }
+        }
+
+        if ($rows === []) {
+            return;
+        }
+
+        $this->newLine();
+        $this->error(sprintf('%d MDT map POI(s) have no map icon type and were NOT imported:', count($rows)));
```
Wotuu: :robot: `count($rows)` is the number of *distinct items* (the rows are grouped by type+spellId+texture just above), not the number of POIs — the per-POI count is already in the `Count` column. So the 19 dropped Murder Row/etc. POIs from the PR description would print as e.g. "7 MDT map POI(s) ... were NOT imported". Either sum the group counts or word it as "distinct MDT map POI(s)".

Wotuu:

app/Logic/MDT/Conversion.php
Wotuu: :robot: Fixed in a3ea1a1e9 — confirmed and it was the right call.

Rather than widening the dedupe window (which would risk merging genuinely distinct same-type items), `importMapPOIs()` now deletes the cloned `genericItem`-derived icons up front and re-creates the complete set from MDT. Every map icon type in `MAP_POI_GENERIC_ITEM_SPELL_ID_MAP_ICON_TYPE_MAPPING` exists for exactly one MDT POI and nothing else, so MDT owning them is the correct semantic — and it makes the ±5 window irrelevant for this class of icon instead of tuning it.

Verified by force-reimporting both dungeons against the worktree DB. Before: `maisara_caverns_hearty_vilebranch_stew` at `-150.9799, 103.8161`. After: exactly one, at `-152.568, 109.128` (MDT's position). Same for Seat of the Triumvirate: one `seat_of_the_triumvirate_void_infusion` at `-141.859, 172.111`. No duplicates.

app/Console/Commands/MDT/ImportMapping.php
Wotuu: :robot: Fixed in a3ea1a1e9 — good catch, that runbook assumption is the actual root cause of #3993.

`SKILL.md` now: (a) says an `exit=0` with no `Found new` matches confirms no *enum* changes were needed and explicitly does **not** mean every POI was imported; (b) adds `UnhandledMapPOI|were NOT imported` to the grep line; (c) has a full bullet in the output-handling list covering the table, the `mapicon:downloadmdtitemicons` step, the FileDataID → wago.tools → zamimg path (and that Wowhead pages 403 the container), the five places a new map icon type must be registered, and that it does not affect the exit code; (d) updates the checklist item and the hand-off summary step.

app/Service/Wago/WagoToolsServiceInterface.php
Wotuu: :robot: Fixed in a3ea1a1e9 — removed from both the interface and the implementation. It was speculative API, and as you note each call would have re-downloaded the full ~9MB CSV; the batch method is the only sane way to use this.

app/Console/Commands/MDT/ImportMapping.php
Wotuu: :robot: Fixed in a3ea1a1e9. Now reports both numbers: `"%d MDT map POI(s) across %d distinct item(s) have no map icon type and were NOT imported"`, with the POI total accumulated before the `groupBy`. The Murder Row case reads as 10 POIs across 6 items rather than a misleading 6.

app/Console/Commands/MapIcon/DownloadMDTItemIcons.php
```diff
@@ -0,0 +1,135 @@
...
+            $fileDataId   = $mdtMapPOI->getTextureFileDataId();
+            $iconFileName = $fileDataId === null ? null : ($iconFileNames[$fileDataId] ?? null);
+
+            $status = 'no texture';
+            if ($iconFileName !== null) {
+                if (isset($downloaded[$iconFileName])) {
+                    $status = $downloaded[$iconFileName];
+                } elseif (file_exists(sprintf('%s/%s.jpg', $targetFolder, $iconFileName))) {
+                    $status = 'already present';
+                } elseif ($wowheadService->downloadIcon($iconFileName, $targetFolder)) {
```
Wotuu, line 97: So you just told me that downloading from Wowhead was not possible, and you are downloading it from another source, but here you are downloading it from Wowhead anyway. Does it work or does it not work?

app/Console/Commands/MapIcon/GenerateItemIcons.php
```diff
@@ -5,6 +5,22 @@
...
+ * Adding one is five registrations, all of which must agree on the same key: a constant + a new id in
+ * {@see \App\Models\MapIconType::ALL}, a `MapIconTypesSeeder` entry, a `lang/en_US/mapicontypes.php` string,
+ * a source => target pair below, and - if MDT draws it - an entry in
+ * {@see \App\Logic\MDT\Conversion::MAP_POI_GENERIC_ITEM_SPELL_ID_MAP_ICON_TYPE_MAPPING}. Note that adding an
+ * MDT mapping for an icon that was already placed by hand duplicates it on the next reimport (#3993).
+ *
+ * To get the raw icon: do **not** scrape Wowhead, its pages 403 the app container. Resolve the icon file name
+ * from a texture FileDataID through wago.tools' ManifestInterfaceData DB2 and download it off Wowhead's CDN -
+ * `mapicon:downloadmdtitemicons` does exactly that for every MDT POI we have no icon for. This command only
+ * runs from the main checkout: the assets repo is not mounted into a worktree's container.
```
Wotuu: You can make a change to mount the assets project into the worktree's container. I'd be fine with you doing that.

app/Service/MDT/MDTMappingImportService.php
```diff
@@ -1090,6 +1098,18 @@ private function importMapPOIs(
                                 $newMappingVersion->dungeonFloorSwitchMarkers()->count(),
                             );
                         }
+                    } elseif (Conversion::isMDTMapPOIUnhandled($mdtMapPOI)) {
+                        // MDT draws something we have no map icon type for - our map will be missing it. This
```
Wotuu: Just keep the first sentence in this comment block.

database/seeders/MapIconTypesSeeder.php
```diff
@@ -631,6 +631,88 @@ public function run(): void
...
+                'height'     => 32,
+                'admin_only' => true,
+            ],
+
+            MapIconType::MAP_ICON_TYPE_THE_BLINDING_VALE_FLOURISHING_STRIDE => [
+                'name'       => 'mapicontypes.the_blinding_vale_flourishing_stride',
+                'width'      => 32,
+                'height'     => 32,
+                'admin_only' => true,
+            ],
```
Wotuu, line 715: File a follow-up ticket to put the contents of this file into a JSON file so we can clean it up like we did with the season seeder.

PR #3996, opened 2026-08-13
#3995 Refuse an MDT import whose NPCs have nothing in common with the current mapping

### LINKED ISSUE #3995: MDT import silently replaced The Blinding Vale's whole mapping with Den of Nalorakk's - guard against a wholesale NPC-set swap
## What happened

MDT **6.2.1 shipped a broken `Midnight/TheBlindingVale.lua`** - it is a duplicate of Den of Nalorakk's
data. The mapping import in #3980 did exactly what it was told and replaced The Blinding Vale's
mapping with Den of Nalorakk's enemies, **silently**: exit 0, no warning, nothing in the summary.

- mv 3 (id 814, MDT 6203) - correct, 276 enemies, 26 NPCs
- mv 4 (id 833, MDT 6205) - correct, 276 enemies, 26 NPCs
- **mv 5 (id 843, MDT 621) - wrong**, 168 enemies, 32 NPCs, all Den of Nalorakk's

This is upstream's bug, not ours - `Nnoggie/master`'s `TheBlindingVale.lua` is still md5-identical to
the broken 6.2.1 file today, and The Blinding Vale's enemy data (245345 Lightgorged Lasher et al.)
exists nowhere in upstream master. It is a duplication, not a swap: Den of Nalorakk's own file is
fine. All 16 Midnight files were compared master vs vendor - The Blinding Vale is the only one
affected. Wotuu is preparing an upstream PR with the correct data.

**Our problem is that nothing caught it.** An import that replaces ~100% of a dungeon's NPC set with a
different dungeon's is never legitimate, and it went unnoticed until someone eyeballed the map.

## The guard

Compare the incoming MDT NPC id set against the previous mapping version's, before committing the
import.

Thresholds are measured, not guessed. Across **all 307 consecutive mapping-version transitions in the
database**:

| | overlap with previous NPC set |
|---|---|
| The Blinding Vale mv 833 -> 843 (the bug) | **0%** |
| lowest *legitimate* transition ever (Seat of the Triumvirate, Scarlet Enclave) | 59% |
| number of legitimate transitions at 0% | **0** |

So:

- **Abort** when the overlap is **0%** - no NPC in common at all. Zero false positives across 307
  transitions, and it cannot happen legitimately: a dungeon cannot be remapped to a completely
  disjoint set of NPCs. Throwing inside `importMappingVersionFromMDT`'s try block already does the
  right thing - the partial mapping version is deleted and `mdt_mapping_hash` stays unstamped, so the
  next run retries instead of silently no-oping.
- **Warn loudly** (ERROR log + a line in the command output) below **50%** - under every legitimate
  transition observed, with 9pp of headroom - so the grey zone is visible without blocking.
- The abort message should name **which dungeon the incoming NPC set actually matches**, if any. That
  is the failure signature here (100% Den of Nalorakk) and turns a confusing abort into a diagnosis.
- An explicit escape hatch for the genuine wholesale-rework case (a new `--force`-style option), since
  a legitimate 0% import must remain possible without editing code.

First imports (no predecessor, or a predecessor with no enemies) skip the check.

## Not in scope

Restoring The Blinding Vale's mapping. mv 4 (id 833) is intact in both the DB and
`database/seeders/dungeondata/midnight/the_blinding_vale/1/enemies.json`, and no dungeon routes
reference any of the three mapping versions, so it is recoverable whenever we choose. Note mv 833 is
stamped `mdt_addon_version=6205` while MDT is on 621, so the clean restore is a new mapping version
cloned from 833 and stamped 621 rather than deleting 843. Tracked separately.

## Key files

- `app/Service/MDT/MDTMappingImportService.php` (`importMappingVersionFromMDT`, `importEnemies`)
- `app/Service/MDT/Logging/MDTMappingImportServiceLogging{,Interface}.php`
- `app/Console/Commands/MDT/ImportMapping.php`
- `.claude/skills/update-mdt-package/SKILL.md` - the runbook must say what to do when it fires

## Definition of done

- An import whose incoming NPC set is disjoint from the previous mapping version's aborts, deletes its
  partial mapping version, leaves the hash unstamped, and says which dungeon the data looks like.
- A sub-50% overlap import succeeds but reports loudly.
- Regression test reproducing the Blinding Vale case.
- Tests green, `composer run analyse` and `composer run fix` clean.

Wotuu:
:robot: Closes #3995

## Why

MDT 6.2.1 shipped a `Midnight/TheBlindingVale.lua` that is a duplicate of Den of Nalorakk's data.
The import in #3980 did exactly what it was told and replaced The Blinding Vale's entire mapping with
Den of Nalorakk's enemies - **exit 0, no warning, nothing in the summary**. It went unnoticed until
Wotuu eyeballed the map weeks later.

Upstream is still broken today: `Nnoggie/master`'s `TheBlindingVale.lua` is md5-identical to the
pinned 6.2.1 file, and The Blinding Vale's enemy data exists nowhere in it. Wotuu is preparing an
upstream PR. This PR is about the half we control - **nothing on our side noticed.**

## The guard

`importMappingVersionFromMDT()` now compares the incoming MDT NPC set against the current mapping
version's, and refuses a total miss.

Thresholds are measured, not guessed. Across **all 307 consecutive mapping-version transitions in the
database**:

| | overlap with previous NPC set |
|---|---|
| The Blinding Vale mv 833 -> 843 (the bug) | **0%** |
| lowest *legitimate* transition ever observed | 59% |
| legitimate transitions at 0% | **0** |

- **Under 50% overlap refuses the import outright** - below every legitimate transition observed,
  with 9pp of headroom, and zero false positives across all 307 transitions.
- The abort **names the dungeon the incoming NPCs actually belong to** (`den_of_nalorakk` in this
  case), turning a confusing failure into a diagnosis.
- **`--force` is the only override** - it already means "I know what I am doing, do it anyway", so
  the guard reuses it rather than adding a second flag.

The check runs **before** `createNewMappingVersionFromMDTMapping()`, so no mapping version is created
and deleted again, and `mdt_mapping_hash` is never touched: the dungeon keeps its previous mapping and
the next run simply retries.

## Verification

Reproduced the real failure by temporarily restoring the good `TheBlindingVale.lua` into the
worktree's vendor and reimporting (vendor restored to the pinned file afterwards - confirmed
md5-identical, hardlink-safe):

```
MDT offers 26 NPC(s) for dungeon the_blinding_vale that have NOTHING in common with the 32 NPC(s)
of its current mapping version - refusing to replace the mapping.
Pass --allow-npc-set-replacement if this really is intended.
```

10 tests covering: disjoint refuses; the 50% line from both sides (20% refused, 60% allowed) with a
controlled NPC mix; the message names the owning dungeon; `--force` proceeds; a matching NPC set
passes; no predecessor passes; both branches of the "whose NPCs are these" lookup; and the call-site
placement itself - that nothing is created and `mdt_mapping_hash` is never stamped, mutation-verified
by deleting the guard call and watching it fail. They drive the guard directly rather than through a full import - it reads nothing but
`$dungeon` and `$currentMappingVersion->enemies()`, so handing it one dungeon's MDT data and another
dungeon's mapping version reproduces the exact condition **without writing a single row**, and without
pinning a test to the upstream bug we want fixed.

## Note for the reviewer

`.claude/skills/update-mdt-package/SKILL.md` gains a bullet at the same anchor #3994 touches, so
expect a small conflict there if #3994 merges first. Trivial to resolve - both are new bullets in the
same list.

## Not in scope

Restoring The Blinding Vale's mapping. mv 4 (id 833) is intact in the DB and in
`the_blinding_vale/1/enemies.json`, and no dungeon routes reference any of the three mapping versions.
Note mv 833 is stamped `mdt_addon_version=6205` while MDT is on 621, so the clean restore is a new
mapping version cloned from 833 rather than deleting 843. Tracked on #3995.

## Cold review

3 findings, all fixed in 71be3e1cc, all threads resolved. Two were real coverage gaps worth naming:
the naming test passed with the `$looksLikeDungeonKey` branch deleted (its arrangement was the mirror
image of the bug, so the lookup correctly returned null), and the call-site placement - the entire
safety claim - was untested because every test reflected into the private method. Both now pinned,
the second mutation-verified.

## Change of behaviour since review

Wotuu's call: sub-50% overlap now **refuses the import** rather than warning and carrying on. The two
thresholds collapsed into one, and `--allow-npc-set-replacement` is gone - `--force` is the only
override.

### DIFF
--- a/.claude/skills/update-mdt-package/SKILL.md
+++ b/.claude/skills/update-mdt-package/SKILL.md
@@ -257,6 +257,19 @@ Read the output for each dungeon and handle:
   `lang/en_US/mapping.php` with a sensible English string (e.g. `'captive' => 'Captive'`),
   then **notify the user** so they can refine the wording. Only ever edit `lang/en_US`. No
   re-import is needed — translations resolve at display time, not import time.
+- **`MDTMappingNpcSetReplacedException` / "refusing to replace the mapping"**: the import aborted
+  because MDT's NPCs for that dungeon have **nothing** in common with its current mapping version.
+  This is almost always MDT shipping the wrong dungeon's data in its `.lua` — 6.2.1 shipped a
+  `TheBlindingVale.lua` that was a duplicate of Den of Nalorakk's, and before this guard existed it
+  silently replaced the whole mapping at exit 0 (#3995). The message names the dungeon the incoming
+  NPCs actually belong to. **Do not reach for `--allow-npc-set-replacement` to make it go away** —
+  verify against the `.lua` first (`grep '\["name"\]' vendor/nnoggie/mythicdungeontools/<Expansion>/<Dungeon>.lua | head`)
+  and report it upstream; the flag is only for a genuine wholesale remap. Nothing was written when
+  this fires: no mapping version is created and `mdt_mapping_hash` is untouched, so the dungeon keeps
+  its previous mapping and the next run retries.
+- **`importMappingVersionFromMDTNpcSetChangedSignificantly` warning**: the import kept going, but
+  under 50% of the previous mapping version's NPCs survived — below every legitimate transition ever
+  observed. Sanity-check that dungeon on the map before handing over.
 - **Other errors:** try to fix them yourself first (the user prefers this). Only pause for the
   user if you get genuinely stuck.
 
@@ -264,7 +277,7 @@ Read the output for each dungeon and handle:
 > lines. Redirect to a file and grep for the markers that matter rather than dumping it all,
 > e.g. `... > /tmp/mdt_<key>.log 2>&1; echo "exit=$?"` then
 > `grep -ciE 'Found new (template|type)' /tmp/mdt_<key>.log` and
-> `grep -iE 'No change detected|MappingChanged|MissingTranslation' /tmp/mdt_<key>.log`.
+> `grep -iE 'No change detected|MappingChanged|MissingTranslation|NpcSetReplaced|NpcSetChangedSignificantly' /tmp/mdt_<key>.log`.
 > A new POI type/template would make the command **exit non-zero**, so an `exit=0` with zero
 > `Found new` matches confirms no enum changes were needed.
 
--- a/.claude/skills/update-mdt-package/SKILL.md
+++ b/.claude/skills/update-mdt-package/SKILL.md
@@ -227,7 +227,8 @@ docker compose exec -T app php artisan mdt:importmapping <dungeonKey> retail
 ```
 
 (Command lives in `app/Console/Commands/MDT/ImportMapping.php`, signature
-`mdt:importmapping {dungeon} {gameVersion} {--force}`. `gameVersion` is always `retail`. A
+`mdt:importmapping {dungeon} {gameVersion} {--force}`.
+`gameVersion` is always `retail`. A
 numeric season ID may be passed as `dungeon` to import the whole season at once, but prefer
 per-dungeon runs so you can observe each output and test it individually.)
 
@@ -258,18 +259,16 @@ Read the output for each dungeon and handle:
   then **notify the user** so they can refine the wording. Only ever edit `lang/en_US`. No
   re-import is needed — translations resolve at display time, not import time.
 - **`MDTMappingNpcSetReplacedException` / "refusing to replace the mapping"**: the import aborted
-  because MDT's NPCs for that dungeon have **nothing** in common with its current mapping version.
+  because under 50% of the current mapping version's NPCs survive the incoming MDT data (the lowest
+  legitimate transition ever observed was 59%).
   This is almost always MDT shipping the wrong dungeon's data in its `.lua` — 6.2.1 shipped a
   `TheBlindingVale.lua` that was a duplicate of Den of Nalorakk's, and before this guard existed it
   silently replaced the whole mapping at exit 0 (#3995). The message names the dungeon the incoming
-  NPCs actually belong to. **Do not reach for `--allow-npc-set-replacement` to make it go away** —
+  NPCs actually belong to. **Do not reach for `--force` to make it go away** —
   verify against the `.lua` first (`grep '\["name"\]' vendor/nnoggie/mythicdungeontools/<Expansion>/<Dungeon>.lua | head`)
-  and report it upstream; the flag is only for a genuine wholesale remap. Nothing was written when
+  and report it upstream; `--force` is only for a genuine wholesale remap. Nothing was written when
   this fires: no mapping version is created and `mdt_mapping_hash` is untouched, so the dungeon keeps
   its previous mapping and the next run retries.
-- **`importMappingVersionFromMDTNpcSetChangedSignificantly` warning**: the import kept going, but
-  under 50% of the previous mapping version's NPCs survived — below every legitimate transition ever
-  observed. Sanity-check that dungeon on the map before handing over.
 - **Other errors:** try to fix them yourself first (the user prefers this). Only pause for the
   user if you get genuinely stuck.
 
@@ -277,7 +276,7 @@ Read the output for each dungeon and handle:
 > lines. Redirect to a file and grep for the markers that matter rather than dumping it all,
 > e.g. `... > /tmp/mdt_<key>.log 2>&1; echo "exit=$?"` then
 > `grep -ciE 'Found new (template|type)' /tmp/mdt_<key>.log` and
-> `grep -iE 'No change detected|MappingChanged|MissingTranslation|NpcSetReplaced|NpcSetChangedSignificantly' /tmp/mdt_<key>.log`.
+> `grep -iE 'No change detected|MappingChanged|MissingTranslation|NpcSetReplaced' /tmp/mdt_<key>.log`.
 > A new POI type/template would make the command **exit non-zero**, so an `exit=0` with zero
 > `Found new` matches confirms no enum changes were needed.
 
--- a/app/Service/MDT/MDTMappingImportService.php
+++ b/app/Service/MDT/MDTMappingImportService.php
@@ -5,6 +5,7 @@ namespace App\Service\MDT;
 use App\Logic\MDT\Conversion;
 use App\Logic\MDT\Data\MDTDungeon;
 use App\Logic\MDT\Entity\MDTMapPOIType;
+use App\Logic\MDT\Entity\MDTNpc;
 use App\Logic\MDT\Entity\MDTPatrol;
 use App\Models\Dungeon;
 use App\Models\DungeonFloorSwitchMarker;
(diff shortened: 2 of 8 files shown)


tests/Feature/App/Service/MDT/MDTMappingImportNpcSetGuardTest.php
```diff
@@ -0,0 +1,196 @@
...
+        try {
+            $this->assertMDTNpcSetIsPlausible($dungeon, $foreignMappingVersion, false);
+        } catch (MDTMappingNpcSetReplacedException $exception) {
+            $message = $exception->getMessage();
+        }
+
+        // Assert - the diagnosis is the whole point: an abort that does not say whose data this is just
+        // sends the next person hunting through .lua files
+        $this->assertNotNull($message, 'The guard should have thrown.');
+        $this->assertStringContainsString(DungeonKey::MURDER_ROW->value, $message);
```
Wotuu: :robot: This test does not pin the behaviour its name claims. I ran the exact arrangement through tinker in the worktree:

```
MDT offers 41 NPC(s) for dungeon murder_row that have NOTHING in common with the 32 NPC(s) of its
current mapping version - refusing to replace the mapping. Pass --allow-npc-set-replacement if this
really is intended.
```

No dungeon is named. The arrangement is the mirror image of #3995: `$dungeon` is `murder_row` and MDT hands back *murder_row's own* NPCs, so `findDungeonKeyOwningNpcs()` filters them out via `whereNot('dungeon_id', $dungeon->id)` and returns `null` — the `$looksLikeDungeonKey` branch of the exception never runs. Both assertions (`murder_row`, `--allow-npc-set-replacement`) are in the base message unconditionally, so this test would still pass if that whole branch were deleted from `MDTMappingNpcSetReplacedException`.

The real direction (foreign MDT data vs. the dungeon's *own* mapping version) can't be reproduced without a `.lua` holding the wrong dungeon's data, which the PR body rightly avoids pinning to. Suggested fix: rename this test to what it actually asserts (that the abort message states the dungeon and the escape hatch), and add a small separate test that constructs `MDTMappingNpcSetReplacedException` directly with a non-null `$looksLikeDungeonKey` to pin the "names the owning dungeon" half of the message. `findDungeonKeyOwningNpcs()` is already covered by its own two tests, so that closes the gap.

(Related, non-blocking: with mv 843 now current and holding Nalorakk NPCs, the eventual *correct* Blinding Vale import will also be 0% overlap and will need `--allow-npc-set-replacement` — and for the same `whereNot` reason its message won't name a dungeon either. Escape hatch covers it and #3995 tracks the restore.)

app/Service/MDT/MDTMappingImportService.php
```diff
@@ -92,6 +102,11 @@ public function importMappingVersionFromMDT(
         if ($forceImport || $currentMappingVersion === null || $currentMappingVersion->mdt_mapping_hash !== $latestMdtMappingHash) {
             $this->log->importMappingVersionFromMDTMappingChanged($currentMappingVersion?->mdt_mapping_hash, $latestMdtMappingHash);
 
+            // Before anything is created: refuse outright if MDT is handing us a different dungeon's NPCs
+            // (#3995). Checking here rather than mid-import means no mapping version is created and deleted
+            // again, and mdt_mapping_hash is never touched - the next run simply retries.
+            $this->assertMDTNpcSetIsPlausible($dungeon, $currentMappingVersion, $allowNpcSetReplacement);
```
Wotuu: :robot: The placement of this call is the PR's headline safety claim — "no mapping version is created and deleted again, and `mdt_mapping_hash` is never touched" — but nothing tests it. All 7 tests in `MDTMappingImportNpcSetGuardTest` invoke `assertMDTNpcSetIsPlausible()` by reflection, so they pass identically whether or not this line exists at all; the placement is verified only by the manual reproduction described in the PR body. A later refactor that moves the call below `createNewMappingVersionFromMDTMapping()` (or drops it) stays green.

A single test through the real `importMappingVersionFromMDT()` that asserts `$dungeon->mappingVersions()->count()` and the previous version's `mdt_mapping_hash` are unchanged after the throw would pin it. `MDTMappingImportCrashRecoveryTest` and `MDTMappingImportGameVersionScopingTest` already drive the full pipeline, so the pattern exists in-repo.

.claude/skills/update-mdt-package/SKILL.md
```diff
@@ -257,14 +257,27 @@ Read the output for each dungeon and handle:
   `lang/en_US/mapping.php` with a sensible English string (e.g. `'captive' => 'Captive'`),
   then **notify the user** so they can refine the wording. Only ever edit `lang/en_US`. No
   re-import is needed — translations resolve at display time, not import time.
+- **`MDTMappingNpcSetReplacedException` / "refusing to replace the mapping"**: the import aborted
```
Wotuu, line 261: :robot: While you're in this file: line 230 still documents the signature as `mdt:importmapping {dungeon} {gameVersion} {--force}`, which is now stale — it omits `--allow-npc-set-replacement`. That's the line the next MDT-update session reads when it hits the new bullet you added here and goes looking for the flag.

Wotuu:

tests/Feature/App/Service/MDT/MDTMappingImportNpcSetGuardTest.php
Wotuu: :robot: Fixed in 71be3e1cc — correct, and the test was worse than useless: it read as coverage for the naming while passing with the `$looksLikeDungeonKey` branch deleted.

Your diagnosis of why is exactly right — the arrangement was the mirror image of #3995 (MDT returning the dungeon's *own* NPCs), so `findDungeonKeyOwningNpcs()` filtered them out via `whereNot('dungeon_id', $dungeon->id)` and returned null. Replaced with `exceptionMessage_givenTheDungeonTheNpcsBelongTo_namesIt`, which constructs the exception both ways and asserts the owning dungeon appears with an owner and is absent without one. The lookup itself stays covered by the two `findDungeonKeyOwningNpcs` tests.

app/Service/MDT/MDTMappingImportService.php
Wotuu: :robot: Fixed in 71be3e1cc. Added `importMappingVersionFromMDT_givenMDTNpcsDisjointFromCurrentMapping_createsNothingAndLeavesTheHashUntouched`, which goes through the public entry point with a temporary mapping version holding another dungeon's NPCs, and asserts both halves of the claim: the dungeon's mapping version count is unchanged, and `mdt_mapping_hash` still holds its sentinel.

Mutation-verified rather than assumed — commenting out the guard call at line 108 makes it fail on the `assertNotNull($exception)` and pass again once restored. The temporary rows are removed in a `finally` and I confirmed zero leftovers in the shared test DB afterwards.

.claude/skills/update-mdt-package/SKILL.md
Wotuu: :robot: Fixed in 71be3e1cc — the documented signature now matches. Note it landed differently than it would have: Wotuu decided sub-50% overlap should refuse the import outright rather than warn, so the two thresholds collapsed into one and the dedicated flag is gone entirely. `--force` is the only override, and the signature is back to `{dungeon} {gameVersion} {--force}`. The runbook bullet now says explicitly not to reach for `--force` to silence the guard.

Issue #3988, opened 2026-08-12
Season 2 dungeon mapping checklist: start icons, floor unions, The Blinding Vale (follow-up to #3980)


Wotuu:
:robot: Follow-up to #3980 / #3984, requested by Wotuu in review.

#3980 bumped the MDT composer package to 6.2.1 and reimported the mappings of all 8 dungeons in
the current retail season (Midnight Season 2), each producing a new mapping version:

| Dungeon | New mapping version |
| --- | --- |
| kingsrest | v4 |
| templeofsethraliss | v3 |
| rubylifepools | v9 |
| altar_of_fangs | v4 |
| den_of_nalorakk | v4 |
| murder_row | v4 |
| the_blinding_vale | v5 |
| voidscar_arena | v4 |

The reimport is mechanical (pulled straight from the upstream MDT Lua export) and was not
manually spot-checked dungeon-by-dungeon for map-editor correctness. Wotuu asked for a per-dungeon
checklist to work through.

## Per-dungeon checklist

For **each** of the 8 dungeons above, check in the map editor:

- [x] kingsrest — dungeon start icon present; floor unions set correctly
- [x] templeofsethraliss — dungeon start icon present; floor unions set correctly
- [x] rubylifepools — dungeon start icon present; floor unions set correctly
- [x] altar_of_fangs — dungeon start icon present; floor unions set correctly
- [x] den_of_nalorakk — dungeon start icon present; floor unions set correctly
- [x] murder_row — dungeon start icon present; floor unions set correctly
- [x] the_blinding_vale — dungeon start icon present; floor unions set correctly; **additionally
      double-check this one specifically** — Wotuu flagged it as possibly wrong in #3980/#3984
      (`database/seeders/dungeondata/midnight/the_blinding_vale/`, mapping version v5). The PR
      added two new dungeon-start map icons on floor 1 (`map_icon_type_id: 10`) and new floor 2
      union data; verify both against the in-game layout.
- [x] voidscar_arena — dungeon start icon present; floor unions set correctly

## What to check per dungeon

1. **Dungeon start icon** — does the dungeon have a start marker (`map_icon_type_id: 10`) on the
   correct floor/position? Missing or misplaced start icons block route creation from working
   correctly.
2. **Floor unions** — for multi-floor dungeons, are `floor_unions`/`floor_union_areas` set so the
   floors overlay correctly on the combined facade view? See the `generate-floor-unions` skill for
   how these are computed/regenerated.
[...]

Issue #3240, opened 2026-06-01
Increase PhpStan level


Wotuu:
To what? I don't know - can use Claude to fix the issues that arise.

PR #3999, opened 2026-08-13
#3997 Give The Blinding Vale enemy packs, derived from real combat logs

### LINKED ISSUE #3997: Assign enemy packs to The Blinding Vale, derived from real combat logs
:robot: The Blinding Vale (dungeon 145, `challenge_mode_id` 584) has **no enemy packs** on its
current mapping version (v4 / mv 833, 276 enemies on floor 408). Without packs the map editor
cannot express "I pull one enemy and the three standing next to it come along", which makes the
dungeon effectively unusable for routing.

MDT ships no usable group data for this dungeon (the v5 import that appeared to carry 63 packs was
in fact Den of Nalorakk's mapping — see #3995 / #3996 — and has been removed), so the packs have to
be derived from real gameplay.

## Approach

Two real Blinding Vale combat logs (Mythic 0, WoW 12.0.1) are used as ground truth. They are run
through the existing result-event pipeline and the auto-route creator, and the resulting
`KillZone`s — each one an actual pull — are used to decide pack membership.

1. `combatlog:outputresultevents <dir> --dungeonOrRaid --force`
2. `combatlog:outputcombatlogroutejson <dir> --dungeonOrRaid`
3. ingest the JSON through the auto-route creator (`api.v1.combatlog.route.store` /
   `combatlog:ingestcombatlogroutejson`) with `settings.debugIcons = true`
4. cross-reference the pulls from both runs, fall back to ingame-distance proximity clustering
   where the two runs disagree, and author `enemy_packs.json` + `enemy_pack_id` on `enemies.json`

## Blocker found up front

The result-event pipeline currently detects **zero enemies** in these logs. `<log>_events.txt`
contains 45 result events: 1 `COMBAT_LOG_VERSION`, 1 `ZONE_CHANGE`, 2 `MAP_CHANGE`,
30 `COMBATANT_INFO` and 11 `UNIT_DIED` — and all 11 `UNIT_DIED` are *player* deaths. The raw logs
contain 151 creature `UNIT_DIED` and 176 `PARTY_KILL`, so `accurateEnemySightings` in
`BaseCombatFilter` is never populated and every creature death exits through a `return false` gate.

Ruled out by inspection:

- the zone gate in `Filters/MappingVersion/CombatFilter` (`ZONE_CHANGE` is line 2 and player deaths
  did come through, so it opened)
- `validNpcIds` — `CombatLogService::getResultEventsForDungeonOrRaid()` never calls
  `setValidNpcIds()`, so both gates that consult it are skipped
- combat log version registration — `RETAIL_12_0_1` is registered and `AdvancedDataV22` is selected
- the summon gate — only 1 of 288 summon target GUIDs overlaps the 157 death GUIDs

Root-causing this is part of the work; the fix ships in the same MR as the pack data.

## Scope

- `app/Logic/CombatLog/**` / `app/Service/CombatLog/**` — parser fix + regression test
- `database/seeders/dungeondata/midnight/the_blinding_vale/1/{enemies,enemy_packs}.json` — the packs

Not in scope: the mv 843 corruption (#3995), floor 459 (no enemies).

Wotuu:
:robot: Closes #3997

The Blinding Vale had **no enemy packs**, so the dungeon could not express "I pull one enemy and
the three standing next to it come along" and was unusable for routing. The packs are derived from
two real Mythic 0 runs of the dungeon rather than guessed from positions alone.

Getting there turned up two blockers that had to be fixed first, so this MR is four commits.

---

## 1. Mapping version 5 held Den of Nalorakk's enemies

`the_blinding_vale/1/enemies.json` mapping version 5 (mv 843, from the MDT 6.2.1 reimport)
contained 168 enemies whose **32 NPC ids are all Den of Nalorakk NPCs**. Version 4 (mv 833, 276
enemies) holds the real roster - and those are exactly the NPC ids present in both combat logs
(245410, 245345, 245527, 245339, 245336, …).

Every other Midnight dungeon has 100% NPC overlap between its last two mapping versions; this one
had 0%, so it is isolated rather than a dungeon-index shift. #3996 adds the guard that would have
refused this import; this MR removes the data it left behind. Wotuu deleted the mapping version
locally and ran `mapping:save`; that output is commit 1.

That `mapping:save` also carries one change that is **not** a mapping-version-5 row: mapping version
4 loses a second `dungeon_start` map icon at `(-146.3333, 75.3852)`, keeping the one at
`(-140.25, 32)`. #3984 added the duplicate to versions 3, 4 and 5; Wotuu deleted it from version 4
in the map editor, which is what #3988 asks for. Version 3 keeps its copy - nothing edits old
mapping versions. This matters beyond tidiness because the pack `group` numbering sweeps from the
start icon's bearing, and it is the surviving icon that was used.

## 2. The combat filter discarded every enemy death (local raw-log path only)

> **Not production ARC.** `api.v1.combatlog.route.store` receives an already-built
> `CombatLogRouteRequestDto` from Raider.IO and goes straight to
> `convertCombatLogRouteToDungeonRoute()`, so `BaseCombatFilter` is never on that path. The affected
> consumers are the ones that read a raw `.zip` themselves: `combatlog:outputresultevents`,
> `CombatLogRouteDungeonRouteService::getCombatLogRoute()` (behind
> `combatlog:outputcombatlogroutejson`) and `ResultEventDungeonRouteService` (the admin tools page
> and `combatlog:createdungeonroutes`).

`BaseCombatFilter::parse()` type checked the death branch against `CombatLogEvent`, but a death
arrives as `UNIT_DIED` (or `PARTY_KILL`, which extends it) and those are `GenericSpecialEvent`s - a
separate hierarchy that happens to expose the same `getGenericData()`. Every creature death fell
through to `parseInvalidCombatLogEvent()` and was dropped.

The failure was silent in the worst way: a run still produced a route, it just had **no enemies in
it**.

| | before | after |
| --- | --- | --- |
| result events (run 1 / run 2) | 45 / 35 | 349 / 335 |
| enemy kills (151 / 144 in the logs) | 0 / 0 | 152 / 144 |
| warnings | one per discarded death | none |

Regressed in a3b5cfb47 (#3240, PhpStan level 4), which replaced a plain `else` with the narrower
check. Both hierarchies now implement a shared `HasGenericData` interface so the check can say
"carries generic data" instead of naming one of them, and a null dest GUID no longer fatals.

Covered by `BaseCombatFilterEnemyDeathTest` - 2 of its 3 tests fail on the old code.

## 3. Floor 408's ingame coordinates were never rotated

The bounds came straight from wago.gg's uiMapAssignment table, in raw world-space order. ARC works
in the `(-worldY, worldX)` convention `AdvancedDataV22` produces, so feeding `-worldY` against a
`worldX`-calibrated `min_x` put every enemy far off the map (lng 1895 on a 0..384 axis) and **all
302 NPCs failed to resolve**.

Fixed with the existing `mapping:rotateingamecoords`. Matching combat log positions against the
mapping's own enemies, median error drops from **14.7 to 6.1 lat/lng**, and the runs resolve into
26 and 21 pulls with 26 of 302 unresolved. Only the ingame calibration moves - no `lat`/`lng`
changes, so nothing renders differently.

> **73 other floors across 34 dungeons share this un-rotated calibration** - they are exactly the
> `sizeX/sizeY == +0.667` set noted in `.claude/CLAUDE.md`, where a correctly calibrated floor is
> `±1.5` (the map's 384/256). Among them are Den of Nalorakk, Murder Row and Altar of Fangs, all
> current-season. Out of scope here; filed as #4000.

## 4. The packs

Both logs went through `combatlog:outputresultevents --dungeonOrRaid` →
`combatlog:outputcombatlogroutejson --dungeonOrRaid` → the auto route creator, and the resolved
enemies were clustered on three signals, strongest first:

- **co-aggro** - enemies engaged within 2s of each other in a run were pulled as one. The
  engage-gap distribution has a clear break there (p50 0.4s, p80 4-7s).
- **separation** - enemies a run pulled in *different* bursts are not one pack. Only trusted beyond
  8 ingame yards; closer than that, enemy resolution picks the nearest enemy with a matching npc id
  and genuinely cannot tell two neighbours apart.
- **proximity** - single-linkage at 12 ingame yards, 30 yard span cap, 5 members max, for the 106
  enemies neither run touched.

Wotuu then merged the packs in the map editor - **106 down to 82**, up to 14 members where the
derivation had capped at 5 - and that is what ships. Sizes `1x33, 2x8, 3x12, 4x6, 5x10, 6x4, 7x3,
8x1, 9x2, 13x1, 14x2` over 271 of 277 enemies; the 6 left out are the encounter NPCs (Lightblossom
Trinity x3, Ikuzz the Light Hunter, Lightwarden Ruia, Ziekett).

For what the logs were actually worth on the clustering: re-running the derivation with the combat
log evidence removed entirely gives 107 packs against 106, **77% of them identical**. The logs
changed the pack-mates of 61 of 270 enemies - 28 vetoes (close enough to merge, but a run
demonstrably pulled them apart) and 36 merges proximity alone would not have made. Only 164 of 270
enemies were ever touched by the two runs, so the rest is proximity with no evidence behind it.

## 5. Group numbering follows the route

The first pass numbered groups with the Den of Nalorakk anti-clockwise sweep. That was only ever
right for that dungeon's first floor, and here it ran almost exactly *opposite* to how the dungeon
is played (16%/9% agreement with the runs' kill order). Groups now follow the route:

| groups | leg |
| --- | --- |
| 1-20 | from the start, south, sweeping anti-clockwise to the 3:00 bridge |
| 21-57 | back to the start, north, clockwise to the same bridge |
| 58-69 | down off the bridge and east, ending in Lightwarden Ruia's arena |
| 70-82 | back west along the corridor to the island, ending beside Ziekett |

The combat logs set the phase boundaries rather than a guess at where the bridge is. Both runs
clear exactly the same packs between Ikuzz and Ruia (the eastward leg, lng 274-286 plus the arena)
and between Ruia and Ziekett (the return corridor, lng 176-247), and the ring's own packs stop at
lng 247, so both cuts are unambiguous. NPC composition agrees independently: Lightwarden Ruia,
Spirit Bear and Spirit Moonkin appear only in the eastern arena, and Lightspawn Lasher and
Bloodthorn Roots only in the last group, beside Ziekett.

Within a phase the two arcs are a bearing sweep and the return corridor is east-to-west. Against
the runs' own kill order that scores **91%/88%** on the northern arc, **91%/93%** going east and
**89%/97%** coming back. A nearest-neighbour walk did worse (84%/82%) - it strands packs it skips
past and jumps back for them later. The southern arc has no route evidence at all: both teams took
the northern path from the start, and the one team that did clear the south came at it backwards
from the east afterwards, so it is ordered geometrically.

group order

Green 1-20 south, blue 21-57 north, yellow 58-69 east, pink 70-82 back west.

## Verification

- `BaseCombatFilterEnemyDeathTest` (3 tests) - fails 2/3 without the filter fix
- `--group=CombatLog`: 308 passed
- `composer run fix` / `composer run analyse`: clean
- Seeded with `db:seedone` and checked in the map editor: 106 packs, 270/276 enemies attached, no
  orphan `enemy_pack_id`, no unused pack

### Screenshots

Mapping editor, mapping version 4. Before is the same page with the packs removed.

| | before | after |
| --- | --- | --- |
| overview | before | after |
| zoomed | before | after |

Single-enemy packs draw a box smaller than the enemy icon, so they are not visible as boxes.


database/seeders/dungeondata/midnight/the_blinding_vale/1/map_icons.json
```diff
@@ -63,44 +63,5 @@
         "permanent_tooltip": 0,
         "seasonal_index": null,
         "is_admin": true
-    },
-    {
-        "mapping_version_id": 833,
```
Wotuu, line 68: :robot: This hunk removes three map icons, but only two of them belong to mapping version 843. The block starting here is **mapping version 833** — the version this MR keeps and attaches the 106 new packs to — so a `dungeon_start` icon (`map_icon_type_id: 10`, lat `-146.3333`, lng `75.3852`) is being deleted from the *current* mapping version. That is outside the stated data scope of the MR ("remove what the bad mv 843 import left behind") and is not mentioned in the body.

Counts before/after for this file:

| mv | before | after |
| --- | --- | --- |
| 642 | 1 | 1 |
| 806 | 1 | 1 |
| 814 | 2 | 2 |
| 833 | **2** | **1** |
| 843 | 2 | 0 |

Every other Midnight dungeon has exactly one `dungeon_start` icon per mapping version, so 1 is very plausibly the correct end state and this may well be an incidental fix — but mv 814 keeps its duplicate, so the file ends up inconsistent, and the MR body's pack numbering explicitly keys off "the dungeon start icon's bearing", which makes *which* icon survives load-bearing for the 1..106 group ordering.

Was this deletion intentional? If so it's worth a line in the body (and arguably mv 814 should be cleaned up the same way); if not, the row should come back.

Wotuu:

database/seeders/dungeondata/midnight/the_blinding_vale/1/map_icons.json
Wotuu: :robot: Good catch - that was real and undocumented. Verified against the source of the export: the main checkout's DB has exactly **one** `dungeon_start` icon for mv 833 (`-140.25, 32`), so Wotuu deleted the duplicate at `-146.3333, 75.3852` in the map editor before running `mapping:save`. #3984 had added that second icon to versions 3, 4 and 5, and removing it from the current version is exactly what #3988 asks for. Version 3 keeps its copy because nothing edits old mapping versions.

The pack numbering used the surviving icon (`-140.25, 32`), so the sweep is anchored on the one Wotuu kept.

Documented in the PR body now, under "Mapping version 5 held Den of Nalorakk's enemies".

Wotuu:
:robot: **Fixed the red CI — it was an id collision the rebase couldn't see.**

All four `php-tests` shards were dying during seeding with:

```
Illuminate\Database\UniqueConstraintViolationException (23000):
Duplicate entry '26383' for key 'enemy_packs_temp.PRIMARY'
```

**Cause.** While this branch was open, master gained #3988's Den of Nalorakk enemy packs (`9b893a82c`, `63f4267b0`), which occupy `enemy_packs` ids **26383-26398** — exactly where this branch had started The Blinding Vale's block. The two dungeons' JSON files never overlap textually, so the rebase merged cleanly and the collision only surfaced at seed time, where both sets land in the same table.

**Fix.** Master's data is authoritative, so only The Blinding Vale moved. Its 82 packs are renumbered contiguously from **26399** (global max + 1) to **26480**, in their existing order, and the 271 `enemy_pack_id` references in `enemies.json` were remapped in the same pass.

Files changed:
- `database/seeders/dungeondata/midnight/the_blinding_vale/1/enemy_packs.json` (82 id lines)
- `database/seeders/dungeondata/midnight/the_blinding_vale/1/enemies.json` (271 `enemy_pack_id` values)

Nothing else in the repo referenced those ids — `enemies.json` was the only remap target, and Den of Nalorakk's files were left untouched.

**Verification.**
- Swept all nine id-bearing Blinding Vale seeder files against every other dungeon before editing, in case #3988 had collided on more than packs. It hadn't: `enemies` (132439-134433) is clear, and `enemy_patrols` / `enemy_forces_checkpoints` / `map_icons` / `floor_unions` / `floor_union_areas` / `mountable_areas` / `dungeon_floor_switch_markers` carry no ids for this dungeon. `enemy_packs` was the only collision, and there are now zero.
- `git diff` contains **only** `"id"` and `"enemy_pack_id"` value changes — no formatting drift, so the next `mapping:save` export stays a clean diff.
- Full `php artisan db:seed --database=migrate --force` completes (`DungeonDataSeeder` DONE in 29.6s).
- `tests/Feature/App/Service/MDT`, `tests/Feature/App/Model/Mapping`, `tests/Feature/Traits` and `BaseCombatFilterEnemyDeathTest`: 99 passed.
- `composer run fix` (0 files changed) and `composer run analyse` (no errors) are clean.

Issue #4005, opened 2026-08-13
Auth forms accessibility: unnamed Google button, unlabelled region select, no dialog names, 38px touch targets (follow-up to #4003)


Wotuu:
:robot: Third of three issues from the login/registration design critique (#4003 failure path, #4004 composition). Browser-measured accessibility and touch findings on `common/forms/{login,register,oauth}.blade.php` + `common/general/modal.blade.php`.

## Findings (all measured in headless Chrome)

1. **The Google sign-in link has no accessible name.** `<a href><div class="google_login_image"></div></a>` — an anchor wrapping an empty div whose only content is a CSS background image. Screen readers announce "link", nothing else. A login method a screen-reader user cannot identify.
2. **The region select is unlabelled and its id duplicated** — `html()->select('region', ...)` emits `id="region"` in both page and modal copies (ignores `$modalClass`); `<label for="register_region">`/`for="modal-register_region"` both point at nothing. (Id fix lands with #4003's region work; the label association belongs here.)
3. **Neither modal has an accessible name**: `role="dialog"` but no `aria-labelledby`/`aria-label`; the `<h3>` inside carries no id.
4. **The six `.btn-bnet` buttons are the only controls with no focus indicator at all** — `outline: none` *and* `box-shadow: none`; every other control shows a ring. Their `border-secondary` separator measures **1.36:1** against a 3:1 non-text gate.
5. **Every interactive control is exactly 38px tall**; both checkboxes 40×38. Nothing meets the 44×44 touch minimum, and the region buttons sit in `row g-0` with zero gutter.
6. **The two `<i class="fas fa-info-circle">` tooltip icons are the entire help system** and are non-focusable (no `tabindex`/`role`) and hover-only — unreachable by keyboard and by touch. The only explanation of what the username is used for is invisible to those users.
7. **The three legal links have no `target="_blank"`** — tapping "terms of service" from the register modal destroys the modal, the form, and the page state.
8. `aria-required` absent on every control; the `required` visual cue is a `#c40000` asterisk at **2.10:1** (colour-only visual cue; the `*` is in the accessible name so not a 1.4.3 failure, but the non-text gate fails). Login's fields have `required` but no asterisk at all; `legal_agreed` is server-required but neither `required` nor marked.
9. Register tab order: Confirm password → terms → privacy → cookies → *then* the consent checkbox (checkbox emitted after its own label). Register email missing `autocomplete="email"`.
10. No password-reveal toggle on any of the three password fields.

[...]

Issue #4004, opened 2026-08-13
Auth forms composition: explicit breakpoints, one OAuth button component, remove dead BS3/BS4 CSS (follow-up to #4003)


Wotuu:
:robot: Second of three issues from the login/registration design critique (see #4003 for the failure-path fixes). This one covers the **composition** of `common/forms/{login,register,oauth}.blade.php` and its CSS. Product decision already made: **Battle.net and email/password stay co-equal** — rebuild the presentation, do not reorder the hierarchy.

## Findings

1. **One partial, three layouts, because no breakpoint exists.** The two `.col`s have no responsive variant; stacking only happens when `min-width: auto` forces `flex-wrap` to give out, so the wrap point drifts with content. Measured at 1440px: login **modal** stacked at 442px wide, register **modal** side-by-side at 842px, login **page** side-by-side at 950px — same file. Width difference: `.modal-dialog-small { max-width: 900px }` (`resources/assets/css/lib/modals.css`) loses specificity to Bootstrap's `.modal-dialog { max-width: 500px }`, while `.register-modal-dialog` (`theme.scss`) wins — only the register modal gets its intended width.
2. **Five dead CSS rules around the divider.** Two target `.border-left` (BS4) while the blade emits `border-start` (BS5); two target `.login-modal-dialog`, a class that exists nowhere (`app.blade.php:105` uses `modal-dialog-small`). Their comment says "hide the separating border … when they're stacked vertically" — so the hardcoded pure-`#fff` divider (`border-white`) has been visible on stacked mobile since the BS4→BS5 migration. It's the brightest element on the surface (16.3:1) and appears nowhere else in the design system.
3. **Three OAuth idioms in one ~200px column**: raster PNG Battle.net logo above a `row g-0` of six text buttons; a bare `<img>` link for Discord; an empty `<div class="google_login_image">` backed by a fixed 191×46 raster — the **dark** variant, wrong on the `lux` light theme. `.btn-bnet` hardcodes `#005c99`/`:hover #31b0d5`, colors in no palette. Three `<hr>`s stack in the register right column.
4. **"World" is offered as a Battle.net login region.** `oauth.blade.php` loops `GameServerRegion::all()` (6 rows incl. `world`); `BattleNetLoginController` guards on table-existence so the row passes and builds an OAuth URL against a region that cannot authenticate anyone.
[...]

PR #4006, opened 2026-08-13
#4003 Render auth failures in place: field errors, modal AJAX 422, reset-form and region fixes

### LINKED ISSUE #4003: Auth failure path: no field-level errors, modal loses input, reset form prefills "1", region -1 written to DB
:robot: From a design critique (dual-agent, browser-measured) of the login/registration surface — `resources/views/common/forms/{login,register,oauth}.blade.php` serving both the standalone `/login`·`/register` pages and the `login_modal`/`register_modal` popups in `layouts/app.blade.php`. This issue covers the **failure path**; composition and accessibility are split into their own follow-ups.

## Findings

1. **No field-level errors anywhere in either form.** Neither partial contains `@error`, `$errors->first()`, or `invalid-feedback`. The only error affordance is the `has-error` class toggle — a Bootstrap 3 relic with **zero CSS anywhere in the repo** (same for `help-block`, used 6× in the password blades). Errors surface only via the page-level `.alert-danger` summary from `common/general/messages.blade.php`.
2. **The modal destroys input on every failure.** Measured: submitting the login modal empty posts to `/login?redirect=%2F` and lands back on `/` with the modal closed, the form gone, and input lost. `LoginController::sendFailedLoginResponse()` redirects to the login page; `RegisterController::register()` catches `ValidationException` and returns a hard-coded `redirect('/register')` — the comment above it acknowledges the modal case and ejects anyway. For a user mid-route-planning (unsaved pulls, live-session presence) a mistyped password is a full context loss. Bitter detail: `$redirect = old('redirect', $redirect)` in the partials was written specifically so a failed attempt survives — the hard-coded redirect throws that away.
3. **Password-reset form prefills the email field with the literal text `1`.** `resources/views/auth/passwords/reset.blade.php:15` — `value="{{ $email or old('email') }}"`. Blade's `or` syntax was removed in Laravel 6; this compiles to the PHP `or` operator and yields a boolean. Browser-confirmed: the link from a reset email renders `value="1"`. The form also posts to `route('password.request')` (the GET route) and only works by path collision with `password.update`.
4. **Region writes `-1` to the database by default.** The form field is `region`; the validator declares `game_server_region_id => nullable|int` — a key never submitted, so the rule is a no-op. `create()` does `$data['region'] ?? DEFAULT_REGION`, but the untouched placeholder submits `"-1"` which is not null, so **`game_server_region_id = -1`** is written — a dangling reference that feeds reset-day/epoch logic. Also: `html()->select('region', ...)` ignores `$modalClass`, so the id is `region` in both copies — duplicated on `/register` and matched by no label.

## Fix

- `@error` blocks + `is-invalid` on inputs + `invalid-feedback` messages in `common/forms/login.blade.php`, `register.blade.php`, and both password blades; delete `has-error`/`help-block`/`control-label`/`form-horizontal` dead markup while in there. Add `aria-invalid` and `role="alert"` on the error output.
- Modal failure in place: AJAX submit returning 422 rendered inside the modal, or at minimum make `RegisterController::register()` respect the originating context instead of hard-coding `/register`.
- `reset.blade.php`: `{{ $email ?? old('email') }}`; action → `route('password.update')`.
- Region: validate the actual field per project convention (`'region' => ['nullable', 'integer', 'exists:game_server_regions,id']`), placeholder `value=""`, prefix the select id with `$modalClass`.

## Verify

- Feature tests for: failed login/register renders field errors; register with untouched region select does not write `-1`; reset form prefilled with the actual email.
- Browser: submit each form empty on page and in modal; confirm inline errors and (modal) no page navigation / input preserved.

Full critique snapshot: `.impeccable/critique/2026-08-13T19-44-17Z__resources-views-common-forms.md` (local).

Wotuu:
:robot: Closes #4003

First of three MRs from the login/registration design critique (#4003 failure path, #4004 composition, #4005 accessibility).

## What

- **Field-level validation errors** in `common/forms/login.blade.php` / `register.blade.php` and both password blades: `is-invalid` on the input, message via the shared `form-error` partial, `aria-invalid` + `role="alert"`. The `form-error` partial itself is upgraded from BS3 `help-block` (0 CSS in the repo — it rendered unstyled everywhere) to BS5 `invalid-feedback d-block`; every existing consumer (admin/profile forms) gets styled errors from the same change.
- **Modal forms submit over AJAX** (new `common/forms/authform` inline script, activated on `shown.bs.modal` only for the modal copies): a validation failure now renders as field errors *inside* the modal with input preserved — previously the page navigated, the modal closed, and all input was lost. Success reloads the current page, so the user stays where they were, now authenticated. Non-422 failures (throttle, expired session) fall back to the existing `defaultAjaxErrorFn` toast. `RegisterController::register()` rethrows the `ValidationException` for JSON requests; the non-AJAX page flow keeps its existing redirect.
- **Password reset form**: the legacy Blade `{{ $email or old('email') }}` compiled to the PHP `or` operator since Laravel 6 and prefilled the email field with the literal text **"1"** (browser-confirmed on master, screenshot below). Now `??`. The form also posts to `password.update` instead of `password.request` — it only worked because the GET/POST routes share a path.
- **Region select**: the validator declared `game_server_region_id` — a key the form never submits, so the rule was a no-op and the placeholder's `-1` was written to `users.game_server_region_id` verbatim (a dangling reference feeding reset-day/epoch logic). Now the actual `region` field is validated (`nullable|integer|exists`), the placeholder submits an empty value, the select id gets the `$modalClass` prefix (it was duplicated on `/register` and matched by no label), and old input is restored on a failed page submit. Note: the option map now uses `+` instead of `array_merge` — `array_merge` renumbers numeric keys, and the ids only lined up by coincidence of seed order. `create()`'s no-region fallback also mapped wrongly: `DEFAULT_REGION` is the *short* `'us'`, not an id — it never fired before because the form always submitted `-1`, and would have 500'd in MySQL strict mode once the placeholder became empty. Now `GameServerRegion::ALL[DEFAULT_REGION]`.

## Verification

13 new feature tests (`tests/Feature/Controller/Auth/`), all green. Browser-verified in headless Chrome (worktree stack):

**Login modal, wrong credentials — error in place, URL unchanged, input preserved:**
login modal error

**Register modal, three invalid fields — per-field messages, `aria-invalid`, modal intact:**
register modal errors

**Password reset form, master (before) — the email field contains the literal "1":**
reset before

**Same page on this branch — actual email prefilled:**
reset after

## Notes for review

- The `.is-invalid ~ .invalid-feedback` display rule is broken site-wide: the theme build prefixes it into `.darkly.is-invalid ~ .invalid-feedback` (theme class required on the *input*). Hence `d-block` on all rendered feedback; fixing the SCSS build prefix is out of scope here.
- The two-column layout, OAuth presentation, "World" region button, and ESC dismissal are #4004; accessibility (names, focus, touch targets) is #4005.

### DIFF
--- a/app/Http/Controllers/Auth/RegisterController.php
+++ b/app/Http/Controllers/Auth/RegisterController.php
@@ -10,6 +10,7 @@ use App\Models\User;
 use Illuminate\Auth\Events\Registered;
 use Illuminate\Contracts\Foundation\Application;
 use Illuminate\Foundation\Auth\RegistersUsers;
+use Illuminate\Http\JsonResponse;
 use Illuminate\Http\RedirectResponse;
 use Illuminate\Http\Request;
 use Illuminate\Http\Response;
@@ -20,6 +21,7 @@ use Illuminate\Support\Facades\Validator;
 use Illuminate\Validation\ValidationException;
 use Random\RandomException;
 use Session;
+use Teapot\StatusCode;
 
 class RegisterController extends Controller implements HasMiddleware
 {
@@ -60,11 +62,11 @@ class RegisterController extends Controller implements HasMiddleware
     protected function validator(array $data): \Illuminate\Contracts\Validation\Validator
     {
         return Validator::make($data, [
-            'name'                  => 'required|alpha_dash|max:32|unique:users',
-            'email'                 => 'required|email|max:255|unique:users',
-            'game_server_region_id' => 'nullable|int',
-            'password'              => 'required|min:8|confirmed',
-            'legal_agreed'          => 'required|accepted',
+            'name'         => 'required|alpha_dash|max:32|unique:users',
+            'email'        => 'required|email|max:255|unique:users',
+            'region'       => 'nullable|integer|exists:game_server_regions,id',
+            'password'     => 'required|min:8|confirmed',
+            'legal_agreed' => 'required|accepted',
         ], [
             'legal_agreed.required' => __('controller.register.legal_agreed_required'),
             'legal_agreed.accepted' => __('controller.register.legal_agreed_accepted'),
@@ -84,11 +86,12 @@ class RegisterController extends Controller implements HasMiddleware
 
         /** @var User $user */
         $user = User::create([
-            'public_key'            => User::generateRandomPublicKey(),
-            'name'                  => $data['name'],
-            'email'                 => $data['email'],
-            'echo_color'            => randomHexColor(),
-            'game_server_region_id' => $data['region'] ?? GameServerRegion::DEFAULT_REGION,
+            'public_key' => User::generateRandomPublicKey(),
+            'name'       => $data['name'],
+            'email'      => $data['email'],
+            'echo_color' => randomHexColor(),
+            // ALL maps the region short to its id - DEFAULT_REGION itself is the short ('us'), not an id
+            'game_server_region_id' => $data['region'] ?? GameServerRegion::ALL[GameServerRegion::DEFAULT_REGION],
             'password'              => Hash::make($data['password']),
             'legal_agreed'          => $data['legal_agreed'],
             'legal_agreed_ms'       => intval($data['legal_agreed_ms']),
@@ -102,7 +105,7 @@ class RegisterController extends Controller implements HasMiddleware
     /**
      * Handle a registration request for the application.
      *
-     * @return Application|RedirectResponse|Response|Redirector
+     * @return Application|JsonResponse|RedirectResponse|Response|Redirector
      *
      * @throws ValidationException
      */
@@ -113,7 +116,13 @@ class RegisterController extends Controller implements HasMiddleware
 
         try {
             $validator->validate();
-        } catch (ValidationException) {
+        } catch (ValidationException $validationException) {
+            // The modal register form submits over AJAX and renders the 422's errors inside the
+            // modal - rethrow so the exception handler builds that JSON response
+            if ($request->expectsJson()) {
+                throw $validationException;
+            }
+
             // We always want to redirect to /register, even if you tried to register from modal anywhere on the side
             return redirect('/register')->withInput()->withErrors($validator->messages()->getMessages());
         }
@@ -127,6 +136,15 @@ class RegisterController extends Controller implements HasMiddleware
         // Set the redirect path if it was set
         $this->redirectTo = $request->get('redirect', '/profile');
 
-        return $this->registered($request, $user) ?: redirect($this->redirectPath());
+        if ($response = $this->registered($request, $user)) {
+            return $response;
+        }
+
+        // The modal form reloads the page itself on success - a 201 instead of a redirect keeps
+        // the flashed status message alive for that reload (jquery would silently follow the
+        // redirect and the hidden GET would consume the flash)
+        return $request->wantsJson()
+            ? new JsonResponse([], StatusCode::CREATED)
+            : redirect($this->redirectPath());
     }
 }
--- a/app/Http/Controllers/Auth/RegisterController.php
+++ b/app/Http/Controllers/Auth/RegisterController.php
@@ -142,8 +142,11 @@ class RegisterController extends Controller implements HasMiddleware
 
         // The modal form reloads the page itself on success - a 201 instead of a redirect keeps
         // the flashed status message alive for that reload (jquery would silently follow the
-        // redirect and the hidden GET would consume the flash)
-        return $request->wantsJson()
+        // redirect and the hidden GET would consume the flash).
+        // expectsJson() rather than wantsJson(), to match the failure path above: a plain XHR that
+        // does not send an explicit JSON Accept header would otherwise get a 422 on failure but a
+        // silently followed redirect on success - exactly the flash-eating hop this avoids
+        return $request->expectsJson()
             ? new JsonResponse([], StatusCode::CREATED)
             : redirect($this->redirectPath());
     }
--- a/resources/assets/js/custom/inline/common/forms/authform.js
+++ b/resources/assets/js/custom/inline/common/forms/authform.js
@@ -1,0 +1,145 @@
+/**
+ @typedef {Object} CommonFormsAuthformOptions
+ @property {string} formSelector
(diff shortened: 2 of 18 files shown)


resources/assets/js/custom/inline/common/forms/authform.js
```diff
@@ -0,0 +1,116 @@
...
+            data: $form.serialize(),
+            // Make Laravel's expectsJson() return true so a ValidationException renders as a
+            // 422 JSON response instead of a redirect (which would eject the user out of the modal)
+            headers: {
+                'Accept': 'application/json',
+            },
+            success: function () {
+                // Reload rather than following the response's redirect - the user logged in from
+                // this page, so this page is where they want to be, now authenticated
+                window.location.reload();
```
Wotuu: :robot: The success path silently eats the success flash message. jQuery follows the 302 from a successful login/register transparently, so the browser fetches the redirect target (a full hidden page render) just to throw it away, and that hidden GET **consumes the flashed session message** — `Session::flash('status', ...registered_successfully)` in `RegisterController::register()` is rendered into the discarded response, so after `window.location.reload()` the user never sees the "registered successfully" confirmation (or any other flashed status). Consider having the server return a JSON `201`/`204` for `wantsJson()` on success (the laravel/ui trait does exactly this) — that also saves the wasted full-page fetch before the reload.

app/Http/Controllers/Auth/RegisterController.php
```diff
@@ -84,11 +84,12 @@ protected function create(array $data): User
...
-            'name'                  => $data['name'],
-            'email'                 => $data['email'],
-            'echo_color'            => randomHexColor(),
-            'game_server_region_id' => $data['region'] ?? GameServerRegion::DEFAULT_REGION,
+            'public_key' => User::generateRandomPublicKey(),
+            'name'       => $data['name'],
+            'email'      => $data['email'],
+            'echo_color' => randomHexColor(),
+            // ALL maps the region short to its id - DEFAULT_REGION itself is the short ('us'), not an id
+            'game_server_region_id' => $data['region'] ?? GameServerRegion::ALL[GameServerRegion::DEFAULT_REGION],
```
Wotuu, line 94: :robot: This fixes `-1` writes for *new* registrations only — every user who registered through this form before this release still carries `users.game_server_region_id = -1` (the dangling reference the PR body says feeds reset-day/epoch logic), and nothing in this diff or the issue trail schedules a cleanup. Worth a follow-up data backfill (`UPDATE users SET game_server_region_id = <default id> WHERE game_server_region_id = -1` as an additive migration, safe under the mixed-deploy rules) or at least a filed follow-up issue, otherwise the underlying bug stays live for the entire existing user base.

tests/Feature/Controller/Auth/LoginControllerTest.php
```diff
@@ -0,0 +1,88 @@
...
+
+    #[Test]
+    public function showLoginForm_givenGuest_rendersFieldErrorMarkup(): void
+    {
+        // Act
+        $response = $this->get(route('login'));
+
+        // Assert
+        $response->assertOk();
+        $response->assertSee('login_form', false);
```
Wotuu: :robot: Test name promises more than the assertion delivers: `showLoginForm_givenGuest_rendersFieldErrorMarkup` only asserts the substring `login_form` is present — that matches the plain form id (and also `modal-login_form` from the header modal), and says nothing about field-error markup. Since a guest GET of `/login` has no errors, there is no error markup to assert on this page at all; either rename it to reflect what it checks (the form renders), or make it a real error-markup test by following a failed POST's redirect and asserting `is-invalid`/`invalid-feedback` in the response.

Wotuu:

resources/assets/js/custom/inline/common/forms/authform.js
Wotuu: :robot: Fixed in 3997b958d — `register()` now returns a 201 `JsonResponse` for `wantsJson()` (mirroring `sendLoginResponse`'s existing 204), so no redirect is followed and the flashed status message survives for the modal's page reload. Covered by `register_givenValidDataExpectingJson_returnsCreatedSoTheFlashSurvivesTheReload`.

app/Http/Controllers/Auth/RegisterController.php
Wotuu: :robot: Fixed in 3997b958d — added an additive data migration (`2026_08_13_210000_fix_dangling_game_server_region_id_on_users_table`) pointing existing `-1` rows at the same default a no-region registration now gets. Down is a documented no-op since the `-1`s were corrupt data.

tests/Feature/Controller/Auth/LoginControllerTest.php
Wotuu: :robot: Fixed in 3997b958d — renamed to `showLoginForm_givenFailedLogin_rendersFieldErrorMarkup`: it now fails a real login and asserts `form-control is-invalid`, `aria-invalid="true"` and `invalid-feedback` on the follow-up render.

resources/assets/js/custom/inline/common/forms/authform.js
```diff
@@ -0,0 +1,116 @@
...
+     * @private
+     */
+    _submit(event) {
+        event.preventDefault();
+
+        let $form = $(this.options.formSelector);
+
+        // The register form's own submit handler also writes this, but handler order is not
+        // guaranteed, so write it before serializing (defined in sitescripts.blade)
+        let $legalAgreedMs = $form.find('input[name="legal_agreed_ms"]');
```
Wotuu, line 35: Create a follow-up ticket to remove the legal agreed MS code.

resources/assets/js/custom/inline/common/forms/authform.js
Wotuu: Can we add an extra check so that if the user registers from a model on the /register page, they get redirected to the home page? It's a bit of an edge case, but let's make sure that this is covered.

Wotuu: Yeah do nr 2.

PR #4399, opened 2026-08-31
#4378 Keep the mobile main menu inside the viewport

### LINKED ISSUE #4378: Mobile main menu overflows and cannot be scrolled, hiding bottom options
On mobile, the main menu can extend below the visible screen when there are enough menu items (e.g. with extra admin options), and the menu itself cannot be scrolled to reach the bottom items. The page behind the menu can still scroll, but scrolling does not work over the open menu, so the bottom options are permanently out of reach in that case. Likely only affects users with additional menu options (e.g. admins).

Wotuu:
Closes #4378

## Problem

The site header is `position: sticky; top: 0` and the Bootstrap navbar collapse expands in normal document flow below the brand row. With enough menu items — extra admin entries, a short/landscape viewport — the bottom of the open menu ends up past the bottom of the viewport, and there is no way to reach it: the page scroll moves the document, not the pinned header, and the collapse itself had no overflow of its own.

## Fix

Below the `lg` breakpoint the open collapse gets a max height and scrolls internally.

- The cap is put on the collapse rather than on the header: capping the header would clip the mobile dungeon-context dropdown, which lives in the brand row and is deliberately `position: absolute` there. Dropdowns *inside* the collapse are static below `lg`, so they simply add scrollable height.
- The cap covers `.collapsing` as well as `.show`, because Bootstrap animates the collapse to its *uncapped* scroll height — gated on `.show` alone the menu swelled well past the viewport and snapped back when the animation finished. The `overflow-y` stays on `.show` only, so `.collapsing`'s own `overflow: hidden` still keeps a scrollbar from flashing mid-animation.
- `siteheader.js` publishes the remaining space as `--ksg-navbar-collapse-max-height`. The exact value comes from the ResizeObserver that already watches the header — it fires as the collapse animates open and again whenever the shrink/unshrink transition moves the menu's top edge, so neither needs its own hook or timer. `show.bs.collapse` only seeds an estimate, and it deliberately does **not** trust the collapse's rect: Bootstrap fires that event *before* it removes `.collapse`, so the element is still `display: none` and reports an all-zero rect. The closed-state fallback is the header's own bottom edge, which errs a few pixels low — and erring low keeps every item reachable.
- Publishing a custom property instead of an inline style means the media query stays in charge: crossing into desktop with the menu open cannot leave a stale cap on a navbar that lays out as a single row. It is never cleared for the same reason — it has no effect while the menu is closed, and is always refreshed before the menu becomes visible again.
- The CSS fallback `calc(100dvh - 5rem)` covers any layout rendering the header without that inline script. `dvh`, not `vh`: on mobile Safari `100vh` is the *large* viewport and would overestimate, reintroducing the bug.

## Testing

**Unit** — `vitest run resources/assets/js/custom/inline/common/general/siteheader.test.js`: 21 passed. New cases cover the pure `calculateNavbarCollapseMaxHeight` (remaining viewport, rounding down, clamping at 0) and the DOM wiring (the closed-menu zero rect falling back to the header bottom, refinement once open, recompute on viewport resize and on the header growing back with the menu open, no-op when the header has no collapse).

**Real browser** — headless Chrome at 390×500 against a harness built from the real `bootstrap.min.css` / `bootstrap.bundle.min.js`, this branch's `header.css` and `siteheader.js`, with 20 menu items:

| | peak height while opening | menu bottom ≤ viewport | scrolls internally | last item reachable |
|---|---|---|---|---|
| without the new rule | 790px | ✗ (849px vs 500px) | ✗ | ✗ |
| capped on `.show` only | 790px, snapping back to 442px | ✓ | ✓ | ✓ |
| as shipped | 442px | ✓ (491px) | ✓ (800px content in 442px) | ✓ |

Closing still animates back to zero, and the run reports no console errors or `pageerror`s (in particular no ResizeObserver loop warning from recomputing inside the observer).

**Build** — `node scripts/build/build.mjs` succeeds and the rule lands in the concatenated custom stylesheet.

Not exercised against the live app (this worktree has no running stack), so a glance at the real header on a phone is still worth doing before merge.

https://claude.ai/code/session_014QVEMcRjxXx1LwczWpqUfc

### DIFF
--- a/resources/assets/css/sections/header.css
+++ b/resources/assets/css/sections/header.css
@@ -115,11 +115,17 @@
    in the brand row unclipped; menu dropdowns are static below lg, so they just add scroll height.
    siteheader.js publishes the exact height; the fallback covers layouts that render the header
    without that inline script. `dvh`, not `vh` - on mobile Safari `100vh` is the *large* viewport
-   and would overestimate, reintroducing the bug. Gated on `.show` so the overflow does not fight
-   `.collapsing { overflow: hidden }` and flash a scrollbar mid-animation. */
+   and would overestimate, reintroducing the bug. The cap covers `.collapsing` too - Bootstrap
+   animates to the *uncapped* scroll height, so without it the menu would swell past the viewport
+   and snap back the moment the animation finishes. The overflow stays off `.collapsing`, whose own
+   `overflow: hidden` then keeps a scrollbar from flashing mid-animation. */
 @media (max-width: 991.98px) {
-    .ksg-header .navbar-second .navbar-collapse.show {
+    .ksg-header .navbar-second .navbar-collapse.show,
+    .ksg-header .navbar-second .navbar-collapse.collapsing {
         max-height: var(--ksg-navbar-collapse-max-height, calc(100dvh - 5rem));
+    }
+
+    .ksg-header .navbar-second .navbar-collapse.show {
         overflow-y: auto;
         overscroll-behavior: contain;
     }
--- a/resources/assets/css/sections/header.css
+++ b/resources/assets/css/sections/header.css
@@ -108,17 +108,9 @@
     line-height: 1;
 }
 
-/* The open mobile main menu must stay reachable (#4378): the header is sticky at the top of the
-   viewport and the collapse expands in normal flow below the brand row, so any item past the
-   viewport bottom cannot be scrolled to - the page scroll moves the document, not the pinned
-   header. Capping the collapse itself (rather than the header) keeps the dungeon-context dropdown
-   in the brand row unclipped; menu dropdowns are static below lg, so they just add scroll height.
-   siteheader.js publishes the exact height; the fallback covers layouts that render the header
-   without that inline script. `dvh`, not `vh` - on mobile Safari `100vh` is the *large* viewport
-   and would overestimate, reintroducing the bug. The cap covers `.collapsing` too - Bootstrap
-   animates to the *uncapped* scroll height, so without it the menu would swell past the viewport
-   and snap back the moment the animation finishes. The overflow stays off `.collapsing`, whose own
-   `overflow: hidden` then keeps a scrollbar from flashing mid-animation. */
+/* Caps the open mobile menu to the viewport (#4378) - the sticky header doesn't scroll with the
+   page. `dvh` avoids Safari's oversized `100vh`; capped during `.collapsing` too so it doesn't
+   overshoot then snap back once the open animation finishes. */
 @media (max-width: 991.98px) {
     .ksg-header .navbar-second .navbar-collapse.show,
     .ksg-header .navbar-second .navbar-collapse.collapsing {
--- a/resources/assets/css/sections/header.css
+++ b/resources/assets/css/sections/header.css
@@ -107,3 +107,20 @@
     font-weight: 700;
     line-height: 1;
 }
+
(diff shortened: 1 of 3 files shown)


resources/assets/css/sections/header.css
```diff
@@ -107,3 +107,26 @@
...
+   viewport and the collapse expands in normal flow below the brand row, so any item past the
+   viewport bottom cannot be scrolled to - the page scroll moves the document, not the pinned
+   header. Capping the collapse itself (rather than the header) keeps the dungeon-context dropdown
+   in the brand row unclipped; menu dropdowns are static below lg, so they just add scroll height.
+   siteheader.js publishes the exact height; the fallback covers layouts that render the header
+   without that inline script. `dvh`, not `vh` - on mobile Safari `100vh` is the *large* viewport
+   and would overestimate, reintroducing the bug. The cap covers `.collapsing` too - Bootstrap
+   animates to the *uncapped* scroll height, so without it the menu would swell past the viewport
+   and snap back the moment the animation finishes. The overflow stays off `.collapsing`, whose own
+   `overflow: hidden` then keeps a scrollbar from flashing mid-animation. */
```
Wotuu: Okay, I'm not even going to read all this. Make this a lot shorter, and if you are going to write a comment, make it one, maybe two lines max. We don't need this giant explanation on every single line of code.

resources/assets/js/custom/inline/common/general/siteheader.js
```diff
@@ -100,6 +133,60 @@ class CommonGeneralSiteheader extends InlineCode {
...
+        document.documentElement.style.setProperty('--ksg-navbar-collapse-max-height', `${maxHeight}px`);
+    }
+
+    /**
+     * Where the menu's top edge is - or will be, while it is still closed.
+     *
+     * A closed collapse is `display: none` and reports an all-zero rect, which would pass for a
+     * menu starting at the very top of the viewport and yield a cap a whole header too large. The
+     * closed header's bottom edge is the same row bottom plus the navbar's bottom padding, so it
+     * errs a few pixels low - and erring low keeps every item reachable.
```
Wotuu: Remove this.

resources/assets/js/custom/inline/common/general/siteheader.js
```diff
@@ -100,6 +133,60 @@ class CommonGeneralSiteheader extends InlineCode {
         }
     }
 
+    /**
+     * Keep the mobile main menu inside the viewport (#4378).
+     *
+     * Only the height is published - whether the cap applies at all is CSS's call (header.css
+     * gates it on the mobile breakpoint and on the menu being open), so the property is never
+     * cleared: it simply has no effect while the menu is closed or the navbar lays out as a
+     * single row, and it is always refreshed before the menu becomes visible again.
```
Wotuu: Remove this.

resources/assets/css/sections/header.css
Wotuu: :robot: Fixed - condensed to two sentences.

resources/assets/js/custom/inline/common/general/siteheader.js
Wotuu: :robot: Removed.

resources/assets/js/custom/inline/common/general/siteheader.js
Wotuu: :robot: Removed.

